PR #6, opened 2019-03-12
doc: Code doc

bruno-garcia:


include/sentry.h
```diff
@@ -24,83 +24,155 @@ extern "C" {
 #endif
 #endif
 
+/*
```
jan-auer: If you use `/**` here, we could use doxygen to generate code docs.

bruno-garcia: That's good to know! Let's do that together with adding doxygen

PR #11, opened 2019-03-21
feat: Add premake configuration

tonyo:
This PR adds initial premake configuration for crashpad on Mac, and for breakpad on Linux.


src/breakpad_wrapper.cpp
```diff
@@ -47,7 +47,7 @@ bool callback(const MinidumpDescriptor &descriptor,
 }
 #endif
 
-ExceptionHandler *handler;
+breakpad::ExceptionHandler *handler;
```
tonyo: Apparently CoreFoundation also has something called "ExceptionHandler"

PR #15, opened 2019-05-16
Set up CI with Azure Pipelines

tonyo:


tonyo:
Azure does this automatically :(

PR #23, opened 2019-07-17
ignore cloned crashpad and in-source visual studio cruft

m1keall1son:
We'd like to ignore the new files generated by setting up and building sentrypad out of the box.


tonyo:
Thanks!

PR #25, opened 2019-07-31
Build crashpad_handler as a "Windows" application

icedtoast:
This stops the command window from being shown when it is executed.

### DIFF
--- a/premake/premake5.crashpad.lua
+++ b/premake/premake5.crashpad.lua
@@ -590,7 +590,7 @@ project "crashpad_minidump"
 
 
 project "crashpad_handler"
-  kind "ConsoleApp"
+  kind "WindowedApp"
   crashpad_common()
   links {
     "crashpad_minichromium_base",


tonyo:
Thanks for the PR!
This seems a bit counter-intuitive to me, though.
To recap, with `WindowedApp` kind it will **not** show the window? What will be shown instead?

icedtoast:
WindowedApp in premake changes the subsystem to be "Windows" rather than "Console"; these are described in more detail here:
https://docs.microsoft.com/en-us/cpp/build/reference/subsystem-specify-subsystem?view=vs-2019

So, to answer your question; no window will be shown unless the crashpad_handler creates one (it doesn't).

For some background
Google's crashpad build would produce two outputs,  one called `crashpad_handler.exe` which is against the Windows subsystem and one called `crashpad_handler.com` which is against the Console subsystem. From what I can tell the purpose was so that if you ran `crashpad_handler` from a command prompt the help text would be shown (as crashpad_handler.com would be invoked).

tonyo:
@icedtoast Thanks for the explanation, and sorry for the delay.
We'll revisit it this or next week.

tonyo:
Thanks! 🏄

PR #51, opened 2019-09-23
feat: Backend implementation for Breakpad

jan-auer:


jan-auer:
@tonyo note that this branch contains a few fixes for vscode and also include dirs of `sentry_breakpad`.

PR #72, opened 2019-10-21
feat: Cmake for Android tests 

tonyo:


Makefile
```diff
@@ -105,20 +110,35 @@ linux-shell:
...
-android-configure: $(PREMAKE_DIR)/$(SOLUTION_NAME)_Application.mk
+$(CMAKE_BUILD_DIR):
+	mkdir -p $(CMAKE_BUILD_DIR)
+
+$(CMAKE_BUILD_DIR)/Makefile: $(CMAKE_BUILD_DIR) CMakeLists.txt
+	cd $(CMAKE_BUILD_DIR) && \
+	cmake \
+		-DANDROID_ABI=x86 \
+		-DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK_HOME}/build/cmake/android.toolchain.cmake \
+		-DANDROID_PLATFORM=android-21 ..
```
bruno-garcia: The platform `android-21` seems a bit hidden in here, is this the way to go?
Would it be possible to move it up to some variable? The goal is that to be easier to spot and modify in the future.

tonyo: Added some variables, but this is just a stub now to run/test at least something.

jan-auer:
We'll revise this in the C version soon.

PR #107, opened 2019-12-05
fix: Enable and fix warnings on macOS

jan-auer:


jan-auer:
Fixing all warnings on windows became a bottomless pit. Probably going to turn on defaults on Windows, and all on Linux & Mac. Anyway, this far should already fix the most pressing issues, and we can run a release after this.

src/value.hpp
```diff
@@ -380,7 +380,7 @@ class Value {
         } else if ((m_repr._bits & TAG_INT32) == TAG_INT32) {
             return (double)as_int32();
         } else {
-            return NAN;
+            return nan("");
```
jan-auer: This is a sensitive change, but I verified that the binary pattern is the same (all zeroes at the end).

PR #124, opened 2020-01-16
WIP: adopt current azure-pipeline for new sentry-native code

Swatinem:
This is very much WIP, and currently only macos works.

I have been experimenting with writing the pipeline from scratch to try different ideas, and challenge things as they are done now:

* Use new pipeline artifacts, seems like to be the new recommended way to transfer files between stages: https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/pipeline-artifacts?view=azure-devops&tabs=yaml
* Skip checkout steps (those are currently duplicated in all the individual builders)
* Related: Not sure why the scripts fetch the `sentry-native` sources again, pipelines already does a git checkout.
* Pipelines has builtin support for docker containers, not sure why we have to roll our own
* … and some general streamlining and cleanup…


tonyo:
Sounds promising 👍 

A small note about "Pipelines has builtin support for docker containers": pretty sure it's possible, but one advantage of running docker commands "manually" is that it's probably a bit closer to what a developer would run in local environment, for dev/testing. Feel free to experiment, of course.

Swatinem:
superceded by #130

PR #125, opened 2020-01-16
feat(android): add the bare minimum to get this compiling inside sentry-android-ndk

Swatinem:
With these changes, I get crash events written to disk, and submitted to sentry via `sentry-android` on restart. However there is no stack yet…


Swatinem:
> abstract over file and in memory I/O

Already thinking about this. Maybe something akin to Rusts `Write` trait… Or maybe *always* writing to disk, but I will write down my ideas and run them by you.

PR #127, opened 2020-01-17
ci: Get tests running on both windows and linux

Swatinem:


jan-auer:
@Swatinem slightly updated the title to conform with the commit guidelines.

tests/test_modulefinder.c
```diff
@@ -4,6 +4,8 @@
 
 SENTRY_TEST(test_module_finder)
 {
+// the module finder is not yet implemented on linux
+#if defined(SENTRY_PLATFORM_DARWIN) || defined(SENTRY_PLATFORM_WINDOWS)
```
jan-auer: Maybe add a TODO comment here, so we can track this?

jan-auer: Also, let's confirm, but iirc this is required for Android.

Swatinem:
changed the comment and created a task to get a module finder on linux going.

PR #128, opened 2020-01-20
feat(unwinding): add support for libunwindstack on android

Swatinem:
This is just a copy of the code from the CPP SDK


src/unix/sentry_unix_unwinder_libunwindstack.cpp
```diff
@@ -0,0 +1,65 @@
...
+        ptrs[rv++] = (void *)regs->pc();
+        unwindstack::MapInfo *const map_info = maps.Find(regs->pc());
+        if (!map_info) {
+            break;
+        }
+
+        // the boolean false parameter disables debugdata loading which we don't
+        // want due to size constraints.  Also that data is unlikely to be
+        // useful anyways.
+        unwindstack::Elf *const elf = map_info->GetElf(process_memory, false);
```
jan-auer: These `const` specifiers are sort of strange. Are you sure you don't mean `const Elf *` (equivalent to `Elf const *`)?

Swatinem: I was just copy-pasting that code, but also pushed your suggestions.

PR #129, opened 2020-01-22
feat(modulefinder): Implement a procfs based module finder

Swatinem:
This reads the loaded modules from `/proc/self/maps` and then enumerates
and parses all ELF modules


src/linux/sentry_procmaps_modulefinder.c
```diff
@@ -180,16 +193,13 @@ module_to_value(const sentry_module_t *module)
 #define MIN(a, b) ((a) < (b) ? (a) : (b))
 
 static void
-try_append_module(sentry_value_t *modules, const sentry_module_t *module)
+try_append_module(sentry_value_t modules, const sentry_module_t *module)
 {
-    if (!is_elf_module((const char *)module->start)) {
+    if (!is_elf_module(module->start) || !module->file) {
```
jan-auer: Might be slightly cheaper if you switch these two conditions.

src/linux/sentry_procmaps_modulefinder.c
```diff
@@ -149,19 +163,18 @@ module_to_value(const sentry_module_t *module)
...
         // do appropriate byte-flipping
-        sentry_value_t code_id
-            = sentry__value_new_hexstring(debug_id, strlen(debug_id));
-        sentry_value_set_by_key(mod_val, "code_id", code_id);
+        sentry_value_set_by_key(mod_val, "code_id",
+            sentry__value_new_hexstring((const char *)code_id, code_id_size));
 
-        sentry_uuid_t uuid = sentry_uuid_from_bytes(debug_id);
-        char *uuid_bytes = (char *)uuid.bytes;
+        sentry_uuid_t uuid = sentry_uuid_from_bytes((char *)code_id);
```
jan-auer: _nit_: Cast to `const char *` here.

src/linux/sentry_procmaps_modulefinder.c
```diff
@@ -206,24 +216,33 @@ load_modules(void)
...
         if (n <= 0) {
             if (errno == EAGAIN || errno == EINTR) {
                 continue;
             }
             break;
         }
-        sentry__stringbuilder_append_buf(&sb, buf, n);
+        if (sentry__stringbuilder_append_buf(&sb, buf, n)) {
+            close(fd);
+            goto done;
```
jan-auer: Instead of all those `goto`s, how about we return the number of modules read, and take care of freezing, logs etc in the caller?

src/linux/sentry_procmaps_modulefinder.c
```diff
@@ -232,14 +251,14 @@ load_modules(void)
         if (!read) {
             break;
         }
-        // skip over anonymous mappings
-        if (!module.file
-            || (module.file[0] == '[' && strcmp("[vdso]", module.file) != 0)) {
+        // skip over anonymous mappings, but keep the one for the linux vdso
+        if (module.start != linux_vdso && !module.file
```
jan-auer: For the vdso image, you still need to overwrite the file name with `"linux-gate.so"`.

src/linux/sentry_procmaps_modulefinder.c
```diff
@@ -53,24 +56,27 @@ sentry__procmaps_parse_module_line(char *line, sentry_module_t *module)
...
+    *size_out = 0;
     if (alignment < 4) {
         alignment = 4;
     } else if (alignment != 4 && alignment != 8) {
         return NULL;
     }
 
-    while (offset < end) {
+    const uint8_t *offset = start;
+    while (offset < (uint8_t *)end) {
```
jan-auer: _nit_: Cast to `const uint8_t *` instead.

src/linux/sentry_procmaps_modulefinder.c
```diff
@@ -246,18 +244,22 @@ load_modules(void)
...
-            && (!module.file || module.file[0] == '[')) {
+
+        // for the vdso, we use the special filename `linux-gate.so`
+        if (module.start == linux_vdso) {
+            sentry_free(module.file);
+            module.file = sentry__string_clone("linux-gate.so");
+        }
+        // skip over anonymous mappings
+        if (!module.file || module.file[0] == '[') {
             continue;
```
jan-auer: One last potential memory leak here, then this is g2g 👍

PR #130, opened 2020-01-23
ci: Create a fresh azure config for the C version

Swatinem:
* This removes most of the old CI config and starts from scratch
* It does cmake builds on all the major platforms
* It actually runs the code in the android simulator
* Fetching all the breakpad/crashpad sources is still a TODO, but I have tasks to do that anyway.


Swatinem:
>     * Actually persisting the cache is super slow, no idea why exactly. If it turns out to be slower (with a warm cache) than installing from scratch, it might be a good idea to just remove this.

So…

https://dev.azure.com/sentry-builds/sentry-native/_build/results?buildId=1222&view=logs&j=131a2263-1bc0-5383-2b51-d89179cf40d9 <- this run needs ~4 minutes to persist the cache
https://dev.azure.com/sentry-builds/sentry-native/_build/results?buildId=1223&view=logs&j=131a2263-1bc0-5383-2b51-d89179cf40d9&t=ce140f78-0333-51cd-8a99-78eea1bf789e <- this run needs ~6 minutes to restore the cache…

While actually fetching the fresh Android SDK components takes ~2 minutes…

Srsly Azure, wtf?
CC https://github.com/microsoft/azure-pipelines-tasks/issues/11864

Swatinem:
> Curious: how will our local-linux-dev-and-testing story look like?

I just spin up a docker container inside which I can just run the normal cmake commands…
I basically never leave the container :-D

Also, I’m still undecided if I might just switch to linux anyway, haha :-D

Swatinem:
btw, the build failure on android is basically because master is broken since #128, should be fixed once #129 is there…

PR #135, opened 2020-01-30
feat: Use custom code to get the linux vdso

Swatinem:
this should compile down to api16

### DIFF
--- a/src/linux/sentry_procmaps_modulefinder.c
+++ b/src/linux/sentry_procmaps_modulefinder.c
@@ -284,7 +284,7 @@ load_modules(sentry_value_t modules)
     }
     char *current_line = contents;
 
-    void *linux_vdso = (void *)get_linux_vdso();
+    void *linux_vdso = get_linux_vdso();
 
     // we have multiple memory maps per file, and we need to merge their offsets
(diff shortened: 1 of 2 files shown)


src/linux/sentry_procmaps_modulefinder.c
```diff
@@ -250,12 +284,11 @@ load_modules(sentry_value_t modules)
     }
     char *current_line = contents;
 
-    // See http://man7.org/linux/man-pages/man7/vdso.7.html
-    void *linux_vdso = (void *)getauxval(AT_SYSINFO_EHDR);
+    void *linux_vdso = (void *)get_linux_vdso();
```
jan-auer: _nit_: This cast is redundant.

src/linux/sentry_procmaps_modulefinder.c
```diff
@@ -275,7 +308,8 @@ load_modules(sentry_value_t modules)
         } else if (!module.file.len
```
jan-auer: Should we be defensive here and bail if `module.start == 0`? I'm assuming it's rather unrealistic, but if `/proc/maps` returns a mapping at `0x0` and `linux_vdso` is null, you'll segfault later on.

PR #139, opened 2020-02-05
feat: Add support for attachments which are read lazily when capturin…

Swatinem:
…g events


src/sentry_core.c
```diff
@@ -175,6 +175,19 @@ sentry_capture_event(sentry_value_t event)
...
     if (opts->transport && !sentry_value_is_null(event)) {
         sentry_envelope_t *envelope = sentry__envelope_new();
+
+        SENTRY_TRACE("adding attachments to envelope");
+        for (sentry_attachment_t *attachment = opts->attachments; attachment;
+             attachment = attachment->next) {
+            sentry_envelope_item_t *item = sentry__envelope_add_from_path(
+                envelope, attachment->path, "attachment");
+            if (!item) {
+                sentry_envelope_free(envelope);
```
jan-auer: If you free the envelope here, then you should probably bail the function. However, it seems that you might just want to skip this item since this can totally happen if the file pointed to doesn't exist. This is to be expected.

src/sentry_core.c
```diff
@@ -175,6 +175,19 @@ sentry_capture_event(sentry_value_t event)
     }
     if (opts->transport && !sentry_value_is_null(event)) {
         sentry_envelope_t *envelope = sentry__envelope_new();
+
```
jan-auer: Check for `envelope == NULL` here?

src/sentry_core.c
```diff
@@ -218,6 +231,17 @@ sentry_options_free(sentry_options_t *opts)
     sentry__path_free(opts->handler_path);
     sentry_transport_free(opts->transport);
     sentry__backend_free(opts->backend);
+
+    sentry_attachment_t *next_attachment = opts->attachments;
+    while (next_attachment) {
+        sentry_attachment_t *attachment = next_attachment;
+        next_attachment = attachment->next;
+
+        sentry__path_free(attachment->path);
```
jan-auer: _nit_: For consistency, move this into `sentry__attachment_free`? Otherwise, it'll be hard to find this spot when adding fields to attachments in the future.

PR #141, opened 2020-02-06
test: Make unwinder test work on windows

Swatinem:
this also avoids pulling in `windows.h` and cleans up some other windows related warnings


src/windows/sentry_windows_path.c
```diff
@@ -189,7 +189,7 @@ sentry__path_clone(const sentry_path_t *path)
     if (!rv) {
         return NULL;
     }
-    rv->path = wcsdup(path->path);
+    rv->path = _wcsdup(path->path);
```
jan-auer: This calls malloc internally and is not "signal" safe, I guess. Are we okay with that?

Issue #142, opened 2020-02-06
Memory leak and status of master


xamix:
Hello,

I'm on a Linux 64bits Ubuntu and I'have tested the current [master](https://github.com/getsentry/sentry-native/tree/master).

As far as I can tell there are some memory leak in this branch.

For exemple the new created object in worker at [line 85](https://github.com/getsentry/sentry-native/blob/master/src/worker.cpp#L85) is never deleted.
Looking at the code it seem it need some reworking to be cleaner.

I just came accross the branch [feature/c-ver](https://github.com/getsentry/sentry-native/tree/feature/c-ver) which seem more active than the master.

Is this branch the successor of the master branch?
What is the state of this branch and will it work on Ubuntu?
Should I prefer to use this new branch instead?

Regards

jan-auer:
Hi @xamix,

> I just came accross the branch feature/c-ver which seem more active than the master.

That's right, we've been working on a full rewrite of the SDK and are planning to replace `master` with it. 
As for the current status (we'll keep this updated going forward): 

- [x] Updated test suite with memory leak detection (using ASAN)
- [x] API for capturing events and managing scope data
- [x] Signal handling, symbolication and module lists on Unix (including Android)
- [x] CURL transport for event submission
- [x] Module lists and symbolication on Windows
- [ ] Exception filters for Windows
- [ ] WinHTTP Transport
- [ ] Crash and attachment uploads for the in-process backend
- [ ] Full Crashpad integration
- [ ] Full Breakpad integration.

> Should I prefer to use this new branch instead?

Once _"Crash and attachment uploads for the in-process backend"_ are implemented, you should be able to use the branch on Ubuntu. If you prefer to send Minidumps to Sentry, then you'll need to wait for the full Breakpad integration.

xamix:
Thank you for your answer!

On master I'm not using breakpad for the moment.
I use a custom "external" backend to dump crash (I register signal handler and use boost to safe dump backtrace to disk and send it by converting it into an Exception event payload when application restart).

This was pretty simple, just some function was missing maybe you can add it into the current dev branch:

API missing **sentry_value_new_addr** -> [Value::new_addr](https://github.com/getsentry/sentry-native/blob/master/src/value.cpp#L247) to generate address to string
API missing **sentry_value_new_level** -> [Value::new_level](https://github.com/getsentry/sentry-native/blob/master/src/value.cpp#L247) to push a level into a good formatted string

jan-auer:
[...]

PR #143, opened 2020-02-07
FIX memory leak

xamix:
This fix 3 memory leaks:
- 1) In background worker -> create new task to process which are never deleted
- 2) In delete of option object, transport was not deleted
- 3) MemoryWriter was unreferencing allocated data and so memory was leaked. Delete was called on wrong nullptr address in destructor.

### DIFF
--- a/src/api.cpp
+++ b/src/api.cpp
@@ -258,7 +258,13 @@ sentry_value_t sentry_envelope_get_event(const sentry_envelope_t *envelope) {
 char *sentry_envelope_serialize(const sentry_envelope_t *envelope,
                                 size_t *size_out) {
     const transports::Envelope *e = (const transports::Envelope *)envelope;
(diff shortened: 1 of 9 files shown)


xamix:
The second commit which FIX MemoryWriter leak  is in my sense ugly and could be better (was just a quick FIX).

Maybe by not freeing memory in MemoryWriter destructor and freeing memory at higher level. If I have time I will look at that.

jan-auer:
@xamix Thanks for this submission. We've just merged the C rewrite into `master`, which should get rid of many leaks. We only ran an LLVM-ASAN based leak detector on macOS so far, so there might still be a few cases we missed. 

I hope you don't mind if I'm closing this PR now. Feel free to let us know in #142 if you find more leaks (as we're going to add more leak detection to CI in near future). Thanks for your efforts.

PR #144, opened 2020-02-10
feat: Enqueue unsent envelopes on startup

Swatinem:
In the case of a crash, envelopes are serialized to disk in a special `run` folder. Those are enumerated on startup and loaded up as `raw` envelopes (avoiding parsing the serialized format), and are then sent as `application/x-sentry-envelope` to the store API.


src/sentry_database.c
```diff
@@ -0,0 +1,80 @@
...
+}
+
+bool
+sentry__run_write_envelope(const sentry_run_t *run, sentry_envelope_t *envelope)
+{
+    // 37 for the uuid, 9 for the `.envelope` suffix
+    char envelope_filename[37 + 9];
+    sentry_uuid_t event_id = sentry__envelope_get_event_id(envelope);
+    sentry_uuid_as_string(&event_id, envelope_filename);
+    strncpy(&envelope_filename[36], ".envelope", 10);
```
jan-auer: _nit_: Use `strcpy` here to avoid length mismatch?

src/sentry_envelope.c
```diff
@@ -17,6 +17,9 @@ typedef enum {
 } endpoint_type_t;
 
 typedef enum {
+    // When loading an envelope from disk, we don’t really want to parse it,
+    // but just pipe it through as a raw buffer
+    ENVELOPE_ITEM_TYPE_ENVELOPE,
```
jan-auer: Let's please find a way to not represent envelopes as items. It's totally fair to have a flag or raw payload, as you do this below with `is_raw_envelope`, and direct it to the store endpoint.

src/sentry_database.c
```diff
@@ -0,0 +1,80 @@
...
+{
+    sentry_pathiter_t *db_iter
+        = sentry__path_iter_directory(options->database_path);
+    const sentry_path_t *run_dir;
+    while ((run_dir = sentry__pathiter_next(db_iter)) != NULL) {
+        sentry_pathiter_t *run_iter = sentry__path_iter_directory(run_dir);
+        const sentry_path_t *envelope_file;
+        while ((envelope_file = sentry__pathiter_next(run_iter)) != NULL) {
+
+            if (options->transport) {
```
jan-auer: Could you check here that the file is actually an envelope (e.g. check for the file extension)?

src/sentry_envelope.c
```diff
@@ -59,11 +68,16 @@ sentry__prepared_http_request_free(sentry_prepared_http_request_t *req)
 static sentry_envelope_item_t *
 envelope_add_item(sentry_envelope_t *envelope)
 {
-    if (envelope->item_count >= MAX_ENVELOPE_ITEMS) {
+    if (envelope->is_raw) {
+        return NULL;
```
jan-auer: _nit_: Should we add a warning message here, since this indicates an implementation bug?

Swatinem: true. Lets keep it in mind. But in theory users should deal with `NULL` already.

PR #147, opened 2020-02-11
feat: Add Crashpad Integration

Swatinem:
This will pull in crashpad via submodule, and hooks it up via CMake.
It also fixes some issues around Scope notifications and Serialization.


src/backends/sentry_crashpad_backend.cpp
```diff
@@ -0,0 +1,199 @@
...
+            = crashpad::CrashpadInfo::GetCrashpadInfo();
+        crashpad_info->set_system_crash_reporter_forwarding(
+            crashpad::TriState::kDisabled);
+    }
+}
+
+static void
+flush_scope(sentry_backend_t *backend, const sentry_scope_t *scope)
+{
+    sentry_value_t event = sentry_value_new_object();
```
jan-auer: _nit_: Comment here that we deliberately do **not** create an event so that the scope payload does not contain default values like the `timestamp` or `event_id`.

src/backends/sentry_crashpad_backend.cpp
```diff
@@ -0,0 +1,199 @@
...
+    SENTRY_WITH_SCOPE (scope) {
+        // we want the scope without any modules or breadcrumbs
+        sentry__scope_apply_to_event(scope, event, SENTRY_SCOPE_NONE);
+    }
+
+    size_t mpack_size;
+    char *mpack = sentry_value_to_msgpack(event, &mpack_size);
```
jan-auer: Please add a null-check and free the buffer at the end.

src/backends/sentry_crashpad_backend.cpp
```diff
@@ -0,0 +1,199 @@
...
+    sentry_value_decref(event);
+
+    crashpad_state_t *data = (crashpad_state_t *)backend->data;
+    int _rv = sentry__path_write_buffer(data->event_path, mpack, mpack_size);
+    // TODO: check rv?
```
jan-auer: Since `flush_scope` doesn't return the error, you could log an error here. Same for the function below.

src/backends/sentry_crashpad_backend.cpp
```diff
@@ -0,0 +1,199 @@
...
+}
+
+static void
+add_breadcrumb(sentry_backend_t *backend, sentry_value_t breadcrumb)
+{
+    crashpad_state_t *data = (crashpad_state_t *)backend->data;
+
+    bool first_breadcrumb = data->num_breadcrumbs % SENTRY_BREADCRUMBS_MAX == 0;
+
+    sentry_path_t *breadcrumb_file
```
jan-auer: _nit_: `const sentry_path_t *`

src/backends/sentry_crashpad_backend.cpp
```diff
@@ -0,0 +1,199 @@
...
+        // substantial time *after* crashpad has done its job.
+        crashpad::CrashpadInfo *crashpad_info
+            = crashpad::CrashpadInfo::GetCrashpadInfo();
+        crashpad_info->set_system_crash_reporter_forwarding(
+            crashpad::TriState::kDisabled);
+    }
+}
+
+static void
+flush_scope(sentry_backend_t *backend, const sentry_scope_t *scope)
```
jan-auer: _nit_: While these are all static functions, they could still be named something with `crashpad_` to aid code search. How about `sentry__crashpad_flush_scope`?

src/sentry_core.c
```diff
@@ -217,6 +218,19 @@ sentry_options_new(void)
...
+    if (exe_dir) {
+        opts->handler_path = sentry__path_join_str(exe_dir,
+#ifdef SENTRY_PLATFORM_WINDOWS
+            "crashpad_handler.exe"
+#else
+            "crashpad_handler"
+#endif
+        );
+    }
+
```
jan-auer: Initialize `system_crash_reporter_enabled` explicitly to `false`/`0`.

src/sentry_scope.h
```diff
@@ -16,20 +16,31 @@ typedef struct sentry_scope_s {
     sentry_level_t level;
 } sentry_scope_t;
 
+typedef enum {
+    SENTRY_SCOPE_NONE = 0x0,
+    SENTRY_SCOPE_BREADCRUMBS = 0x1,
+    SENTRY_SCOPE_MODULES = 0x2,
+    // TODO: SENTRY_SCOPE_STACKTRACES = 0x4,
+    SENTRY_SCOPE_ALL = 0x7,
```
jan-auer: _nit_: `~0`?

src/unix/sentry_unix_path.c
```diff
@@ -20,6 +25,50 @@ struct sentry_pathiter_s {
...
+    if (err) {
+        sentry_free(buf);
+        return NULL;
+    }
+    return sentry__path_from_str_owned(buf);
+#elif defined(SENTRY_PLATFORM_LINUX)
+    // inspired by:
+    // https://github.com/rust-lang/rust/blob/0176a9eef845e7421b7e2f7ef015333a41a7c027/src/libstd/sys/unix/os.rs#L328-L337
+    char buf[4096];
+    readlink("/proc/self/exe", buf, 4096);
```
jan-auer: This does not terminate the link properly. See https://linux.die.net/man/2/readlink:
> `readlink()` does not append a null byte to _buf_.

```suggestion
    ssize_t len = readlink("/proc/self/exe", buf, sizeof(buf) - 1);
    if (len < 0) {
        return NULL;
    }
    buf[len] = 0;
```

Can you please also add a test for this?

src/unix/sentry_unix_path.c
```diff
@@ -20,6 +25,50 @@ struct sentry_pathiter_s {
...
+    return sentry__path_from_str(buf);
+#endif
+    return NULL;
+}
+
+sentry_path_t *
+sentry__path_dir(sentry_path_t *path)
```
jan-auer: Since this internally modifies the path, shouldn't you pass a `const sentry_path_t *path` and clone it internally?

src/unix/sentry_unix_path.c
```diff
@@ -20,6 +25,50 @@ struct sentry_pathiter_s {
...
+{
+    // dirname can modify its argument, and may return pointers to static memory
+    // that we are not allowed to free.
+    char *buf = path->path;
+    char *dir = dirname(buf);
+    path->path = sentry__string_clone(dir);
+    if (!path->path) {
+        sentry__path_free(path);
```
jan-auer: This branch leaks `buf`.

src/windows/sentry_windows_path.c
```diff
@@ -35,6 +35,24 @@ path_with_len(size_t len)
...
 }
 
+sentry_path_t *
+sentry__path_current_exe(void)
+{
+    // inspired by:
+    // https://github.com/rust-lang/rust/blob/183e893aaae581bd0ab499ba56b6c5e118557dc7/src/libstd/sys/windows/os.rs#L234-L239
+    sentry_path_t *path = path_with_len(MAX_PATH);
+    size_t _len = GetModuleFileNameW(NULL, path->path, MAX_PATH);
+    // TODO: check _len
```
jan-auer: Yes, could log an error here. `_len` will be zero on error.

src/backends/sentry_crashpad_backend.cpp
```diff
@@ -0,0 +1,199 @@
...
+    if (!g_db || !g_db->GetSettings()) {
+        return;
+    }
+    g_db->GetSettings()->SetUploadsEnabled(!sentry__should_skip_upload());
+}
+
+static void
+startup_crashpad_backend(sentry_backend_t *backend)
+{
+    // TODO: backends should really get the options as argument
```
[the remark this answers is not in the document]
Swatinem: I would rather do this as a followup. As I shortly mentioned, the usage of options is *super* confusing and unsafe right now.

src/sentry_core.c
```diff
@@ -217,6 +218,19 @@ sentry_options_new(void)
     }
     memset(opts, 0, sizeof(sentry_options_t));
     opts->database_path = sentry__path_from_str("./.sentry-native");
+
+    sentry_path_t *exe_dir = sentry__path_current_exe();
```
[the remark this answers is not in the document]
Swatinem: I do agree. But then it is not clear whether the user *explicitly* sets the `options->handler_path` to `NULL`, or if its just the default. So the users intention is not clear.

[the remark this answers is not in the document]
jan-auer: What behavior would you expect from an active Crashpad backend if the handler is null?

CMakeLists.txt
```diff
@@ -33,10 +39,21 @@ if(CURL_FOUND)
...
+
 #set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror=implicit-function-declaration -Werror=incompatible-function-pointer-types -Wall -fvisibility=hidden")
 if(NOT WIN32)
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -fvisibility=hidden")
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+endif()
+
+if(WITH_CRASHPAD AND WIN32)
+	# Disable duplicate-define warnings, as the crashpad build defines all the
+	# "slim windows.h" defines, which sentry_boot also defines.
```
jan-auer: Can't you do an `#ifndef` then, instead?

Makefile
```diff
@@ -22,14 +22,19 @@ test: update-test-discovery build/sentry_tests
...
-	@$(MAKE) -C leak-build sentry_tests
+	@cmake --build leak-build --target sentry_tests --parallel
 	@ASAN_OPTIONS=detect_leaks=1 LSAN_OPTIONS=suppressions=leak-suppressions.txt ./leak-build/sentry_tests
 .PHONY: test-leaks
 
 clean: build/Makefile
 	@$(MAKE) -C build clean
 .PHONY: clean
 
+setup: setup-git
```
jan-auer: `.PHONY: setup`

src/sentry_core.c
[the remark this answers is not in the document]
jan-auer: Discussed offline: Let's move it as suggested. The reason is that the crashpad backend only really makes sense if there is a handler defined and so we don't want to offer users the possibility to explicitly set it to `NULL`.

PR #152, opened 2020-02-26
fix: Generate and install debug symbols

Swatinem:


CMakeLists.txt
```diff
@@ -33,6 +37,23 @@ if(BACKEND_INPROC AND WIN32)
 	message(FATAL_ERROR "The in-process backend is not supported on Windows")
 endif()
 
+# use -O3 when doing `RelWithDebInfo` builds
+string(REPLACE "-O2" "-O3" CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO}")
+string(REPLACE "-O2" "-O3" CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
```
jan-auer: Could you also define this for ASM? I'm not sure where, but I remember that we had some ASM files in one of the C++ builds. It might have been breakpad. If you want, you can reduce code duplication with a macro (see [here](https://github.com/Kitware/CMake/blob/3ec986ce8e8df269eb6b6b9f37e12b02194168fd/Modules/Compiler/Intel.cmake#L19) for an example) or with a `foreach` block as you use below.

Also, this will work for all compilers except MSVC, which uses a [different syntax](https://github.com/Kitware/CMake/blob/2a26a6de012ff12c343456abc2236e60e95e15c5/Modules/Platform/Windows-MSVC.cmake#L408).

PR #153, opened 2020-02-26
test: Move test suite into pytest

Swatinem:
This moves unit tests into pytest, in preparation for integration testing


tests/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+
+def enumerate_unittests():
+    regexp = re.compile("XX\((.*?)\)")
+    tests = []
+    # TODO: actually generate the `tests.inc` file with python
+    with open('tests/unit/tests.inc', 'r') as testsfile:
+        for line in testsfile:
+            match = regexp.match(line)
+            if match:
+                tests.append(match.group(1))
```
untitaker: you can use generator functions here: `yield match.group(1)`

azure-pipelines.yml
```diff
@@ -66,11 +62,8 @@ stages:
             fetchDepth: 5
             submodules: recursive
           - script: |
-              mkdir build && cd build
-              cmake ..
-              cmake --build . --parallel --target sentry_tests
-            displayName: Build
-          - script: .\build\sentry_tests.exe
+              pip install --upgrade --requirement integration-test-requirements.txt
```
jan-auer: Does this also need `UsePythonVersion@0`? Might be worth factoring this out?

tests/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+    subprocess.run(configcmd, cwd=cwd, check=True)
+
+    buildcmd = ["cmake", "--build", ".", "--parallel"]
+    for target in targets:
+        buildcmd.extend(["--target", target])
+    subprocess.run(buildcmd, cwd=cwd, check=True)
+
+# Adapted from: https://raw.githubusercontent.com/getsentry/sentry-python/276acae955ee13f7ac3a7728003626eff6d943a8/sentry_sdk/envelope.py
+
+class Envelope(object):
```
jan-auer: Let's import this from `sentry_sdk` as soon as it is released.

tests/test_integration.c
```diff
@@ -0,0 +1,94 @@
...
+    if (has_arg(argc, argv, "overflow-breadcrumbs")) {
+        for (size_t i = 0; i < 101; i++) {
+            char buffer[4];
+            sprintf(buffer, "%zu", i);
+            sentry_add_breadcrumb(sentry_value_new_breadcrumb(0, buffer));
+        }
+    }
+
+    if (has_arg(argc, argv, "crash")) {
+        memset((char *)0x0, 1, 100);
```
jan-auer: Having this in the example would be really great, too.

tests/test_integration.py
```diff
@@ -0,0 +1,97 @@
...
+def assert_meta(envelope):
+    event = envelope.get_event()
+
+    expected = {
+        "platform": "native",
+        "release": "integration-test-release",
+        "user": { "id": 42, "username": "some_name" },
+        "transaction": "test-transaction",
+        "tags": { "expected-tag": "some value" },
+        "extra": { "extra stuff": "some value" },
```
jan-auer: Let's also text `contexts` and `environment`, please.

tests/test_integration.c
Swatinem: I actually plan to remove the example completely, and replace it with the integration test file ;-)

PR #158, opened 2020-03-03
feat: Add WinHTTP support

Swatinem:


src/sentry_sync.c
```diff
@@ -122,7 +121,7 @@ sentry__bgworker_shutdown(sentry_bgworker_t *bgw, uint64_t timeout)
     uint64_t started = sentry__msec_time();
     while (true) {
         sentry__mutex_lock(&bgw->task_lock);
-        bool done = bgw->task_count == 0 || !bgw->running;
+        bool done = bgw->task_count == 0 && !bgw->running;
```
jan-auer: Seems like the idea was that we can break out of our loop sooner and start to `join` the thread, since it's going to shut down anyway. However, this is also correct, so we can leave it like this, imho.

PR #159, opened 2020-03-03
ci: Clean up more files from source archive

Swatinem:


jan-auer:
Running `cmake -B build` now gives this:
```
CMake Error at CMakeLists.txt:214 (add_executable):
  Cannot find source file:

    ./tests/test_integration.c
```

I think we'll have to bring back the tests, unless we figure out a way to exclude tests from the main CMake file.

Swatinem:
*insert random rage at cmake gif here*

PR #160, opened 2020-03-03
feat: Add sdk information to scope

jan-auer:
Adds SDK version to the scope:

image


tests/test_integration.py
```diff
@@ -35,6 +35,16 @@ def assert_meta(envelope):
         "transaction": "test-transaction",
         "tags": { "expected-tag": "some value" },
         "extra": { "extra stuff": "some value" },
+        "sdk": {
```
Swatinem: ❤️ ah how nice to have these tests!

PR #161, opened 2020-03-03
fix: Memory leaks

jan-auer:
Updates leak tests and fixes a leak in the crashpad backend.


src/backends/sentry_crashpad_backend.cpp
```diff
@@ -70,6 +71,7 @@ sentry__crashpad_backend_startup(sentry_backend_t *backend)
 
     base::FilePath database(options->database_path->path);
     base::FilePath handler(handler_path->path);
+    sentry__path_free(handler_path);
```
Swatinem: This is not safe in the case when the user explicitly sets it, because `g_options` owns it. So rather defensively copy it?

jan-auer: Excellent point, completely missed that.

PR #168, opened 2020-03-11
build: Add installation support and fix cmake files

madebr:
This pr:
- refactors the cmake build script a bit by:
   - reordering,
   - adding a cmake install configuration script,
   - installing all libraries for shared and dynamic builds of the sentry library
- fix the enumeration in the readme
- fix a warning about redefinitions of some compile definitions
- fix error about [wrong calling convention](https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/hosting/lpthread-start-routine-function-pointer)

I tested this on Linux and MSVC


Amphaal:
Changing WIN32 flags to MSVC are indeed required since Linux>Windows cross-compilation will fail to build Crashpad right now.

CMakeLists.txt
```diff
@@ -51,58 +66,87 @@ if(APPLE)
...
+	if(APPLE)
+		sentry_install(FILES "$<TARGET_FILE:sentry>.dSYM" DESTINATION "${CMAKE_INSTALL_LIBDIR}")
+	elseif(MSVC)
+		sentry_install(FILES "$<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:$<TARGET_PDB_FILE:sentry>>"
+			DESTINATION "${CMAKE_INSTALL_BINDIR}")
+	endif()
 endif()
 
-FIND_PACKAGE(CURL)
-if(CURL_FOUND)
```
Swatinem: I think you should keep this `if`, and just override the `SENTRY_CURL_SUPPORT` var with the result.
CI fails on platforms that don’t have curl available.

CMakeLists.txt
```diff
@@ -111,107 +155,96 @@ elseif(NOT WIN32)
...
-	ENDFOREACH(f)
-	SET(${var} "${listVar}" PARENT_SCOPE)
-ENDFUNCTION(PREPEND)
-
 if(BACKEND_CRASHPAD)
 	add_subdirectory(external/crashpad EXCLUDE_FROM_ALL)
-	include_directories(external/crashpad external/crashpad/third_party/mini_chromium/mini_chromium)
-	add_definitions(-DSENTRY_WITH_CRASHPAD_BACKEND)
-	set(LINK_LIBRARIES ${LINK_LIBRARIES} crashpad_client crashpad_util)
-	install(TARGETS "crashpad_handler")
```
Swatinem: is this missing?

CMakeLists.txt
```diff
@@ -51,58 +66,87 @@ if(APPLE)
 	if(DSYMUTIL_PROGRAM)
 		foreach(lang C CXX)
 			foreach(var LINK_EXECUTABLE CREATE_SHARED_LIBRARY)
-				set(CMAKE_${lang}_${var} "${CMAKE_${lang}_${var}}" "${DSYMUTIL_PROGRAM} <TARGET>")
+				#set(CMAKE_${lang}_${var} "${CMAKE_${lang}_${var}}" "${DSYMUTIL_PROGRAM} <TARGET>")
```
Swatinem: I think we still need this, because it is not the default on mac :-(

CMakeLists.txt
```diff
@@ -51,58 +66,87 @@ if(APPLE)
...
+	#set_property(TARGET sentry PROPERTY C_VISIBILITY_PRESET hidden)
+	target_compile_options(sentry PUBLIC $<BUILD_INTERFACE:-Wall>)
 endif()
 
-if(BACKEND_CRASHPAD AND WIN32)
+if(BACKEND_CRASHPAD AND MSVC)
 	# Disable duplicate-define warnings, as the crashpad build defines all the
 	# "slim windows.h" defines, which sentry_boot also defines.
-	add_definitions(/wd4005)
+	target_compile_options(sentry PUBLIC $<BUILD_INTERFACE:/wd4005>)
```
Swatinem: since you explicitly check for these with `#if !defined` now in code, maybe remove this from the cmake file.

CMakeLists.txt
```diff
@@ -111,107 +155,96 @@ elseif(NOT WIN32)
...
-)
-get_target_property(SENTRY_DEFS sentry COMPILE_DEFINITIONS)
-target_compile_definitions("sentry_test_unit" PUBLIC SENTRY_UNITTEST ${SENTRY_DEFS})
-target_link_libraries("sentry_test_unit" ${LINK_LIBRARIES})
+get_target_property(SENTRY_COMPILE_DEFINITIONS sentry COMPILE_DEFINITIONS)
+get_target_property(SENTRY_INTERFACE_INCLUDE_DIRECTORIES sentry INTERFACE_INCLUDE_DIRECTORIES)
+get_target_property(SENTRY_INCLUDE_DIRECTORIES sentry INCLUDE_DIRECTORIES)
+get_target_property(SENTRY_LINK_LIBRARIES sentry LINK_LIBRARIES)
+
+add_executable(sentry_test_unit EXCLUDE_FROM_ALL ${SENTRY_NATIVE_ALL_SOURCES})
```
gocarlos: eventually make this optional e.g. add option and set it to `ON`. this allows consumers of this lib to not have to compile the tests...

same goes for the example

gocarlos:
@madebr are you working on https://github.com/conan-io/conan-center-index/issues/1089?

CMakeLists.txt
Swatinem: good comment! I thought the `EXCLUDE_FROM_ALL` was enough. But cmake still processes all the files. We would ideally want to exclude the `./tests/` dir from the source file, but ran into cmake problems.

CMakeLists.txt
madebr: My opinion on this matter is that it's better to fail explicitly when a desired configuration cannot be set.
By failing, you know the binary will be as configured.
By overriding, you have to check the logs whether CURL has been found.
Maybe those platforms that do not have curl, should set `SENTRY_CURL_SUPPORT=OFF`?

CMakeLists.txt
madebr: Yes, I removed it.
But I will add it again, since it might be useful for consumers.

CMakeLists.txt
madebr: Oops, I removed it while modifying the cmake script.
(Enabling the CXX language conditionally when using the crashpad backend, will fail cmake)
This line can be re-enabled again.

CMakeLists.txt
madebr: Yeah, I did not check what was the reason for adding this warning disabler option.

CMakeLists.txt
madebr: Same for the example(s)?
SENTRY_BUILD_TESTS
SENTRY_BUILD_EXAMPLES

madebr:
> @madebr are you working on [conan-io/conan-center-index#1089](https://github.com/conan-io/conan-center-index/issues/1089)?

Looks like it :smile: 
Let's first have upstream a nice cmake build script before adding it to CCI.

CMakeLists.txt
madebr: In the above comment, I meant I removed the `install(TARGETS "crashpad_handler")`.
[The rest is still available](https://github.com/madebr/sentry-native/blob/77f721a870e17c121560077b21e8ba13383f896f/CMakeLists.txt#L191-L198)

madebr: I guess you want to unconditionally install `crashpad_handler`?

madebr:
@Amphaal 

> Changing WIN32 flags to MSVC are indeed required since Linux>Windows cross-compilation will fail to build Crashpad right now.

Indeed.

After a small fix to sentry's crashpad cmake scripts (replacing WIN32 with MSVC), the build fails at:
https://github.com/getsentry/crashpad/blob/8a2ddfc0a475bff9e25bfc7af693bac373adebe3/compat/win/sys/types.h#L18-L19

CMakeLists.txt
Swatinem: well yes. the crashpad backend is basically useless without that executable.

CMakeLists.txt
Swatinem: I think that makes sense

CMakeLists.txt
madebr: I disabled CURL in my latest batch of patches.

CMakeLists.txt
madebr: This is fixed in the latest patches.

CMakeLists.txt
madebr: I could remove them.

CMakeLists.txt
madebr: Done

CMakeLists.txt
madebr: @Swatinem
I present some alternatives:
- only enable curl on those platforms where you know CURL is installed (such as `sys.platform == "darwin"`)?
- use a package manager to install curl
e.g. conan:


<details>
<summary>Build using conan</summary>

Create a `conanfile.py` file in the root of `sentry-native` with the following contents:
```python
from conans import ConanFile, CMake, tools
from conans.errors import ConanInvalidConfiguration


class F(ConanFile):
    settings = "os", "arch", "compiler", "build_type"
    options = {
        "shared": [True, False],
        "fPIC": [True, False],
        "backend": ["default", "none", "crashpad", "inproc"],
        "with_curl": [True, False],
    }
    default_options = {
        "shared": False,
        "fPIC": True,
        "backend": "default",
        "with_curl": True,
    }

    generators = "cmake_find_package"

    def config_options(self):
        if self.settings.os == "Windows":
            del self.options.fPIC

    def configure(self):
        if self.options.shared:
            del self.options.fPIC
        if self.options.backend == "default":
            if self.settings.os == "Windows" or tools.is_apple_os(self.settings.os):
                self.options.backend = "crashpad"
            elif self.settings.os == "Linux":
                self.options.backend = "inproc"
            else:
                raise ConanInvalidConfiguration("Unsupported os")
        if self.settings.os == "Linux" and self.options.backend == "crashpad":
            raise ConanInvalidConfiguration("crashpad backend is unsupported on Linux")
        elif self.settings.os == "Windows" and self.options.backend == "inproc":
            raise ConanInvalidConfiguration("inproc backend is unsupported on Windows")

    def requirements(self):
        if self.options.with_curl:
            self.requires("libcurl/7.68.0")

    def build(self):
        cmake = CMake(self)
        cmake.definitions["SENTRY_CURL_SUPPORT"] = self.options.with_curl
        cmake.definitions["SENTRY_BACKEND"] = str(self.options.backend).lower()
        cmake.configure()
        cmake.build()

    def package(self):
        cmake = CMake(self)
        cmake.install()

    def test(self):
        cmake = CMake()
        cmake.test()
```

and run the following commands:
```sh
python -m pip install --user conan
mkdir build
cd build
python -m conans.conan install ..
python -m conans.conan build ..
```

</details>

madebr:
I have some questions
- Until now, I haven't touched libunwindstack and libstacktrace. Are these backends, equivalent to `inproc` and `crashpad`?
- Is libunwindstack a backend or can it be used together with another backend?
- It looks like libbacktrace is not used. There is only mention of a `WITH_LIBBACKTRACE` variable, but nothing else.
- Is CURL support usable on all platforms? Windows, Macos, Linux, Android, IPhone, ...?

Amphaal:
> @Amphaal
> 
> > Changing WIN32 flags to MSVC are indeed required since Linux>Windows cross-compilation will fail to build Crashpad right now.
> 
> Indeed.
> 
> After a small fix to sentry's crashpad cmake scripts (replacing WIN32 with MSVC), the build fails at:
> https://github.com/getsentry/crashpad/blob/8a2ddfc0a475bff9e25bfc7af693bac373adebe3/compat/win/sys/types.h#L18-L19

...And that's only the tip of the iceberg 😆 I'll try to push a patch for the Sentry's Crashpad fork once I get the whole process working with MinGW.

src/CMakeLists.txt
```diff
@@ -1,127 +1,126 @@
...
+	list(APPEND SENTRY_SOURCES
+		unwinder/sentry_unwinder_dbghelp.c
+	)
 endif()
 
 # make paths absolute
 PREPEND(SENTRY_SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/" ${SENTRY_SOURCES})
-get_target_property(SENTRY_OLD_SOURCES sentry SOURCES)
-target_sources(sentry PRIVATE ${SENTRY_OLD_SOURCES} ${SENTRY_SOURCES})
+target_sources(sentry PRIVATE ${SENTRY_SOURCES})
```
Swatinem: does this work correctly?
I had the problem that just using `target_sources` overrides the `vendor/mpack.c` set in the main cmake file, thats why I did this dance here. But CI is passing, so I hope everything is correct?

tests/__init__.py
```diff
@@ -15,10 +15,17 @@ def check_output(*args, **kwargs):
     stdout = stdout.replace(b"\r\n", b"\n")
     return stdout
 
-def cmake(cwd, targets, options=[]):
+def cmake(cwd, targets, options=None):
+    if options is None:
+        options = {}
+    options.update({
+        "CMAKE_RUNTIME_OUTPUT_DIRECTORY": cwd,
```
Swatinem: 👍 this indeed looks like the better place for this than inside the cmake file!

tests/conftest.py
```diff
@@ -23,7 +20,7 @@ class Unittests:
     def __init__(self, dir):
         # for unit tests, the backend does not matter, and we want to keep
         # the compile-times down
-        cmake(dir, ["sentry_test_unit"], ["SENTRY_BACKEND=none"])
+        cmake(dir, ["sentry_test_unit"], {"SENTRY_BACKEND":"none", "SENTRY_CURL_SUPPORT":"OFF",})
```
Swatinem: 👍

CMakeLists.txt
```diff
@@ -57,161 +85,164 @@ if(APPLE)
...
-	${CMAKE_CURRENT_SOURCE_DIR}/tests/unit/*.c
+target_include_directories(sentry PUBLIC
+	"$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>"
+	"$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>"
+	"$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>"
 )
 
+# Without this, dladdr can't find functions
+target_link_libraries(sentry INTERFACE
+	"$<$<OR:$<PLATFORM_ID:Linux>,$<PLATFORM_ID:Android>>:-Wl,--build-id=sha1,-E>")
```
Swatinem: We had this only for the unit tests at first, since they need to have their functions findable by `dladdr`, but I guess this is fine to add in general

jan-auer: `-E` was really intended for tests only, and I would prefer to keep it on the test target. 

We do need `--build-id=sha1` on the library in case we want to support building with `clang` on Linux. Otherwise, `clang` does not generate a build note which Sentry needs to compute a debug identifier. So, we can keep that flag here.

Swatinem:
I just merged breakpad support in #169 , which will cause conflicts here.

Sorry for the churn, and I hope I can learn some cmake skills by looking at how you adapt this PR.

src/CMakeLists.txt
madebr: `target_sources` only adds sources (as a set), it does not remove.
All sources are now added as an absolute path, so `vendor/mpack.c` cannot be overridden.

I can imagine the problem you're describing happening if there are 2 `vendor/mpack.c` sources in different folders, and the sources are added as a relative path. But that is not the case here.

tests/__init__.py
madebr: The only disadvantage is that a quick `cmake .. && make` will create the test executable in the `test/unit` subfolder.
But you can always run `make test`.

CMakeLists.txt
madebr: So keep it as is?
This link flag will also be exported when packaging `sentry-native`.

Swatinem: Maybe copy-paste the code, once without the `-E` for normal libsentry, and once with `-E` for the unit tests.

madebr:
A word of caution, all unit tess have now CURL disabled.
But with some pytest magic, I think it can be easily conditionally enabled.

CMakeLists.txt
madebr: Yup, that's what is done now.

Swatinem:
> A word of caution, all unit tess have now CURL disabled.

I think thats fine for now. The tests don’t test http yet anyway. I want to get there eventually.

Swatinem:
Thank you very much @madebr , awesome work!

PR #181, opened 2020-03-16
feat: Add fallback buildid hashing

Swatinem:


src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -178,45 +236,41 @@ module_to_value(const sentry_module_t *module)
...
     size_t code_id_size;
     const uint8_t *code_id = get_code_id_from_elf(module->start, &code_id_size);
+    sentry_uuid_t uuid = sentry_uuid_nil();
     if (code_id) {
         // the usage of these is described here:
         // https://getsentry.github.io/symbolicator/advanced/symbol-server-compatibility/#identifiers
-        // in particular, the debug_id is a `little-endian GUID`, so we have to
-        // do appropriate byte-flipping
+        // in particular, the debug_id is a `little-endian GUID`, so we
+        // have to do appropriate byte-flipping
```
jan-auer: _nit_: The comment probably belongs after the `if` block?

src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -162,8 +165,63 @@ get_code_id_from_elf(void *base, size_t *size_out)
...
+        const Elf64_Ehdr *elf = base;
+        const Elf64_Shdr *strheader = (const Elf64_Shdr *)(addr + elf->e_shoff
+            + elf->e_shentsize * elf->e_shstrndx);
+
+        const char *names = (const char *)(addr + strheader->sh_offset);
+        for (int i = 0; i < elf->e_shnum; i++) {
+            const Elf64_Shdr *header = (const Elf64_Shdr *)(addr + elf->e_shoff
+                + elf->e_shentsize * i);
+            const char *name = names + 1 + header->sh_name;
+            if (header->sh_type == SHT_PROGBITS && strcmp(name, "text") == 0) {
```
jan-auer: Adding `+1` to the name and then comparing to `"text"` seems a bit obscure. Could you compare to `".text"` directly?

src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -12,6 +12,9 @@
 #include <string.h>
 #include <unistd.h>
 
+#define MAX(a, b) ((a) > (b) ? (a) : (b))
```
jan-auer: _nit_: Works in this case, but note that this would evaluate the parameters twice if they were function calls.

src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -162,8 +165,63 @@ get_code_id_from_elf(void *base, size_t *size_out)
...
+                break;
+            }
+        }
+    }
+
+    sentry_uuid_t uuid = sentry_uuid_nil();
+
+    // adapted from
+    // https://github.com/getsentry/symbolic/blob/8f9a01756e48dcbba2e42917a064f495d74058b7/debuginfo/src/elf.rs#L100-L110
+    size_t max = MAX(text_size, _SC_PAGESIZE);
```
jan-auer:
```suggestion
    size_t max = MIN(text_size, 4096);
```

- This needs to take the smaller of both, to ensure we read _at most_ 4k bytes.
- `_SC_PAGESIZE` is not guaranteed to be 4k, but Breakpad has 4k hardcoded.

src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -293,7 +272,33 @@ try_append_module(sentry_value_t modules, const sentry_module_t *module)
...
+    // we would be poking into invalid memory. To be safe, we load the file
+    // from disk, so we have the on-disk layout, and are independent of how the
+    // runtime linker would load or re-order any sections. The exception here is
+    // the linux-gate, which is not an actual file on disk, so we actually poke
+    // at its memory.
+    sentry_value_t mod = sentry_value_new_null();
+
+    if (sentry__slice_eq(module->file, LINUX_GATE)) {
+        mod = sentry__procmaps_module_to_value(module, module->start);
+    } else {
```
jan-auer: This means we're now copying every file into memory now on every platform. This is potentially a *lot* and might seriously halt the application for a while. Can we potentially detect that the module is not loaded properly and only do this conditionally? Alternatively, could we memory-map here to reduce the impact of this a bit?

src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -293,7 +272,33 @@ try_append_module(sentry_value_t modules, const sentry_module_t *module)
...
+        char *filename = sentry__slice_to_owned(module->file);
+        sentry_path_t *file = sentry__path_from_str_owned(filename);
+        if (file) {
+            char *buf = sentry__path_read_to_buffer(file, NULL);
+            sentry__path_free(file);
+            if (buf) {
+                mod = sentry__procmaps_module_to_value(module, buf);
```
jan-auer: Could create a fake `sentry_module_t` structure here instead of introducing an `addr` parameter? Also, this is potentially unsafe, since the `module->end` pointer no longer fits the address. Probably something like this would be better:

```c
            size_t buf_len;
            char *buf = sentry__path_read_to_buffer(file, &buf_len);
            sentry__path_free(file);
            if (buf) {
                sentry_module_t loaded_module = { buf, buf + buf_len, module->file };
                mod = sentry__procmaps_module_to_value(loaded_module);
```

Swatinem: I thought about this, but the `start`/`end` pairs are actually used and put into the resulting `value_t`. Not sure about the naming, but we need to have two different things, one for the module as it is really loaded, and one just for the elf parsing -_-

src/modulefinder/sentry_modulefinder_linux.c
Swatinem: I would prefer mmap, since the OS should be smart enough to not create a separate copy for the already mapped libraries. I will try that tomorrow.

src/modulefinder/sentry_modulefinder_linux.c
Swatinem: maybe have one function that populated a `sentry_value_t` from a raw pointer to an ELF file.

src/modulefinder/sentry_modulefinder_linux.c
jan-auer: Sounds good. Even though mmap introduces some complexity, I think it's well worth it. Thanks!

src/modulefinder/sentry_modulefinder_linux.c
jan-auer: I see, thanks for clarifying. Let's keep your version then. 👍

PR #186, opened 2020-03-19
fix: Avoid some compile warnings

Swatinem:


examples/example.c
```diff
@@ -4,8 +4,9 @@
 #include <string.h>
 
 static void
-print_envelope(sentry_envelope_t *envelope, void *_data)
+print_envelope(sentry_envelope_t *envelope, void *unused_data)
```
jan-auer: Can you use `UNUSED` here?

Swatinem: The example only uses public API, which `UNUSED` isn’t…

PR #187, opened 2020-03-19
tests: Test HTTP transports

Swatinem:


tests/__init__.py
```diff
@@ -79,6 +90,11 @@ def cmake(cwd, targets, options=None):
 
 # Adapted from: https://raw.githubusercontent.com/getsentry/sentry-python/276acae955ee13f7ac3a7728003626eff6d943a8/sentry_sdk/envelope.py
 
+def event_envelope(jsonstr):
+    j = json.loads(jsonstr)
+    item = Item(headers={"type":"event"}, payload=PayloadRef(json=j))
+    return Envelope(items=[item])
```
jan-auer: `sentry-sdk` is going to be released soon with their envelope implementation, then this code can be removed.

PR #191, opened 2020-03-23
feat: Dump in-flight requests in case of a crash

Swatinem:
The inproc and breakpad backends will invoke the default transports dump method,
that method will first check that it is actually the transport responsible,
and will then dump all the envelopes in matching tasks.
The bg worker is changed in a way that it only pops tasks *after* they are completed.
That way, in case of a race, we end up with duplicate events rather than dropped events.

### DIFF
--- a/src/backends/sentry_backend_breakpad.cpp
+++ b/src/backends/sentry_backend_breakpad.cpp
@@ -6,7 +6,9 @@ extern "C" {
 #include "sentry_core.h"
 #include "sentry_envelope.h"
 #include "sentry_path.h"
+#include "sentry_sync.h"
 #include "sentry_transport.h"
+#include "sentry_unix_pageallocator.h"
 }
 
 #ifdef __GNUC__
--- a/src/backends/sentry_backend_breakpad.cpp
+++ b/src/backends/sentry_backend_breakpad.cpp
@@ -96,6 +96,9 @@ sentry__breakpad_backend_callback(
     const google_breakpad::MinidumpDescriptor &descriptor,
     void *UNUSED(context), bool succeeded)
 {
+    sentry__page_allocator_enable();
+    sentry__enter_signal_handler();
+
     const sentry_options_t *options = sentry_get_options();
     const char *dump_path = descriptor.path();
 
@@ -116,6 +119,7 @@ sentry__breakpad_backend_callback(
         sentry__transport_dump_queue(transport);
     }
 
+    sentry__leave_signal_handler();
     return succeeded;
 }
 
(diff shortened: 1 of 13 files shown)


src/backends/sentry_backend_breakpad.cpp
```diff
@@ -102,13 +102,18 @@ sentry__breakpad_backend_callback(
...
     sentry__enforce_breakpad_transport(options, dump_path);
 
     // after the transport is set up, we will capture an event, which will
     // create an envelope with all the scope, attachments, etc.
     sentry_value_t event = sentry_value_new_event();
     sentry_capture_event(event);
 
+    // after capturing the crash event, try to dump all the in-flight data of
+    // the previous transport
+    sentry__transport_dump_queue(transport);
```
jan-auer: Is it possible that `transport` can be `NULL` here (or is there always a dummy transport)? If so, we should only execute this conditionally.

src/sentry_sync.c
```diff
@@ -200,6 +199,51 @@ sentry__bgworker_submit(sentry_bgworker_t *bgw,
     return 0;
 }
 
+int
+sentry__bgworker_foreach_matching(sentry_bgworker_t *bgw,
+    sentry_task_function_t exec_func,
+    bool (*callback)(void *task_data, void *data), void *data)
+{
+    sentry__mutex_lock(&bgw->task_lock);
```
jan-auer: Could this deadlock if the mutex is held at the time of the crash?

src/transports/sentry_transport_winhttp.c
```diff
@@ -286,3 +286,8 @@ sentry__transport_new_default(void)
 
     return transport;
 }
+
+void
+sentry__transport_dump_queue(sentry_transport_t *UNUSED(transport))
+{
+}
```
jan-auer: Is this intentionally empty?

Swatinem: well, kind of. Windows doesn’t have any inproc backend, so this is basically useless. I can implement it, its basically the same as for the curl backend, so there are no surprises once windows will get an inproc backend.

src/sentry_sync.c
Swatinem: Hm, now that you are mentioning this. The inproc backend has special code that switches the mutex implementation (and malloc) in case of a signal handler. According to the comment, it’s supposed to avoid deadlocks, and give the signal handler thread access.

I’m a bit surprised that there are no issues with the breakpad backend. Isn’t that running inside the signal handler?

src/transports/sentry_transport_winhttp.c
jan-auer: 👍 that would be great. Otherwise it might be too easy to miss this.

src/sentry_sync.c
jan-auer: Yes, Breakpad registers signal handlers which eventually invoke your callback. I also missed this in the initial review -- we probably have to add this lock override, too.

I was looking for it at the right spot, but didn't realize that it was `sentry__block_for_signal_handler` what I was looking at :)

PR #192, opened 2020-03-23
fix: Properly clean up global module cache

Swatinem:
And also fix some crashpad related issues:
* fix crash on invalid DSN
* fix crash on invalid handler_path


src/backends/sentry_backend_crashpad.cpp
```diff
@@ -134,8 +140,8 @@ sentry__crashpad_backend_startup(sentry_backend_t *backend)
     sentry__crashpad_backend_user_consent_changed(backend);
 
     crashpad::CrashpadClient client;
-    char *minidump_url = sentry__dsn_get_minidump_url(&options->dsn);
-    std::string url(minidump_url);
+    auto minidump_url = sentry__dsn_get_minidump_url(&options->dsn);
```
jan-auer: Let's not use `auto` please, since this makes it harder to read the code and miss mistakes (e.g. `auto` vs `auto *`).

src/modulefinder/sentry_modulefinder_apple.c
```diff
@@ -132,10 +132,25 @@ sentry__modules_get_list(void)
     sentry__mutex_lock(&g_mutex);
     if (!g_initialized) {
         g_modules = sentry_value_new_list();
+        // TODO: maybe use `_dyld_image_count` and `_dyld_get_image_header`?
+        // Those functions are documented to not be thread-safe, though using
+        // the `register_X` functions are also unsafe because they lack a
+        // corresponding `unregister` function, and will thus crash when sentry
+        // itself is unloaded.
```
jan-auer: Does this even apply here? You're running inside a mutex lock, so thread safety shouldn't be a concern.

Swatinem: I think its rather that you can `dlopen` in other threads, so the count/get-image get out of sync… per docs, get-image should gracefully return NULL on out-of-bounds, but with concurrent changes, there is a chance to miss some modules.

jan-auer: I see, thanks for clarifying 👍

PR #196, opened 2020-03-26
[CMake] Warning options should not propagate

Amphaal:
Using sentry-native as a submodule with `add_subdirectory` enforces compilers options about warnings handling, which are then inherited in the main project. This should not happen.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,5 +9,5 @@
 	url = https://chromium.googlesource.com/linux-syscall-support
 [submodule "external/crashpad"]
 	path = external/crashpad
-	url = https://github.com/Amphaal/crashpad
+	url = https://github.com/getsentry/crashpad.git
 	branch = getsentry
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,5 +9,4 @@
 	url = https://chromium.googlesource.com/linux-syscall-support
 [submodule "external/crashpad"]
 	path = external/crashpad
-	url = https://github.com/getsentry/crashpad.git
-	branch = getsentry
+	url = https://github.com/Amphaal/crashpad
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,5 +9,5 @@
 	url = https://chromium.googlesource.com/linux-syscall-support
 [submodule "external/crashpad"]
 	path = external/crashpad
-	url = https://github.com/getsentry/crashpad.git
+	url = https://github.com/Amphaal/crashpad
 	branch = getsentry
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,7 +209,7 @@ if(MSVC)
 		set(CMAKE_ASM_MASM_FLAGS "${CMAKE_ASM_MASM_FLAGS} /safeseh")
 	endif()
 else()
-	target_compile_options(sentry PUBLIC $<BUILD_INTERFACE:-Wall -Wextra -Wpedantic>)
+	target_compile_options(sentry PRIVATE $<BUILD_INTERFACE:-Wall -Wextra -Wpedantic>)
 endif()
 
 target_include_directories(sentry PUBLIC
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,7 +187,7 @@ endif()
 if(BUILD_SHARED_LIBS)
 	target_compile_definitions(sentry PRIVATE SENTRY_BUILD_SHARED)
 else()
-	target_compile_definitions(sentry PUBLIC SENTRY_BUILD_STATIC)
+	target_compile_definitions(sentry PRIVATE SENTRY_BUILD_STATIC)
 endif()
 
 if(SENTRY_TRANSPORT_CURL)
@@ -258,7 +258,6 @@ if(SENTRY_BACKEND_CRASHPAD)
 		"$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/external/crashpad/third_party/mini_chromium/mini_chromium>"
 	)
 	# we include a crashpad header, which itself needs C++14 support
-	target_compile_features(sentry PUBLIC cxx_std_14)
 	target_link_libraries(sentry PRIVATE crashpad_client crashpad_util)
 	if(NOT BUILD_SHARED_LIBS)
 		sentry_install(TARGETS crashpad_client crashpad_util crashpad_compat getopt mini_chromium zlib EXPORT sentry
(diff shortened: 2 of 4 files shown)


Swatinem:
https://github.com/getsentry/sentry-native/blob/e67c29349261044c1154f5708e09b8daf9b3ccb1/CMakeLists.txt#L187-L191

lol. I srsly have no idea what this `PUBLIC`/`PRIVATE` business is all about…

Amphaal:
> https://github.com/getsentry/sentry-native/blob/e67c29349261044c1154f5708e09b8daf9b3ccb1/CMakeLists.txt#L187-L191
> 
> lol. I srsly have no idea what this `PUBLIC`/`PRIVATE` business is all about…

Yeah, the documentation is not that explicit on the meaning. To make it simple, there are 3 types of visibility you can set on a CMake target : `INTERFACE`, `PRIVATE `and `PUBLIC`. 
- `INTERFACE`, which means the arguments following are NOT used on the target itself, but ONLY the others targets that use this one. It is fairly specific, and means everything that will use this specific target inherits the arguments. This is what I mean by "propagation".
- `PRIVATE`, which is what you should use by default, means the arguments following are ONLY used in the context of the specific target you bind them to, and are not propagated. It's the safest option and generally what you are looking for.
- `PUBLIC`, which combines PRIVATE and PUBLIC behavior.

Swatinem:
> by the others targets that are linked to this one.

But only direct dependants, not transitive ones?

Seeing that the build problems right now are related to crashpad…
So this means that when sentry links to all the crashpad libs, when those have c++14 set as `PUBLIC` target, sentry itself will *inherit* those settings? Because we need them to actually include the crashpad headers. Right?

And while you are at it, we override some other CFLAGS and the linker cmd to actually generate dSym directories on macos. How is that best done to not mess with other projects that are including sentry?

Amphaal:
> But only direct dependants, not transitive ones?

I think it goes all the way up the dependency tree. 


> Seeing that the build problems right now are related to crashpad…

Is there any way to see the error logs of pytest when it runs a subprocess ? I had no clue on why those tests failed :P

> So this means that when sentry links to all the crashpad libs, when those have c++14 set as `PUBLIC` target, sentry itself will _inherit_ those settings? Because we need them to actually include the crashpad headers. Right?

In that case, sentry would acknoledge the information for sure. I think something like (or its `target_set_properties` equivalent)

```
SET(CMAKE_CXX_STANDARD 14)
SET(CMAKE_CXX_STANDARD_REQUIRED ON)
```

defined in the CMakeLists.txt of Crashpad is the recommanded way of telling this thing, as it sets the compilers flags depending on the compiler used and checks the "inheritor" compatibility. Since `CMAKE_CXX_STANDARD` is set in sentry-native, there should be no need for manuals `target_compile_features`.

> And while you are at it, we override some other CFLAGS and the linker cmd to actually generate dSym directories on macos. How is that best done to not mess with other projects that are including sentry?

Sure, I'll check it. Small disclosure, I'm only a self-taught CMake user and thus can still do silly / useless things... CMake is really flexible, and IMO it's its strength and weakness.

Swatinem:
> Is there any way to see the error logs of pytest when it runs a subprocess ? I had no clue on why those tests failed :P

You can pass `--capture=no`. But for building only, you don’t need pytest, you can just run cmake manually. The pytest suite however runs quite a few permutations, depending on the platform.

> Small disclosure, I'm only a self-taught CMake user and thus can still do silly / useless things... CMake is really flexible, and IMO it's its strength and weakness.

Don’t sweat it! I only started learning CMake with this project. And coming from the JS ecosystem, I was wondering if something could be worse than dealing with webpack and JS tooling. CMake certainly did not disappoint there, haha. I still have regular nightmares.

CMakeLists.txt
```diff
@@ -187,7 +187,7 @@ endif()
 if(BUILD_SHARED_LIBS)
 	target_compile_definitions(sentry PRIVATE SENTRY_BUILD_SHARED)
 else()
-	target_compile_definitions(sentry PUBLIC SENTRY_BUILD_STATIC)
+	target_compile_definitions(sentry PRIVATE SENTRY_BUILD_STATIC)
```
madebr: You might want to keep this option public.
On Windows, this definition is used to set `__declspec(dllimport)` if it is shared and `extern` if it is static.

.gitmodules
```diff
@@ -9,5 +9,5 @@
 	url = https://chromium.googlesource.com/linux-syscall-support
 [submodule "external/crashpad"]
 	path = external/crashpad
-	url = https://github.com/getsentry/crashpad.git
+	url = https://github.com/Amphaal/crashpad
```
madebr: Are you sure you want to move crashpad's repo from getsentry to Amphaal's?

CMakeLists.txt
Swatinem: That is a good point

.gitmodules
Amphaal: Nah my bad, this was supposed to be only for master :') I'm preparing an upgrade for crashpad MinGW compatibility (for Linux XCompilation), i'm reverting this asap.

CMakeLists.txt
Amphaal: > You might want to keep this option public.
> On Windows, this definition is used to set `__declspec(dllimport)` if it is shared and `extern` if it is static.

Unless I understand `PUBLIC` / `PRIVATE` wrong ; as long as `sentry.h` (which is the only file that I found that uses `SENTRY_BUILD_STATIC`) is in sentry's `target_sources`, `PRIVATE `is enough and will produce the expected behavior.

Amphaal:
Alright, Github, VSCode git implementation and my own skills are not mixing well together, this pull request is now a mess. I'll put a hold on this massacre, and will come back with a cleaner draft once I'll get my Jenkins CI Linux XCompilation working. Sorry for the noise and the time lost.

CMakeLists.txt
madebr: > Unless I understand `PUBLIC` / `PRIVATE` wrong ; as long as `sentry.h` (which is the only file that I found that uses `SENTRY_BUILD_STATIC`) is in sentry's `target_sources`, `PRIVATE `is enough and will produce the expected behavior.

It's the combination that is important. When a dependency includes `senty.h`, it needs to know whether it is a `shared` or `static` sentry, that is in use.
You can also `generate` a header that will automagicaly set the correct defines.
See [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html).

`target_sources` is a way to add sources to the target itself and/or to its dependencies.

Swatinem:
No problem. I’m very happy to encourage external contributors ;-)

PR #198, opened 2020-03-27
feat: Implement sample rate to randomly throw away events

Swatinem:


src/sentry_core.c
```diff
@@ -200,6 +201,15 @@ event_is_considered_error(sentry_value_t event)
 sentry_uuid_t
 sentry_capture_event(sentry_value_t event)
 {
+    const sentry_options_t *opts = sentry_get_options();
+    uint64_t rnd;
+    if (opts->sample_rate < 1.0 && !sentry__getrandom(&rnd, 8)
```
jan-auer: _nit_: 
```suggestion
    if (opts->sample_rate < 1.0 && !sentry__getrandom(&rnd, sizeof(rnd))
```

src/sentry_core.c
```diff
@@ -364,6 +374,18 @@ sentry_options_get_dsn(const sentry_options_t *opts)
     return opts->raw_dsn;
 }
 
+void
+sentry_options_set_sample_rate(sentry_options_t *opts, double sample_rate)
+{
+    opts->sample_rate = sample_rate;
```
jan-auer: _nit_: Maybe normalize it to `(0.0, 1.0)` so that `get_sample_rate` always returns a valid rate?

PR #201, opened 2020-03-27
Handle Exception API

cammm:
Allows manually triggering of the backend given a completed ucontext

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -653,8 +653,10 @@ SENTRY_API sentry_uuid_t sentry_capture_event(sentry_value_t event);
 
 /*
  * Captures an exception to be handled by the backend.
+ *
+ * This is safe to be called from a crashing thread and may not return.
  */
-SENTRY_EXPERIMENTAL_API void sentry_handle_exception(sentry_ucontext_t *context);
+SENTRY_EXPERIMENTAL_API void sentry_handle_exception(sentry_ucontext_t *uctx);
 
 /*
  * Adds the breadcrumb to be sent in case of an event.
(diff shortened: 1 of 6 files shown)


src/sentry_backend.h
```diff
@@ -6,10 +6,12 @@
 #include "sentry_scope.h"
 
 struct sentry_backend_s;
+struct sentry_ucontext_s;
```
Swatinem: doesn’t the `sentry_boot.h` above pull in `sentry_ucontext_t` already?

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -243,6 +243,18 @@ sentry__crashpad_backend_free(sentry_backend_t *backend)
...
 }
 
+static void sentry__crashpad_backend_except(
+    sentry_backend_t *UNUSED(backend), sentry_ucontext_t *context)
+{
+#ifdef SENTRY_PLATFORM_WINDOWS
+    crashpad::CrashpadClient::DumpAndCrash(&context->exception_ptrs);
+#else
+    // TODO: Crashpad has the ability to do this on linux / mac but the
+    // method interface is not exposed for it, a patch would be required
```
Swatinem: that is unfortunate :-(

include/sentry.h
```diff
@@ -650,6 +651,11 @@ SENTRY_API sentry_user_consent_t sentry_user_consent_get(void);
  */
 SENTRY_API sentry_uuid_t sentry_capture_event(sentry_value_t event);
 
+/*
+ * Captures an exception to be handled by the backend.
```
Swatinem: Like I mentioned in chat, it would be good to better document the expectations for this function.

src/sentry_backend.h
cammm: Yes it does :) I'll remove this

PR #202, opened 2020-03-27
Fixes initializing random on windows

cammm:
The SENTRY_CTOR isn't working for me and afaik it is not safe to call LoadLibraryA outside of main() ie during static initialization or DLLMain. Switching this to initialize on the first call to random instead.

### DIFF
--- a/src/sentry_random.c
+++ b/src/sentry_random.c
@@ -54,19 +54,20 @@ getrandom_devurandom(void *dst, size_t bytes)
 #ifdef SENTRY_PLATFORM_WINDOWS
 typedef BOOLEAN(WINAPI *sRtlGenRandom)(PVOID Buffer, ULONG BufferLength);
 
-static sRtlGenRandom pRtlGenRandom;
-
-SENTRY_CTOR (init_winapi) {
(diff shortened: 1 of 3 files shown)


src/sentry_random.c
```diff
@@ -54,19 +54,20 @@ getrandom_devurandom(void *dst, size_t bytes)
 #ifdef SENTRY_PLATFORM_WINDOWS
 typedef BOOLEAN(WINAPI *sRtlGenRandom)(PVOID Buffer, ULONG BufferLength);
 
-static sRtlGenRandom pRtlGenRandom;
-
-SENTRY_CTOR (init_winapi) {
```
Swatinem: I think this was the only usage of `SENTRY_CTOR`, was it? In which case would be good to also remove it.

cammm: Yeah good idea, removed

jan-auer:
@Swatinem what do you think about creating a random number in `sentry_init` just to eager-initialize and avoid overheads when the first event is created?

Swatinem:
We basically do that already, when creating a random uuid for the run folder.

PR #203, opened 2020-03-28
Fixes for legacy Windows versions

Mixaill:
* `cmake`: fix `CMAKE_SYSTEM_VERSION` comparison
* `cmake`: add support for `5.2` version (2003 / 2003 R2 / XP 64-bit)
* `cmake`: disable `crashpad` build with XP toolchain
* `cmake`: link with `pathcch.lib` only for Windows 8.0+
---
* `src/path/windows`: include `<pathcch.h>` only for Windows 8.0+
* `src/transports/winhttp`: use `WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY` only for Windows 8.1+
* `src/transports/winhttp`: fallback to `WINHTTP_ACCESS_TYPE_DEFAULT_PROXY` instead of `WINHTTP_ACCESS_TYPE_NO_PROXY`
* `src/sync`: implement `InitOnceExecuteOnce` for Windows XP
* `src/unwinder/dbghelp`: do not use`CaptureStackBackTrace` on Windows XP 
---
* `readme`: add info about `CMAKE_SYSTEM_VERSION` into readme


CMakeLists.txt
```diff
@@ -151,19 +154,33 @@ endfunction()
...
-    elseif(${CMAKE_SYSTEM_VERSION} EQUAL 6.1)
-        add_definitions(-D_WIN32_WINNT=0x0601)
-    elseif(${CMAKE_SYSTEM_VERSION} EQUAL 6.0)
-        add_definitions(-D_WIN32_WINNT=0x0600)
-    else()
-        add_definitions(-D_WIN32_WINNT=0x0501)
-    endif()
+	if(${CMAKE_SYSTEM_VERSION} MATCHES "^10")
+		add_definitions(-D_WIN32_WINNT=0x0A00)
+		set(SYSTEM_LIBRARIES "pathcch")
```
Swatinem: don’t we need both `pathcch` and `shlwapi`?

Other than that, please move this whole block down below the `add_library` definition, so you can actually use the `target_X` set of functions to not set the definitions globally.

CMakeLists.txt
```diff
@@ -53,7 +53,10 @@ endif()
 
 option(SENTRY_ENABLE_INSTALL "Enable sentry installation" "${SENTRY_MAIN_PROJECT}")
 
-if(APPLE OR WIN32)
+if(MSVC AND CMAKE_GENERATOR_TOOLSET MATCHES "_xp$")
+	message(WARNING "Backend is set to 'none' for MSVC with XP toolset")
```
Swatinem: hm, I was thinking if we should remove this warning. But since crashpad in the only supported backend on windows right now, it is good to be more vocal about this, to not raise false expectations for users.

src/sentry_sync.h
```diff
@@ -8,8 +8,57 @@
 
 // define a recursive mutex for all platforms
 #ifdef SENTRY_PLATFORM_WINDOWS
-#    include <synchapi.h>
+#    if _WIN32_WINNT >= 0x0600
+#        include <synchapi.h>
+#    else
+#        include <windows.h>
```
Swatinem: do we need this? I think `sentry_boot` already pulls in a minimal version of `windows.h`.

src/unwinder/sentry_unwinder_dbghelp.c
```diff
@@ -7,7 +7,11 @@ sentry__unwind_stack_dbghelp(
     void *addr, const sentry_ucontext_t *uctx, void **ptrs, size_t max_frames)
 {
     if (!uctx && !addr) {
+#if _WIN32_WINNT && _WIN32_WINNT < 0x0600
+        return 0;
```
Swatinem: The rest of the unwinder still works though? Would be good to document these limitations in the README as well.

CMakeLists.txt
Mixaill: moved and changed to `target_`

CMakeLists.txt
Mixaill: changed wording a bit to be a more concrete.

src/sentry_sync.h
Mixaill: deleted

src/unwinder/sentry_unwinder_dbghelp.c
Mixaill: Looks like `RtlCaptureStackBackTrace` is available on XP's version of `ntdll.dll`.
Replaced it with call via `GetProcAddress`.

Testing pending.

Mixaill:
@Swatinem now it pass all the unit tests on Windows XP SP3

Issue #204, opened 2020-03-29
Error building a static version in macOS 


david-antiteum:
Hi,

While doing a port for Vcpkg I have found a build problem in macOS (10.15.4 Xcode 11.4) that can be reproduce with this commands:

```
git clone --recurse-submodules https://github.com/getsentry/sentry-native.git
cd sentry-native
cmake -B build -DSENTRY_BACKEND=crashpad -DSENTRY_TRANSPORT=curl -DSENTRY_BUILD_TESTS=OFF -DSENTRY_BUILD_EXAMPLES=OFF -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF
```

CMake error is:

```
CMake Error at CMakeLists.txt:136 (install):
  install TARGETS given target "getopt" which does not exist.
Call Stack (most recent call first):
  CMakeLists.txt:264 (sentry_install)
```

Shared build compiles with no issues.

Thanks,

David

Swatinem:
Ah… So these are only defined/built on windows/msvc:

https://github.com/getsentry/crashpad/blob/4ec46c337905a10f1690ef57594daa11d6e08644/CMakeLists.txt#L91-L94

So need to make it conditional here as well:

https://github.com/getsentry/sentry-native/blob/58b541ff2cd47ad7545282753e4c9b0de25eacd2/CMakeLists.txt#L263-L267

I will take care of this tomorrow…

Swatinem:
CMake!!! 😭 

Seems like this is not as easy as I thought. I now run into the following error:

```
CMake Error in external/crashpad/compat/CMakeLists.txt:
  Target "crashpad_compat" INTERFACE_SOURCES property contains path:

    "/Users/swatinem/Coding/sentry-native/external/crashpad/compat/non_elf/elf.h"

  which is prefixed in the source directory.
```

The crashpad/compat sources are prefix with the absolute directory, because if they are not, I get other errors. (oh, and I also saw that `list transform` is also not supported on our low cmake 3.10 target anyway -_-)
I fiddled around with this a bit, but I’m completely lost, and don’t really want to waste too much time with this.

@madebr @Amphaal You two seem to have a bit more cmake experience than me, maybe one of you can help me out here?

david-antiteum:
I know, my plan was to fix it in a PR :) but I give up.

Ah, the vcpkg port is here:
https://github.com/david-antiteum/vcpkg-ports/tree/master/ports/sentry

[...]

PR #205, opened 2020-03-30
cmake: Remove Windows-only install targets from Mac

jblazquez:
This change fixes one part of #204. With this change, you can now build with `-DBUILD_SHARED_LIBS=NO` on macOS, as long as you _also_ disable the install target using `-DSENTRY_ENABLE_INSTALL=NO`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -277,10 +277,16 @@ if(SENTRY_BACKEND_CRASHPAD)
 	target_compile_features(sentry PUBLIC cxx_std_14)
 	target_link_libraries(sentry PRIVATE crashpad_client crashpad_util)
 	if(NOT BUILD_SHARED_LIBS)
-		sentry_install(TARGETS crashpad_client crashpad_util crashpad_compat getopt mini_chromium zlib EXPORT sentry
+		sentry_install(TARGETS crashpad_client crashpad_util crashpad_compat mini_chromium EXPORT sentry
 			LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
 			ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
 		)
+		if(WIN32)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -277,10 +277,16 @@ if(SENTRY_BACKEND_CRASHPAD)
 	target_compile_features(sentry PUBLIC cxx_std_14)
 	target_link_libraries(sentry PRIVATE crashpad_client crashpad_util)
 	if(NOT BUILD_SHARED_LIBS)
-		sentry_install(TARGETS crashpad_client crashpad_util crashpad_compat getopt mini_chromium zlib EXPORT sentry
+		sentry_install(TARGETS crashpad_client crashpad_util crashpad_compat mini_chromium EXPORT sentry
 			LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
 			ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
 		)
+		if(WIN32)
```
madebr: https://github.com/getsentry/crashpad/blob/4ec46c337905a10f1690ef57594daa11d6e08644/CMakeLists.txt#L91-L94

I think it should be:
```suggestion
		if(MSVC)
```

jblazquez: Good catch. I was using the version of crashpad's `CMakeLists.txt` from the `0.2.2` release and that used to say `IF(WIN32)` there. I will update to say `IF(MSVC)` here for consistency with `master`.

jblazquez:
Looks like two CI builders failed with timeouts. How does one retry Azure Pipelines builds?

Swatinem:
> How does one retry Azure Pipelines builds?

by having admin priviledges I guess :-D

Also, I don’t think this PR is necessary anymore, @madebr made very good progress restructuring the crashpad CMake files. I will review all this today.

Swatinem:
Closing this in favor of #207, but thanks a lot for looking into this!

PR #206, opened 2020-03-30
build: add -fPIC flag when building static lib on Linux

aykamko:
Without this, we get the following linker error:
```
make: Entering directory '.../build'
  SOLINK_MODULE(target) Debug/obj.target/addon.node
/usr/bin/ld: .../lib/linux/libsentry.a(sentry_core.c.o): relocation R_X86_64_PC32 against symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
addon.target.mk:211: recipe for target 'Debug/obj.target/addon.node' failed
make: *** [Debug/obj.target/addon.node] Error 1
make: Leaving directory '.../build'
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,6 +182,8 @@ if(BUILD_SHARED_LIBS)
 		sentry_install(FILES "$<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:$<TARGET_PDB_FILE:sentry>>"
 			DESTINATION "${CMAKE_INSTALL_BINDIR}")
 	endif()
+elseif(LINUX)
+	target_compile_options(sentry PRIVATE -fPIC)
 endif()
 
 if(BUILD_SHARED_LIBS)


CMakeLists.txt
```diff
@@ -182,6 +182,10 @@ if(BUILD_SHARED_LIBS)
 		sentry_install(FILES "$<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:$<TARGET_PDB_FILE:sentry>>"
 			DESTINATION "${CMAKE_INSTALL_BINDIR}")
 	endif()
+else()
+	if(LINUX)
```
Swatinem: Please merge this into a `elseif`

CMakeLists.txt
```diff
@@ -182,6 +182,10 @@ if(BUILD_SHARED_LIBS)
...
+		set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
```
Swatinem: please use `target_compile_options` here, instead of messing with global definitions

aykamko:
comments addressed! thanks for the quick review

madebr:
This should be handled by the [POSITION_INDEPENDENT_CODE](https://cmake.org/cmake/help/v3.0/prop_tgt/POSITION_INDEPENDENT_CODE.html) cmake property.

Imho, sentry should not care about PIC and let users define `CMAKE_POSITION_INDEPENT_CODE` if a static library is desired that support PIC (and can be embedded in a shared library).

Swatinem:
> I'll open a pr once my other cmake pr is merged.

Thank you very much for looking into this ;-)

PR #209, opened 2020-03-31
fix(init): Correctly lock run directories

### LINKED ISSUE #220: On Windows, sentry__process_old_runs can remove the wrong folders
The call to `SHFileOperationW` in `sentry__path_remove_all` when cleaning up can go **disastrously** wrong. Per [documentation](https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shfileoperationa), it must never be called with relative paths because it is not thread safe. Second, I have seen old [comments online](https://szsu.wordpress.com/2010/12/20/forward-slashes-in-the-paths-passed-to-shfileoperation/) about it not handling forward slashes.

How disastrous is disastrous? I was running my test app from a folder `c:\projects\keyman\app\windows\bin\engine` and it deleted everything in `c:\projects` up until it failed with a file in use: `sentry.dll` itself. That was many thousands of folders and hundreds of thousands of files -- I was wondering why it was so slow to start. I am not sure how much else I've lost from that Projects folder; most everything is on GitHub but I don't remember the 60-70 projects there (there are 5 left). Of course, when this happened I had no idea what had caused it. To trace the cause I had to experience this several times.

🤕

I had not specified `database_path` in my test app, relying on the default of `"./.sentry-native"`.

Thoughts about fixing this properly:

1. Never use `SHFileOperationW` (or any shell file operation style functions). It's not thread safe and introduces GUI dependencies, unwanted for console apps or services:
    * [oldnewthing ref 1](https://devblogs.microsoft.com/oldnewthing/20131206-00/?p=2443), Raymond Chen says "inadvisable".
    * [oldnewthing ref 2](https://devblogs.microsoft.com/oldnewthing/20141121-00/?p=43563), he extends that to "flat-out wrong (with caveat)".
    * [oldnewthing ref 3](https://devblogs.microsoft.com/oldnewthing/20150522-00/?p=45054), he says still wrong but if you are going to ignore his advice, at least use a sterilized needle.
2. The practice of using a relative path, anywhere, is not robust. Relative paths must be converted to absolute paths on first use to avoid further disaster when the app changes directory during run.
3. On Windows, you should really be using a local appdata folder (e.g. `%LOCALAPPDATA%` or, if you don't mind depending on shell32.dll, `SHGetKnownFolderPath(FOLDERID_LocalAppData)`) as the default for any writable data. In normal use, most apps installed to Program Files will end up with a non-writable path as their working directory. Don't forget to [expand the environment string](https://docs.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-expandenvironmentstringsa).
4. The default database path is  `"./.sentry-native"`.  `sentry_path_t` should be converting all forward slashes to backslashes on Windows as forward slash is inconsistently supported.

### LINKED ISSUE #223: sentry_options_set_database_path wipes existing directory contents
(This might be a documentation bug, reporting it here but if you want to WONTFIX and document it, I understand)

I got bitten today by the fact that sentry_options_set_database_path wants to take control of the directory you pass. I was expecting it make its own sub-dir, but leave other contents untouched. So I passed ~/Library/Application Support/myApp/ as the database dir, this directory already contains some other user data, which then got erased by sentry-native.

The fix for me was just to append 'sentry-db' to the path, but it caused some confusion wondering why my settings file (in the same dir) kept being wiped each launch.

Swatinem:
This allows multi-process usage, and also fixes file deletion on windows.

fixes #220 
fixes #223


src/path/sentry_path_unix.c
```diff
@@ -25,6 +27,84 @@ struct sentry_pathiter_s {
...
+    size_t offset = 0;
+    size_t remaining = buf_len;
+
+    while (true) {
+        if (remaining == 0) {
+            break;
+        }
+        ssize_t n = write(fd, buf + offset, remaining);
+        if (n <= 0) {
+            if (errno == EAGAIN || errno == EINTR) {
```
jan-auer, line 42: `errno` is only defined for `n == -1`:

```c
        ssize_t n = read(fd, buf, 4096);
        if (n < 0 && (errno == EAGAIN || errno == EINTR)) {
            continue;
        } else if (n <= 0) {
            break;
        }
```

src/path/sentry_path_unix.c
```diff
@@ -25,6 +27,84 @@ struct sentry_pathiter_s {
...
+            // check if the pid inside the lockfile still exists
+            char *contents = sentry__path_read_to_buffer(lock->path, NULL);
+            if (contents) {
+                int filepid = strtol(contents, NULL, 10);
+                bool process_exists
+                    = !(kill(filepid, 0) == -1 && errno == ESRCH);
+
+                if (!process_exists) {
+                    sentry__path_remove(lock->path);
+                    return sentry__filelock_try_lock(lock);
```
jan-auer, line 71: Can you loop around `fd < 0` instead of recursing here?

src/path/sentry_path_unix.c
```diff
@@ -25,6 +27,84 @@ struct sentry_pathiter_s {
...
+    if (fd < 0) {
+        if (errno == EEXIST) {
+            // check if the pid inside the lockfile still exists
+            char *contents = sentry__path_read_to_buffer(lock->path, NULL);
+            if (contents) {
+                int filepid = strtol(contents, NULL, 10);
+                bool process_exists
+                    = !(kill(filepid, 0) == -1 && errno == ESRCH);
+
+                if (!process_exists) {
```
jan-auer, line 69: Can you remove the double-negation? It makes it a bit harder than necessary to understand what's happening here.

src/path/sentry_path_unix.c
```diff
@@ -25,6 +27,84 @@ struct sentry_pathiter_s {
...
+        S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH);
+
+    if (fd < 0) {
+        if (errno == EEXIST) {
+            // check if the pid inside the lockfile still exists
+            char *contents = sentry__path_read_to_buffer(lock->path, NULL);
+            if (contents) {
+                int filepid = strtol(contents, NULL, 10);
+                bool process_exists
+                    = !(kill(filepid, 0) == -1 && errno == ESRCH);
```
jan-auer, line 67: `strtol` returns `0` if the conversion fails. This would mean that the signal is sent to *every* process. Please check for `filepid == 0` explicitly and assume that someone has the lock. This can happen if the file was just created but not written to by another process.

This also shows a race condition. I think you need to require the lock *before* reading from it.

src/path/sentry_path_unix.c
```diff
@@ -25,6 +27,84 @@ struct sentry_pathiter_s {
...
+                    sentry__path_remove(lock->path);
+                    return sentry__filelock_try_lock(lock);
+                }
+            }
+        }
+        lock->is_locked = false;
+        return false;
+    }
+
+    flock(lock->fd, LOCK_EX | LOCK_NB);
```
jan-auer, line 79: This can fail. We should check for the result and bail out.

examples/example.c
```diff
@@ -3,6 +3,12 @@
 #include <stdlib.h>
 #include <string.h>
 
+#ifdef SENTRY_PLATFORM_WINDOWS
+#    include <Windows.h>
```
madebr, line 7: Use lower case includes for mingw.

```suggestion
#    include <windows.h>
```

examples/example.c
```diff
@@ -106,6 +112,14 @@ main(int argc, char **argv)
         }
     }
 
+    if (has_arg(argc, argv, "sleep")) {
+#ifdef SENTRY_PLATFORM_WINDOWS
+        Sleep(10 * 1000);
+#else
+        sleep(10);
+#endif
```
madebr, line 120: Maybe move these platform dependent defines to a separate file/header that contains e.g.:
```
        #define sentry_sleep_s(SECONDS) Sleep((SECONDS) * 1000)
#else
        #define sentry_sleep_s(SECONDS) sleep(SECONDS)
#endif
```

Swatinem: hm, I don’t really want to expose/maintain these things as a *public API*.

Swatinem: but I moved it to the top of the file, so the platform check is not duplicated ;-)

madebr: I didn't see this was code from the example.
Good solution!

cammm:
Same with windows, you don't need to write the PID, just make a call to the _locking function

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/locking?view=vs-2019

The example Jan linked is great and has a windows version too

https://github.com/benediktschmitt/py-filelock/blob/b30bdc4fb998f5f4350257235eb51147f9e81862/filelock.py#L338

src/path/sentry_path_unix.c
```diff
@@ -25,6 +27,76 @@ struct sentry_pathiter_s {
...
+
+    return remaining;
+}
+
+bool
+sentry__filelock_try_lock(sentry_filelock_t *lock)
+{
+    lock->is_locked = false;
+
+    int fd = open(lock->path->path, O_RDWR | O_CREAT | O_TRUNC,
```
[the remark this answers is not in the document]
Swatinem: I think its both simpler and easier to understand without `O_EXCL`. The problem is with leftover files that were not correctly cleaned up. Like in the case of a crash.

src/backends/sentry_backend_inproc.c
```diff
@@ -147,7 +147,7 @@ make_signal_event(
 }
 
 static void
-handle_ucontext(sentry_ucontext_t* uctx)
+handle_ucontext(sentry_ucontext_t *uctx)
```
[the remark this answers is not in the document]
Swatinem: I guess it can. Lets do this in a different PR though.

src/path/sentry_path_unix.c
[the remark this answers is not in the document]
jan-auer: Good point, in case of a crash, we might consider to actually clean those up, wdty?

src/path/sentry_path_windows.c
```diff
@@ -30,20 +30,18 @@ struct sentry_pathiter_s {
...
+        } else if (n < buf_len) {
             break;
         }
-        offset += n;
-        remaining -= n;
+        buf += n;
+        buf_len -= n;
     }
-    return remaining;
+    fflush(f);
```
jan-auer, line 43: Is calling `fflush` after an error is still valid or do we need to do that inside of the loop?

src/sentry_database.c
```diff
@@ -121,7 +149,40 @@ sentry__process_old_runs(const sentry_options_t *options)
...
+        sentry__stringbuilder_append(&sb, filename);
+#endif
+        // `<db>/<uuid>.run.lock`
+        sentry__stringbuilder_append(&sb, ".lock");
+        char *lockfile_s = sentry__stringbuilder_into_string(&sb);
+        if (!lockfile_s) {
+            continue;
+        }
+        sentry_path_t *lockfile
+            = sentry__path_join_str(options->database_path, lockfile_s);
```
jan-auer, line 176: _nit_: This entire part where you build the path might be nice to extract into a function for readability. The code itself looks quite good to me, but it's a bit overwhelming in the context of this function with all the `continue`.

src/path/sentry_path_windows.c
Swatinem: Docs don’t mention anything about it. I also think the flush should not be in a loop, since its potentially slow.

src/sentry_database.c
Swatinem: BTH, all this string manipulation, also with the fact that we have normal, and wstr variants makes this a great pain in the rear. Lets land it like this and rethink it another time…

PR #210, opened 2020-03-31
fix(modulefinder): Reset errno before read() to prevent infinite loop

aykamko:
The `while` loop within `load_modules()` would forever read 0 bytes -- `errno` was set to `EINTR` at some point but was never cleared.

### DIFF
--- a/src/modulefinder/sentry_modulefinder_linux.c
+++ b/src/modulefinder/sentry_modulefinder_linux.c
@@ -403,6 +403,7 @@ load_modules(sentry_value_t modules)
     sentry_stringbuilder_t sb;
     sentry__stringbuilder_init(&sb);
     while (true) {
+        errno = 0;
         ssize_t n = read(fd, buf, 4096);
         if (n <= 0) {
             if (errno == EAGAIN || errno == EINTR) {


aykamko:
grepping for `errno = 0` shows 28 hits within vendored libs, like crashpad and curl

so possibly yes, although i'm no a C expert :sweat_smile:

Swatinem:
I had a talk with @jan-auer just now. The root cause here is that the if condition is wrong. `0` is a perfectly valid return value, and `errno` is only set if the return value is `< 0`.

I searched the code, and we have a couple of places checking for errno, and doing reads. Would you be willing to update your PR with all of the other cases, once #209 is merged to avoid merge conflict, because that is also moving some related code around. Or I can take this over tomorrow.

aykamko:
Sure! I'll wait for #209

Swatinem:
@aykamko Thanks for the patience! Finally the great #209 has landed. So please go ahead and update your patch. Please do look at the unix path implementation, that has calls to read. I can’t remember any other from the top of my head.

aykamko:
I'll take a look, thanks!

PR #216, opened 2020-04-02
fix: Expose a method to clear the module cache

### LINKED ISSUE #213: sentry_modulefinder_windows does not cater for changes in dynamically loaded modules
If a dynamically loaded module is not loaded until after the first event (with `SENTRY_SCOPE_MODULES`) is captured, it will never be seen in the module list, and therefore will also lose symbolication.

Loading the module cache is expensive so understand we don't want to do that on every event. Would it be possible to surface methods to invalidate the cache in the API, and/or always invalidate the cache for exceptions (where current module data is pretty essential and the time penalty is worthwhile)?

Version: 0.2.2+ (eb73904a8700fcf67e632a8c7d71174d214a344c)

### LINKED ISSUE #214: `sentry_options_t` parameter of function `sentry_options_get_http_proxy` is not declared const
In all of the functions used for retrieving options from sentry client, a `const sentry_options_t*` variable is passed in as the only parameter, _except_ for [`sentry_options_get_http_proxy`](https://github.com/getsentry/sentry-native/blob/eb73904a8700fcf67e632a8c7d71174d214a344c/include/sentry.h#L541), where the variable is not declared `const`.

Since the function [`sentry_get_options`](https://github.com/getsentry/sentry-native/blob/eb73904a8700fcf67e632a8c7d71174d214a344c/include/sentry.h#L640) returns a const reference, the result of this issue is that you can only fetch the `http_proxy` value from a `sentry_options_t` reference that has been created with [`sentry_options_new`](https://github.com/getsentry/sentry-native/blob/eb73904a8700fcf67e632a8c7d71174d214a344c/include/sentry.h#L456) (non-const), rather than the client's _active_ options.

I have checked out the master branch and changed the reference to const, compiled (successfully) and ran the tests where 51 passed and 3 skipped (none failed). Running a basic usage test **with this change** yields the desired output.

**Test:**
```c
sentry_options_set_http_proxy(options, "https://example.com");
sentry_init(options);

const sentry_options_t* options_active = sentry_get_options();
const char* http_proxy = sentry_options_get_http_proxy(options_active);
printf("sentry active http_proxy: %s\n", http_proxy);
```

**Output:**
```
sentry active http_proxy: https://example.com
```

I am happy to submit a pull request if desired. (Sorry if that's what I should have done initially!)

Swatinem:
fixes #213

Also make `sentry_options_get_http_proxy` const-correct, fixes #214

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -538,7 +538,8 @@ SENTRY_API void sentry_options_set_http_proxy(
 /*
  * returns the configured http proxy
  */
-SENTRY_API const char *sentry_options_get_http_proxy(sentry_options_t *opts);
+SENTRY_API const char *sentry_options_get_http_proxy(
+    const sentry_options_t *opts);
 
 /*
  * configures the path to a file containing ssl certificates for
@@ -633,6 +634,15 @@ SENTRY_API int sentry_init(sentry_options_t *options);
  * Shuts down the sentry client and forces transports to flush out.
  */
 SENTRY_API void sentry_shutdown(void);
+
+/**
+ * For performance reasons, sentry will cache the list of loaded libraries when
+ * capturing events. This cache can get out-of-date when loading or unloading
+ * libraries at runtime. It is therefore recommended to call
+ * `sentry_clear_modulecache` when doing so, to make sure that the next call to
+ * `sentry_capture_event` will have an up-to-date module list.
+ */
+SENTRY_EXPERIMENTAL_API void sentry_clear_modulecache(void);
 
 /*
  * Returns the client options.
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -69,7 +69,6 @@ stages:
             displayName: Installing Android SDK Dependencies
           - script: |
               bash scripts/start-android.sh
-            timeoutInMinutes: 5 # because gettings things wrong here can hang the job for up to an hour…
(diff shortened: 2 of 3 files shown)


Swatinem:
> going by the comment being removed it seems the default timeout is 1h. so i guess that still stops it from being completely stuck

yes. Originally, I added the timeout and the comment because I messed up the start script, which did basically nothing for an hour before being timed out. Now that the script at least does what it should, its rather a matter of how quick the CI machine running the job actually is. We have simulator starts in a minute, or 6 minutes, its *very* random. :-(

include/sentry.h
```diff
@@ -538,7 +538,8 @@ SENTRY_API void sentry_options_set_http_proxy(
 /*
  * returns the configured http proxy
  */
-SENTRY_API const char *sentry_options_get_http_proxy(sentry_options_t *opts);
+SENTRY_API const char *sentry_options_get_http_proxy(
+    const sentry_options_t *opts);
```
jan-auer, line 542: Can you make such bug fixes in separate PRs? By looking at the git history, it gets easier to see changes and when writing the changelog we don't miss things (especially when someone else writes it).

Swatinem: Sure. I was just mashing together a few oneliners in this PR to avoid the overhead ;-) But yes, that overhead actually serves a very good purpose.

PR #217, opened 2020-04-02
ci: Set up a linting job, and add a precommit hook

Swatinem:


scripts/run-clang-format.py
```diff
@@ -0,0 +1,382 @@
+#!/usr/bin/env python
+
+# This is taken from: https://github.com/Sarcasm/run-clang-format
```
flub, line 3: License for both is MIT, but the copyright info would still be different.  I'm not sure how this should normally be handled.

Makefile
```diff
@@ -60,15 +63,19 @@ setup-venv: .venv/bin/python
...
 		src/*/*.c \
+		src/*/*.cpp \
 		src/*/*.h \
 		tests/unit/*.c \
 		tests/unit/*.h
 	@.venv/bin/black tests
 .PHONY: format
 
 style: setup-venv
+	@.venv/bin/python ./scripts/run-clang-format.py -r examples include src tests/unit
```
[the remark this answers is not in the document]
jan-auer: We generally do that so that people can just run `make check` and it sets up all dependencies.

scripts/run-clang-format.py
Swatinem: Hm, I wasn’t really paying that much attention to licensing so far TBH.
I was under the impression that as long as we don’t re-distribute or link to these things, it doesn’t really matter. It is a dev-only dependency.

Makefile
jan-auer, line 79: _nit_: Can we call this `check-clang-format`?  When reading this for the first time, I got the impression that this would actually format the files :)

scripts/run-clang-format.py
jan-auer: Maybe add the necessary copyright notice, then we're fine.

src/sentry_sync.h
```diff
@@ -38,7 +41,9 @@ InitOnceExecuteOnce(
     for (;;) {
         switch ((ULONG_PTR)InitOnce->Ptr) {
         case 0: // not started
-            if (InterlockedCompareExchangePointer(&InitOnce->Ptr, (PVOID)1, (PVOID)0) != 0) {
+            if (InterlockedCompareExchangePointer(
+                    &InitOnce->Ptr, (PVOID)1, (PVOID)0)
+                != 0) {
```
jan-auer, line 46: _nit_: Please reformat this for readability.

src/unwinder/sentry_unwinder_dbghelp.c
```diff
@@ -19,11 +19,13 @@ sentry__unwind_stack_dbghelp(
         if (!(ntdll = LoadLibraryW(L"ntdll.dll"))) {
             return 0;
         }
-        if (!(proc = (RtlCaptureStackBackTraceProc)GetProcAddress(ntdll, "RtlCaptureStackBackTrace"))) {
+        if (!(proc = (RtlCaptureStackBackTraceProc)GetProcAddress(
```
jan-auer, line 22: _nit_: Please reformat this for readability.

Swatinem:
The things you mentioned are actually auto-formatted that way. The auto-formatting sure is weird in some cases.

PR #218, opened 2020-04-02
cmake: add SENTRY_PIC option

madebr:
Alternative to #206 

@aykamko


Swatinem:
Please also document the newly added options in the README.
I can do that as well when I clean that stuff up a bit before tagging the next release.

PR #225, opened 2020-04-07
fix: Use absolute paths for the database

Swatinem:
This way, sentry is resilient to cwd changes, as mentioned in #220.

### DIFF
--- a/src/sentry_core.c
+++ b/src/sentry_core.c
@@ -61,6 +61,17 @@ sentry_init(sentry_options_t *options)
     sentry_shutdown();
     sentry__mutex_lock(&g_options_mutex);
     g_options = options;
+
+    sentry_path_t *database_path = options->database_path;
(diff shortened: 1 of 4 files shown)


src/sentry_core.c
```diff
@@ -61,7 +61,18 @@ sentry_init(sentry_options_t *options)
...
     sentry__mutex_lock(&g_options_mutex);
     g_options = options;
+
     sentry__path_create_dir_all(options->database_path);
+    sentry_path_t *database_path = options->database_path;
+    options->database_path = sentry__path_absolute(database_path);
+    if (options->database_path) {
+        sentry__path_free(database_path);
+    } else {
+        options->database_path = database_path;
```
jan-auer, line 71: _nit_: I'd add an error log here, for now, since this case indicates a major error. In the future, we should leave the database path `NULL`, and handle that explicitly in the transports, user consent, and run handling.

PR #229, opened 2020-04-07
fix: Clean up read loops and errno handling

Swatinem:
Supercedes and closes #210


aykamko:
Thanks so much for this, sorry I never got around to it!

Issue #221, opened 2020-04-06
CMake imported target leaks C++14 requirement


zakalawe:
The generated sentry-targets.cmake specifies 'INTERFACE_COMPILE_FEATURES "cxx_std_14"', which is causing me some difficulties. It's no problem for me to build the sentry libs using a C++14 compiler, but I don't want to switch my entire application to use it (currently on C++11)

Is it intentional that C++14 requirement (from Crashpad) is exposed as an interface feature? Given that sentry.h exposes a C API, this seems strange.

zakalawe:
As an experiment, I tried removing the INTERFACE_COMPILE_FEATURES from the cmake file, and at least on macOS (Xcode 11.4, Apple clang version 11.0.3 (clang-1103.0.32.29)) everything builds and links fine. I'm about to start testing the Windows integration

Swatinem:
I think this is already fixed in master, can you try with that?

CC @madebr

zakalawe:
Thanks, I was using the latest stable release, will try master and report back.

madebr:
I think it's still present.
I removed it in sentry-native in https://github.com/getsentry/sentry-native/commit/7b18bef0b7cdc258c4baf17a4903fa0a44b905c3#diff-af3b638bc2a3e6c650974192a53c7291
And added it to crashpad: https://github.com/getsentry/crashpad/blob/23474e346aa7c24a27506ee038b22bae74a5ce75/client/CMakeLists.txt#L60

It can be fixed by surrounding cxx_std_14 with the BUILD_INTERFACE generator expression

zakalawe:
@madebr I'm seeing a different issue on master, which is blocking me reporting on this. What I have in my CmakeList for my target is:

  find_package(sentry)
  target_link_libraries(myFooTarget sentry::sentry)

This was working with the stable release (0.2.2). With master, CMake fails to process, with the following message:

> CMake Error at src/Main/CMakeLists.txt:83 (add_executable):
  Target "myFooTarget" links to target "sentry_crashpad::client" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?

Of course there is no such thing as 'sentry_crashpad::client' defined in any of the exported Cmake files.

zakalawe:
Ah figure out the issue, sentry_targets.cmake contains the following:

> # Create imported target sentry::sentry
add_library(sentry::sentry SHARED IMPORTED)

[...]

PR #231, opened 2020-04-09
daily cmake updates

madebr:
- No need to prepend compile definitions with `/D` or `-D` in cmake
- Link privately to crashpad client (#221)

- tests don't need c++14
- compile options of sentry can private + only include directory should be public


CMakeLists.txt
```diff
@@ -284,7 +284,7 @@ if(SENTRY_BACKEND_CRASHPAD)
 			set(CRASHPAD_ENABLE_INSTALL ON CACHE BOOL "Enable crashpad installation" FORCE)
 		endif()
 		add_subdirectory(external/crashpad crashpad_build)
-		target_link_libraries(sentry PUBLIC
+		target_link_libraries(sentry PRIVATE
```
Swatinem, line 287: When building a shared lib, this makes total sense.
When building a static lib, how can we make sure that users will transitively get crashpad as well? Or does that happen already?

madebr: CMake does this automatically.
I checked the generated targets and it contains a `$<LINK_ONLY:...>` in the generated targets file (in the `INTERFACE_LINK_LIBRARIES` property).
This is similar to the following situation:
```
add_library(liba STATIC ...)
add_library(libb STATIC ...)
target_link_libraries(libb PRIVATE liba)
add_executable(main ...)
target_link_libraries(main PRIVATE libb)
```
The `main` executable must link to `liba ` and `libb`.

madebr: cmake knows static libraries should be made transparent.

madebr: @Swatinem 
I just noticed I introduced a bug in crashpad yesterday.
I used `CMAKE_C_FLAGS` twice as source for the regex.
I'll open a pr...

madebr: Ok, I"ll bump crashpad after my fix to crashpad is merged.

madebr:
Question: what c standard are you targeting c99 or c11?

CMakeLists.txt
madebr: See getsentry/crashpad#9

Swatinem:
> Question: what c standard are you targeting: c99 or c11?

If it were up to me personally, I would go with the most recent one ;-)
I guess since we have c++14, we should be safe to require c11. I think @jan-auer known more about this than I do though.

madebr:
I've updated the crashpad git submodule.

I ask about the c standard so all compilers now and in the future will know what standard to use.
This standard will also be set PRIVATE and not visible to consumers.

Swatinem:
> I've updated the crashpad git submodule.

would you mind bumping again, to also pull in the changes from @Amphaal in one go?

Amphaal:
Btw many thanks to @madebr who basically did what I wanted to do a while ago :') I can now safely delete my forks 👍

Swatinem:
Thanks to both of you! I am very grateful for these contributions.

PR #236, opened 2020-04-17
fix: Avoid reading procmaps memory offsets directly

Swatinem:
This is the only place I can think of that could potentially segfault. The only thing that is still accessed directly though is the vdso.

### DIFF
--- a/src/modulefinder/sentry_modulefinder_linux.c
+++ b/src/modulefinder/sentry_modulefinder_linux.c
@@ -333,7 +333,8 @@ sentry__procmaps_module_to_value(const sentry_module_t *module)
         sentry_mmap_t mm;
         if (!sentry__mmap_file(&mm, filename)) {
             sentry_free(filename);
-            return mod_val;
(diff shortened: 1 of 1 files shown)


src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -348,11 +349,14 @@ sentry__procmaps_module_to_value(const sentry_module_t *module)
 static void
 try_append_module(sentry_value_t modules, const sentry_module_t *module)
 {
-    if (!module->file.ptr || !is_elf_module(module->start)) {
+    if (!module->file.ptr) {
```
jan-auer, line 352: Should you call `is_elf_module` in `sentry__procmaps_module_to_value` now?

Swatinem: yes, absolutely!

Swatinem:
just note: the logs I added are super chatty, as they print every single module, even those that are not ELF files.

PR #241, opened 2020-04-22
fix: Send sessions as envelopes to the correct endpoint

Swatinem:
Lesson learned: Don’t merge code without e2e tests ;-)
Also, I’m very much tempted to just rip out all the endpoint and multipart-related code and just send envelopes *always*. As it is now, regular `capture_event` works with non-relay servers still.

One question I have: A crashed session won’t increase the `errors` count. Is that expected?

### DIFF
--- a/src/sentry_envelope.c
+++ b/src/sentry_envelope.c
@@ -11,6 +11,7 @@
 #define MAX_HTTP_HEADERS 5
 
 typedef enum {
+    ENDPOINT_TYPE_ENVELOPE,
     ENDPOINT_TYPE_STORE,
     ENDPOINT_TYPE_MINIDUMP,
     ENDPOINT_TYPE_ATTACHMENT,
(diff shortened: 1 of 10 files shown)


src/sentry_envelope.c
```diff
@@ -450,7 +454,7 @@ sentry__envelope_for_each_request(const sentry_envelope_t *envelope,
     sentry_uuid_t event_id = sentry__envelope_get_event_id(envelope);
 
     if (envelope->is_raw) {
-        req = prepare_http_request(&event_id, ENDPOINT_TYPE_STORE,
+        req = prepare_http_request(&event_id, ENDPOINT_TYPE_ENVELOPE,
             "application/x-sentry-envelope", envelope->contents.raw.payload,
```
jan-auer, line 458: Would it be possible to introduce a constant for this?

Swatinem: Will do when I refactor all of this ;-)

PR #243, opened 2020-04-22
fix: Avoid global static in crashpad backend

Swatinem:


src/backends/sentry_backend_crashpad.cpp
```diff
@@ -32,27 +32,28 @@ extern "C" {
 extern "C" {
 
 typedef struct {
+    std::unique_ptr<crashpad::CrashReportDatabase> db;
```
jan-auer, line 35: This won't get cleaned up in `sentry__crashpad_backend_free` since the destructor is not invoked when running `free`. Without looking it up, I'm not sure if there's a straight-forward way to do this, so I would propose to use an explicit `new` and `delete` of a raw pointer. Note that this may cause problems when run in the signal handler.

Swatinem: the backend shutdown hook is called before free. I think that should be sufficient?

jan-auer: Ah, I missed that. Alright, two asks here, please:
 1. Could you use a raw pointer instead? Having a `unique_ptr` in a C struct may indicate wrong intentions at first sight.
 2. Move the `delete` call to `sentry__crashpad_backend_free` so that it runs along all other deallocations.

PR #245, opened 2020-04-27
fix: Correctly send session init flag

Swatinem:
This also fixes an issue with reading sessions serialized in the run-folder.


src/sentry_utils.c
```diff
@@ -376,3 +376,43 @@ sentry__msec_time_to_iso8601(uint64_t time)
...
+    tm.tm_min = m;
+    tm.tm_sec = s;
+    // a negative value means `mktime` should infer it
+    tm.tm_isdst = -1;
+    time_t time = mktime(&tm);
+    if (time == -1) {
+        return 0;
+    }
+    // add the UTC offset, since mktime assumes local time
+    time += tm.tm_gmtoff;
```
jan-auer, line 415: `tm_gmtoff` is a nonstandard extension and not available in MSVC.

jan-auer: Looks like on MSVC this should call [`_mkgmtime`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mkgmtime-mkgmtime32-mkgmtime64?view=vs-2019).

jan-auer: There's a similar check in the function above. I believe this should check for the compiler rather than the plaform.

jan-auer: You could probably do this (substituting with a check for the compiler instead):

```c
time_t timestamp = _mkgmtime(&tm);
#else
time_t timestamp = mktime(&tm) - _timezone;
#endif
```

Swatinem: > I believe this should check for the compiler rather than the plaform.

I think this rather depends on the platform libc, or does it?

src/sentry_utils.c
```diff
@@ -376,3 +376,47 @@ sentry__msec_time_to_iso8601(uint64_t time)
...
+    int len = strlen(iso);
+    // The code is adapted from: https://stackoverflow.com/a/26896792
+    int y, M, d, h, m, s, msec = 0;
+    int consumed = 0;
+    if (sscanf(iso, "%d-%d-%dT%d:%d:%d%n", &y, &M, &d, &h, &m, &s, &consumed)
+        < 6) {
+        return 0;
+    }
+    // we must have a tail, either `Z` or starting with `.`
+    if (consumed >= len || (iso[consumed] != 'Z' && iso[consumed] != '.')) {
```
jan-auer, line 392: _nit_: Can you move the check for `Z` below consuming the decimal part, and remove it from the `sscanf` below?

src/sentry_utils.c
```diff
@@ -376,3 +376,47 @@ sentry__msec_time_to_iso8601(uint64_t time)
...
+    tm.tm_mday = d;
+    tm.tm_hour = h;
+    tm.tm_min = m;
+    tm.tm_sec = s;
+    // a negative value means `mktime` should infer it
+    tm.tm_isdst = -1;
+#ifdef SENTRY_PLATFORM_WINDOWS
+    time_t time = _mkgmtime(&tm);
+#else
+    // add the UTC offset, since mktime assumes local time
```
jan-auer, line 413: Could you please add a bit more explanation here regarding the sign of `tm_gmtoff`.

jan-auer: Just noticing that there's also [`timegm`](http://man7.org/linux/man-pages/man3/timegm.3.html). Should have looked earlier, sorry about that.

PR #248, opened 2020-04-29
fix: Make sure to use an absolute crashpad handler path

Swatinem:


src/sentry_core.c
```diff
@@ -71,9 +71,20 @@ sentry_init(sentry_options_t *options)
         SENTRY_DEBUG("falling back to non-absolute database path");
         options->database_path = database_path;
     }
-    SENTRY_DEBUGF("Using database path \"%" SENTRY_PATH_PRI "\"",
+    SENTRY_DEBUGF("using database path \"%" SENTRY_PATH_PRI "\"",
         options->database_path->path);
 
+    sentry_path_t *handler_path = options->handler_path;
+    if (handler_path) {
```
jan-auer, line 78: Would it make sense to move this block into the crashpad backend and add a comment why we're doing this? That's also where we check if the file exists, and so we can error out properly.

src/sentry_core.c
```diff
@@ -71,9 +71,20 @@ sentry_init(sentry_options_t *options)
...
+        options->handler_path = sentry__path_absolute(handler_path);
+        if (options->handler_path) {
+            sentry__path_free(handler_path);
+        } else {
+            SENTRY_DEBUG("falling back to non-absolute handler path");
+            options->handler_path = handler_path;
```
jan-auer, line 84: This fallback will cause the `execvp` crash. I believe we should make sure that the crashpad backend logs an error and skips initialization in this case. This would be another argument for moving this block into the backend.

src/sentry_core.c
```diff
@@ -303,6 +314,8 @@ sentry_options_new(void)
     memset(opts, 0, sizeof(sentry_options_t));
     opts->database_path = sentry__path_from_str(".sentry-native");
     sentry_options_set_dsn(opts, getenv("SENTRY_DSN"));
+    const char *debug = getenv("SENTRY_DEBUG");
+    opts->debug = debug && strlen(debug) > 0;
```
jan-auer, line 318: Can we check for `"1"` explicitly? It's quite common to disable by setting `SENTRY_DEBUG=0`.

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -84,19 +84,27 @@ sentry__crashpad_backend_startup(sentry_backend_t *backend)
         }
     }
 
-    if (!handler_path || !sentry__path_is_file(handler_path)) {
+    // Starting the crashpad client needs either an absolute path, or an
+    // executable which can be resolved via $PATH, or otherwise the client will
+    // crash if it fails to execute the given handler executable.
```
jan-auer, line 89: Technically, shell lookup rules apply. It can also be a relative path, but that would have to be prefixed with `./` to work in all cases.

Swatinem: I updated the wording a bit, thanks

PR #249, opened 2020-04-29
ref: Always send envelopes

Swatinem:
This removes all the multipart handling code and is a *soft break*,
since the API stays the same, but the SDK now *really* requires a
Relay-enabled sentry server.


src/sentry_envelope.c
```diff
@@ -400,75 +318,41 @@ prepare_http_request(const sentry_uuid_t *event_id,
...
         const sentry_envelope_t *, void *data),
     const sentry_rate_limiter_t *rl, void *data)
 {
     sentry_prepared_http_request_t *req;
-    sentry_uuid_t event_id = sentry__envelope_get_event_id(envelope);
 
     if (envelope->is_raw) {
-        req = prepare_http_request(&event_id, ENDPOINT_TYPE_ENVELOPE,
-            "application/x-sentry-envelope", envelope->contents.raw.payload,
+        req = prepare_http_request(envelope->contents.raw.payload,
```
jan-auer, line 343: Would it make sense to change the function signature to just return a `sentry_prepared_http_request_t` now that there will only be one?

src/sentry_envelope.c
```diff
@@ -177,7 +96,7 @@ envelope_item_get_category(const sentry_envelope_item_t *item)
     } else if (sentry__string_eq(ty, "transaction")) {
         return SENTRY_RL_CATEGORY_TRANSACTION;
     }
-    // NOTE: the `type` here can be `event`, `minidump` or `attachment`.
+    // NOTE: the `type` here can be `event` or `attachment`.
     // Ideally, attachments should have their own RL_CATEGORY.
```
jan-auer, line 100: Note for later: This comment is true, but the logic is complicated and not fully specified yet. It's good if it stays like this for now.

src/sentry_envelope.c
```diff
@@ -477,98 +361,18 @@ sentry__envelope_for_each_request(const sentry_envelope_t *envelope,
...
-            break;
-        }
-        case ENVELOPE_ITEM_TYPE_UNKNOWN: {
-            other[other_count++] = item;
-            break;
-        }
-        default:
-            continue;
-        }
+        sentry__envelope_serialize_item_into_stringbuilder(item, &sb);
```
jan-auer, line 364: Can we split this into two steps with separate testable functions?
 1. Remove rate limited items from this envelope.
 2. Serialize the envelope in one go

Between these, the transport would simply check if the envelope is empty.

src/sentry_envelope.c
jan-auer, line 343: Would this function make more sense in a base transport file rather than `envelope.c`? It seems that most of what's happening here is transport-specific. See comments below regarding splitting this up.

Swatinem: Hm, I think we are pretty much guaranteed that we get ether 0 or 1 requests for a single envelope, right? In that case, it makes sense.

Swatinem: both envelope and envelope_item are private to that file right now. Not quite sure if we should expose it right now?

Swatinem: Either way, we can do this incrementally anyway. First making a breaking change to hide the transport struct, and then expose the http request interface at some point later.

src/sentry_envelope.c
Swatinem: Oh nevermind, we don’t need to expose envelope internals if we have rate limiting in the envelope module, and http generation in the transport…

src/sentry_envelope.c
jan-auer: I would say so. Theoretically, the SDK could also split up an envelope into multiple envelopes, but I wouldn't know what the use case would be. Rather combining multiple into one.

jan-auer: Okay, that sounds good to me. 

Eventually we might have to expose the envelope for the sake of custom transports. At that, it might be nice to have the two functions separate.

src/sentry_core.c
```diff
@@ -202,6 +202,8 @@ sentry__capture_envelope(sentry_envelope_t *envelope)
     if (opts->transport) {
         SENTRY_TRACE("sending envelope");
         opts->transport->send_envelope_func(opts->transport, envelope);
+    } else {
+        sentry_envelope_free(envelope);
```
jan-auer, line 206: Good catch! Should we add a trace log here, too? Might be useful to see that we're discarding an envelope.

src/sentry_envelope.c
```diff
@@ -614,6 +338,21 @@ sentry__envelope_serialize_into_stringbuilder(
     }
 }
 
+char *
+sentry_envelope_serialize_consume(sentry_envelope_t *envelope, size_t *size_out)
```
jan-auer, line 342: Is there a reason to have this function consume? It seems a bit restrictive. Worst case, the caller can still call free.

src/transports/sentry_transport_curl.c
```diff
@@ -110,16 +111,24 @@ header_callback(char *buffer, size_t size, size_t nitems, void *userdata)
...
+static void
+task_exec_func(void *data)
 {
     struct task_state *ts = data;
+
+    sentry_prepared_http_request_t *req
+        = sentry__prepare_http_request(ts->envelope, ts->transport_state->rl);
+    ts->envelope = NULL;
+    if (!req) {
+        return;
```
jan-auer, line 123: Consider logging an error here (same for WinHTTP)?

Swatinem: It’s not really an error, this means the whole envelope was rate limited…

src/sentry_envelope.c
Swatinem: since the previous code was using callbacks, it could reuse the buffer of a raw envelope. otherwise we would have to make a copy.

src/transports/sentry_transport_curl.c
jan-auer: You are right, I've missed that the rate limiting call inside. This means there are two significantly different error modes (rate limit vs http request preparation failed). In that case one of two suggestions:
1. Split up `prepare_http_request` from `rate_limit`, call them separately and log accordingly.
2. Log inside of `prepare_http_request`, but there are quite a few places where you exit.

Swatinem:
I had to change the implementation quite a bit. The problem here was a conflict with the dumping code. Envelopes need to be immutable for the lifetime of the task, otherwise things are super unsafe. So I removed all the mutation, and instead I have a *serialize with rate limiter* function which can also reuse the allocation of raw envelopes.

PR #256, opened 2020-05-08
ref: Make transport opaque

Swatinem:
`sentry_transport_t` was made opaque, to make it extensible in the future.
The `sentry_new_function_transport` API was replaced by `sentry_transport_new`, with one important change: The send function is now responsible to free the envelope.
The shutdown function now takes a timeout parameter.


jan-auer:
Can you summarize the changes and the reasoning in the PR description?

include/sentry.h
```diff
@@ -477,32 +477,60 @@ SENTRY_API int sentry_envelope_write_to_file(
...
-typedef struct sentry_transport_s {
-    void (*send_envelope_func)(
-        struct sentry_transport_s *, sentry_envelope_t *envelope);
-    void (*startup_func)(struct sentry_transport_s *);
-    void (*shutdown_func)(struct sentry_transport_s *);
-    void (*free_func)(struct sentry_transport_s *);
-    void *data;
-} sentry_transport_t;
+SENTRY_API sentry_transport_t *sentry_transport_new(
+    void (*send_func)(void *data, sentry_envelope_t *envelope), void *data);
```
jan-auer, line 510: Can you swap the parameter order in `send_func`? IIRC, most of our APIs pass `data` as secondary parameter, and often times it's even ignored since it's only optional.

include/sentry.h
```diff
@@ -477,32 +477,60 @@ SENTRY_API int sentry_envelope_write_to_file(
...
+ *
+ * This hook will receive a millisecond-resolution timeout; it should return
+ * `true` in case all the pending envelopes have been sent within the timeout,
+ * or `false` if the timeout was hit.
  */
-SENTRY_API sentry_transport_t *sentry_new_function_transport(
-    void (*func)(sentry_envelope_t *envelope, void *data), void *data);
+SENTRY_API void sentry_transport_set_shutdown_func(
+    sentry_transport_t *transport,
+    bool (*shutdown_func)(void *data, uint64_t timeout));
```
jan-auer, line 533: Same here, let's pass data last.

include/sentry.h
```diff
@@ -477,32 +477,60 @@ SENTRY_API int sentry_envelope_write_to_file(
...
+ *   This hook receives a millisecond-resolution `timeout` parameter and should
+ *   return `true` when the transport was flushed and shut down successfully.
+ *   In case of `false`, sentry will log an error, but continue with freeing the
+ *   transport.
+ * * `free_func`: Frees the provided `data`. This hook might be called even
+ *   though `shudown_func` returned `false` previously.
+ *
+ * The transport interface might be extended in the future with hooks to flush
+ * its internal queue without shutting down, and to dump its internal queue to
+ * disk in case of a crash.
```
jan-auer, line 499: Should we also mention `data` here and what it should be used for? Also, I'm wondering if we can give this a different name, like `state`, that indicates the intent a bit better.

**Edit**: Looks like internally we call this `sentry_transport_state` already, so naming it `state` sounds good.

src/sentry_core.c
```diff
@@ -114,7 +114,15 @@ sentry_shutdown(void)
     if (options) {
         if (options->transport && options->transport->shutdown_func) {
             SENTRY_TRACE("shutting down transport");
-            options->transport->shutdown_func(options->transport);
+            // TODO: make this configurable
```
jan-auer, line 117: How about we add this to the transport struct and expose a setter, just like the `send_envelope_func`, etc? Looks like a small enough change to do it in this PR. Note that there should also be a way to disable the timeout completely.

include/sentry.h
```diff
@@ -477,32 +477,60 @@ SENTRY_API int sentry_envelope_write_to_file(
...
+struct sentry_transport_s;
+typedef struct sentry_transport_s sentry_transport_t;
 
 /**
- * This represents an interface for user-defined transports.
+ * Creates a new transport.
  *
- * This type is *deprecated*, and will be replaced by an opaque pointer type and
- * builder methods in a future version.
+ * This sets both the send hook, as well as the user-provided `data`.
```
jan-auer, line 507: It's probably worth mentioning that if `data` is passed, a free func **must** be provided. This gets me wondering if we should move this, instead:

- `new`: only create a transport without data.
- `set_data` or probably better `set_state`: set the state and provide a free func in one go.

There may of course be an edge case when the data is static and doesn't need to be freed, but in this case passing an explicit `NULL` to the free func sounds much more defensive.

include/sentry.h
jan-auer, line 510: To confirm: This enforces that there is always a `send_envelope_func`. Are we happy with this constraint going forward? If in the future we choose to deprecate it, we'd either make a breaking change, or allow passing `NULL`.

IMHO, there's no mid-term reason to make `send_envelope_func` optional or replace it with another function, so this is fine.

src/sentry_transport.h
```diff
@@ -5,6 +5,19 @@
 
 typedef struct sentry_rate_limiter_s sentry_rate_limiter_t;
 
+/**
+ * The transport interface of sentry.
+ */
+// TODO: make this completely opaque, and rather use explicit functions from
+// core
```
jan-auer, line 12: Can we resolve this TODO in this PR already? It would be great to start reducing the number of TODOs.

src/sentry_core.c
Swatinem: According to unified API, this lives in the generic init options. In either case I would rather do that as a followup.

include/sentry.h
Swatinem: I was actually thinking about pairing up functions, but I think its better this way. This way the function signature is also the same as with the previous function transport. I will add a comment about free.

include/sentry.h
Swatinem: Yes, this is the way forward.

src/sentry_core.c
jan-auer: Ah, interesting. I had only compared with Rust where this was on the transport.

include/sentry.h
jan-auer: Since we're breaking anyway, I think it's worth reconsidering this. My fear is that with the current signature (especially when migrating!) it is way too easy to produce a memory leak.

PR #257, opened 2020-05-12
fix: Avoid crashes on sentry_init failure

Swatinem:


src/sentry_core.c
```diff
@@ -75,10 +76,24 @@ sentry_init(sentry_options_t *options)
     SENTRY_DEBUGF("using database path \"%" SENTRY_PATH_PRI "\"",
         options->database_path->path);
 
+    // try to create and lock our run folder as early as possibly, since it is
+    // fallible. since it does locking, it will not interfere with run folder
+    // enumeration.
+    options->run = sentry__run_new(options->database_path);
+    if (!options->run) {
+        sentry_options_free(options);
+        return 1;
```
jan-auer, line 85: Potentially there could be a way to use part of the SDK even without a run folder. It's fair to assume that we need it, though.

src/sentry_core.c
```diff
@@ -191,6 +206,10 @@ sentry_user_consent_t
 sentry_user_consent_get(void)
 {
     sentry__mutex_lock(&g_options_mutex);
+    if (!g_options) {
+        sentry__mutex_unlock(&g_options_mutex);
+        return SENTRY_USER_CONSENT_UNKNOWN;
+    }
     sentry_user_consent_t rv = g_options->user_consent;
     sentry__mutex_unlock(&g_options_mutex);
     return rv;
```
jan-auer, line 215: This function could probably be simplified by initializing with `rv = SENTRY_USER_CONSENT_UNKNOWN` and then accessing options to overwrite it.

src/sentry_core.c
```diff
@@ -288,9 +310,13 @@ sentry_capture_event(sentry_value_t event)
...
 {
+    const sentry_options_t *opts = sentry_get_options();
+    if (!opts) {
+        return;
+    }
     SENTRY_DEBUG("handling exception");
-    if (g_options->backend && g_options->backend->except_func) {
-        g_options->backend->except_func(g_options->backend, uctx);
+    if (opts->backend && opts->backend->except_func) {
+        opts->backend->except_func(opts->backend, uctx);
```
jan-auer, line 319: This didn't change, but the signature makes me wonder who takes ownership of `uctx`. Since it's not `const`, it seems like we might need to free in branches where we don't pass it on. Otherwise, can we make it `const` by any chance?

src/sentry_session.c
```diff
@@ -112,10 +112,12 @@ sentry__session_to_json(
 
     sentry__jsonwriter_write_key(jw, "attrs");
     sentry__jsonwriter_write_object_start(jw);
-    sentry__jsonwriter_write_key(jw, "release");
-    sentry__jsonwriter_write_str(jw, sentry_options_get_release(opts));
-    sentry__jsonwriter_write_key(jw, "environment");
-    sentry__jsonwriter_write_str(jw, sentry_options_get_environment(opts));
+    if (opts) {
+        sentry__jsonwriter_write_key(jw, "release");
+        sentry__jsonwriter_write_str(jw, sentry_options_get_release(opts));
```
jan-auer, line 117: This would create an invalid session, since sessions must contain a release. In that case, we should not write the session at all.

src/sentry_core.c
Swatinem: hm… crashpad needs it for the msgpack files, breakpad needs it in case of a hard crash, as does the inproc backend. Not sure it makes sense to proceed without a run folder.

src/sentry_core.c
Swatinem: nice catch! made it const

PR #258, opened 2020-05-13
style: Make format for all files

philipphofmann:
The vendor folder was not included in the list of folders for
clang-format and therefore not formatted. Use find to search
for all .h, .c and .cpp files and call clang-format on them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,16 +62,11 @@ setup-venv: .venv/bin/python
 	.venv/bin/pip install --upgrade --requirement tests/requirements.txt
 
 format: setup-venv
-	@clang-format -i \
-		examples/*.c \
-		include/*.h \
-		src/*.c \
-		src/*.h \
-		src/*/*.c \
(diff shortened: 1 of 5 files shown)


jan-auer:
That was intentional, since we didn't want to reformat vendored files. What's the reason for formatting them?

Swatinem:
IMO its a lot easier to keep track of upstream changes if you just include them verbatim without any reformatting.

philipphofmann:
We are also introducing Clang-Format in sentry-cocoa and I checked out this repo to find out if you use it. I stumbled across the makefile and wanted to format all files in cocoa. Therefore I came up with using `find` and I thought maybe it improves the code here as well. It only took me 10 minutes. If you don't find it useful we can close this PR. I could have also asked before.

Swatinem:
I definitely appreciate the effort ;-)

jan-auer:
Thanks for bringing this up, @philipphofmann!

PR #259, opened 2020-05-13
fix: Avoid creating invalid sessions

Swatinem:


src/sentry_session.c
```diff
@@ -47,7 +47,19 @@ status_from_string(const char *status)
...
+        return NULL;
+    }
+    const char *release = sentry_options_get_release(opts);
+    const char *environment = sentry_options_get_environment(opts);
+    if (!release) {
+        return NULL;
+    }
+
     sentry_session_t *rv = SENTRY_MAKE(sentry_session_t);
+    rv->release = release;
```
jan-auer, line 61: Just pointing out that this is safe because we don't allow to change the release after initialization.

PR #267, opened 2020-05-31
feat: implement logger hook

### LINKED ISSUE #148: add logger callback for SENTRY_LOGF
add sentry_logger_function_t

Thanks!

### LINKED ISSUE #244: Logging
Is it possible to enable debug logging to a log file in the file system?

Mixaill:
add ability to create logger hook

```c
void logger_hook(sentry_level_t level, const char* message, va_list args)
{
    char buf[1024]{};
    vsprintf(buf, message, args);
    OutputDebugStringA(buf);
}

...

void init_sentry() 
{
    sentry_options_t* _options = sentry_options_new();
    sentry_options_set_logger(_options, logger_hook);

   ...
}

```

fixes #148 
references #244

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -70,6 +70,7 @@ extern "C" {
 #endif
 
 #include <inttypes.h>
+#include <stdarg.h>
 #include <stdbool.h>
 #include <stddef.h>
 
@@ -716,6 +717,14 @@ SENTRY_API void sentry_options_set_debug(sentry_options_t *opts, int debug);
  * Returns the current value of the debug flag.
  */
 SENTRY_API int sentry_options_get_debug(const sentry_options_t *opts);
+
+/**
+ * Sets the sentry-native logger function.
+ * Used for logging debug events when the `debug` option is set to true.
+ */
+SENTRY_API void sentry_options_set_logger(sentry_options_t *opts,
+    void (*logger_func)(
+        sentry_level_t level, const char *message, va_list args));
 
 /**
  * Enables or disables user consent requirements for uploads.
--- a/src/sentry_logger.c
+++ b/src/sentry_logger.c
@@ -1,0 +1,90 @@
+#include <stdio.h>
+
+#include "sentry_logger.h"
+#include "sentry_options.h"
+
+#if defined(SENTRY_PLATFORM_ANDROID)
+
+#    include <android/log.h>
+void
+sentry__logger_defaultlogger(
+    sentry_level_t level, const char *message, va_list args)
+{
+    android_LogPriority priority = ANDROID_LOG_UNKNOWN;
+    switch (level) {
+    case SENTRY_LEVEL_DEBUG:
+        priority = ANDROID_LOG_DEBUG;
+        break;
+    case SENTRY_LEVEL_INFO:
+        priority = ANDROID_LOG_INFO;
+        break;
+    case SENTRY_LEVEL_WARNING:
+        priority = ANDROID_LOG_WARN;
+        break;
+    case SENTRY_LEVEL_ERROR:
+        priority = ANDROID_LOG_ERROR;
+        break;
+    case SENTRY_LEVEL_FATAL:
+        priority = ANDROID_LOG_FATAL;
+        break;
+    default:
+        break;
+    }
+    __android_log_vprint(priority, "sentry-native", message, args);
+}
+
+#else
+
+void
+sentry__logger_defaultlogger(
+    sentry_level_t level, const char *message, va_list args)
+{
+    const char *prefix = "[sentry] ";
+    const char *priority = sentry__logger_describe(level);
+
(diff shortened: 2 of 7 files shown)


include/sentry.h
```diff
@@ -717,6 +718,13 @@ SENTRY_API void sentry_options_set_debug(sentry_options_t *opts, int debug);
  */
 SENTRY_API int sentry_options_get_debug(const sentry_options_t *opts);
 
+/**
+ * Sets the sentry-native logger hook.
+ */
+SENTRY_API void sentry_options_set_loghook(sentry_options_t *opts,
```
Swatinem, line 724: I wonder if we should rather name this `logger`, and explain in the comment that it is used for sentries own logging output, and how it interacts with the `debug` option. Speaking of that, I’m not sure if setting a logger should automatically override `debug`? probably not?

src/sentry_log.c
```diff
@@ -0,0 +1,38 @@
...
+{
+    __android_log_vprint(ANDROID_LOG_INFO, "sentry-native", message, args);
+}
+
+#else
+
+void
+sentry_log_hook_default(sentry_level_t level, const char *message, va_list args)
+{
+    vfprintf(stderr, message, args);
```
Swatinem, line 20: can we still have the `[sentry] ` prefix here?

src/sentry_log.c
```diff
@@ -0,0 +1,38 @@
...
+}
+
+#endif
+
+void
+sentry_log(sentry_level_t level, const char *message, ...)
```
Swatinem, line 26: In general, we would like to prefix internal functions with double underscores, like this:

```suggestion
sentry__log(sentry_level_t level, const char *message, ...)
```

The same is true for the default function above.

src/sentry_log.c
```diff
@@ -0,0 +1,38 @@
...
+{
+    const sentry_options_t *_options = sentry_get_options();
```
Swatinem, line 28: also here, we don’t have a convention to prefix with underscores, so:

```suggestion
    const sentry_options_t *options = sentry_get_options();
```

Ah oh, now I see that this comes from within the macro. Still a good idea to change it though.

src/sentry_log.h
```diff
@@ -0,0 +1,15 @@
...
+
+void sentry_log(sentry_level_t level, const char *message, ...);
+
+#define SENTRY_DEBUGF(message, ...)                                            \
+    sentry_log(SENTRY_LEVEL_DEBUG, message "\n", __VA_ARGS__)
+
+#define SENTRY_DEBUG(message) sentry_log(SENTRY_LEVEL_DEBUG, message "\n")
+
+// for compatibility with existent macros
+#define SENTRY_TRACE SENTRY_DEBUG
```
Swatinem, line 14: hm, I’m wondering if we would want to actually make use of more log levels. Would be nice to actually have proper `WARN` there, and have everything else rather as `INFO`, and trace as `DEBUG`. But that should be a followup.

include/sentry.h
Mixaill: renamed, comment changed

src/sentry_log.h
Mixaill: added `SENTRY_WARN` and output severity level in default logger.

src/sentry_logger.c
```diff
@@ -0,0 +1,69 @@
...
+#include "sentry_options.h"
+
+#if defined(SENTRY_PLATFORM_ANDROID)
+
+#    include <android/log.h>
+void
+sentry__logger_default(
+    sentry_level_t level, const char *message, va_list args)
+{
+    __android_log_vprint(ANDROID_LOG_INFO, "sentry-native", message, args);
```
Swatinem, line 13: please use appropriate `ANDROID_LOG_XXX` levels here.

src/sentry_logger.c
```diff
@@ -0,0 +1,69 @@
...
+#else
+
+void
+sentry__logger_defaultlogger(
+    sentry_level_t level, const char *message, va_list args)
+{
+    const char *prefix = "[sentry]";
+    const char *severity = sentry__logger_describe(level);
+
+    char *format = malloc(strlen(prefix) + strlen(severity) + strlen(message) + 1);
```
Swatinem, line 25: We have a `sentry_malloc` which is signal safe, since we also do some logging inside signal handlers.

https://github.com/getsentry/sentry-native/blob/5020ff55ac457b02951dd56909424d0fbcda57d0/include/sentry.h#L97

Also, the prefixes need some spacing. How about a format like `[sentry] WARN …` or something like that?

Mixaill:
clang-format support in MSVS is the worst thing I've ever seen. :(

src/sentry_logger.c
Mixaill: replaced with sentry, spacing added: `[sentry] priority message`

PR #270, opened 2020-06-02
fix: Avoid memory unsafety when loading session from disk

Swatinem:
The session now owns its `release`/`environment`, to avoid borrowing problems when loading from disk. And also add a test that ensures loading from disk and flagging as abnormal works.

### DIFF
--- a/src/sentry_session.c
+++ b/src/sentry_session.c
@@ -51,15 +51,21 @@ sentry__session_new(void)
     if (!opts) {
         return NULL;
     }
-    const char *release = sentry_options_get_release(opts);
-    const char *environment = sentry_options_get_environment(opts);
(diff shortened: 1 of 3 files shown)


src/sentry_session.c
```diff
@@ -51,15 +51,21 @@ sentry__session_new(void)
...
+    if (!rv) {
+        return NULL;
+    }
+
+    char *environment
+        = sentry__string_clone(sentry_options_get_environment(opts));
+
     rv->release = release;
-    rv->environment = environment;
+    rv->environment = sentry__string_clone(environment);
```
jan-auer, line 68: You're cloning twice here, is this intended?

Swatinem: lol, my bad

PR #271, opened 2020-06-03
fix: Two issues uncovered by GCC analyzer

Swatinem:
I ran gcc with `-fanalyzer`, which uncovered some potential null dereferences. 

It does however have some false positives as well, such as:

https://github.com/getsentry/sentry-native/blob/b48c21d244092658d6e2d1bb243b705fd968b9f7/src/sentry_envelope.c#L90

Plus it complains about some code in `mpack` which we vendor.

### DIFF
--- a/src/sentry_json.c
+++ b/src/sentry_json.c
@@ -411,6 +411,10 @@ tokens_to_value(jsmntok_t *tokens, size_t token_count, const char *buf,
     jsmntok_t *root = POP();
     sentry_value_t rv = sentry_value_new_null();
 
+    if (!root) {
+        goto error;
+    }
+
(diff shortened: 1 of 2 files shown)


Swatinem:
> Could you run that analyzer on CI?

Not sure if gcc >10 is available there. Also, it does have some false positives, and we don’t run with warnings as errors on CI, since we still have some warnings there.

Also, running this locally, gcc 10 actually errored on this code here:

https://github.com/getsentry/sentry-native/blob/b48c21d244092658d6e2d1bb243b705fd968b9f7/tests/unit/test_symbolizer.c#L28

Apparently it does not like "functions being treated as arrays"?

PR #273, opened 2020-06-04
Add option to disable linking Sentry with pthread

blinkov:
It allows to externally provide different `pthread` implementation, for example, compatible with older `glibc` in outdated Linux distributions.


Mixaill:
By the way, why it explicitly links to `pthread` instead of 
```cmake
find_package(Threads REQUIRED)
target_link_libraries(sentry PRIVATE Threads::Threads)
``` 
?

https://cmake.org/cmake/help/latest/module/FindThreads.html

Swatinem:
Hm, that would make sense I guess.

Swatinem:
I merged this as is. We can start using the suggested `find_package` as a followup.

PR #275, opened 2020-06-04
build: Publish with craft

jan-auer:


.craft.yml
```diff
@@ -0,0 +1,29 @@
...
+  repo: sentry-native
+changelogPolicy: simple
+
+statusProvider:
+  name: github
+artifactProvider:
+  name: zeus
+
+targets:
+  - name: github
```
Swatinem, line 13: so this pushes the generated zip file also to the github release, right? How does it do the naming?

jan-auer:
Ok, this looks good now. Zeus revision is [here](https://zeus.ci/gh/getsentry/sentry-native/revisions/e19565959e651fa8e923ef934e4f3adf80a0de1e).

.craft.yml
jan-auer: This publishes to GitHub and applies the Changelog, correct. Naming of the file is always `sentry-native.zip`. Do you think it's worth keeping the version? Would have to figure out how to get the version in Azure.

PR #277, opened 2020-06-05
fix: Run through clang-analyzer and fail on warnings

Swatinem:
* correct some issues found by clang-analyzer (dead stores, one of them
  seems to have been a bug in the page allocator all along)
* fix other warnings on clang
* try to turn of warnings-as-errors on CI, and run through `scan-build`

### DIFF
--- a/src/sentry_sync.h
+++ b/src/sentry_sync.h
@@ -226,9 +226,7 @@ typedef pthread_cond_t sentry_cond_t;
         } while (0)
 #    define sentry__cond_wake pthread_cond_signal
 #    define sentry__thread_spawn(ThreadId, Func, Data)                         \
-        (pthread_create(ThreadId, NULL, (void *(*)(void *))Func, Data) == 0    \
-                ? 0                                                            \
-                : 1)
+        (pthread_create(ThreadId, NULL, Func, Data) == 0 ? 0 : 1)
 #    define sentry__thread_join(ThreadId) pthread_join(ThreadId, NULL)
 #    define sentry__threadid_equal pthread_equal
 #    define sentry__current_thread pthread_self
--- a/src/sentry_unix_pageallocator.c
+++ b/src/sentry_unix_pageallocator.c
@@ -85,6 +85,7 @@ sentry__page_allocator_alloc(size_t size)
 
     sentry__spinlock_lock(&g_lock);
 
+    // current page is large enough
     if (g_alloc->current_page
         && g_alloc->page_size - g_alloc->page_offset >= size) {
         rv = g_alloc->current_page + g_alloc->page_offset;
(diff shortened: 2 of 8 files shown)


src/sentry_logger.c
```diff
@@ -43,11 +43,12 @@ sentry__logger_defaultlogger(
...
-        strlen(prefix) + strlen(priority) + strlen(message) + 1);
-    strcpy(format, prefix);
-    strcat(format, priority);
-    strcat(format, message);
+    size_t len = strlen(prefix) + strlen(priority) + strlen(message) + 1;
+    char *format = sentry_malloc(len);
+
+    strncpy(format, prefix, strlen(prefix));
+    strncat(format, priority, strlen(priority));
+    strncat(format, message, strlen(message));
```
jan-auer, line 51: Where is the difference to a trivial `strcat`? Since you just use `strlen`, this should be exactly equivalent.

src/sentry_scope.c
```diff
@@ -209,7 +209,8 @@ sentry__symbolize_stacktrace(sentry_value_t stacktrace)
         if (!addr) {
             continue;
         }
-        sentry__symbolize((void *)addr, sentry__symbolize_frame, &frame);
+        sentry__symbolize(
+            (void *)(size_t)addr, sentry__symbolize_frame, &frame);
```
jan-auer, line 213: Can you make `addr` a `size_t` directly, rather than casting?

src/sentry_sync.h
```diff
@@ -258,7 +258,7 @@ static inline int
 sentry__atomic_fetch_and_add(volatile int *val, int diff)
 {
 #ifdef SENTRY_PLATFORM_WINDOWS
-    return InterlockedExchangeAdd(val, diff);
+    return (int)InterlockedExchangeAdd((volatile LONG *)val, (LONG)diff);
```
jan-auer, line 261: Usually, `int` and `long` are both 32-bit, but generally I'm afraid that this might cause problems. If you're compiling this on a platform where `int` and `long` have different size, an overflow may write into memory after `int val`.

I'm actually glad that the linter caught that.

src/sentry_unix_pageallocator.c
```diff
@@ -93,21 +100,24 @@ sentry__page_allocator_alloc(size_t size)
...
+            = (size + sizeof(struct page_header) + g_alloc->page_size - 1)
+            / g_alloc->page_size;
+        rv = get_pages(pages);
+
+        if (rv) {
+            g_alloc->page_offset
+                = (g_alloc->page_size
+                      - (g_alloc->page_size * pages
+                          - (size + sizeof(struct page_header))))
+                % g_alloc->page_size;
```
jan-auer, line 115: This is all quite hard to read, could you maybe introduce some intermediary variables and -- if necessary -- explain what they mean?

src/sentry_logger.c
Swatinem: This is actually quite frustrating. So msvc suggests to use `strcat_s` instead, as a save alternative. But even though that is supposed to be in C11, neither clang nor gcc have it for whatever reason. There is `strlcat` only on BSDs apparently, and `strscat` only in the linux kernel.

src/sentry_sync.h
Swatinem: windows 32-bit builds had a compiler warning here. TBH, I’m not quite sure what the correct thing to do here is.

src/sentry_logger.c
jan-auer: Right, `strcat_s` is marked as optional in the C standard and not in the C++ standard afaik. In that case, maybe add a comment above that block explaining what the intention is.

src/sentry_sync.h
jan-auer: It seems that we only use this for refcounting, so you have two options:
1. Make it a `long` all the way. `__sync_fetch_and_add` also supports `long`.
2. Assert in the Windows case that `sizeof(int) == sizeof(long)`, then cast.

Issue #190, opened 2020-03-20
Provide Breakpad and inproc backends for Windows


sseyod:
Hi!

My remaining problem now is that I'm not seeing any crashes reported in the Browser.

NB I'm provoking a crash in response to a menu item within the app, which calls this code:

```
      // Force crash report for sentry.io
      int* crashPtr = nullptr;
      *std::launder(crashPtr) = 0;
```

The app does indeed crash as expected, but for the avoidance of doubt, I see no report about this on the sentry.io website.

I've enabled this line as requested:
sentry_options_set_debug(options, 1);

NB I don't know if this is relevant, but I've noticed that, whether or not I crash the app, or have the sentry_options_set_debug call, I'm no longer seeing anything reported on the sentry.io site sent from sentry_capture_event(...) ... *unless* I modify the "message" parameter in the call below.
In otherwords, only the first call is reported on the site. Subsequent calls aren't reported on the site.

  sentry_capture_event(sentry_value_new_message_event(
    /*   level */ SENTRY_LEVEL_INFO,
    /*  logger */ "custom",
    /* message */ "Prepare to crash..."
  ));

Hopefully you see something useful on the server! :)

Best wishes,

Pete

jan-auer:
Thanks for the information. I can see two separate issues here.

1. Your crashes are not arriving in Sentry. What do the debug logs say when you start up the app? I'm assuming you're running on Windows with the Crashpad backend (default). Provided you've set up the SDK with a valid DSN, crashpad should be uploading immediately after the crash. Based on the SDK's debug logs, we can investigate further.

2. Subsequent calls are not reported independently of whether the app crashes. There is a known issue that if the app crashes while sending an event (which is done in a background thread), you might lose the event based on the progress. However, this seems unrelated, as you're only ever getting one event through. Can you observe similar behavior when building the [`example`](https://github.com/getsentry/sentry-native/blob/c86a6271de131e8599e684c2fee83e8389916ddf/examples/example.c) program, included in the package?

sseyod:
Hi!

With respect to 1) I presume that is what I'm doing... I'm just running with a vanilla configuration.
cmake . -DBUILD_SHARED_LIBS=OFF

I'm sure the DSN must be correct, as otherwise I presume I wouldn't see *anything* from sentry_capture_event ...?!

[...]

Issue #170, opened 2020-03-11
Windows 10 Store App - C++ - how to configure project to use Sentry crash reporting?


sseyod:
Hi!

Now that you've just released updated the Native / C++ version (thanks!), I've been able to download it, install Make for Windows 10, and start trying to integrate with one of my C++ projects for Windows 10.

1) I've run cmake like this (please note that it wasn't clear that this is how it needs to be run for Windows 10 native - I hope it is correct!)

cmake . -DBUILD_SHARED_LIBS=OFF

2) I've updated my C++ project to include calls as per your documentation (note use of extern "C",  which I presume is correct):

extern "C" {
#include <sentry.h>
}

int main(void) {
...
  sentry_options_t *options = sentry_options_new();
  sentry_options_set_dsn(options, "https://blah/blah");
  sentry_init(options);

  /* ... */

  // make sure everything flushes
  sentry_shutdown();
}

3) I've updated my project to import the same .lib files that your test project (sentry_example) uses, including sentry.lib (and all the other ones).

4) However, I'm getting this:

1>Main.obj : error LNK2001: unresolved external symbol __imp_sentry_value_new_message_event
1>Main.obj : error LNK2001: unresolved external symbol __imp_sentry_init
1>Main.obj : error LNK2001: unresolved external symbol __imp_sentry_shutdown
1>Main.obj : error LNK2001: unresolved external symbol __imp_sentry_options_set_dsn
1>Main.obj : error LNK2001: unresolved external symbol __imp_sentry_capture_event
1>Main.obj : error LNK2001: unresolved external symbol __imp_sentry_options_new
1>C:\...blah...\myapp.exe : fatal error LNK1120: 6 unresolved externals

Can you please advise?

Pete

sseyod:
Aha

Looking in CMakeLists.txt:

if(BUILD_SHARED_LIBS)
        add_definitions(-DSENTRY_BUILD_SHARED)
else()
        add_definitions(-DSENTRY_BUILD_STATIC)
endif()

... so that makes me think I need to do something like this instead:

cmake . -DBUILD_SHARED_LIBS=OFF -DSENTRY_BUILD_STATIC=ON

Clearly, I'm struggling with the documentation right now!

Pete

sseyod:
Looks like the solution was to do this in my app:

#define SENTRY_BUILD_STATIC 1
#include <sentry.h>

sseyod:
[...]

PR #278, opened 2020-06-05
feat: Add Breakpad support for Windows

Mixaill:
Kinda working for me. Tested on example project.

Several questions:
- Is it intended that breakpad backend sends data on second run?
- Is it OK to create `.envelope` even when DSN is not set? In such case `.dmp` is not available until DSN will be set and envelope will be sended to the sentry.
- Is there any utility to inspect/unpack `.envelope` ?

references #170
references #190


jan-auer:
Thanks, @Mixaill. Answering your questions first:

> * Is it intended that breakpad backend sends data on second run?

This is intended, yes. Since breakpad runs everything in-process after the crash, there are only limited possibilities. Particularly, it is no longer possible to run reliable network communication, much less to upload an entire minidump. This is why we only write it to disk and upload it on the next run.

> * Is it OK to create `.envelope` even when DSN is not set? In such case `.dmp` is not available until DSN will be set and envelope will be sended to the sentry. 

Excellent question. The current backend implementations (both Crashpad and Breakpad) do create the envelope and write it into the run folder. Ideally, however, you would not create the envelope. Not setting a DSN is intended to disable most of the SDK's functionality, and as such should also skip writing an envelope. This is also something for @Swatinem to chime in on.

> Is there any utility to inspect/unpack `.envelope` ?

There is `sentry__envelope_from_path`, which loads an envelope. That only stores the raw payload, though, and you cannot inspect its contents. So far, we have not had the need to read an envelope once it's been serialized -- what would be your use case here?

external/CMakeLists.txt
```diff
@@ -1,51 +1,96 @@
...
+    breakpad/src/client/windows/crash_generation/crash_generation_client.cc
+    breakpad/src/client/windows/crash_generation/crash_generation_client.h
+    breakpad/src/client/windows/crash_generation/crash_generation_server.cc
+    breakpad/src/client/windows/crash_generation/crash_generation_server.h
+    breakpad/src/client/windows/crash_generation/minidump_generator.cc
+    breakpad/src/client/windows/crash_generation/minidump_generator.h
+    breakpad/src/client/windows/handler/exception_handler.cc
+    breakpad/src/client/windows/handler/exception_handler.h
+    breakpad/src/client/windows/sender/crash_report_sender.cc
+    breakpad/src/client/windows/sender/crash_report_sender.h
```
jan-auer, line 68: Can we compile without the sender since we have our own?

Mixaill: removed unnedded files for Windows from CMakeLists

Mixaill:
>what would be your use case here

In my case I want to display something like that I had with CrashRpt before:
image
image

Sorry for Russian, but the common sense should be clear: 
* the first window with the general information and link to the Privacy Policy is displayed after crash
* the second one with submission content is displayed on `Inspect` button click

So, I plan to implement it something like that:
- Application crash
- Breakpad `UnhandledExceptionFilter` execution
- Start other process which waits for original application termination
- After original application termination, initialize sentry in second application with the same DB folder
- Display window like the first screenshot and display envelope content on inspect button
- Send envelope on `Send` button and delete envelope on `Close` button.

jan-auer:
That sounds excellent! In case you would like to show the files contained in the envelope, you can find a reference of its structure [here](https://develop.sentry.dev/sdk/envelopes/#serialization-format). 

If your background-uploader works well and you would like to share it, I would love to take a look at it. Uploading from a background process sounds like a useful addition to the SDK in general, even without a UI.

src/backends/sentry_backend_breakpad.cpp
```diff
@@ -37,38 +42,43 @@ sentry__breakpad_backend_send_envelope(
...
         sentry_envelope_free(envelope);
         return;
     }
     sentry__envelope_item_set_header(
         item, "attachment_type", sentry_value_new_string("event.minidump"));
-    sentry__envelope_item_set_header(item, "filename",
-        sentry_value_new_string(sentry__path_filename(dump_path)));
+
+#ifdef SENTRY_PLATFORM_WINDOWS
+    char *file_path = sentry__string_from_wstr(state->dump_path->path);
```
Swatinem, line 62: not quite sure why you need this ifdef here. I think `sentry__path_filename` should be implemented on windows as well, and we only want the basename, not the absolute path.

Swatinem:
btw, the CI failure in `Create Release Archive Job` is our fault and we will try to fix this today.

Mixaill:
>I think changing this condition here should be enough to run the relevant breakpad tests also on windows:

Enabled

src/backends/sentry_backend_breakpad.cpp
Mixaill: Changed

src/sentry_path.h
```diff
@@ -85,7 +85,7 @@ sentry_path_t *sentry__path_clone(const sentry_path_t *path);
 void sentry__path_free(sentry_path_t *path);
 
 /**
- * This will return a copy of the last path segment, which is typically the file
+ * This will return a pointer to the last path segment, which is typically the file
```
Swatinem, line 88: 👍

src/backends/sentry_backend_breakpad.cpp
```diff
@@ -37,38 +42,41 @@ sentry__breakpad_backend_send_envelope(
...
         sentry_envelope_free(envelope);
         return;
     }
     sentry__envelope_item_set_header(
         item, "attachment_type", sentry_value_new_string("event.minidump"));
-    sentry__envelope_item_set_header(item, "filename",
-        sentry_value_new_string(sentry__path_filename(dump_path)));
+
+#ifdef SENTRY_PLATFORM_WINDOWS
+    const char *file_name = sentry__path_filename_str(state->dump_path);
```
Swatinem, line 62: ah, I completely forgot about the wstr on windows. I think we should rather do it like this here:

https://github.com/getsentry/sentry-native/blob/621ee38b0c6fa24a5bd035f66f8accda3ac1910f/src/sentry_core.c#L299-L305

instead of defining a new `_str` API.

src/sentry_path.h
```diff
@@ -211,13 +211,19 @@ sentry_path_t *sentry__path_from_wstr(const wchar_t *s);
  */
 sentry_path_t *sentry__path_join_wstr(
     const sentry_path_t *base, const wchar_t *other);
+
+/**
+ * This will return a copy of the last path segment, which is typically the file
+ * or directory name. The returns will utf8 string
+ */
+const char *sentry__path_filename_str(const sentry_path_t *path);
```
Swatinem, line 219: this should technically be a `char *` without the const, but as commented above, I would rather avoid having this function at all.

src/backends/sentry_backend_breakpad.cpp
Mixaill: replaces

src/sentry_path.h
Mixaill: removed

Mixaill:
Well, there are two problems

* `Assertion failed: !bgw->running && bgw->task_count == 0, file D:\a\1\s\src\sentry_sync.c, line 159` WinHTTP transport was destroyed before all tasks were executed?
* CI hangs on assert instead of fail.

Swatinem:
>     * WinHTTP transport was destroyed before all tasks were executed?

oh, its possible we are running into this:

https://github.com/getsentry/sentry-native/blob/9651561d45e4d00e9fe708275c086a3cfeb496bd/src/sentry_core.c#L127-L133

I remember I was actually on windows when adding that comment.

That may be related to this:

https://github.com/getsentry/sentry-native/blob/9651561d45e4d00e9fe708275c086a3cfeb496bd/src/sentry_sync.c#L131-L148

I remember we did observe time going backwards, I think that was actually on windows @jan-auer? Maybe @mitsuhiko can also chime in, since he wrote that original thread joining code.

I cant imagine why a HTTP request to localhost may take up more than 5s, so maybe our timer is jumping wildly.

Swatinem:
so, what should we be doing in that case anyway? In the very least, we will leak memory from the background thread. Not quite sure if we should trigger a hard assert there?

Mixaill:
>I cant imagine why a HTTP request to localhost may take up more than 5s
Firewall configuration on test machine / VM?

src/sentry_sync.h
```diff
@@ -258,7 +258,7 @@ static inline long
 sentry__atomic_fetch_and_add(volatile long *val, long diff)
 {
 #ifdef SENTRY_PLATFORM_WINDOWS
-    return InterlockedExchangeAdd(val, diff);
+    return InterlockedExchangeAdd((unsigned int *)val, diff);
```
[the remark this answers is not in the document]
Mixaill: Fixed, although `long` is usually 32-bit on Windows.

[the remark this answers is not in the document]
eakoli: Yea, i'm still confused by the warning that it gives, as it makes no sense.
I can only assume there is something going on depending on if its visual c, clang or gcc doing the compilation.

Mixaill:
@Swatinem I temporary merge #284 here, and it looks like fix the issue.

The other issue, that it fails on https://github.com/getsentry/sentry-native/blob/master/tests/test_integration_http.py#L188

So, why it event expects two request from client?

Swatinem:
> I temporary merge #284 here, and it looks like fix the issue.

That is good to hear.

So the envelope logic is explained here:

https://github.com/getsentry/sentry-native/blob/51acc00c90ce60c72d583abd4164aad6b27971c5/src/backends/sentry_backend_breakpad.cpp#L104-L124

You should get one session, and one event. Your content-length suggests you are getting the event, but not the session.

https://github.com/getsentry/sentry-native/blob/51acc00c90ce60c72d583abd4164aad6b27971c5/tests/test_integration_http.py#L207-L221

This test here should make sure that the dump-to-disk logic works as expected.

Mixaill:
@Swatinem so the problem was it tries to send session via WinHTTP instead of saving to disk.

The problem is enforcing of disk transport fixes `test_breakpad_crash_http` test but breaks `test_breakpad_dump_inflight` (because of 00000000-0000-0000-0000-000000000000.envelope ?)

Mixaill:
@Swatinem fixes, now it passes all the tests

src/backends/sentry_backend_breakpad.cpp
```diff
@@ -118,13 +149,18 @@ sentry__breakpad_backend_callback(
     sentry_capture_event(event);
 
     // after capturing the crash event, try to dump all the in-flight data of
-    // the previous transport
-    if (transport) {
-        sentry__transport_dump_queue(transport);
+    // the previous transports
+    if (transport_disk) {
+        sentry__transport_dump_queue(transport_disk);
```
Swatinem, line 154: While this is not *wrong*, you don’t need it since the disk transport is completely sync.

src/backends/sentry_backend_breakpad.cpp
```diff
@@ -89,17 +97,40 @@ sentry__enforce_breakpad_transport(
...
+    dump_path = sentry__path_join_wstr(dump_path, minidump_id);
+    dump_path = sentry__path_append_str(dump_path, ".dmp");
+#else
+    dump_path = sentry__path_new(descriptor.path());
+#endif
+
+    // since we can’t use HTTP in crash handlers, we will swap out the
+    // transport here to one that serializes the envelope to disk
+    sentry_transport_t *transport_original = options->transport;
+    sentry__enforce_disk_transport();
```
Swatinem, line 133: 👍 Why didn’t I think of that back when I did this. Seems so obvious in hindsight.

src/backends/sentry_backend_breakpad.cpp
Mixaill: Yeah, I put it here just in case the logic of callback sometime gets more complicated.

eakoli:
Ill have to look through this pr more closely as I didnt run into any of these issues. But I'm also based off of 0.3.1 not master.

also whenever I see a ```00000000-0000-0000-0000-000000000000.envelope``` its due to a second crash while processing the current crash. Which was normally because of memory corruption etc.

Mixaill:
`00000000-0000-0000-0000-000000000000.envelope` contains session information. It is also created on first run if you start `sentry_example` with `start-session` argument. It is full of zeroes because session does not contains event ID.

eakoli:
@Swatinem It seems like it would be a good idea to add a way for internals to add an attachment via the event, so that the the Breakpad transport would no longer be needed.

Swatinem:
Indeed it would! We discussed this briefly to have a more fine-grained attachment support, but we didn’t reach a conclusion there yet. The native SDK is actually the first one which has attachment support AFAIK, and one of the few that works with envelopes internally.

Mixaill:
@Swatinem rebased on top of master with merged monotonic time

Swatinem:
🎉 Amazing stuff, thanks for sticking with it this long ;-)

src/backends/sentry_backend_breakpad.cpp
```diff
@@ -89,17 +97,40 @@ sentry__enforce_breakpad_transport(
...
+#endif
 
     const sentry_options_t *options = sentry_get_options();
     sentry__write_crash_marker(options);
-    const char *dump_path = descriptor.path();
+
+    sentry_path_t *dump_path = nullptr;
+#ifdef SENTRY_PLATFORM_WINDOWS
+    dump_path = sentry__path_new(breakpad_dump_path);
+    dump_path = sentry__path_join_wstr(dump_path, minidump_id);
```
eakoli, line 124: @Mixaill @Swatinem I think this introduces a leak, as the join and append return a new path object.

Swatinem: Good catch! I think you are right.

PR #283, opened 2020-06-09
feat: add SENTRY_BUILD_RUNTIMESTATIC flag

### LINKED ISSUE #282: Build dll with static runtime
Hello,

I want to build sentry.dll with static runtime (/MT).
I saw this mentioned in issue 170: https://github.com/getsentry/sentry-native/issues/170,
but couldn't figure out what the conclusion of that issue was.

I added to CMakeLists.txt, as suggested in that issue:
if(WIN32)
set(SENTRY_DEFAULT_TRANSPORT "winhttp")
# ??? FL: Added flags
# https://stackoverflow.com/questions/14172856/compile-with-mt-instead-of-md-using-cmake
set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} /MT")
set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} /MTd")
 
Can we get that as build option? Or is it already possible? I'm not that good at CMake..

We are not a Windows Store app, btw, but we have a plugin version, so I don't want to have dependency issues when we're loaded in another app.

Thanks,
Fredrik

Mixaill:
Allows to build sentry-native with the static runtime using MSVC compiler.
Fixes #282 

```
mkdir build
cd build
cmake .. -DSENTRY_BUILD_RUNTIMESTATIC=ON
cmake --build .
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,12 @@
-cmake_minimum_required (VERSION 3.10)
+if(WIN32)
+	cmake_minimum_required (VERSION 3.15)
+
+	# enables support for CMAKE_MSVC_RUNTIME_LIBRARY
+	cmake_policy(SET CMP0091 NEW)
+else()
+	cmake_minimum_required (VERSION 3.10)
+endif()
+
 project(Sentry-Native LANGUAGES C CXX ASM)
 
 set(SENTRY_MAIN_PROJECT OFF)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,10 @@
 cmake_minimum_required (VERSION 3.10)
+
+#MSVC ABI (CMake 3.15+)
```
Swatinem, line 3:
```suggestion
# MSVC ABI (CMake 3.15+)
```
can you also change the other comments to have a leading space?

Also, we do have a minimum requirement of `3.10`, mostly because of android and old linux though. I think its fine to require a higher version on windows. Is it possible to make the `cmake_minimum_required` above conditional?

Mixaill: changed

PR #291, opened 2020-06-10
Support for 32 bit linux builds and gcc 4.9

eakoli:
Adds option to enable 32 bit linux builds when on a 64bit host.
Links librt explicit in test suites (needed by older gcc chains that don't automatically manage it.)


tests/unit/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ target_include_directories(sentry_test_unit PRIVATE
 target_link_libraries(sentry_test_unit PRIVATE
 	${SENTRY_LINK_LIBRARIES}
 	${SENTRY_INTERFACE_LINK_LIBRARIES}
+	"$<$<PLATFORM_ID:Linux>:rt>"
```
Swatinem, line 48: can you add a comment what this means? I have no idea TBH

CMakeLists.txt
```diff
@@ -23,6 +23,15 @@ option(SENTRY_PIC "Build sentry (and dependent) libraries as position independen
 option(SENTRY_BUILD_TESTS "Build sentry-native tests" "${SENTRY_MAIN_PROJECT}")
 option(SENTRY_BUILD_EXAMPLES "Build sentry-native example(s)" "${SENTRY_MAIN_PROJECT}")
 
+if( LINUX )
```
Swatinem, line 26:
```suggestion
if(LINUX)
```

Can you properly format this code block so that the whitespaces are consistent please?

tests/unit/CMakeLists.txt
Mixaill: Links `rt` library if target platform is Linux

https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html

https://docs.oracle.com/cd/E88353_01/html/E37842/librt-3lib.html

Swatinem: well yes, but why is that suddenly needed for the unit tests only? Is that a restriction of old GCC?

eakoli: Yes, older gccs don't automatically link librt when needed. Im not sure what version started handling it automatically.

tests/__init__.py
```diff
@@ -102,6 +102,8 @@ def cmake(cwd, targets, options=None):
         configcmd.append("-D{}={}".format(key, value))
     if sys.platform == "win32" and os.environ.get("TEST_X86"):
         configcmd.append("-AWin32")
+    else if sys.platform == "linux" and os.environ.get("TEST_X86"):
```
Swatinem, line 105:
```suggestion
    elif sys.platform == "linux" and os.environ.get("TEST_X86"):
```

python 🤦‍♂️

PR #294, opened 2020-06-12
ContextRecord incorrectly passed to StackWalk64

eakoli:
The ContextRecord parameter should be a PCONTEXT (not a PCONTEXT* as was coded)
A local copy of context made as StackWalk64 can alter it.


eakoli:
It works correctly when correctly with the stack of the current thread ( not sure why ).
I actually found it as part of capturing the stack of other threads (another PR that will be pending soon). If the context is for another thread it only works for the first frame.

Swatinem:
Interesting. I wonder if we can create a unit test for this?

PR #296, opened 2020-06-15
improve 'sentry_options_set_logger' add 'userdata'

irov:
Hello, i some improve sentry logger, add 'userdata' for callback.

Thanks!


irov:
> looking good, but this is a breaking change technically, now that the logger has been released.

I do not think that this slight improvement can be called a major change. Maybe this could be expanded as an improvement in the next release? ;)

Mixaill:
> do not think that this slight improvement can be called a major change. 

It breaks API compatibility with the 0.3.2 so it is a major chnage :)

irov:
> It breaks API compatibility with the 0.3.2 so it is a major change which requires version bump to 0.4.0 at least :)

Give thanks! And I hope that my gray hair will not cover my whole head by then! In the meantime, I'll make a crutch through a global variable

Swatinem:
Lol, don’t worry. While those are not semver-breaking, I think the addition of windows breakpad + inproc backend are large enough changes to warrant a `0.4` release ;-) I hope to get those in during the week.

Swatinem:
I was preparing a separate struct and global for this: https://github.com/getsentry/sentry-native/commit/5bff34a4018979d4fcc42cd472258ac8452aceb6
Please rebase on top of that.

irov:
maybe I will create a new pull request, my first rebase failed :(

Swatinem:
I merged this now as part of #325 , thanks!

PR #299, opened 2020-06-16
ci: Run analyzers and leak-check on CI

Swatinem:
* run `scan-build` (clang analyzer) on linux (although currently it only produces warnings and not hard errors)
* switch to gcc-10 and clang-10 on linux
* run asan leak-check on clang builds on linux

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,17 +21,24 @@ stages:
               CC: gcc-7
               CXX: g++-7
               TEST_X86: 1
-            Linux (gcc 9):
-              vmImage: "ubuntu-latest"
-              CC: gcc-9
-              CXX: g++-9
-            Linux (clang 9):
-              vmImage: "ubuntu-latest"
-              CC: clang-9
-              CXX: clang++-9
+            Linux (gcc 10):
+              vmImage: "ubuntu-20.04"
+              CC: gcc-10
+              CXX: g++-10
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -21,17 +21,24 @@ stages:
...
-            Linux (clang 9):
-              vmImage: "ubuntu-latest"
-              CC: clang-9
-              CXX: clang++-9
+            Linux (gcc 10):
+              vmImage: "ubuntu-20.04"
+              CC: gcc-10
+              CXX: g++-10
+              # GCC does not respect the `-Wvariadic-macros` suppression right now
+              # ERROR_ON_WARNINGS: 1
```
jan-auer, line 29: How would you feel if we stay on GCC 9 until this is resolved?

src/sentry_logger.c
```diff
@@ -41,6 +41,7 @@ sentry__logger_defaultlogger(
     sentry_level_t level, const char *message, va_list args)
 {
     const char *prefix = "[sentry] ";
+    size_t len_prefix = strlen(prefix);
```
jan-auer: This is dead code.

azure-pipelines.yml
Swatinem: If I remember correctly, it had the same problems, lol.

PR #301, opened 2020-06-16
fix: Avoid using strncat and a trailing newline in logs

### LINKED ISSUE #298: Remove final new line from message in sentry_options_set_logger
Thanks for the feature but, can you send the message without a final new line? Other loggers already add a new line (in my case spdlog).

Thanks,

Swatinem:
The usage of `strcat` bothers gcc-10 and clang analyzer.

fixes #298

### DIFF
--- a/src/sentry_logger.c
+++ b/src/sentry_logger.c
@@ -41,17 +41,11 @@ sentry__logger_defaultlogger(
     sentry_level_t level, const char *message, va_list args)
 {
     const char *prefix = "[sentry] ";
-    size_t len_prefix = strlen(prefix);
     const char *priority = sentry__logger_describe(level);
-    size_t len_priority = strlen(priority);
-    size_t len_msg = strlen(message);
 
-    char *format = sentry_malloc(len_prefix + len_priority + len_msg + 2);
-    memcpy(format, prefix, len_prefix);
-    memcpy(format + len_prefix, priority, len_priority);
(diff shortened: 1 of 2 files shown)


Mixaill:
>The only disadvantage that I can see is that stderr can be unbuffered, but I think even that should not be an issue here.

It also could lead to messed output if application thread is also trying to write to the stderr at the same moment.

Swatinem:
> It also could lead to messed output if application thread is also trying to write to the stderr at the same moment.

That’s what I wanted to say. Even our example itself is quite spammy with the http thread in the background, and you do have interspersed logging there, which is quite difficult to follow as it is. Either way, while the allocation is suboptimal, I suspect people won’t have logging on in production anyway, and if they do they will probably integrate with their own system.

jan-auer:
Excellent point, I knew something was up with this :) Alright, let me take another stab at this. How about:

```c
const char *prefix = "[sentry] ";
const char *priority = sentry__logger_describe(level);

size_t total_len = strlen(prefix) + strlen(priority) + strlen(message) + 1;
char *format = sentry_malloc(total_len);
snprintf("%s%s%s", total_len, prefix, priority, message);

vfprintf(stderr, format, args);
```

Of course, prefix could also be inlined, but that's probably more prone to errors when changing the prefix.

Swatinem:
I also thought about that before, but yes, lets do it ;-)

PR #302, opened 2020-06-17
fix: Start logger as early as possible

### LINKED ISSUE #300: missing logging in sentry_init
some log in sentry_init does not show because g_options is not configured

like this

    SENTRY_DEBUGF("using database path \"%" SENTRY_PATH_PRI "\"",

Swatinem:
fixes #300

### DIFF
--- a/src/sentry_core.c
+++ b/src/sentry_core.c
@@ -61,10 +61,12 @@ sentry_init(sentry_options_t *options)
 {
     sentry_shutdown();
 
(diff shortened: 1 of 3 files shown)


Swatinem:
The `debug` flag enables or suppresses *all* the logging that sentry does internally. I think someone would have to create their own logger and gate internally on the level, with `debug` set to true in the options.

PR #304, opened 2020-06-17
ci: Run analysis using CodeChecker

Swatinem:
See https://github.com/Ericsson/codechecker

This basically is a frontend for `clang-tidy` and `clang-static-analyzer`.


Swatinem:
Some examples of false positives I see with CTU running locally (had to disable that on CI for now):

```
[HIGH] /home/swatinem/Coding/sentry/sentry-native/src/sentry_string.c:39:5: Null pointer passed to 1st parameter expecting 'nonnull' [core.NonNullParamChecker]
    memcpy(sb->buf + sb->len, s, len);
    ^
  Report hash: bbf569e0c283fab9b521cb403de0903c
  Macro expansions:
    1, sentry_envelope.c:360:1: Macro 'MUST_USE' expanded to '__attribute__((warn_unused_result))'
  Steps:
     1, sentry_envelope.c:382:14: Calling 'sentry_envelope_write_to_path'
     2, sentry_envelope.c:360:1: Entered call from 'sentry_envelope_write_to_file'
     3, sentry_envelope.c:367:17: Calling 'sentry_envelope_serialize'
     4, sentry_envelope.c:348:1: Entered call from 'sentry_envelope_write_to_path'
     5, sentry_envelope.c:352:5: Calling 'sentry__stringbuilder_init'
     6, sentry_string.c:8:1: Entered call from 'sentry_envelope_serialize'
     7, sentry_string.c:11:5: Null pointer value stored to 'sb.buf'
     8, sentry_envelope.c:352:5: Returning from 'sentry__stringbuilder_init'
     9, sentry_envelope.c:354:5: Calling 'sentry__envelope_serialize_into_stringbuilder'
    10, sentry_envelope.c:291:1: Entered call from 'sentry_envelope_serialize'
    11, sentry_envelope.c:295:9: Assuming field 'is_raw' is true
    12, sentry_envelope.c:296:9: Calling 'sentry__stringbuilder_append_buf'
    13, sentry_string.c:53:1: Entered call from 'sentry__envelope_serialize_into_stringbuilder'
    14, sentry_string.c:57:12: Calling 'append'
    15, sentry_string.c:16:1: Entered call from 'sentry__stringbuilder_append_buf'
    16, sentry_string.c:20:9: Assuming 'needed' is <= field 'allocated'
    17, sentry_string.c:39:5: Null pointer passed to 1st parameter expecting 'nonnull'
```

I don’t think this can ever happen, as `allocated` will be `0` in that case; and `needed` is always `>= 1` even if you want to append an empty string. Anyway, I will add another condition to it.

```
[MEDIUM] /home/swatinem/Coding/sentry/sentry-native/src/sentry_json.c:258:1: Potential leak of memory pointed to by 'formatted' [unix.Malloc]
}
^
  Report hash: d6367b9302465f90c90a3ab8eb9535cf
  Steps:
     1, sentry_json.c:255:23: Calling 'sentry__msec_time_to_iso8601'
     2, sentry_utils.c:327:1: Entered call from 'sentry__jsonwriter_write_msec_timestamp'
     3, sentry_utils.c:342:9: Assuming 'msecs' is 0
     4, sentry_utils.c:347:12: Calling 'sentry__string_clone'
     5, sentry_string.c:107:1: Entered call from 'sentry__msec_time_to_iso8601'
     6, sentry_string.c:110:18: Calling 'sentry__string_clonen'
     7, sentry_string.c:113:1: Entered call from 'sentry__string_clone'
     8, sentry_string.c:117:16: Calling 'sentry_malloc'
     9, sentry_alloc.c:13:1: Entered call from 'sentry__string_clonen'
    10, sentry_alloc.c:21:12: Memory is allocated
    11, sentry_string.c:117:16: Returned allocated memory
    12, sentry_string.c:118:9: Assuming 'rv' is non-null
    13, sentry_string.c:110:18: Returned allocated memory
    14, sentry_utils.c:347:12: Returned allocated memory
    15, sentry_json.c:255:23: Returned allocated memory
    16, sentry_json.c:258:1: Potential leak of memory pointed to by 'formatted'
```

Relevant code:

```
void
sentry__jsonwriter_write_msec_timestamp(sentry_jsonwriter_t *jw, uint64_t time)
{
    char *formatted = sentry__msec_time_to_iso8601(time);
    sentry__jsonwriter_write_str(jw, formatted);
    sentry_free(formatted);
}
```

lol, we just called `sentry_free` on it, how can that leak?

Looking at a lot more of these cases, I think the analyzer does not correctly handle ownership transfer of `void*`.
There are tons of false positives as well with `sentry_path_free`.

src/sentry_value.c
```diff
@@ -308,7 +308,11 @@ sentry_value_new_list(void)
     list_t *l = SENTRY_MAKE(list_t);
     if (l) {
         memset(l, 0, sizeof(list_t));
-        return new_thing_value(l, THING_TYPE_LIST);
+        sentry_value_t rv = new_thing_value(l, THING_TYPE_LIST);
+        if (sentry_value_is_null(rv)) {
```
Swatinem, line 312: These cases here are interesting. Sure, `new_thing_value` allocates internally and can fail. But `CTU` can’t look through the `sentry_value_is_null` assertion and will still flag all of these as false positives.

src/sentry_value.c
```diff
@@ -525,12 +547,8 @@ sentry__value_clone(sentry_value_t value)
         const obj_t *obj = thing->payload;
         sentry_value_t rv = sentry__value_new_object_with_size(obj->len);
         for (size_t i = 0; i < obj->len; i++) {
-            char *key = sentry__string_clone(obj->pairs[i].k);
```
Swatinem, line 528: this is just lol, why? Also, another piece of code that is apparently never run, or at least not in leak-checked code?

tests/cmake.py
```diff
@@ -81,15 +96,73 @@ def cmake(cwd, targets, options=None):
...
+
+        disable = [
+            "readability-magic-numbers",
+            "cppcoreguidelines-avoid-magic-numbers",
+            "readability-else-after-return",
+        ]
+        disables = []
+        for d in disable:
+            disables.append("--disable")
+            disables.append(d)
```
flub, line 139: minor: is `disables = ["--disable={}".format(d) for d in disable]` valid?

tests/cmake.py
```diff
@@ -81,15 +96,73 @@ def cmake(cwd, targets, options=None):
...
+            *disables,
+            "--print-steps",
+            "--ignore",
+            os.path.join(source_dir, ".codechecker-ignore"),
+            "--logfile",
+            "compilation.json",
+        ]
+        print("{} > {}".format(cwd, " ".join(checkcmd)), flush=True)
+        child = subprocess.run(checkcmd, stdout=subprocess.PIPE, cwd=cwd, check=True)
+        sys.stdout.buffer.write(child.stdout)
```
flub, line 159: TIL about `.buffer`

tests/cmake.py
```diff
@@ -81,15 +96,73 @@ def cmake(cwd, targets, options=None):
...
+        marker = b"Total number of reports: "
+        errors = child.stdout[child.stdout.rfind(marker) + len(marker) :]
+        errors = int(errors[: errors.find(b"\n")])
+        if errors > 0:
+            pytest.fail("code-checker analysis failed")
+
```
flub, line 165: minor: personally this function is now too big and long for me to remember all variables in scope etc and i would break this up.  at this point it's not problematic though, so i don't mind.

src/sentry_value.c
jan-auer: Potentially we didn't use `sentry_value_set_by_key` initially. To be fair, it would probably be faster to use something lower level, given that we know the size up-front and can skip a lot of checks. But please double-check why this never came up in any tests, we should at least have unit test coverage for this (preferred over integration tests, actually).

Otherwise, please add a comment that `sentry_value_set_by_key` clones the key internally.

tests/__init__.py
```diff
@@ -76,7 +76,16 @@ def run(cwd, exe, args, env=dict(os.environ), **kwargs):
...
+    try:
+        failed = False
+        return subprocess.run([*cmd, *args], cwd=cwd, env=env, **kwargs)
+    except subprocess.CalledProcessError:
+        cmd = " ".join(cmd)
+        args = " ".join(args)
+        cmd = f"{cmd} {args}"
+        failed = True
+    if failed:
+        pytest.fail(f'running command failed: "{cmd}"')
```
jan-auer, line 88: Why not just put `pytest.fail` the `except` block?

src/sentry_envelope.c
```diff
@@ -86,8 +86,11 @@ static sentry_envelope_item_t *
 envelope_add_from_owned_buffer(
     sentry_envelope_t *envelope, char *buf, size_t buf_len, const char *type)
 {
+    if (!buf) {
+        return NULL;
```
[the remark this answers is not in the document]
Swatinem: I think moving the checks into the function simplifies it a bit. The failure condition is basically when failing to copy or allocate for serialization. And when reading a file. That last bit might be worth printing a warning separately.

src/sentry_value.c
Swatinem: The problem why I do all the checks outside is that `new_thing_value` takes ownership of whatever its wrapping, but it doesn’t know how to free. Maybe with a separate function pointer argument, this becomes clearer.

src/sentry_string.h
```diff
@@ -90,7 +90,7 @@ static inline void
 sentry__string_ascii_lower(char *s)
 {
     for (; *s; s++) {
-        *s = (char)tolower((char)*s);
+        *s = (char)tolower((unsigned char)*s);
```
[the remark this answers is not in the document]
Swatinem: I don’t remember, and I can’t reproduce it either, so I just reverted this.

src/sentry_value.c
jan-auer: Usually, there are attributes you can put on the function to declare this behavior. Since we encode the pointers in a strange way, it's understandable that tools don't follow the path. 

Let's skip this for now.

PR #305, opened 2020-06-17
test: Restructure tests to deduplicate cmake runs

Swatinem:


tests/cmake.py
```diff
@@ -0,0 +1,102 @@
...
+
+class CMake:
+    def __init__(self, factory):
+        self.runs = dict()
+        self.factory = factory
+
+    def compile(self, targets, options=None):
+        key = (
+            ";".join(targets),
+            ";".join(["{}={}".format(k, options[k]) for k in options]),
```
flub, line 14: Pretty sure `options` can be `None` here and that can not be iteratred.

flub: I think `options` is meant to be a `**kwargs` passed through, so is a dictionary.  Much better to write `";".join("{}={}".format(k, v) for k, v in options.items())`.  (Also note you don't need to wrap into an extra list, you can directly use a generator expression instead of a list comprehension first.  Further you could try and use f-strings - but I have no idea if we depends on recent enough python versions for that here.  It might be worth it)

tests/cmake.py
```diff
@@ -0,0 +1,102 @@
...
+        # MP = object level parallelism, WX = warnings as errors
+        cpus = os.cpu_count()
+        env["CFLAGS"] = env["CXXFLAGS"] = "/WX /MP{}".format(cpus)
+    if "gcc" in os.environ.get("RUN_ANALYZER", ""):
+        env["CFLAGS"] = env["CXXFLAGS"] = "{} -fanalyzer".format(env["CFLAGS"])
+
+    print("\n{} > {}".format(cwd, " ".join(configcmd)), flush=True)
+    try:
+        subprocess.run(configcmd, cwd=cwd, env=env, check=True)
+    except CalledProcessError:
```
flub: doesn't this need to be `subprocess.CalledProcessError`?  Python doesn't actually try to evaluate that thing until you actually trigger the error.  So it could be that this succeeds for you because you've not triggered the exception case.

tests/conftest.py
```diff
@@ -2,6 +2,7 @@
 import pytest
 import re
 from . import cmake, run
+from .cmake import CMake
```
flub, line 5: low|nit|whatever: I'm fairly sure pep8 style says something akin to "import modules, not things within them"

tests/conftest.py
```diff
@@ -20,23 +21,11 @@ def pytest_generate_tests(metafunc):
...
+    cmake = CMake(tmp_path_factory)
 
+    def compile(*args, **kwargs):
+        return cmake.compile(*args, **kwargs)
 
-@pytest.fixture(scope="session")
-def unittests(tmp_path_factory):
-    tmpdir = tmp_path_factory.mktemp("unittests")
-    return Unittests(tmpdir)
+    return compile
```
flub, line 31: You could just `return cmake.compile` and not create the intermediate function?

tests/cmake.py
Swatinem: haha yes, linux tests actually ran into this one.

tests/conftest.py
Swatinem: does that automatically curry? hm, I will try locally ;-)

tests/cmake.py
```diff
@@ -8,10 +8,10 @@ def __init__(self, factory):
         self.runs = dict()
         self.factory = factory
 
-    def compile(self, targets, options=None):
+    def compile(self, targets, options={}):
```
flub, line 11: Shouldn't do this either.  Using `None` in the signature was correct.  Problem is that you now have a mutable object attached to the keyword argument, this is evaluated at import time and any modifications to this mutable object is seen my all calls to this function.
The classical, boilerplaty, pattern is to use `None` in the signature and then have:
```
if options is None:
    options = {}
```

flub: Also, totally minor nit, but since we have both set and dict literals I prefer to use the init functions for them when creating an empty one as that's much clearer: `options = dict()`

PR #310, opened 2020-06-19
fix: Avoid invalid memory access in the logger

### LINKED ISSUE #309: `set_debug(true)` crashes
Apparently simply doing this:
```cpp
```

Crashes the application. This doesn't happen in 0.3.2, so maybe it has something to do with #302?

Swatinem:
This also turns on the logger for the HTTP-based tests.

Fixes #309


daxpedda:
Just tried it, works perfectly 👌.

Swatinem:
> Are there still tests running without logs? We need to ensure that both paths are properly tested.

we test both with the standard http transport, with logs, and we also test with a specific stdout transport without logs, since I think that interferes with the way we capture the output, although stderr should be separate.

PR #313, opened 2020-06-22
meta: Add Contribution guide and issue/pr templates

Swatinem:
* Creates a contribution guide, which documents the testing infrastructure and options.
* Adds issue/PR templates, which ask for relevant versions/options, and instruct contributors to run tests.
* Configures CI to not run for every push ;-)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,11 +15,12 @@ Building and testing `sentry-native` currently requires the following tools:
 
 `pytest` and `black` are installed as virtualenv dependencies automatically.
 
-## Setting up Git Hooks
+## Setting up Environment
 
     $ make setup
 
-Installs a pre-commit hook, and initializes/updates the necessary git submodules.
+This sets up both git, including a pre-commit hook and submodules, and installs
+a python virtualenv which is used to run tests and formatting.
 
 ## Formatting Code
 
@@ -34,11 +35,7 @@ be done manually.
 
     $ make test
 
-Creates a virtualenv with python, and runs all the tests through `pytest`.
-
-The Makefile currently assumes python 3.7, but this can be overridden manually:
-
-    $ make SENTRY_NATIVE_PYTHON_VERSION=python3.8 setup-venv
+Creates a python virtualenv, and runs all the tests through `pytest`.
 
 **Running integration-tests manually**:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,123 @@
+# Contribution guidelines
+
+We love and welcome contributions!
+
+In order to maintain a high quality, we run a great number of checks on
+different OS / Compiler combinations, and using different analysis tools.
+
+## Prerequisites
+
+Building and testing `sentry-native` currently requires the following tools:
+
+- **CMake** and a supported C/C++ compiler, to actually build the code.
+- **python** and **pytest**, to run integration tests.
+- **clang-format** and **black**, to format the C/C++ and python code respectively.
+
+## Setting up Git Hooks
+
+    $ make setup
+
+Installs a pre-commit hook, and initializes/updates the necessary git submodules.
+
+## Formatting Code
+
+    $ make format
+
+This should be done automatically as part of the pre-commit hook, but can also
+done manually.
+
+    $ black tests
+
+## Running Tests
+
+    $ make test
+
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+- `ERROR_ON_WARNINGS`: Turns on `-Werror` for gcc compatible compilers.
+  This is also the default for `MSVC` on windows.
+- `RUN_ANALYZER`: Runs the code with/through one or more of the given analyzers.
+  This accepts a comma-separated list, and currently has support for:
+  - `asan`: Uses clangs AddressSanitizer and runs integration tests with the
+    `detect_leaks` flag.
+  - `scan-build`: Runs the build through the `scan-build` tool.
+  - `code-checker`: Uses the [`CodeChecker`](https://github.com/Ericsson/codechecker)
+    tool for builds.
+  - `kcov`: Uses [`kcov`](https://github.com/SimonKagstrom/kcov) to collect
```
Swatinem, line 86: This does not *yet* exist, but lets just put it in the docs already ;-)

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+their own.
+
+## How to interpret CI failures
+
+The way that tests are run unfortunately does not make immediately obvious from
+the summary what the actual failure is, especially for compile-time failures.
+In such cases, it is good to scan the test output _from top to bottom_ and find
+the offending compile error.
+
+When running tests locally, one can use the `--maxfail=1` / `-x` parameter to
```
flub, line 67: `maxfail=1 -x` parameters

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+In order to maintain a high quality, we run a great number of checks on
+different OS / Compiler combinations, and using different analysis tools.
+
+## Prerequisites
+
+Building and testing `sentry-native` currently requires the following tools:
+
+- **CMake** and a supported C/C++ compiler, to actually build the code.
+- **python** and **pytest**, to run integration tests.
+- **clang-format** and **black**, to format the C/C++ and python code respectively.
```
jan-auer, line 14: `black` and `pytest` should be installed in the virtualenv. Do we need to call them out here?

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+Installs a pre-commit hook, and initializes/updates the necessary git submodules.
+
+## Formatting Code
+
+    $ make format
+
+This should be done automatically as part of the pre-commit hook, but can also
+done manually.
+
+    $ black tests
```
jan-auer, line 29: That should be `make format-python`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+    $ cmake -B build -D CMAKE_RUNTIME_OUTPUT_DIRECTORY=$(pwd)/build
+    $ cmake --build build --target sentry_test_unit
+    $ ./build/sentry_test_unit
+
+The unit-tests are a separate executable target and can be built and run on
+their own.
+
+## How to interpret CI failures
+
+The way that tests are run unfortunately does not make immediately obvious from
```
jan-auer, line 62:
```suggestion
The way that tests are run unfortunately does not make it immediately obvious from
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+
+    $ cmake -B build -D CMAKE_RUNTIME_OUTPUT_DIRECTORY=$(pwd)/build
+    $ cmake --build build --target sentry_example
+    $ ./build/sentry_example log capture-event
+
+The example can be run manually with a variety of flags to test different
+scenarios. Additionally, it will use the `SENTRY_DSN` env-variable, and can thus
+also be used to capture events/crashes directly to sentry.
+
+The example currently supports the following flags:
```
jan-auer, line 107: Maybe clarify to pass those flags without a `--`, as this is what I expected in the past. Maybe, call them "commands" instead?

CONTRIBUTING.md
Swatinem: We don’t have separate format targets. Also since I still haven’t managed to properly set up make/bash on windows, I tend to invoke `black` manually there. I also invoke pytest there manually. So I would like to give the commands without using any make/bash specifics.

PR #317, opened 2020-06-24
test: Add more testcases for previously untested code

Swatinem:
# Tests added:

* Object cloning and overwriting keys.
* Serializing more msgpack values.
* Forcing captured events to have a uuid event_id.
* JSON escapes and surrogate parsing, also fixing a bug for that.
* Discarding events when missing consent.
* Test and fix recording errors on current session.
* Sessions with username did.
* Lots of abnormal sessions on startup.

# Bugs fixed:

* Wrong JSON decoding with surrogate pairs.
* Memory Leak parsing unmatched surrogate pairs.
* Never recorded error counts on sessions.


Swatinem:
The failing test (memory leak) is totally expected and is fixed in https://github.com/getsentry/sentry-native/pull/304

examples/example.c
```diff
@@ -149,6 +149,24 @@ main(int argc, char **argv)
...
+    if (has_arg(argc, argv, "capture-exception")) {
+        // TODO: Create a convenience API to create a new exception object,
+        // and to attach a stacktrace to the exception.
+        // See also https://github.com/getsentry/sentry-native/issues/235
+        sentry_value_t event = sentry_value_new_event();
+        sentry_value_t exception = sentry_value_new_object();
+        sentry_value_set_by_key(
+            exception, "type", sentry_value_new_string("ExceptionType"));
+        sentry_value_set_by_key(exception, "value",
+            sentry_value_new_string("Some exception message"));
```
jan-auer, line 161: _nit_: Can we make this a little more "realistic"?

tests/unit/test_value.c
```diff
@@ -248,7 +253,29 @@ SENTRY_TEST(value_json_parsing)
...
+        "formfeed: \\f, linefeed: \\n, carriage: \\r, tab: \\t\","
+        "\"surrogates\":\"𐐷\"}");
+    sentry_value_decref(rv);
+
+    // unmatched surrogates don’t parse
+    rv = sentry__value_from_json(STRING("\"\\uD801\""));
+    TEST_CHECK(sentry_value_is_null(rv));
+    rv = sentry__value_from_json(
+        STRING("{\"valid key\": true, \"invalid key \\uD801\": false}"));
+    TEST_CHECK_JSON_VALUE(rv, "{\"valid key\":true}");
```
jan-auer, line 279: _nit_: I'd prefer to have these in separate tests with a low number of checks. If one of them fails, it's easier to know what regressed (independently).

PR #323, opened 2020-06-29
fix: Avoid assertion when hitting shutdown_timeout

Swatinem:
This reworks the task locking and background worker shutdown logic.
Tasks are now refcounted so they can be _executed_ concurrently while another thread removes them from the queue. In case the background worker can’t be cleanly shut down (due to slow network), the task queue is being dumped, but the background worker itself is not freed from the main thread, to avoid potential use-after-free bugs.
Also, the background worker itself is refcounted and owns its state.

The transport state is now owned by the background worker, and passed explicitly to the task send function. The task state itself was removed in favor of providing the envelope directly.


jan-auer:
Notes from an offline discussion: When the crash happens close to startup, it is more likely that it interrupts an ongoing upload. In this case, it would be better to avoid dropping it. 

If we refcount the envelopes in the transport queue, we could have multiple threads read and clean the queue concurrently.

src/transports/sentry_transport_curl.c
```diff
@@ -17,8 +17,8 @@ typedef struct curl_transport_state_s {
     bool initialized;
     bool debug;
     CURL *curl_handle;
-    const char *http_proxy;
-    const char *ca_certs;
+    char *http_proxy;
+    char *ca_certs;
```
jan-auer, line 21: All these should be owned by the background task and freed when the task shuts down. As such, how about we move them to their own `sentry_curl_bgw_s` struct and pass it as `data`/`state` parameter into `sentry__bgworker_new()`?

src/sentry_sync.c
```diff
@@ -148,31 +182,41 @@ sentry__bgworker_shutdown(sentry_bgworker_t *bgw, uint64_t timeout)
...
+    sentry__mutex_lock(&bgw->task_lock);
+    sentry_bgworker_task_t *task = bgw->first_task;
     while (task) {
-        struct sentry_bgworker_task_s *next_task = task->next_task;
-        if (task->cleanup_func) {
-            task->cleanup_func(task->data);
-        }
-        sentry_free(task);
+        sentry_bgworker_task_t *next_task = task->next_task;
+        sentry__task_decref(task);
```
jan-auer, line 203: _nit_: This loop can be moved out of the locked region. Probably irrelevant, since free can be lock-free once we implement refcounting.

src/sentry_sync.c
```diff
@@ -91,23 +124,24 @@ worker_thread(void *data)
...
             sentry__mutex_unlock(&bgw->task_lock);
-
-            if (task->cleanup_func) {
-                task->cleanup_func(task->data);
-            }
-            sentry_free(task);
         }
     }
+    if (bgw->free_self) {
+        sentry_free(bgw);
```
jan-auer, line 143: Considering that the thread continues to run and send envelopes while the owning thread already timed out, this looks like the right place to free the background worker. However, this should probably free unconditionally, essentially calling a `decref` on the state.

src/sentry_sync.c
```diff
@@ -21,10 +33,11 @@ struct sentry_bgworker_s {
     sentry_mutex_t done_signal_lock;
     sentry_mutex_t task_lock;
     sentry_threadid_t thread_id;
-    struct sentry_bgworker_task_s *first_task;
-    struct sentry_bgworker_task_s *last_task;
+    sentry_bgworker_task_t *first_task;
+    sentry_bgworker_task_t *last_task;
     size_t task_count;
     bool running;
+    bool free_self;
```
jan-auer, line 40: The entire bgworker state needs to be owned by both the main thread (where `sentry_init` and `sentry_shutdown` is run), and the background worker. Thus, we need to employ refcounting.

Once we release both, it can clean up, which replaces `sentry__bgworker_free` with `sentry__bgworker__decref`. The nice side effect of this is that `decref` no longer needs to lock :)

src/sentry_sync.c
```diff
@@ -41,78 +35,95 @@ typedef struct sentry_bgworker_task_s {
...
+    sentry_cond_t task_signal;
+    sentry_mutex_t task_lock;
     sentry_bgworker_task_t *first_task;
     sentry_bgworker_task_t *last_task;
-    size_t task_count;
     void *state;
     void (*free_state)(void *state);
     long refcount;
-    bool running;
+    long running;
```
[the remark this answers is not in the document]
jan-auer: I did some more digging, and apparently it is discouraged. In this case, I'm thinking that `volatile bool` should be sufficient to toggle a flag in another thread *eventually* (that is, without atomicity constraints), but since it is discouraged, let's maybe stick with your atomic long. WDYT?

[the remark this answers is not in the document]
jan-auer: The comments on [this](https://stackoverflow.com/a/4558024/4228225) stackoverflow answer pretty much capture it. Funny enough, the figure that our case is the one where `volatile` would make sense, but `atomic` is the better choice. Given that we have atomic in now, let's keep it, even though it looks more complicated.

PR #324, opened 2020-06-30
ci: Run a linux build through valgrind

Swatinem:


src/path/sentry_path_unix.c
```diff
@@ -156,23 +156,19 @@ sentry_path_t *
...
     if (!path) {
         return NULL;
     }
-    rv = sentry__path_from_str_owned(path);
-    if (rv) {
-        return rv;
-    }
-    sentry_free(path);
-    return NULL;
+    // NOTE: function will free `path` an error
```
flub, line 162: That doesn't really read like english... "function takes ownership of `path`, meaning future `free(path)` calls would be an error" is what is meant?

tests/__init__.py
```diff
@@ -76,7 +76,18 @@ def run(cwd, exe, args, env=dict(os.environ), **kwargs):
             coverage_dir,
             *cmd,
         ]
-    return subprocess.run([*cmd, *args], cwd=cwd, env=env, **kwargs)
+    if "valgrind" in os.environ.get("RUN_ANALYZER", ""):
+        cmd = ["valgrind", "--leak-check=yes", *cmd]
+    try:
+        failed = False
+        return subprocess.run([*cmd, *args], cwd=cwd, env=env, **kwargs)
+    except subprocess.CalledProcessError:
```
flub, line 84: Personally I frown on this because the control flow becomes really hard and it's not really obvious what happens in all cases.  E.g. here the `if failed` statement below is rather abiguous.  I think this is better expressed as
```
try:
    ret = subprocess.run(....)
except subprocess.CalledProcessError:
    ....
    pytest.fail(...)
else:
    return ret
```

tests/__init__.py
```diff
@@ -76,7 +76,18 @@ def run(cwd, exe, args, env=dict(os.environ), **kwargs):
...
+        cmd = " ".join(cmd)
+        args = " ".join(args)
+        cmd = f"{cmd} {args}"
```
flub, line 87: I get that f-strings are cute.  But `pytest.fail("running command failed: {cmd} {args}".format(cmd=" ".join(cmd), args=" ".join(args))` is perhaps as or more readable?

tests/cmake.py
```diff
@@ -102,9 +103,23 @@ def cmake(cwd, targets, options=None):
...
+            "scan-build",
+            *(["--use-cc", cc] if cc else []),
+            *(["--use-c++", cxx] if cxx else []),
+            "--status-bugs",
+            "--exclude",
+            os.path.join(source_dir, "external"),
+            "cmake",
+        ]
+
     configcmd = [*cmake]
```
flub, line 123: `cmake` here is always a list, why the weird `[*cmake]` syntax instead of directly `configcmd = cmake`?

tests/cmake.py
```diff
@@ -136,19 +150,80 @@ def cmake(cwd, targets, options=None):
 
     print("\n{} > {}".format(cwd, " ".join(configcmd)), flush=True)
     try:
+        failed = False
         subprocess.run(configcmd, cwd=cwd, env=env, check=True)
     except subprocess.CalledProcessError:
+        failed = True
+    if failed:
         pytest.fail("cmake configure failed")
```
flub, line 158: I'm not sure what the reasoning behind this `failed = False` pattern is?  Am I missing a reason why you can't do
```
try:
    subprocess.run(...)
except subprocess.CalledProcessError:
    pytest.fail(...)
```
?  You're clearly just adding this, so there must be a reason.  To me it looks very weird though.

tests/cmake.py
```diff
@@ -136,19 +150,80 @@ def cmake(cwd, targets, options=None):
...
+        ]
+
     print("{} > {}".format(cwd, " ".join(buildcmd)), flush=True)
     try:
+        failed = False
         subprocess.run(buildcmd, cwd=cwd, check=True)
     except subprocess.CalledProcessError:
+        failed = True
+    if failed:
         pytest.fail("cmake build failed")
```
flub, line 183: same

src/path/sentry_path_unix.c
Swatinem: `on error` ;-) Meaning yes, it will take ownership and free if necessary.

tests/__init__.py
Swatinem: I should have added a comment about this. This is basically to work around pythons automatic error chaining. The backtrace is horribly unreadable as it is, and this tries to clean it up a bit.

tests/__init__.py
Swatinem: I’m actually quite confused about f-strings. According to the docs, they are supposed to support expressions, but in reality that is not the case?

tests/cmake.py
Swatinem: I use it another time below, I wasn’t sure if `configcmd = cmake` just creates a reference (that I mutate) or actually creates a copy. I basically force it to be a copy, so I can mutate without having problems later.

tests/__init__.py
flub: https://pyformat.info/ is a reasonable good reference for what you can do.  It's only some basic item and attribute access, not full python expressions.

tests/cmake.py
flub: Oh, yes this would be just a reference.  To force a copy `configcmd = cmake.copy()` is much clearer and more pythonic imho.

PR #330, opened 2020-07-03
ref: Remove name parameter from sentry_options_add_attachment

Swatinem:
This is now unused with the latest crashpad changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 **Breaking Changes**:
 
 - The `sentry_options_set_logger` function now accepts a `userdata` parameter.
+- The `name` parameter of `sentry_options_add_attachment(w)` was removed, it will
+  now be inferred from the filename of `path`.
 
 **Features**:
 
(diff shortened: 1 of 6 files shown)


tests/test_integration_crashpad.py
```diff
@@ -2,19 +2,14 @@
...
-#   - breadcrumbs, attachments, etc
-#   - crash
-#   - expect report via http
+# Actually assert that we get a correct event/breadcrumbs payload
 
 
-@pytest.mark.skipif(
-    not has_http or not has_crashpad, reason="test needs crashpad backend"
-)
+@pytest.mark.skipif(not has_crashpad, reason="test needs crashpad backend")
```
flub, line 12: FYI, I think you only use this twice but if you need it more you can:
```
needs_crashpad = pytest.mark.skipif(not has_crashpad, reason="test needs crashpad backend")

@needs_crashpad
def test_crashpad_capture(...):
    ....
```

Swatinem: Oh, that is nice. We do have quite a few repetitive `skipif` in the tests.

PR #333, opened 2020-07-08
fix: Lock global options whenever they are accessed

### LINKED ISSUE #280: `sentry_shutdown` can break things in a multi-threaded environment.
I discovered this while trying to call `sentry_init`, `sentry_shutdown` and `sentry_capture_event` multiple times from multiple threads at the same time during unit tests, not a valid use-case scenario of this library.

But I can imagine a valid use case to be when `sentry_shutdown` is called while other threads are still busy using `sentry_capture_event`.

---

Basically what happens is that functions like `sentry_capture_event` do this:
https://github.com/getsentry/sentry-native/blob/9651561d45e4d00e9fe708275c086a3cfeb496bd/src/sentry_core.c#L257
They check if it's valid here:
https://github.com/getsentry/sentry-native/blob/9651561d45e4d00e9fe708275c086a3cfeb496bd/src/sentry_core.c#L258
And then use it multiple times here:
https://github.com/getsentry/sentry-native/blob/9651561d45e4d00e9fe708275c086a3cfeb496bd/src/sentry_core.c#L282-L285

The issue arises when on a different thread `sentry_shutdown` is called which invalidates `g_options` and makes access to it access freed memory.
So the missing part here to me seems a missing lock around `g_options_mutex`.

---

Another example is `sentry_start_session`:
https://github.com/getsentry/sentry-native/blob/9651561d45e4d00e9fe708275c086a3cfeb496bd/src/sentry_session.c#L206-L214
So `SENTRY_WITH_SCOPE_MUT` locks `g_lock`:
https://github.com/getsentry/sentry-native/blob/9651561d45e4d00e9fe708275c086a3cfeb496bd/src/sentry_scope.c#L85-L90
But `sentry_shutdown` locks a different mutex, `g_options_mutex`:
https://github.com/getsentry/sentry-native/blob/9651561d45e4d00e9fe708275c086a3cfeb496bd/src/sentry_core.c#L142-L145
Which doesn't block `sentry_start_session` from using already freed memory:
https://github.com/getsentry/sentry-native/blob/9651561d45e4d00e9fe708275c086a3cfeb496bd/src/sentry_session.c#L54 https://github.com/getsentry/sentry-native/blob/9651561d45e4d00e9fe708275c086a3cfeb496bd/src/sentry_session.c#L64-L65

---

There are more offenders similar to those.
Is there an intention to account for this scenario?

Swatinem:
Wraps all accesses of the global objects object in a lock. This also changes the DSN to be a heap allocated and refcounted type, since the http workers need to separately own a reference.

fixes #280

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -595,12 +595,6 @@ SENTRY_API void sentry_transport_free(sentry_transport_t *transport);
 SENTRY_API sentry_transport_t *sentry_new_function_transport(
     void (*func)(const sentry_envelope_t *envelope, void *data), void *data);
 
-/**
- * Type of the callback for modifying events.
- */
-typedef sentry_value_t (*sentry_event_function_t)(
-    sentry_value_t event, void *hint, void *closure);
-
 /* -- Options APIs -- */
 
 /**
@@ -630,10 +624,22 @@ SENTRY_API void sentry_options_set_transport(
     sentry_options_t *opts, sentry_transport_t *transport);
 
 /**
+ * Type of the `before_send` callback.
+ *
+ * The callback taken ownership of the `event`, and will usually return that
+ * same event. In the case the event should be discarded, the callback needs to
+ * call `sentry_value_decref` on the provided event, and return a
+ * `sentry_value_new_null` instead.
+ * This function may be invoked inside of a signal handler and must be safe for
+ * that purpose.
+ */
+typedef sentry_value_t (*sentry_event_function_t)(
+    sentry_value_t event, void *hint, void *closure);
+
+/**
  * Sets the `before_send` callback.
  *
- * The `before_send` hook may be invoked inside of a signal handler, and must
- * therefore be signal safe.
+ * See the `sentry_event_function_t` typedef above for more information.
  */
 SENTRY_API void sentry_options_set_before_send(
     sentry_options_t *opts, sentry_event_function_t func, void *data);
@@ -904,15 +910,6 @@ SENTRY_API int sentry_shutdown(void);
  * `sentry_capture_event` will have an up-to-date module list.
  */
 SENTRY_EXPERIMENTAL_API void sentry_clear_modulecache(void);
-
-/**
- * Returns the currently active client options.
- *
- * This might return NULL if sentry is not yet initialized.
- * The returned options pointer is not synchronized, and calling
- * `sentry_shutdown` concurrently can lead to use-after-free situations.
- */
-SENTRY_API const sentry_options_t *sentry_get_options(void);
 
 /**
  * Gives user consent.
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -626,12 +626,12 @@ SENTRY_API void sentry_options_set_transport(
 /**
  * Type of the `before_send` callback.
  *
- * The callback taken ownership of the `event`, and will usually return that
- * same event. In the case the event should be discarded, the callback needs to
+ * The callback takes ownership of the `event`, and should usually return that
+ * same event. In case the event should be discarded, the callback needs to
  * call `sentry_value_decref` on the provided event, and return a
- * `sentry_value_new_null` instead.
+ * `sentry_value_new_null()` instead.
  * This function may be invoked inside of a signal handler and must be safe for
- * that purpose.
+ * that purpose, see https://man7.org/linux/man-pages/man7/signal-safety.7.html.
  */
 typedef sentry_value_t (*sentry_event_function_t)(
     sentry_value_t event, void *hint, void *closure);
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -906,9 +906,11 @@ SENTRY_API int sentry_shutdown(void);
 SENTRY_EXPERIMENTAL_API void sentry_clear_modulecache(void);
 
 /**
- * Returns the client options.
+ * Returns the currently active client options.
  *
  * This might return NULL if sentry is not yet initialized.
+ * The returned options pointer is not synchronized, and calling
+ * `sentry_shutdown` concurrently can lead to use-after-free situations.
  */
 SENTRY_API const sentry_options_t *sentry_get_options(void);
 
--- a/src/sentry_core.c
+++ b/src/sentry_core.c
@@ -61,7 +61,6 @@ sentry__should_skip_upload(void)
     SENTRY_WITH_OPTIONS (options) {
         skip = options->require_user_consent
             && sentry__atomic_fetch((long *)&options->user_consent)
-
                 != SENTRY_USER_CONSENT_GIVEN;
     }
     return skip;
@@ -212,7 +211,11 @@ static void
 set_user_consent(sentry_user_consent_t new_val)
 {
     SENTRY_WITH_OPTIONS (options) {
-        sentry__atomic_store((long *)&options->user_consent, new_val);
+        if (sentry__atomic_store((long *)&options->user_consent, new_val)
+            == new_val) {
+            // nothing was changed
+            break; // SENTRY_WITH_OPTIONS
+        }
 
         if (options->backend && options->backend->user_consent_changed_func) {
             options->backend->user_consent_changed_func(options->backend);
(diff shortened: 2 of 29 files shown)


src/sentry_utils.c
```diff
@@ -240,40 +243,50 @@ sentry__dsn_parse(sentry_dsn_t *dsn_out, const char *dsn)
...
+        sentry_free(dsn->secret_key);
+        sentry_free(dsn);
+    }
 }
 
 char *
 sentry__dsn_get_auth_header(const sentry_dsn_t *dsn)
 {
-    if (!dsn || dsn->empty) {
+    if (!dsn) {
```
Swatinem, line 289: re-add the `is_valid` checks here

tests/test_integration_http.py
```diff
@@ -33,19 +33,17 @@ def test_capture_http(cmake, httpserver):
     httpserver.expect_oneshot_request(
         "/api/123456/envelope/", headers={"x-sentry-auth": auth_header},
     ).respond_with_data("OK")
+    env = dict(os.environ, SENTRY_DSN=make_dsn(httpserver), SENTRY_RELEASE="🤮🚀")
 
-    with httpserver.wait(raise_assertions=True, stop_on_nohandler=True) as waiting:
```
flub, line 37: Is removing the waiting here safe?

Swatinem: Yes, I actually had problems with the wait before. When the subprocess exits, things are guaranteed to be sent.

src/sentry_core.c
```diff
@@ -125,11 +150,10 @@ sentry_shutdown(void)
...
 
-    sentry__mutex_lock(&g_options_mutex);
-    sentry_options_t *options = g_options;
-    sentry__mutex_unlock(&g_options_mutex);
-
-    if (options) {
+    SENTRY_WITH_OPTIONS (options) {
+        // we unset the global here, to try to minimize the chance of errors
+        // with unsynchronized concurrent accesses
+        g_options = NULL;
```
jan-auer, line 156: This lock scope can live a bit shorter. After setting `g_options` to NULL, remember the old reference and the exit the scope.

src/sentry_core.c
```diff
@@ -163,39 +184,29 @@ sentry_clear_modulecache(void)
...
-    case SENTRY_USER_CONSENT_REVOKED:
-        sentry__path_write_buffer(consent_path, "0\n", 2);
-        break;
-    case SENTRY_USER_CONSENT_UNKNOWN:
-        sentry__path_remove(consent_path);
-        break;
-    }
-    sentry__path_free(consent_path);
+    SENTRY_WITH_OPTIONS (options) {
+        options->user_consent = new_val;
```
jan-auer, line 191: Let's also shorten this lock a bit. For instance, run the callback first, and then outside of the scope write the consent file.

src/sentry_core.c
```diff
@@ -323,34 +338,36 @@ sentry_capture_event(sentry_value_t event)
...
+        if (options->backend && options->backend->except_func) {
+            options->backend->except_func(options->backend, uctx);
+        }
     }
 }
 
-void
-sentry__enforce_disk_transport(void)
+sentry_transport_t *
+sentry__enforce_disk_transport(sentry_options_t *options)
```
jan-auer, line 363: _nit_: 
```suggestion
sentry__swap_disk_transport(sentry_options_t *options)
```

src/sentry_core.c
```diff
@@ -264,57 +278,58 @@ event_is_considered_error(sentry_value_t event)
...
 
-    SENTRY_WITH_SCOPE (scope) {
-        SENTRY_TRACE("merging scope into event");
-        sentry_scope_mode_t mode = SENTRY_SCOPE_ALL;
-        if (!opts->symbolize_stacktraces) {
-            mode &= ~SENTRY_SCOPE_STACKTRACES;
+        SENTRY_WITH_SCOPE (scope) {
+            SENTRY_TRACE("merging scope into event");
+            sentry_scope_mode_t mode = SENTRY_SCOPE_ALL;
+            if (!options->symbolize_stacktraces) {
```
jan-auer, line 296: I'm afraid that we're locking too long in some of these functions. Particularly symbolicating the stack trace, or some of the backend functions can take considerable time. Sometimes, they even lock internally, again.

In a concurrent application, this could lead to significant delays. Would it be worth investigating alternative approaches here, for instance immutable options?

src/transports/sentry_transport_curl.c
```diff
@@ -65,9 +67,10 @@ sentry__curl_transport_start(
     sentry_bgworker_t *bgworker = (sentry_bgworker_t *)transport_state;
     curl_bgworker_state_t *state = sentry__bgworker_get_state(bgworker);
 
-    state->debug = options->debug;
+    state->dsn = options->dsn ? sentry__dsn_incref(options->dsn) : NULL;
```
jan-auer, line 70: _nit_: The check for NULL could also happen in `incref` for convenience.

Swatinem:
After discussing this a little bit, I completely reworked this again.
Instead of holding the lock for a long period of time, the locked section is now super small, just loading the global options and incrementing their refcount.
the `SENTRY_WITH_OPTIONS` macro now does that and decrements the refcount afterwards.

The only mutable access to options now is the user consent, which was changed to an atomic.

Instead of swapping out transports, capturing an event was split up into preparing the event, and then capturing it with an explicit transport. This also gets rid of the special case breakpad transport because the minidump can now be attached directly to an envelope instead of making roundtrips via that special transport.

While at it, inproc/breakpad now only send a single envelope on crash, with the session in it, which also now has an error count that covers the crash itself.

include/sentry.h
```diff
@@ -629,11 +623,23 @@ SENTRY_API void sentry_options_free(sentry_options_t *opts);
 SENTRY_API void sentry_options_set_transport(
     sentry_options_t *opts, sentry_transport_t *transport);
 
+/**
+ * Type of the `before_send` callback.
+ *
+ * The callback taken ownership of the `event`, and will usually return that
```
jan-auer, line 629:
```suggestion
 * The callback takes ownership of the `event`, and should usually return that
```

include/sentry.h
```diff
@@ -629,11 +623,23 @@ SENTRY_API void sentry_options_free(sentry_options_t *opts);
...
+ * same event. In the case the event should be discarded, the callback needs to
```
jan-auer, line 630:
```suggestion
 * same event. In case the event should be discarded, the callback needs to
```

include/sentry.h
```diff
@@ -629,11 +623,23 @@ SENTRY_API void sentry_options_free(sentry_options_t *opts);
...
+ * call `sentry_value_decref` on the provided event, and return a
+ * `sentry_value_new_null` instead.
```
jan-auer, line 632:
```suggestion
 * `sentry_value_new_null()` instead.

```

include/sentry.h
```diff
@@ -629,11 +623,23 @@ SENTRY_API void sentry_options_free(sentry_options_t *opts);
...
+ * This function may be invoked inside of a signal handler and must be safe for
+ * that purpose.
```
jan-auer, line 634: Maybe link to https://man7.org/linux/man-pages/man7/signal-safety.7.html ?

src/sentry_core.c
```diff
@@ -192,39 +208,30 @@ sentry_clear_modulecache(void)
...
-    case SENTRY_USER_CONSENT_REVOKED:
-        sentry__path_write_buffer(consent_path, "0\n", 2);
-        break;
-    case SENTRY_USER_CONSENT_UNKNOWN:
-        sentry__path_remove(consent_path);
-        break;
-    }
-    sentry__path_free(consent_path);
+    SENTRY_WITH_OPTIONS (options) {
+        sentry__atomic_store((long *)&options->user_consent, new_val);
```
jan-auer, line 215: _nit_: You could compare the return value with `new_val` and then skip everything below if it didn't change.

```suggestion
        prev_val = sentry__atomic_store((long *)&options->user_consent, new_val);
```

PR #335, opened 2020-07-14
fix: Flush session and send queue in crashpad backend

Swatinem:
This uses crashpads `SetFirstChanceExceptionHandler` on Linux,
and an additional exception filter on Windows, and a heuristic on macOS.

### DIFF
--- a/src/sentry_database.c
+++ b/src/sentry_database.c
@@ -136,7 +136,7 @@ sentry__run_clear_session(const sentry_run_t *run)
 }
 
 void
-sentry__process_old_runs(const sentry_options_t *options)
+sentry__process_old_runs(const sentry_options_t *options, uint64_t last_crash)
 {
     sentry_pathiter_t *db_iter
         = sentry__path_iter_directory(options->database_path);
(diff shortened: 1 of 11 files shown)


tests/leaks.txt
```diff
@@ -0,0 +1,5 @@
+# Not really sure if this is a real leak or just ASAN not correctly handling the autorelease pool.
```
Swatinem, line 1: I think this might be a false positive after some fiddling around with this.

src/sentry_database.c
```diff
@@ -173,15 +173,35 @@ sentry__process_old_runs(const sentry_options_t *options)
...
+                }
                 sentry_session_t *session = sentry__session_from_path(file);
                 if (session) {
+                    // this is just a heuristic: whenever the session was not
+                    // closed properly, and we do have a crash that happened
+                    // *after* the session was started, we will assume that the
+                    // crash corresponds to the session and flag it as crashed.
+                    // this should only happen when using crashpad, and there
+                    // should normally be only a single unclosed session at a
+                    // time.
```
flub, line 187: Are we already in a new session when this is invoked?

tests/test_integration_crashpad.py
```diff
@@ -1,15 +1,22 @@
 import pytest
 import os
+import sys
 import time
 from . import make_dsn, run, Envelope
 from .conditions import has_crashpad
 from .assertions import assert_session
 
+if not has_crashpad:
+    pytest.skip("test needs crashpad backend", allow_module_level=True)
```
flub, line 10: Did you consider the module-level skipif mark instead as the docs suggest?

https://docs.pytest.org/en/stable/example/markers.html#marking-whole-classes-or-modules

Was that not suitable for some reason?

src/sentry_database.c
Swatinem: No, we auto-start the session as the last thing, after processing the old runs. Either way this should be completely independent.

PR #336, opened 2020-07-14
feat: Stabilize sessions and enable by default

Swatinem:
This also updates the changelog and documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,20 +7,29 @@
 - The `sentry_options_set_logger` function now accepts a `userdata` parameter.
 - The `name` parameter of `sentry_options_add_attachment(w)` was removed, it will now be inferred from the filename of `path`.
 
+```c
+// before
+sentry_options_set_logger(options, my_custom_logger);
+sentry_options_add_attachment(options, "some-attachment", "/path/to/some-attachment.txt");
+// after
+sentry_options_set_logger(options, my_custom_logger, NULL);
+sentry_options_add_attachment(options, "/path/to/some-attachment.txt");
+```
+
 **Features**:
 
-- Release Health (Sessions) is now stable, and automatic session handling is enabled by default. Use `sentry_options_set_auto_session_tracking` to change the behavior.
+- [Release Health](https://docs.sentry.io/workflow/releases/health/) support is now stable and enabled by default. After the update, you will see the number of crash free sessions and crash free users on the Releases page in Sentry. To disable automatic session tracking, use `sentry_options_set_auto_session_tracking`.
 - Breakpad support for Windows. This allows you to use `sentry-native` even on Windows XP! ([#278](https://github.com/getsentry/sentry-native/pull/278))
 - Add an in-process backend for Windows. As opposed to Breakpad, stack traces are generated on the device and sent to Sentry for symbolication. ([#287](https://github.com/getsentry/sentry-native/pull/287))
-- Support for the crashpad backend was fixed and enabled for Linux. ([#320](https://github.com/getsentry/sentry-native/pull/320))
+- Support for the Crashpad backend was fixed and enabled for Linux. ([#320](https://github.com/getsentry/sentry-native/pull/320))
 - A new `SENTRY_BREAKPAD_SYSTEM` CMake option was added to link to the system-installed breakpad client instead of building it as part of sentry.
 
 **Fixes**:
 
-- Reworked thread synchronization code and logic in `sentry_shutdown`, avoiding memory unsafety in case of an unclean shutdown. ([#323](https://github.com/getsentry/sentry-native/pull/323))
+- Reworked thread synchronization code and logic in `sentry_shutdown`, avoiding an abort in case of an unclean shutdown. ([#323](https://github.com/getsentry/sentry-native/pull/323))
 - Similarly, reworked options locking, avoiding thread safety issues. ([#333](https://github.com/getsentry/sentry-native/pull/333))
 - Fixed errors not being properly recorded in sessions. ([#317](https://github.com/getsentry/sentry-native/pull/317))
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -5,13 +5,22 @@
 **Breaking Changes**:
 
 - The `sentry_options_set_logger` function now accepts a `userdata` parameter.
-- The `name` parameter of `sentry_options_add_attachment(w)` was removed, it will
-  now be inferred from the filename of `path`.
+- The `name` parameter of `sentry_options_add_attachment(w)` was removed, it will now be inferred from the filename of `path`.
```
jan-auer, line 8: For the breaking changes, what if we add little code examples for the migration or new usage? I think this could help in upgrading, or do you feel that this would be too much?

CHANGELOG.md
```diff
@@ -5,13 +5,22 @@
...
 
 **Features**:
 
+- Release Health (Sessions) is now stable, and automatic session handling is enabled by default. Use `sentry_options_set_auto_session_tracking` to change the behavior.
```
jan-auer, line 12: This is a major new feature, and we could phrase this even more user-oriented and link to our Release Health docs at https://docs.sentry.io/workflow/releases/health/. How about:

```suggestion
- [Release Health](https://docs.sentry.io/workflow/releases/health/) support is now stable and enabled by default. After the update, you will see the number of crash free sessions and crash free users on the Releases page in Sentry. To customize sessions, use: <code block underneath> To disable automatic session tracking, use `sentry_options_set_auto_session_tracking`.
```

CHANGELOG.md
```diff
@@ -5,13 +5,22 @@
...
 - Breakpad support for Windows. This allows you to use `sentry-native` even on Windows XP! ([#278](https://github.com/getsentry/sentry-native/pull/278))
 - Add an in-process backend for Windows. As opposed to Breakpad, stack traces are generated on the device and sent to Sentry for symbolication. ([#287](https://github.com/getsentry/sentry-native/pull/287))
+- Support for the crashpad backend was fixed and enabled for Linux. ([#320](https://github.com/getsentry/sentry-native/pull/320))
```
jan-auer, line 15:
```suggestion
- Support for the Crashpad backend was fixed and enabled for Linux. ([#320](https://github.com/getsentry/sentry-native/pull/320))
```

CHANGELOG.md
```diff
@@ -5,13 +5,22 @@
...
+- A new `SENTRY_BREAKPAD_SYSTEM` CMake option was added to link to the system-installed breakpad client instead of building it as part of sentry.
+
+**Fixes**:
+
+- Reworked thread synchronization code and logic in `sentry_shutdown`, avoiding memory unsafety in case of an unclean shutdown. ([#323](https://github.com/getsentry/sentry-native/pull/323))
```
jan-auer, line 20:
```suggestion
- Reworked thread synchronization code and logic in `sentry_shutdown`, avoiding an abort in case of an unclean shutdown. ([#323](https://github.com/getsentry/sentry-native/pull/323))
```

CHANGELOG.md
```diff
@@ -5,13 +5,22 @@
...
+- Similarly, reworked options locking, avoiding thread safety issues. ([#333](https://github.com/getsentry/sentry-native/pull/333))
+- Fixed errors not being properly recorded in sessions. ([#317](https://github.com/getsentry/sentry-native/pull/317))
+- Enabled code coverage and static analyzers, and fixed potential issues that have been uncovered. ([#304](https://github.com/getsentry/sentry-native/pull/304) and others)
```
jan-auer, line 23: I would not mention coverage and static analyzers, just mention the fixes.

README.md
```diff
@@ -239,6 +239,10 @@ Legend:
 
 ## Known Issues
 
+- The crashpad backend currently has no support for notifying the crashing
+  process, and can thus not properly terminate sessions or call the registered
+  `before_send` hook. It will also lose any events that have been queued for
+  sending at time of crash.
 - Attachments are currently in _Preview_ and may not be available to your
   organization. Please see [Event Attachments] for more information.
```
jan-auer, line 247: We can remove the attachments disclaimer now. Also, what if we rename this section to "Known Limitations"?

include/sentry.h
```diff
@@ -732,6 +735,23 @@ typedef void (*sentry_logger_function_t)(
 SENTRY_API void sentry_options_set_logger(
     sentry_options_t *opts, sentry_logger_function_t func, void *userdata);
 
+/**
+ * Enables or disables automatic session tracking.
+ *
+ * Automatic session tracking is enabled by default and is equivalent to calling
+ * `sentry_start_session` after startup. A running session will always be closed
+ * implicitly by `sentry_shutdown` if not done manually by calling
+ * `sentry_end_session`.
```
jan-auer, line 744: Should we also explain that there can only be one concurrent session and that `sentry_start_session` closes the previous one?

src/sentry_options.c
```diff
@@ -32,6 +32,7 @@ sentry_options_new(void)
     opts->environment = sentry__string_clone(getenv("SENTRY_ENVIRONMENT"));
 #endif
     opts->user_consent = SENTRY_USER_CONSENT_UNKNOWN;
+    opts->auto_session_tracking = true;
```
jan-auer, line 35: Shouldn't this be false on Android, or would they just set `false` in the Android SDK wrapper?

Swatinem: good catch!

jan-auer: Thinking about this again, setting `false` in the wrapper is probably the best solution. That way, it's still enabled when running the native SDK alone.

Swatinem: I agree. Heads up @marandaneto

marandaneto: ok, we can disable on SDK init, thanks for the heads up

PR #341, opened 2020-07-16
fix: Initialize curl in the correct order

Swatinem:
This also checks more thoroughly for both curl and winhttp state.

Also reworks the transport startup to be fallible, and re-orders the startup path a little bit.


src/sentry_core.c
```diff
@@ -89,39 +90,61 @@ sentry_init(sentry_options_t *options)
...
+    SENTRY_WITH_SCOPE_MUT (scope) {
+        (void)scope;
+    }
+    if (backend && backend->user_consent_changed_func) {
+        backend->user_consent_changed_func(backend);
     }
 
     // after initializing the transport, we will submit all the unsent envelopes
     // and handle remaining sessions.
     sentry__process_old_runs(options);
```
jan-auer, line 133: The old order of execution was: _options_ -> _transport_ -> _old runs_ -> _backend_.
The new order of execution is: _transport_ -> _backend_ -> _options_ -> _old runs_.

Are we sure that it is fine to clean up old runs after initializing the backend?

src/sentry_core.c
```diff
@@ -89,39 +90,61 @@ sentry_init(sentry_options_t *options)
...
-        sentry__transport_startup(transport, options);
+    // *after* setting the global options, trigger a scope and consent flush,
+    // since at least crashpad needs that.
+    // the only way to get a reference to the scope is by locking it, the macro
+    // does all that at once, including invoking the backends scope flush hook
+    SENTRY_WITH_SCOPE_MUT (scope) {
+        (void)scope;
+    }
+    if (backend && backend->user_consent_changed_func) {
+        backend->user_consent_changed_func(backend);
```
jan-auer, line 128: Shouldn't we do this in the backend then? It seems like it is the responsibility of the backend to make sure that the initial configuration is applied.

src/sentry_core.c
```diff
@@ -89,39 +90,61 @@ sentry_init(sentry_options_t *options)
...
     }
 
     return 0;
+
+fail:
+    SENTRY_WARN("`sentry_init` failed");
+    if (transport) {
+        sentry__transport_shutdown(transport, 0);
+    }
+    sentry_options_free(options);
```
jan-auer, line 146: Should we also free the backend here if it is set?

src/sentry_core.c
```diff
@@ -134,22 +157,22 @@ sentry_shutdown(void)
...
         if (options->transport) {
             // TODO: make this configurable
-            clean_shutdown = sentry__transport_shutdown(
-                options->transport, SENTRY_DEFAULT_SHUTDOWN_TIMEOUT);
+            if (!sentry__transport_shutdown(
+                    options->transport, SENTRY_DEFAULT_SHUTDOWN_TIMEOUT)) {
+                SENTRY_WARN("transport did not shut down cleanly");
+            }
+            dumped_envelopes = sentry__transport_dump_queue(
+                options->transport, options->run);
```
jan-auer, line 168: Could you even move this into the conditional block above?

src/sentry_transport.c
```diff
@@ -10,11 +10,12 @@
 
 typedef struct sentry_transport_s {
     void (*send_envelope_func)(sentry_envelope_t *envelope, void *state);
-    void (*startup_func)(const sentry_options_t *options, void *state);
+    bool (*startup_func)(const sentry_options_t *options, void *state);
     bool (*shutdown_func)(uint64_t timeout, void *state);
     void (*free_func)(void *state);
     size_t (*dump_func)(sentry_run_t *run, void *state);
     void *state;
+    bool was_started;
```
jan-auer, line 18: I'm curious if this should be `running`, and we set this to `false` after shutting down. The other option would be to empty out the transport so that everything afterwards becomes a noop.

src/transports/sentry_transport_winhttp.c
```diff
@@ -207,16 +222,18 @@ sentry__winhttp_send_task(void *_envelope, void *_state)
...
-        SENTRY_DEBUGF("http request failed with error: %d", GetLastError());
+        SENTRY_DEBUGF(
+            "`WinHttpSendRequest` failed with code `%d`", GetLastError());
     }
 
+    uint64_t now = sentry__monotonic_time();
+    SENTRY_TRACEF("request handled in %llums", now - started);
+
+exit:
     WinHttpCloseHandle(request);
```
jan-auer, line 233: Two questions on this:
1. Does this need to be guarded against `NULL`? In examples, I usually see that calls to this method are guarded.
2. In the [WinHttpCloseHandle documentation](https://docs.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpclosehandle) it says:

> An application should never call WinHttpCloseHandle on a synchronous request. 

Should we actually be calling this? Examples seem to suggest that this is fine, can you just check what this refers to?

src/sentry_core.c
Swatinem: The backend is freed inside of `options_free`. I thought about doing the shutdown, but the backend start is the last thing that can fail, so doing a shutdown does not make sense.

src/sentry_core.c
Swatinem: We are pretty much guaranteed to have 0 dumped envelopes on successful shutdown. But lets be defensive here anyway.

src/transports/sentry_transport_winhttp.c
Swatinem: I think it refers to concurrently calling this on a request that is called in a different thread? All the examples are closing the request.

src/sentry_core.c
Swatinem: The transport previously required the global options but that was fixed already. The problem with the backend is that it currently uses "public" API internally, which requires the global options.
The other option would be to set the options, and afterwards *unset* them again in case the backends fails to init, which would be observable from the outside, which is also weird.
The cleanest thing to do would be to explicitly pass the options around, but that is hard to do right now.

All-in-all, I think options should only be set once when the SDK was successfully initialized and is ready to be used.

include/sentry.h
```diff
@@ -505,15 +504,21 @@ typedef struct sentry_options_s sentry_options_t;
  * This represents an interface for user-defined transports.
  *
  * Transports are responsible for sending envelopes to sentry and are the last
- * step in the event pipeline. A transport has the following hooks, all of which
+ * step in the event pipeline.
+ *
+ * Envelopes will be submitted to the transport in a _fire and forget_ fashion,
+ * and the transport must send those envelopes _in order_.
```
daxpedda, line 510: Envelopes have to be sent in order now? So the transport can't send envelopes in parallel then.
This should be noted in the CHANGELOG as a breaking change as well.

Swatinem: this has been the case ever since release health, since that relies on getting the session updates in order. But sure, I can add it to the changelog as well.

PR #344, opened 2020-07-20
fix: Flag crashed sessions using macOS crashpad

Swatinem:
This is based on a heuristic comparing the timestamps of the latest crashpad report with the start time of the last session.

### DIFF
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -5,9 +5,14 @@ extern "C" {
 #include "sentry_backend.h"
 #include "sentry_core.h"
 #include "sentry_database.h"
+#include "sentry_envelope.h"
 #include "sentry_options.h"
 #include "sentry_path.h"
+#include "sentry_sync.h"
+#include "sentry_transport.h"
+#include "sentry_unix_pageallocator.h"
 #include "sentry_utils.h"
+#include "transports/sentry_disk_transport.h"
 }
 
 #include <map>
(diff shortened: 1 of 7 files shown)


src/backends/sentry_backend_crashpad.cpp
```diff
@@ -312,6 +312,54 @@ sentry__crashpad_backend_except(
...
+    if (data->db->GetCompletedReports(&reports)
+        == crashpad::CrashReportDatabase::kNoError) {
+        for (const crashpad::CrashReportDatabase::Report &report : reports) {
+            report_crash_time(&crash_time, report);
+        }
+    }
+
+    // clean the database afterwards, according to docs, this is only relevant
+    // on linux.
+    int cleaned = data->db->CleanDatabase(3600);
```
jan-auer, line 355: It sounds like this might be called by the handler for you, so I think you can skip this entirely.

src/sentry_database.c
```diff
@@ -178,8 +178,23 @@ sentry__process_old_runs(const sentry_options_t *options)
...
+                    // this should only happen when using crashpad, and there
+                    // should normally be only a single unclosed session at a
+                    // time.
                     if (session->status == SENTRY_SESSION_STATUS_OK) {
-                        session->status = SENTRY_SESSION_STATUS_ABNORMAL;
+                        bool was_crash
+                            = last_crash && last_crash > session->started_ms;
+                        if (was_crash) {
+                            session->duration_ms
+                                = last_crash - session->started_ms;
```
jan-auer, line 193: We should probably set `last_crash = 0` in this branch to avoid marking multiple sessions as crashed.

Swatinem:
merged this into #335

PR #345, opened 2020-07-27
feat: breakpad support for OSX

eakoli:
Add support for breakpad on OSX.

Only minor issue is with `sentry__breakpad_backend_except` as the SignalHandler method used by the linux/windows paths is private in the osx version of the ExceptionHandler


src/backends/sentry_backend_breakpad.cpp
```diff
@@ -23,10 +22,19 @@ extern "C" {
...
+#elif defined SENTRY_PLATFORM_DARWIN
+#    include "client/mac/handler/exception_handler.h"
+#    include <sys/sysctl.h>
 #else
 #    include "client/linux/handler/exception_handler.h"
 #endif
 
+#ifdef SENTRY_PLATFORM_UNIX
+extern "C" {
+#include "sentry_sync.h"
```
Swatinem, line 34: Is there a specific reason this was moved down below the breakpad includes? Would be nice to add a comment about why?

eakoli: In my code base I added support for macOS 10.6, which needs to emulate a recursive mutex, and the way the current apple code path manages setting up the mutex is all with defines, so the simplest solution was to just redefine the pthread routines to recursive variants, so it causes issues with some of the c++ includes that have inline code using pthread_mutext_init.

Ideally I would have refactored sentry_sync so it didn't have everything done via macros and instead implemented all the routines in the c files, but I was trying not to rewrite all of that unit.

tests/conditions.py
```diff
@@ -14,7 +14,7 @@
 has_breakpad = (
     not is_valgrind
     and not is_kcov
-    and (sys.platform == "linux" or sys.platform == "win32")
+    and (sys.platform == "linux" or sys.platform == "win32" or sys.platform == "darwin")
```
Swatinem, line 17:
```suggestion
    and not is_android
```

src/backends/sentry_backend_breakpad.cpp
Swatinem: Oh, so you are saying that you have your own breakpad fork that overrides this here?:

https://github.com/getsentry/sentry-native/blob/cbd7b964b6721b622fb4eb44b77f719aa944e7d4/src/sentry_sync.h#L202

Can you solve that differently by specifying some custom CFLAGS? It sounds like quite a hack that is hard to get right, and also not a case I would like to optimize for really.

src/backends/sentry_backend_breakpad.cpp
```diff
@@ -23,10 +22,19 @@ extern "C" {
 
 #ifdef SENTRY_PLATFORM_WINDOWS
 #    include "client/windows/handler/exception_handler.h"
+#elif defined SENTRY_PLATFORM_DARWIN
```
Swatinem, line 25:
```suggestion
#elif defined(SENTRY_PLATFORM_DARWIN)
```

Can you change these to be consistent with the rest of the codebase please.

src/backends/sentry_backend_breakpad.cpp
eakoli: effectively, I'm not sure how a CFLAG would help the situation. 
The implementation is basically the same principle that SDL has used for ages for recursive mutex support.

eakoli: the proper way would be to to clean up sentry_sync.h so its not all header based, as its rather fragile to begin with and still has lots of warnings depending on where its included.

Swatinem: I think thats a lot of work, and also has some potential for breakage and perf regressions. But admittedly, it would be cleaner.

Jake-Shadle:
Is there still interest in getting this PR in? Hasn't had any activity in a while, but would be really nice to have.

Swatinem:
Rebased and merged this in https://github.com/getsentry/sentry-native/commit/989782943a5b42ad938760654cc7b7a1040e2839

PR #348, opened 2020-08-10
fix: Lock scope for duration of flush hook

Swatinem:
This fixes the threadsafety issues identified in https://github.com/daxpedda/sentry-contrib-native/pull/13#issuecomment-669610951


Swatinem:
copy-pasting from https://github.com/daxpedda/sentry-contrib-native/pull/13#issuecomment-671248874:

That should take care of the `set_extra` and `set_context` issues. And yes, I think using the crashpad backend introduced a little delay here that triggered the threadsafety issue:

1. ` set_context` and friends change the scope while locked, triggering a flush
2. the flush is propagated to the crashpad backend, which does its own locking so should be safe.
3. the session which is attached to the scope is also flushed while *not* holding the lock, which is probably triggering the crashes you see.

src/sentry_scope.h
```diff
@@ -55,9 +55,10 @@ void sentry__scope_cleanup(void);
 
 /**
  * This will notify any backend of scope changes, and persist session
- * information to disk.
+ * information to disk. This function must be called while holding the scope
+ * lock, and it will be unlocked internally.
```
flub, line 59: that seems a bit easy to get wrong, but ok.  i guess it'll just crash and not even pass the test suite if you get it wrong?

Swatinem: I don’t really know what happens when you try to unlock an unlocked lock :-D The only use of that function is from within the scope macro.

src/sentry_scope.c
```diff
@@ -96,17 +96,26 @@ sentry__scope_unlock(void)
 }
 
 void
-sentry__scope_flush(const sentry_scope_t *scope)
+sentry__scope_flush_unlock(const sentry_scope_t *scope)
 {
+    bool did_unlock = false;
```
Swatinem, line 101: hm, this is a bit fragile. Ideally, I would like to put the scope and the session onto the options, since that is reference counted, and all this is quite interconnected anyway. I don’t know why we had the scope separate in the first place.

PR #350, opened 2020-08-11
fix: Fix NaN-tagging of pointers and document things

Swatinem:
Not really sure if we need to care about https://en.wikipedia.org/wiki/X86-64#Canonical_form_addresses ?


Swatinem:
> This is an excellent candidate for thorough unit-testing.

Well yes, but the relevant functions are static to that file, so we currently can’t :-D

I will merge this as is and revisit later maybe.

PR #356, opened 2020-08-19
docs: Better document the database-path requirements

### LINKED ISSUE #355: Windows 10 - breakpad - folder location problem and solution
**Description**

I got my Windows 10 app built to use breakpad crash reporting.

It worked great, up until the time the app is bundled into a package for the Windows 10 Store - at which point, I don't see any reports on sentry.io

I found that the reason, is that sentry-native was building the crash folder immediately under the app folder (!), i.e. this specific (hidden) sub-folder.
 .sentry-native

That failed silently in the packaged Windows 10 app for the Windows 10 Store.

So, I changed the code to do something like this to use a non-magic folder, that persists across crashes.

```
  #define SENTRY_BUILD_STATIC 1

  #include <sentry.h>

  #include <ShlObj.h>
  #include <WinError.h>

  //...

  //...

  // Don't rely on the default folder - or the crashes will not be uploadable in a *packaged* application,
  // even though it works with non-packaged application!
  char tempFolder[2048];
  tempFolder[0] = 0;
  GetTempPathA((DWORD)sizeof(tempFolder), tempFolder);
  strcat(tempFolder, "\\myapp_crashes");
  auto result = SHCreateDirectoryExA(nullptr, tempFolder, nullptr);

  sentry_options_set_database_path(options, tempFolder);

  //...

  sentry_init(options);
```

That now works!

I'd suggest you document this very clearly somewhere!

**Environment**

- OS: Windows 10
- Compiler: MSVC 19
- CMake version and config: 3.16.5
cmake -B build -DBUILD_SHARED_LIBS=OFF -DSENTRY_BACKEND=breakpad -SENTRY_BUILD_RUNTIMESTATIC=ON

**Steps To Reproduce**

Just build and run the app, and make it crash. The crash uploading works, even without the above change.

Bundle the app for app store, and deploy it. The crash reports aren't uploaded, *unless* you do something like I suggested.

I hope this helps!

Swatinem:
I think `XDG_CACHE_HOME` fits the usecase the best, as this is *non-essential* data and can be removed at any time without any real negative effects (apart from missing abnormal/crashed sessions). Not sure about user consent though.

closes #355 CC @sseyod how do you like the explanation?

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Please see [Known Limitations](#known-limitations).
 - [Building and Installation](#building-and-installation)
   - [Compile-Time Options](#compile-time-options)
   - [Build Targets](#build-targets)
+- [Runtime Configuration](#runtime-configuration)
 - [Known Limitations](#known-limitations)
 - [Development](#development)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -237,6 +238,26 @@ Legend:
...
+// your application code …
+
+sentry_shutdown();
+```
+
+Other important configuration options include:
+
+- `sentry_options_set_database_path`: Sentry needs to persist some cache data across application restarts, especially for proper handling of release health sessions. It is recommended to set an explicit absolute path corresponding to the applications cache directory (equivalent to `AppData/Local` on Windows, and `XDG_CACHE_HOME` on Linux). When not set explicitly, sentry will create and use the `.sentry-native` directory inside of the current working directory.
+- `sentry_options_set_handler_path`: When using the crashpad backend, sentry will look for a `crashpad_handler` executable in the same directory as the running executable. It is recommended to set this as an explicit absolute path based on the applications install location.
+- `sentry_options_set_release`: Some features in sentry, including release health need to have a release version set. This corresponds to the applications version and needs to be set explicitly. See [Releases](https://docs.sentry.io/product/releases/) for more information.
```
jan-auer, line 259:
```suggestion
- `sentry_options_set_release`: Some features in sentry, including release health, need to have a release version set. This corresponds to the application's version and needs to be set explicitly. See [Releases](https://docs.sentry.io/product/releases/) for more information.
```

jan-auer (approved):

sseyod:
Am I right in saying that the folder you supply is emptied automatically by the crashed software?

In other words, you'd might want to clarify that the developer ensures that the path name is unique to your app  (i.e. not just rely on the folder path returned by the system.

Swatinem:
> ensures that the path name is unique to your app

oh absolutely. We had this case with our own android SDK which bundles the native SDK. I will add another clarification to it.

PR #359, opened 2020-08-25
build: Use the correct output directory for Makefile target

Swatinem:


flub (approved):

jan-auer (changes requested):
Can we instead set the binary directory of the crashpad `add_subdirectory` according to [the documentation](https://cmake.org/cmake/help/latest/command/add_subdirectory.html), and then also fix the crashpad CMake files instead? 

This approach is not as nice, as it forces everyone importing the CMake project to make the same adjustments. We can do this for the leak-build, as that is an internal test, but better not for a public asset.

Swatinem:
As I wrote in chat, the contents and paths of the build directory are intermediate assets and the build system can put them wherever it sees fit. The only real guarantee that we have is if we actually use the `install` target I think.

jan-auer:
I disagree with that, as the `binary_dir` argument is clearly meant for placing final outputs in a certain location. If you really prefer to keep the current structure, however, then let's instead work on the `install` target. Either way, the current patch is insufficient.

PR #369, opened 2020-09-07
fix: Work around mpack failing the build

Swatinem:


jan-auer:
@Swatinem This effectively reimplements `static_assert` in a funny way. Have you figured out why this linking error happens in the first place?

Swatinem:
So I *think* this is what happens:

CMake 3.18.2 has the following patch:
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5115/diffs

which adds the correct `-std:c11` switch to MSVC.

which then I guess provides the correct defines but triggers this code:

https://github.com/getsentry/sentry-native/blob/6cd81be00dc53b321b3388b279d8fb8e8bb717bb/vendor/mpack.h#L763-L765

which just assumes that msvc has a `_Static_assert` which it apparently does not.

So I’m trying to just patch the mpack header to just use whatever `assert.h` provides…

flub (approved):

vendor/mpack.h
```diff
@@ -754,6 +754,9 @@ MPACK_HEADER_START
 
 /* Static assert */
 
+#include <assert.h>
+#define MPACK_STATIC_ASSERT static_assert
```
jan-auer, line 758: Can you instead fix the ifdefs below to check for `_MSC_VER`?

jan-auer (approved):

PR #372, opened 2020-09-15
fix: Initialize all static data

Swatinem:
potentially fixes https://github.com/getsentry/sentry-java/issues/904


marandaneto (approved):
LGTM, CI is not happy though

PR #373, opened 2020-09-16
feat: Add on-device symbolication on Android

sfionov:
It is important to symbolize as much as possible on Android devices because every
device has own libc and libart versions, so symbols can't be resolved by addrs remotely.

This patch adds symbolication and demangling into inproc handler used in Android.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -62,8 +62,16 @@ else()
 		sentry_unix_pageallocator.h
 		sentry_unix_spinlock.h
 		path/sentry_path_unix.c
-		symbolizer/sentry_symbolizer_unix.c
 	)
+	if(ANDROID)
+		sentry_target_sources_cwd(sentry
+			symbolizer/sentry_symbolizer_unix_demangled.cc
+		)
+	else()
+		sentry_target_sources_cwd(sentry
+			symbolizer/sentry_symbolizer_unix.c
+		)
+	endif()
 endif()
 
 # module finder
--- a/src/backends/sentry_backend_inproc.c
+++ b/src/backends/sentry_backend_inproc.c
@@ -7,6 +7,7 @@
 #include "sentry_envelope.h"
 #include "sentry_options.h"
 #include "sentry_scope.h"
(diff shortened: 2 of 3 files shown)


sfionov:
Anybody here? :)

src/backends/sentry_backend_inproc.c
```diff
@@ -212,8 +227,12 @@ make_signal_event(
     sentry_value_t frames = sentry__value_new_list_with_size(frame_count);
     for (size_t i = 0; i < frame_count; i++) {
         sentry_value_t frame = sentry_value_new_object();
-        sentry_value_set_by_key(frame, "instruction_addr",
-            sentry__value_new_addr((uint64_t)backtrace[frame_count - i - 1]));
+        void *ip = backtrace[frame_count - i - 1];
+#if defined SENTRY_PLATFORM_ANDROID
+        sentry__symbolize(ip, fill_frame, &frame);
```
Swatinem, line 232: I don’t think you need these changes here.
We already have an option to symbolicate on-device here:

https://github.com/getsentry/sentry-native/blob/bac5d366593e68f240b4446c3bf1ab1d1556a7bd/include/sentry.h#L789-L798
That option is on-by-default on android.

src/CMakeLists.txt
```diff
@@ -62,8 +62,16 @@ else()
...
 		path/sentry_path_unix.c
-		symbolizer/sentry_symbolizer_unix.c
 	)
+	if(ANDROID)
+		sentry_target_sources_cwd(sentry
+			symbolizer/sentry_symbolizer_unix_demangled.cc
+		)
+	else()
+		sentry_target_sources_cwd(sentry
+			symbolizer/sentry_symbolizer_unix.c
```
Swatinem, line 72: I wonder if we should maybe just always do demangling on unix? Or rather introduce another CMake option for this, instead of using it just for android.

Swatinem (commented):
sorry, this slipped my radar.
can you double-check if you actually need the changes to the inproc backend?

PR #378, opened 2020-09-23
fix: Change NaN-boxing to Pointer Tagging

Swatinem:
There is already this note here about `sentry_value_t`:

https://github.com/getsentry/sentry-native/blob/425b803d53acc2d598633fd64ac969508487007e/include/sentry.h#L127-L128

I think in a future version we can just typedef this to a `uint64_t`.

This now moves double values to a heap allocation with a refcount.

### DIFF
--- a/src/sentry_value.c
+++ b/src/sentry_value.c
@@ -26,38 +26,50 @@
 #include "sentry_value.h"
 
 /**
- * NaN-boxing of sentry_value_t
+ * Pointer Tagging of `sentry_value_t`
  *
- * We do encode 32-bit values as well as some constants and pointers into the
- * NaN-space of 64-bit floating-point numbers, like so:
+ * We expect all of the pointers we deal with to be at least 4-byte aligned,
+ * which means we can use the least significant 2 bits for tagging.
+ * We only ever save pointers to `thing_t`, which has an `alignof >= 4`, and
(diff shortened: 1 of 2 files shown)


src/sentry_value.c
```diff
@@ -26,38 +26,47 @@
 #include "sentry_value.h"
 
 /**
- * NaN-boxing of sentry_value_t
+ * Pointer Tagging of `sentry_value_t`
  *
- * We do encode 32-bit values as well as some constants and pointers into the
- * NaN-space of 64-bit floating-point numbers, like so:
+ * We expect all of the pointers we deal with to be at least 4-byte aligned,
+ * which means we can use the least significant 2 bits for tagging.
```
flub, line 32: do we have any way of checking this is true at build time rather than mysteriously fail at runtime?

Swatinem: We only ever use pointers to `thing_t`, which we can check has an `alignof` of `>= 4` I guess.
Other than that, both our own allocator, as well as the system allocator is pretty much guaranteed to give us 8-byte alignment I think.

jan-auer: Since we both asked this, maybe write this into the comment :)

jan-auer (approved):

PR #383, opened 2020-09-28
ci: Switch to GitHub Actions

Swatinem:
* Includes a workaround for the linux dependency installation
* replaces downloading an external tool on windows (which takes 2 minutes) to just look into the PE header directly
* (still waiting to verify that artifact upload to zeus actually works)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,178 @@
...
+        if: ${{ contains(env['RUN_ANALYZER'], 'cov') }}
+        uses: codecov/codecov-action@v1
+        with:
+          directory: coverage
+
+  archive:
+    name: Create Release Archive
+    runs-on: ubuntu-latest
+    needs: [lint, test]
+    if: ${{ needs.test.result == 'success' && github.repository_owner == 'getsentry' }}
```
jan-auer, line 150: Should we just run this on release branches? Also, do we really need to wait for tests with this?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,178 @@
...
+      - name: Upload source artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: sentry-native-source
+          # meh: https://github.com/actions/upload-artifact#zipped-artifact-downloads
+          path: sentry-native-source.zip
+
+      - name: Upload archive to zeus
+        run: |
+          npx -p "@zeus-ci/cli" zeus job update --build ${{ github.run_id }} --job ${{ github.job }} --ref ${{ github.sha }} --status passed
```
jan-auer, line 175: Please run this without a status first, then upload the artifact, and then set the status to avoid flakey `craft publish` runs. Otherwise, craft may exit with an error that the required artifact is not available.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,178 @@
...
+          npx -p "@zeus-ci/cli" zeus upload --build ${{ github.run_id }} --job ${{ github.job }} --type "application/zip" --name "sentry-native.zip" "sentry-native-source.zip"
+        env:
+          ZEUS_HOOK_BASE: ${{ secrets.ZEUS_HOOK_BASE }}
```
jan-auer, line 178: This is interesting. To my recollection, this requires the `ZEUS_API_TOKEN`, too. However, apparently it works.

Zeus build can be viewed here: https://zeus.ci/gh/getsentry/sentry-native/builds/243

jan-auer: Note that this always builds on the ugly merge commit, but for release builds where we actually need this, this will always be the actual commit.

.github/workflows/ci.yml
Swatinem: its what we did before, and I think its better to run an extra set of tests, just in case. I remember times when the version bump script did not correctly update the test assertions and caused failures.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,179 @@
...
+
+      - name: Remove unneeded files
+        shell: bash
+        run: rm -rf build .c* .e* .git* scripts Makefile external/breakpad/src/tools external/breakpad/src/processor
+
+      - name: Create source archive
+        shell: python
+        run: |
+          import shutil
+          shutil.make_archive("sentry-native-source", "zip", ".")
```
jan-auer, line 164: fwiw, `zip` is installed on GitHub runners.

jan-auer (approved):

.github/workflows/ci.yml
Swatinem: hm… right, we don’t need this to be cross-platform, but its only running on linux…
:-D I kind of forget that for CI… anyhow, I think its fine for now, I don’t really want to waste more time on it…

PR #385, opened 2020-10-01
fix: Use a u32 value for the signum on Windows

Swatinem:
`EXCEPTION_ACCESS_VIOLATION` is defined as `0xC0000005`, which overflows
an i32 and will turn negative, which causes processing problems later on.


flub (approved):

src/sentry_value.c
```diff
@@ -767,6 +791,8 @@ sentry_value_is_true(sentry_value_t value)
         return 0;
     case SENTRY_VALUE_TYPE_INT32:
         return sentry_value_as_int32(value) != 0;
+    case SENTRY_VALUE_TYPE_UINT32:
+        return sentry_value_as_uint32(value) > 0;
```
jan-auer, line 795: _nit_: Keep it consistent with the previous one.

```suggestion
        return sentry_value_as_uint32(value) != 0;
```

jan-auer (changes requested):
Have you considered simply representing those as double?

There's one obvious problem with representing `uint32_t `: They are not interchangeable with `int32_t`. Now, instead of just checking for a numeric value, you always have to check for two possibilities:
- Is it an int32_t? Then do one thing.
- Is it a uint32_t? Then do another thing.

We could at least shim this in functions like `sentry_value_as_uint32` and then return a value if it fits, and vice versa. For example, if the value is a positive `int32_t`, still return the `uint32_t` counterpart. However, this may lead to strange behavior if suddenly an unexpected value lands in one of the fields.

Furthermore, the JSON parser does not support deserializing `uint32_t` yet. If you were to do a roundtrip, you'd deserialize as double. Any caller who depends on `as_uint32` would fail.

Swatinem:
superceded by #392

PR #386, opened 2020-10-01
feat: Expose the backend as an SDK integration

Swatinem:


tests/assertions.py
```diff
@@ -33,16 +33,15 @@ def assert_meta(envelope, release="test-example-release"):
...
-        },
+    }
+    expected_sdk = {
+        "name": "sentry.native",
+        "version": "0.4.2",
+        "packages": [{"name": "github:getsentry/sentry-native", "version": "0.4.2"},],
     }
 
     assert matches(event, expected)
+    assert matches(event["sdk"], expected_sdk)
```
flub, line 44: I don't really get the test.  You're not actually seeing the backend thing you just added but just re-wrote this a bit?

Swatinem: I basically want to exclude that from the match, as the assertion fails if we have additional attributes. And I don’t really want to pass in that flag explicitly for each testcase.

flub: Is it possible to add a test that checks it's present but `in ["crashpad", "breakpad", "inproc"]`?

Swatinem: its not present for "none" ;-)

jan-auer: Can you then at least test it in a separate assertion?

jan-auer (approved):

PR #387, opened 2020-10-01
fix: Unwind from local stack if ucontext fails

### LINKED ISSUE #384: No frames found when unwinding backtrace on Linux after SIGSEGV
**Description**

Hello! I seem to be having an issue when Sentry tries to get stack frames after a segmentation fault on Linux where it ends up finding 0 frames. I debugged `libsentry.so` and found that sentry-native always skips getting the backtrace from libbacktrace because we always have a valid `sentry_ucontext_t` at the following trace:
```
(gdb) bt
#0  sentry__unwind_stack_libbacktrace (addr=0x0, uctx=0x55555559e310, ptrs=0x55555559de60, max_frames=128) at /home/mischa/dev/mutate/conan-sentry-native/source_subfolder/src/unwinder/sentry_unwinder_libbacktrace.c:9
#1  0x00007ffff7c90cfd in unwind_stack (addr=0x0, uctx=0x55555559e310, ptrs=0x55555559de60, max_frames=128) at /home/mischa/dev/mutate/conan-sentry-native/source_subfolder/src/unwinder/sentry_unwinder.c:27
#2  0x00007ffff7c90d64 in sentry_unwind_stack_from_ucontext (uctx=0x55555559e310, stacktrace_out=0x55555559de60, max_len=128) at /home/mischa/dev/mutate/conan-sentry-native/source_subfolder/src/unwinder/sentry_unwinder.c:45
#3  0x00007ffff7c94b07 in make_signal_event (sig_slot=0x7ffff7f86a98 <SIGNAL_DEFINITIONS+120>, uctx=0x55555559e310) at /home/mischa/dev/mutate/conan-sentry-native/source_subfolder/src/backends/sentry_backend_inproc.c:209
#4  0x00007ffff7c9471d in handle_ucontext (uctx=0x55555559e310) at /home/mischa/dev/mutate/conan-sentry-native/source_subfolder/src/backends/sentry_backend_inproc.c:263
#5  0x00007ffff7c9465f in handle_signal (signum=11, info=0x55555559e470, user_context=0x55555559e340) at /home/mischa/dev/mutate/conan-sentry-native/source_subfolder/src/backends/sentry_backend_inproc.c:308
#6  <signal handler called>
```

Because of the following code at https://github.com/getsentry/sentry-native/blob/master/src/unwinder/sentry_unwinder_libbacktrace.c#L15:
```c
    else if (uctx) {
        return 0;
    } else {
        return backtrace(ptrs, max_frames);
    }
```

Changing that `return 0` to `return backtrace(ptrs, max_frames)` works as expected, but I guess there's more going on here that I don't know about.

- [ ] During build
- [ ] During run-time

**Environment**

- OS: Archlinux x86-64
- Compiler: Clang 10
- CMake version and config: 3.18.3, SENTRY_BACKEND=inproc

**Steps To Reproduce**

```c++
#include <iostream>
#include <sentry.h>
#include <thread>
#include <chrono>

int main(int argc, char** argv)
{
	sentry_user_consent_give();
	sentry_options_set_dsn(options, !!!!);
	sentry_options_set_environment(options, "development");
	if (sentry_init(options)) {
		std::cerr << "Failed to initialize Sentry, crashes will not be reported!" << std::endl;
	}

    std::this_thread::sleep_for(std::chrono::seconds(3));

	*(reinterpret_cast<int*>(0) + 1) = 0;

	sentry_shutdown();

	return 0;
}
```

This reproduces it 100% of the time for me.

**Log output**

```
[sentry] INFO entering signal handler
[sentry] DEBUG captured backtrace with 0 frames
[sentry] DEBUG inspecting module "/tmp/repro"
[sentry] DEBUG inspecting module "/usr/lib/libresolv-2.32.so"
[sentry] DEBUG inspecting module "/usr/lib/libnss_dns-2.32.so"
[sentry] DEBUG inspecting module "/usr/lib/libnss_resolve.so.2"
[sentry] DEBUG inspecting module "/usr/lib/libnss_myhostname.so.2"
[sentry] DEBUG inspecting module "/usr/lib/libnss_mymachines.so.2"
[sentry] DEBUG inspecting module "/usr/lib/libnss_files-2.32.so"
[sentry] DEBUG inspecting module "/usr/lib/liblzma.so.5.2.5"
[sentry] DEBUG inspecting module "/usr/lib/libdl-2.32.so"
[sentry] DEBUG inspecting module "/usr/lib/libpthread-2.32.so"
[sentry] DEBUG inspecting module "/usr/lib/librt-2.32.so"
[sentry] DEBUG inspecting module "/usr/lib/libc-2.32.so"
[sentry] DEBUG inspecting module "/usr/lib/libgcc_s.so.1"
[sentry] DEBUG inspecting module "/usr/lib/libm-2.32.so"
[sentry] DEBUG inspecting module "/usr/lib/libstdc++.so.6.0.28"
[sentry] DEBUG inspecting module "/usr/lib/libunwind.so.8.0.1"
[sentry] DEBUG inspecting module "/tmp/libsentry.so"
[sentry] DEBUG inspecting module "/usr/lib/ld-2.32.so"
[sentry] DEBUG inspecting module "linux-gate.so"
[sentry] DEBUG read 19 modules from /proc/self/maps
[1]    21361 segmentation fault (core dumped)  LD_LIBRARY_PATH=. ./repro
```

Swatinem:
Not quite sure if we should do this in the backend, or rather patch the unwinder itself…

fixes #384

### DIFF
--- a/src/backends/sentry_backend_inproc.c
+++ b/src/backends/sentry_backend_inproc.c
@@ -207,6 +207,12 @@ make_signal_event(
     void *backtrace[MAX_FRAMES];
     size_t frame_count
         = sentry_unwind_stack_from_ucontext(uctx, &backtrace[0], MAX_FRAMES);
+    // if unwinding from a ucontext didn't yield any results, try again with a
+    // direct unwind. this is most likely the case when using `libbacktrace`,
+    // since that does not allow to unwind from a ucontext at all.
+    if (!frame_count) {
+        frame_count = sentry_unwind_stack(NULL, &backtrace[0], MAX_FRAMES);
+    }
     SENTRY_TRACEF("captured backtrace with %lu frames", frame_count);
 
     sentry_value_t frames = sentry__value_new_list_with_size(frame_count);


Mixaill:
When we use direct unwind, do we need to skip 1 (or more) frame to hide sentry?

GenuineAster:
image
5 frames including the signal handler.

flub (approved):

flub:
I'm not sure if we should hide frames.  If this should be done at all perhaps just collapsing them by default in the UI is probably better (which I think might already happen)?

Swatinem:
We don’t flag frames as `in_app` on native (yet). But yes, I would also say we should not truncate anything on the SDK side. Maybe we will set `in_app` in the future at some point.

jan-auer (approved):

PR #388, opened 2020-10-01
fix: Add sentry_reset_backend API

GenuineAster:
We have a third-party dynamic library (.NET Core runtime) that is loaded far into our application's runtime which, on load, steals all signal handlers. On Linux this works out fine, they are responsible and call whatever signal handler was there before them. No such luck on Windows. Forking and patching the .NET Core runtime is quite a bit more work, and we may run into this kind of issue in the future, so our internal fix is to add this API, and hopefully we can get this upstreamed and avoid forking sentry-native altogether.

I'm not really sure how to write tests for this, but I can definitely add tests for it with a bit of guidance!

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -925,6 +925,16 @@ SENTRY_API int sentry_shutdown(void);
 SENTRY_EXPERIMENTAL_API void sentry_clear_modulecache(void);
 
 /**
+ * Re-initializes the Sentry backend
+ *
+ * This is mainly meant for if a third-party library hijacks Sentry's signal
+ * handler and we need to reinstall it.
+ *
+ * Returns 0 on success.
+ */
+SENTRY_API int sentry_reset_backend(void);
+
+/**
  * Gives user consent.
  */
 SENTRY_API void sentry_user_consent_give(void);
--- a/src/sentry_core.c
+++ b/src/sentry_core.c
@@ -213,6 +213,22 @@ sentry_clear_modulecache(void)
     sentry__modulefinder_cleanup();
 }
 
+int
+sentry_reset_backend(void)
+{
+    sentry_backend_t *backend = g_options->backend;
+    if (backend && backend->shutdown_func) {
+        backend->shutdown_func(backend);
+    }
+
+    if (backend && backend->startup_func) {
+        if (g_options->backend->startup_func(g_options->backend, g_options)) {
+            return 1;
+        }
+    }
+    return 0;
+}
+
 static void
 set_user_consent(sentry_user_consent_t new_val)
 {


src/sentry_core.c
```diff
@@ -213,6 +213,22 @@ sentry_clear_modulecache(void)
     sentry__modulefinder_cleanup();
 }
 
+int
+sentry_reset_backend(void)
+{
+    sentry_backend_t *backend = g_options->backend;
```
Swatinem, line 219: You have to wrap this in `SENTRY_WITH_OPTIONS (options) {`
Otherwise you are not thread safe, and another thread can just free `g_options` from under you.
Also note that you can’t early-return from that macro :-(

include/sentry.h
```diff
@@ -924,6 +924,16 @@ SENTRY_API int sentry_shutdown(void);
  */
 SENTRY_EXPERIMENTAL_API void sentry_clear_modulecache(void);
 
+/**
+ * Re-initializes the Sentry backend
```
Swatinem, line 928:
```suggestion
 * Re-initializes the Sentry backend.
```

include/sentry.h
```diff
@@ -924,6 +924,16 @@ SENTRY_API int sentry_shutdown(void);
...
+ *
+ * This is mainly meant for if a third-party library hijacks Sentry's signal
```
Swatinem, line 930: I would maybe phrase this more positively:

This is needed if a third-party library overrides the previously installed signal handler, and reinstalls the Sentry handler.

include/sentry.h
```diff
@@ -924,6 +924,16 @@ SENTRY_API int sentry_shutdown(void);
...
+ * handler and we need to reinstall it.
+ *
+ * Returns 0 on success.
+ */
+SENTRY_API int sentry_reset_backend(void);
```
Swatinem, line 935:
```suggestion
SENTRY_EXPERIMENTAL_API int sentry_reset_backend(void);
```

Swatinem (changes requested):

Swatinem:
I really like this!

Also, I was running into this case recently, investigating why native crashes when using a .NET won’t be captured; instead .NET handles them somehow.
I’m a bit concerned as for example the crashpad backend is not really meant to be restarted.

Also, an e2e test would be nice, but I think its a bit too complex to integrate into our example (which we use for tests)

bruno-garcia:
Could this be achieved by the call to `init`?
Basically close any previous clients, flushes stuff, and initializes again?

It's what we do in .NET:

https://github.com/getsentry/sentry-dotnet/blob/42b0f921f0503c561046b96c3ad486511487e7cb/src/Sentry/SentrySdk.cs#L92-L94

Ref swap the current thing with the new one, the close out the previous one once it's not bound anymore/

GenuineAster:
I am quite busy this week (worked on this as part of a hackathon project) but I will fix your comments early next week, as well as testing with crashpad and looking into testing this a bit more!

Swatinem:
> Could this be achieved by the call to `init`?

@bruno-garcia Thats already what `sentry_init` does. In theory, one could just call `init` again, but that re-initializes the *whole* SDK. Here we just want to reset the backend, and make sure the signal handlers are correct.

Not quite sure whats the case with .NET, what part of the code runs when, and who wins, with respect to which signal handlers are active.

bruno-garcia:
Fwiw I was able to capture minidumps on CoreCLR on all platforms: 

https://github.com/getsentry/sentry-dotnet-minidump

Swatinem:
Rebased and merged in https://github.com/getsentry/sentry-native/commit/2a55499c3bc2f991687aef10d2b7c05e4d692cc8

PR #390, opened 2020-10-02
ci: Simplify and run archive job only on main repo

Swatinem:
This now runs the archiving job only for pushes, so they don’t get run for PRs and have them fail if they originate from a foreign branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,21 +147,16 @@ jobs:
     name: Create Release Archive
     runs-on: ubuntu-latest
     needs: [lint, test]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -147,21 +147,16 @@ jobs:
     name: Create Release Archive
     runs-on: ubuntu-latest
     needs: [lint, test]
-    if: ${{ needs.test.result == 'success' && github.repository_owner == 'getsentry' }}
+    if: ${{ needs.test.result == 'success' && github.event_name == 'push' }}
```
jan-auer, line 150: Can you explain in the description how this only runs on the main repo?

Swatinem:
continuing in #426

PR #398, opened 2020-10-09
fix: add log message when sentry-native fails to create/check directory

Mixaill:
Found it when trying to add sentry-native support to the COM server (COM servers have default working directory in `C:/Windows/System32/`)


Swatinem (approved):
you have to run this through clang-format it seems

PR #403, opened 2020-10-21
fix: Avoid closing automatic session on `start_session`

Swatinem:
This fixes a pitfall when calling `sentry_start_session` with automatic
session tracking enabled, which would have previously flushed a clean
zero-duration session.


flub (approved):

Swatinem:
not gonna do this, the thinking is that things work as intended, people should just not call `start_session` when auto-session-tracking is enabled.

PR #413, opened 2020-11-02
feat: name sentry background worker thread

### LINKED ISSUE #411: Hook to call when background task worker start
Hello,

I'm using Sentry Native 0.4.3 on Linux.

I'm naming all my thread in my application and the only thread I cannot name is the sentry background task worker.

Is it possible to add a hook, (something like the one called by sentry inside of `sentry_init`) but here called by the background thread worker in order to get possibility to name the thread inside this hook?

Or a function to return the thread handle?

Regards.

Mixaill:
Add "Sentry BgWorker" description to the Sentry background worker thread.

Fixes #411 

---

Fun fact: in Windows, Thread Handle != Thread Id.
But in sentry-native, `sentry_threadid_t` is actually thread handle, not id :)

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -729,6 +729,18 @@ SENTRY_API const char *sentry_options_get_ca_certs(
     const sentry_options_t *opts);
 
 /**
+ * Configures the name of background worker thread
+ */
+SENTRY_API void sentry_options_set_bgworker_name(
+    sentry_options_t *opts, const char *name);
+
+/**
+ * Returns the configured background worker thread name.
+ */
+SENTRY_API const char *sentry_options_get_bgworker_name(
+    const sentry_options_t *opts);
+
+/**
  * Enables or disables debug printing mode.
  */
 SENTRY_API void sentry_options_set_debug(sentry_options_t *opts, int debug);
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -729,15 +729,15 @@ SENTRY_API const char *sentry_options_get_ca_certs(
     const sentry_options_t *opts);
 
 /**
- * Configures the name of background worker thread
+ * Configures the name of the http transport thread.
  */
-SENTRY_API void sentry_options_set_bgworker_name(
+SENTRY_API void sentry_options_set_transport_thread_name(
     sentry_options_t *opts, const char *name);
 
 /**
- * Returns the configured background worker thread name.
+ * Returns the configured http transport thread name.
  */
-SENTRY_API const char *sentry_options_get_bgworker_name(
+SENTRY_API const char *sentry_options_get_transport_thread_name(
     const sentry_options_t *opts);
 
 /**
--- a/src/sentry_options.c
+++ b/src/sentry_options.c
@@ -24,6 +24,7 @@ sentry_options_new(void)
     opts->debug = debug && sentry__string_eq(debug, "1");
     sentry_logger_t logger = { sentry__logger_defaultlogger, NULL };
     opts->logger = logger;
+    opts->bgworker_name = sentry__string_clone("Sentry BgWorker");
 #ifdef SENTRY_PLATFORM_WINDOWS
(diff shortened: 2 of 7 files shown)


Mixaill:
It also possible to add ability to set specific name via `sentry_options` (but I do not known if we will have at least one user of such feature :D )

Mixaill:
So, we have at least one user, so I've added ability to set custom thread name :D

src/sentry_sync.h
```diff
@@ -324,6 +324,12 @@ int sentry__bgworker_start(sentry_bgworker_t *bgw);
  */
 int sentry__bgworker_shutdown(sentry_bgworker_t *bgw, uint64_t timeout);
 
+/**
+ * This will set a preferable thread name for background worker.
+ * Should be executed before worker start
+ */
+void sentry__bgworker_setname(sentry_bgworker_t *bgw, const char *thread_name);
```
Swatinem, line 331: maybe instead of an explicit method, make this a new parameter of `sentry__bgworker_new`

src/sentry_options.c
```diff
@@ -24,6 +24,7 @@ sentry_options_new(void)
     opts->debug = debug && sentry__string_eq(debug, "1");
     sentry_logger_t logger = { sentry__logger_defaultlogger, NULL };
     opts->logger = logger;
+    opts->bgworker_name = sentry__string_clone("Sentry BgWorker");
```
Swatinem, line 27:
```suggestion
    opts->bgworker_name = sentry__string_clone("sentry-transport");
```

Swatinem (commented):
Wow, this looks a lot more complicated than it needs to be, lol. I will look into this in more detail again tomorrow

src/sentry_sync.h
Mixaill: I tried, in this case it becomes even more complicated.
 
For example, WinHTTP transport creates bgworker in `sentry__transport_new_default()` which called by `sentry_options_new()`. With this approach we have no ability to set name via options_set_... because bgworker was already created.

src/sentry_options.c
Mixaill: Need to test it, according to @xamix, it could be only 15 chars (without null-terminator)

Mixaill:
> Wow, this looks a lot more complicated than it needs to be, lol. I will look into this in more detail again tomorrow

Not so complicated. Sentry-related code is fairly small. Optical illusion caused by MSVC tricks :D.

src/sentry_options.c
xamix: Yes it will fail on Linux and returning error code `E_RANGE`.

Swatinem: `sentry-http` is 11 chars which should work and looks a bit nicer

include/sentry.h
```diff
@@ -728,6 +728,18 @@ SENTRY_API void sentry_options_set_ca_certs(
 SENTRY_API const char *sentry_options_get_ca_certs(
     const sentry_options_t *opts);
 
+/**
+ * Configures the name of background worker thread
+ */
+SENTRY_API void sentry_options_set_bgworker_name(
```
Swatinem, line 734:
```suggestion
 * Configures the name of the http transport thread.
 */
SENTRY_API void sentry_options_set_transport_thread_name(
```
… and the same below. Just in case we introduce other threads in the future.

src/sentry_options.h
```diff
@@ -36,6 +36,7 @@ typedef struct sentry_options_s {
     char *dist;
     char *http_proxy;
     char *ca_certs;
+    char *bgworker_name;
```
Swatinem, line 39:
```suggestion
    char *transport_thread_name;
```

Swatinem (approved):
Ah nice. Minor suggestions around naming, otherwise good to go, and thanks for doing this ;-)

PR #416, opened 2020-11-04
fix: Explicitly free lock objects

Swatinem:
This fixes an error reported by Windows Application Verifier.


flub (approved):
Is there anything that can be done to observe this in tests in some way?

Swatinem:
> Is there anything that can be done to observe this in tests in some way?

I don’t think so, the application verifier seems to be some kind of magic that runs at `free` time.

PR #428, opened 2020-11-20
fix: Update crashpad and suppress C5105

Swatinem:
This warning was newly introduced in VS 16.8, but breaks with the
Windows SDKs own `winbase.h` header. We will just suppress it for now.

### DIFF
--- a/src/path/sentry_path_windows.c
+++ b/src/path/sentry_path_windows.c
@@ -9,7 +9,6 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <io.h>
-#include <shlwapi.h>
 #include <stdlib.h>
 #include <sys/locking.h>
 #include <sys/stat.h>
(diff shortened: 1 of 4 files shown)


jan-auer (approved):

flub (approved):

src/path/sentry_path_windows.c
```diff
@@ -125,7 +124,13 @@ sentry__path_dir(const sentry_path_t *path)
     if (!dir_path) {
         return NULL;
     }
-    PathRemoveFileSpecW(dir_path->path);
+
+    // find the filename part and truncate just in front of it if possible
+    sentry_pathchar_t *filename
+        = (sentry_pathchar_t *)sentry__path_filename(dir_path);
+    if (filename > dir_path->path) {
+        *(filename - 1) = L'\0';
```
flub, line 132: C is so scary

flub (approved):

src/path/sentry_path_windows.c
Swatinem: Hahaha, I constantly facepalming why we have to write all this horrible code over and over again just because C does not have a sufficiently good std lib, such as Rusts `Path(Buf)` / `&str/String` 🤦‍♂️

PR #429, opened 2020-11-24
fix: Use correct logic to detect macOS target

Swatinem:


tests/cmake.py
```diff
@@ -210,13 +210,7 @@ def cmake(cwd, targets, options=None):
...
         ]
         print("{} > {}".format(cwd, " ".join(checkcmd)), flush=True)
-        child = subprocess.run(checkcmd, stdout=subprocess.PIPE, cwd=cwd, check=True)
-        sys.stdout.buffer.write(child.stdout)
-        marker = b"Total number of reports: "
-        errors = child.stdout[child.stdout.rfind(marker) + len(marker) :]
-        errors = int(errors[: errors.find(b"\n")])
-        if errors > 0:
-            pytest.fail("code-checker analysis failed")
+        child = subprocess.run(checkcmd, cwd=cwd, check=True)
```
flub, line 213: why is this changing?

flub (approved):

lmdsp:
Thanks for fixing this. 
May I suggest using 

```cpp
if (__builtin_available(macOS 10.14, *)) 
{
    return backtrace_from_fp(addr, ptrs, max_frames);
}
```

This allows the best of both worlds, as it will detect the presence of the function at runtime,
depending on the host OS version instead of the minimum macOS deployment target.

Swatinem:
@lmdsp thanks for the suggestion. I added this and also an explicit test for unwinding from a frame pointer that also triggers this.

tests/cmake.py
Swatinem: ah, code-checker now actually has a return code, so we don’t need to parse its stdout anymore ;-)

flub (approved):

PR #430, opened 2020-11-24
test: Add new rate-limit reason-code tests

Swatinem:


flub (approved):
Would be nice to have something in the commit message (or even on this PR ;)) where this comes from and why it needs to be added

PR #433, opened 2020-12-01
feat: Support modifying attachments after init

Swatinem:
Moves the attachments to the scope, and adds `sentry_add_attachment` and
`sentry_remove_attachment` and wstr variants that modify this attachment
list after calling init. Attachments are identified by their path.

Attachments are still based on file paths and loaded lazily, this does *not* add support for attachments based on in-memory buffers.

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -13,6 +13,17 @@
  * encoding, typically ANSI on Windows, UTF-8 macOS, and the locale encoding on
  * Linux; and they provide wchar-compatible alternatives on Windows which are
  * preferred.
+ *
+ * NOTE on attachments:
+ *
+ * Attachments are read lazily at the time of `sentry_capture_event` or at time
+ * of a hard crash. Relative attachment paths will be resolved according to the
+ * current working directory at the time of envelope creation.
+ * When adding and removing attachments, they are matched according to their
+ * given `path`. No normalization is performed.
+ * When using the `crashpad` backend, the list of attachments that will be added
+ * at the time of a hard crash will be frozen at the time of `sentry_init`, and
+ * later modifications will not be reflected.
  */
 
 #ifndef SENTRY_H_INCLUDED
@@ -821,6 +832,8 @@ SENTRY_API int sentry_options_get_symbolize_stacktraces(
  * `path` is assumed to be in platform-specific filesystem path encoding.
  * API Users on windows are encouraged to use `sentry_options_add_attachmentw`
  * instead.
+ *
+ * See the NOTE on attachments above for restrictions of this API.
  */
 SENTRY_API void sentry_options_add_attachment(
     sentry_options_t *opts, const char *path);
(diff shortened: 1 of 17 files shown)


philipphofmann (commented):
Thanks @Swatinem 💯. The API seems sufficient for synching attachments from Java to native. A method for removing all attachments would make it even easier for us. Currently, we wouldn't even need to remove a specific attachment. Sorry that I didn't mention this earlier.

philipphofmann:
I'm sorry I was wrong. I think we need some more changes to this PR. 
In Python and Java, we have two different types of attachments reflected in the same class. An attachment can either have a byte array or a path. Each attachment also has a filename or and a content type, see 
* https://github.com/getsentry/sentry-java/blob/feat/attachments/sentry/src/main/java/io/sentry/Attachment.java
*  https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/attachments.py

In the SDK API Evolution meeting, we decided it is fine to just copy the byte array to native with the tradeoff of increasing the memory footprint.

include/sentry.h
```diff
@@ -1050,6 +1063,39 @@ SENTRY_API void sentry_remove_transaction(void);
...
 
+/**
+ * Adds a new attachment to be sent along.
+ *
+ * `path` is assumed to be in platform-specific filesystem path encoding.
+ * API Users on windows are encouraged to use `sentry_add_attachmentw` instead.
+ *
+ * See the NOTE on attachments above for restrictions of this API.
+ */
+SENTRY_API void sentry_add_attachment(const char *path);
```
flub, line 1074: Will we always want this to be void?  Does a status return make any sense now or in the future?

include/sentry.h
```diff
@@ -1050,6 +1063,39 @@ SENTRY_API void sentry_remove_transaction(void);
...
+/**
+ * Removes a previously added attachment.
+ *
+ * `path` is assumed to be in platform-specific filesystem path encoding.
+ * API Users on windows are encouraged to use `sentry_remove_attachmentw`
+ * instead.
+ *
+ * See the NOTE on attachments above for restrictions of this API.
+ */
+SENTRY_API void sentry_remove_attachment(const char *path);
```
flub, line 1085: While an idempotent API is good, would a return from which you can tell whether anything happened or not be useful?

include/sentry.h
```diff
@@ -1050,6 +1063,39 @@ SENTRY_API void sentry_remove_transaction(void);
...
 SENTRY_API void sentry_set_level(sentry_level_t level);
 
+/**
+ * Adds a new attachment to be sent along.
+ *
+ * `path` is assumed to be in platform-specific filesystem path encoding.
+ * API Users on windows are encouraged to use `sentry_add_attachmentw` instead.
+ *
+ * See the NOTE on attachments above for restrictions of this API.
+ */
```
flub, line 1073: Does it make sense to explicitly say something about the ownership of `*path`?  (same for remove attachment)

src/sentry_envelope.c
```diff
@@ -364,7 +364,9 @@ sentry_envelope_serialize(const sentry_envelope_t *envelope, size_t *size_out)
 
     sentry__envelope_serialize_into_stringbuilder(envelope, &sb);
 
-    *size_out = sentry__stringbuilder_len(&sb);
+    if (size_out) {
```
flub, line 367: nice 😄

jpnurmi:
Hi @Swatinem, this PR can be closed to get it off the list. Thanks for laying the foundations for modifying attachments at runtime! We incorporated these changes into https://github.com/getsentry/sentry-native/pull/1266 and the feature was released in [0.9.1](https://github.com/getsentry/sentry-native/releases/tag/0.9.1). 🎉

PR #439, opened 2020-12-27
ref(build): generate pdbs in release builds

bruno-garcia:
If built in debug the flags would be duplicated but there's no real negative side effect that I'm aware of.

I've tested in debug and release mode.

Changed the docs to explicitly pass `release` getsentry/sentry-docs#2810

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,8 +47,10 @@ option(SENTRY_BUILD_EXAMPLES "Build sentry-native example(s)" "${SENTRY_MAIN_PRO
 
 if(MSVC)
 	option(SENTRY_BUILD_RUNTIMESTATIC "Build sentry-native with static runtime" OFF)
+	# By default Release builds don't create debug files:
+	set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /Zi")
+	set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "${CMAKE_SHARED_LINKER_FLAGS_RELEASE} /DEBUG /OPT:REF /OPT:ICF")
 endif()
-
 if(LINUX)
 	option(SENTRY_BUILD_FORCE32 "Force a 32bit compile on a 64bit host" OFF)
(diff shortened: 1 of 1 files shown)


Mixaill:
>If built in debug the flags would be duplicated but there's no real negative side effect that I'm aware of.

They should not, because of `_RELEASE` suffix

CMakeLists.txt
```diff
@@ -47,8 +47,10 @@ option(SENTRY_BUILD_EXAMPLES "Build sentry-native example(s)" "${SENTRY_MAIN_PRO
 
 if(MSVC)
 	option(SENTRY_BUILD_RUNTIMESTATIC "Build sentry-native with static runtime" OFF)
+	# By default Release builds don't create debug files:
+	set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /Zi")
+	set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "${CMAKE_SHARED_LINKER_FLAGS_RELEASE} /DEBUG /OPT:REF /OPT:ICF")
```
Mixaill, line 52: This have no negative impact, but `/OPT:REF /OPT:ICF` part was not announced in PR summary :D

Mixaill:
By the way, do we really need it? 

CMake provides `RelWithDebInfo` target, which should be used in case if you want to get debug information.

bruno-garcia:
Foes thst work with VS properly? Curious because VS can be built in debug or release after cake build through VS Co configuration.

I dint have any preference other than I'd like optimized builds (release) but get pdbs generated so I can upload to sentry.

Mixaill:
>Curious because VS can be built in debug or release after cake build through VS Co configuration.

It works perfectly for CMake and MSVS.  CMake generates 4 configurations for MSVS:
* Debug
* Release
* RelWithDebInfo
* MinSizeRel

which can be specified command simillar to:
`cmake --build <cmake_build_dir> --config RelWithDebInfo --target INSTALL `

bruno-garcia:
Works for me.
I'll update the docs PR to add this instead. 
Thanks

PR #449, opened 2021-01-05
feat: Add sentry_reinstall_backend API

Swatinem:
This is a rebase of #388, ready to be merged.


jan-auer (approved):
The name is potentially misleading, should we go with something like `sentry_reinstall_backend`? 

We should make it clear that this is potentially dangerous and should only be called when someone is **sure** that this needs to be done.

PR #451, opened 2021-01-05
ci(gha): Switch to release workflow

jan-auer:
Switches the release process to the centralized release action. Craft is run directly on the host.


Swatinem (approved):

BYK (approved):
Looks good. Did not check line-by-line tho :)

PR #453, opened 2021-01-06
Add minimal Qt integration

torarnv:
The integration turns Qt logging via e.g. qDebug() into breadcrumbs.

Enable the integration by passing -DSENTRY_INTEGRATION_QT=ON during
configure time. This will add a runtime dependency on the QtCore
library to sentry.

### DIFF
--- a/src/integrations/sentry_integration_qt.cpp
+++ b/src/integrations/sentry_integration_qt.cpp
@@ -1,0 +1,59 @@
+#include "sentry_integration_qt.h"
+#include "sentry_boot.h"
+
+#include <QtCore/qglobal.h>
+#include <QtCore/qstring.h>
+
+static QtMessageHandler originalMessageHandler = nullptr;
+
+static const char *
+logLevelForMessageType(QtMsgType msgType)
+{
+    switch (msgType) {
+    case QtDebugMsg:
+        return "debug";
+    case QtWarningMsg:
+        return "warning";
+    case QtCriticalMsg:
+        return "error";
+    case QtFatalMsg:
+        return "fatal";
+    case QtInfoMsg:
+        Q_FALLTHROUGH();
+    default:
+        return "info";
+    }
+}
+
+static void
+sentry_qt_messsage_handler(
+    QtMsgType type, const QMessageLogContext &context, const QString &message)
+{
+    sentry_value_t crumb
+        = sentry_value_new_breadcrumb("default", qUtf8Printable(message));
+
+    sentry_value_set_by_key(
+        crumb, "category", sentry_value_new_string(context.category));
+    sentry_value_set_by_key(
+        crumb, "level", sentry_value_new_string(logLevelForMessageType(type)));
+
+    sentry_value_t location = sentry_value_new_object();
+    sentry_value_set_by_key(
+        location, "file", sentry_value_new_string(context.file));
+    sentry_value_set_by_key(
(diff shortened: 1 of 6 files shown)


Mixaill:
Maybe,

`sentry_integration_qt` --> `sentry_logger_qt`
SENTRY_INTEGRATION_QT --> `SENTRY_LOGGER_DEFAULT=qt`

torarnv:
> Maybe,
> 
> `sentry_integration_qt` --> `sentry_logger_qt`
> `SENTRY_INTEGRATION_QT` --> `SENTRY_LOGGER_DEFAULT=qt`

Thanks for you review!

I tried to follow the existing pattern/prefixes of `SENTRY_BACKEND_X`, `SENTRY_TRANSPORT_Y`, as well as the existing use of "integrations" for collectively referring to optional add-ons for frameworks, eg: https://docs.sentry.io/platforms/javascript/configuration/integrations/custom/

If we add more of these integrations (and even if we don't), I think `SENTRY_INTEGRATION_` is a good prefix.

torarnv:
> Maybe,
> 
> `SENTRY_INTEGRATION_QT` --> `SENTRY_LOGGER_DEFAULT=qt`
> 
> and place it inside `sentry_logger.c`
> 
> ```
> #if SENTRY_LOGGER_DEFAULT == "qt"
> ...
> #elif
>  ...
> ```

I think you may have misunderstood what the patch does 😊 It's not adding a logging backend to sentry to log sentry-messages via Qt's logging system. It's adding an integration that feeds any logs emitted from Qt as breadcrumbs in a future crash report.

Mixaill:
Yeah, that's why I deleted my comment :)

torarnv:
😊 Thanks though!!

src/integrations/sentry_integration_qt.cpp
```diff
@@ -0,0 +1,59 @@
...
+    default:
+        return "info";
+    }
+}
+
+static void
+sentry_qt_messsage_handler(
+    QtMsgType type, const QMessageLogContext &context, const QString &message)
+{
+    sentry_value_t crumb
```
bruno-garcia, line 32: I wonder if we can have a better value for `type` here.
Perhaps even `NULL` is better, the UI adjusts to deal with the data it has and it'll have category and level already.

```
type (optional)
The type of breadcrumb. By default, all breadcrumbs are recorded as default, which makes them appear as a Debug entry, but Sentry provides other types that influence how the breadcrumbs are rendered. For more information, see the description of recognized breadcrumb types.

By default, all breadcrumbs are recorded as default, which makes them appear as a Debug entry, but Sentry provides other types that influence how the breadcrumbs are rendered. For more information, see the description of recognized breadcrumb types.
```

There are some docs here: https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/#breadcrumb-types

src/integrations/sentry_integration_qt.h
```diff
@@ -0,0 +1,15 @@
+#ifndef SENTRY_INTEGRATION_QT_H_INCLUDED
+#define SENTRY_INTEGRATION_QT_H_INCLUDED
+
+#ifdef __cplusplus
```
Swatinem, line 4: I think its fine this way, but would rather wrap the function in the `.cpp` file in an `extern "C"` block, similar to how the other parts that depend on C++ work.

Swatinem (approved):
This looks very good!

PR #456, opened 2021-01-11
fix: Ensure collection methods on null do not crash

Swatinem:
This tests all the value methods intended for list/object do
not crash when called on a null value, and fixes a faulty return
value while at it.

In addition, this changes a few things around handling of NaN.

CC @marandaneto


marandaneto (approved):
LGTM

PR #457, opened 2021-01-11
Space between comments

Astrodevil:
By giving space between comments and code details are clearly visible


Swatinem:
I will just go ahead and close this, at it actually regresses readability in some cases, and doesn’t even pass clang-format.

PR #461, opened 2021-01-15
feat: expose sentry-native version to CMake project and package

### LINKED ISSUE #460: [CMake] Utilize PROJECT_VERSION
It would be great if the current `sentry-native` version was easely accessible through CMake, by applying the `VERSION` property to `project()` (https://cmake.org/cmake/help/latest/command/project.html).
In my current usecase, it would allow me to automatically detect each release change and update the associated package on my remote CDN.

Mixaill:
* expose sentry-native version to CMake project, fixes #460 

* expose sentry-native version to CMake package, it allows things like

CMakeLists.txt
```
...
find_package(sentry 0.4.5)
...
```

CMake output
```
CMake Warning at CMakeLists.txt:4 (find_package):
  Could not find a configuration file for package "sentry" that is compatible with requested version "0.4.5".
  The following configuration files were considered but not accepted:
    D:/Projects/VCS/github/getsentry.sentry-native/~install/lib/cmake/sentry/sentry-config.cmake, version: 0.4.4
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,17 @@ else()
 	cmake_minimum_required (VERSION 3.10)
 endif()
 
-project(Sentry-Native LANGUAGES C CXX ASM)
+#read sentry-native version
+file(READ "include/sentry.h" _SENTRY_HEADER_CONTENT)
+string(REGEX MATCH "#define SENTRY_SDK_VERSION \"([0-9\.]+)\"" _SENTRY_VERSION_MATCH "${_SENTRY_HEADER_CONTENT}")
+set(SENTRY_VERSION "${CMAKE_MATCH_1}")
+unset(_SENTRY_HEADER_CONTENT)
+unset(_SENTRY_VERSION_MATCH)
+
+project(Sentry-Native 
+	LANGUAGES C CXX ASM 
+	VERSION ${SENTRY_VERSION}
+)
 
 set(SENTRY_MAIN_PROJECT OFF)
 if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -443,6 +453,9 @@ endif()
 include(CMakePackageConfigHelpers)
 configure_package_config_file(sentry-config.cmake.in sentry-config.cmake
 	INSTALL_DESTINATION "${CMAKE_INSTALL_CMAKEDIR}")
+write_basic_package_version_file(sentry-config-version.cmake 
+	VERSION ${SENTRY_VERSION}
+	COMPATIBILITY AnyNewerVersion)
```
Swatinem, line 458: Haven’t looked at the docs for this yet, but is **any** semver compatible?

Swatinem (approved):
Nice! Another alternative would have been to just hardcode the number, and update our version bump script accordingly.

CMakeLists.txt
Mixaill: for example, `find_package(sentry 1.2.3)`

* `AnyNewerVersion` --> any version >= 1.2.3
* `SameMajorVersion` --> 1.x.x
* `SameMinorVersion` --> 1.2.x
* `ExactVersion` --> 1.2.3

Swatinem: I think in that case we should opt for `SameMinorVersion` as long as we are `< 1.0.0`

Mixaill: changed, with a reminder for >=1.0.0

Swatinem (approved):

Swatinem:
🤦‍♂️ so `SameMinorVersion` is only supported in CMake 3.11, and we have to support 3.10 for usage in sentry-java (android)

Mixaill:
Then we should unconditionally switch to `SameMajorVersion` 🤷‍♂️

PR #467, opened 2021-01-26
feat: Add os context to events automatically

Swatinem:
This captures OS context for the major OS (windows, macos, linux).

Q:

* [x] for macOS big sur, it gives `version: 11.1`, should I append a trailing `.0` if the patch version is missing?
* [x] for linux/android, an example version is `5.4.61-android11-0-00791-gbad091cc4bf3-ab6833933`, should I split that off into a `build` identifier rather?


Mixaill:
By the way, maybe is it worth for Linux to parse `/etc/*-release` to get info about distro?

Swatinem:
@Mixaill I agree that the linux kernel version on its own has little meaning, however this is consistent with the information we can extract from breakpad/crashpad minidumps.

src/sentry_os.c
```diff
@@ -0,0 +1,179 @@
...
+
+    size_t num_dots = 0;
+    for (size_t i = 0; i < buf_len; i++) {
+        if (buf[i] == '.') {
+            num_dots += 1;
+        }
+    }
+    if (num_dots < 2 && buf_len + 3 < sizeof(buf)) {
+        strcat(buf, ".0");
+    }
```
flub, line 91: I think I would opt not to do this, it puts the SDK into the business of trying to interpret what the OS wants to describe itself as and that seems like a game you can never win.

jan-auer:
I've seen crashpad include a custom section with lsb-release and friends, so at least there would be a _possibility_ to do this. 
Generally, we do not have a good answer for how to report Linux consistently across platforms. In Electron, we parse these files, but on any other platform I'm aware of we do not.

src/sentry_os.c
```diff
@@ -0,0 +1,179 @@
...
+
+sentry_value_t
+sentry__get_os_context(void)
+{
+    sentry_value_t os = sentry_value_new_object();
+    if (sentry_value_is_null(os)) {
+        return os;
+    }
+
+    sentry_value_set_by_key(os, "type", sentry_value_new_string("os"));
```
jan-auer, line 16: You can omit the `type` if you attach it as `"os"` into contexts.

src/sentry_os.c
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+    char *build_start = build;
+    if (build[0] == '-' || build[0] == '.') {
+        build_start++;
+    }
+
+    if (build_start[0] != '\0') {
+        sentry_value_set_by_key(
+            os, "build", sentry_value_new_string(build_start));
```
flub, line 154: why not put all of `uts.release` into `build`?

src/sentry_os.c
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    build[0] = '\0';
+
+    sentry_value_set_by_key(os, "name", sentry_value_new_string(uts.sysname));
+    sentry_value_set_by_key(
+        os, "version", sentry_value_new_string(uts.release));
```
flub, line 161: why not `uts.version`?

jan-auer (approved):

src/sentry_os.c
Swatinem: a lot of tools, including crashpad, and our own normalization try to parse 3 semver components. also big sur is the first version that defaults to 2 components, and these things are super stable on mac at least.

Swatinem:
@flub uname gives the following:
```
uts.release: 5.10.6-arch1-1
uts.version: #1 SMP PREEMPT Sat, 09 Jan 2021 18:22:35 +0000
```

The `version` part is pretty useless IMO.

src/sentry_os.c
```diff
@@ -0,0 +1,175 @@
...
+    sentry_value_t os = sentry_value_new_object();
+    if (sentry_value_is_null(os)) {
+        return os;
+    }
+
+    sentry_value_set_by_key(os, "name", sentry_value_new_string("Windows"));
+
+    void *ffibuf = NULL;
+
+    DWORD size = GetFileVersionInfoSizeW(L"kernel32.dll", NULL);
```
Mixaill, line 20: Why `kernel32.dll` version is used instead of `GetVersionExW()`?

Libraries versions does not match the Windows build version.

image

Swatinem: Again, I copied whatever crashpad is doing, to be consistent with those tools. I was a bit surprised myself that the version does not reflect the 20H2 release number.

I will discuss this internally.

Mixaill: Also, you can use `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion` (that's what winver actually use)

* Version: `<CurrentMajorVersionNumber>.<CurrentMinorVersionNumber>`
* Full Build `<CurrentMajorVersionNumber>.<CurrentMinorVersionNumber>.<CurrentBuildNumber>.<UBR>`
* Edition: `<EditionID>`
* Update name: `<DisplayVersion>`

Mixaill: Also, you can combine `GetVersionExW` and registry to detect if application is running under compatibility mode.

Mixaill:
@Swatinem any further work is planned in a short time or can I improve behavior of Windows version detection?

Swatinem:
> any further work is planned in a short time

Well, I call this solved from my end.

The thing is that I built this to have the same behavior as crashpad. They also have a comment related to this here:

https://github.com/getsentry/crashpad/blob/cb520bd41fdea18a4d842efabeac61a2a5beaa61/snapshot
/win/system_snapshot_win.cc#L92-L112

There should be a compelling reason to deviate from that. I haven’t looked at the specific code in breakpad yet, but I expect it to do something similar.

Swatinem:
Breakpad is using windbg APIs for this: https://docs.microsoft.com/en-us/windows/win32/api/minidumpapiset/nf-minidumpapiset-minidumpwritedump

I will see what that gives me in terms on windows version…

Swatinem:
So that minidump gives me `10.0.19042` as the OS version, but honestly I don’t think its worth putting too much effort into this.

Mixaill:
But, in any way, if I provide PR with improvements in this area, it will be accepted?

Swatinem:
@jan-auer was a bit hesitant, maybe you can elaborate your concerns a bit more?

I have no objections if its equally well supported and a clear improvement.

PR #470, opened 2021-02-02
fix: Avoid memory leak when adding breadcrumbs

### LINKED ISSUE #469: potential mem leak when calling breadcrumb
I found our software start mem leaking 1~2 Mb/hour. After troubleshooting and speculation, we fixated on breadcrumb. I have the below code: 
```
void syb_sentry_add_breadcrumb(const std::string& message) {
    sentry_add_breadcrumb(sentry_value_new_breadcrumb(nullptr, message.c_str()));
 }

```

After artifically trigger it 1000 times more, we could see memory increase dramatically:
```
void syb_sentry_add_breadcrumb(const std::string& message) {
  static int cnt = 0;
  std::cout << "add breadcrumb " << cnt++ << std::endl;
  for (int i = 0; i < 1000; ++i) {
    sentry_add_breadcrumb(sentry_value_new_breadcrumb(nullptr, message.c_str()));
  }
}
```

I am saying `potential` since it could be something bad I have done or I am using sentry in the wrong way, however, the interface is pretty easy, just pass a string. 
I also see your example.c where it is how sentry_add_breadcrumb is used. 

//////////////////////////////////////////////////////////////////////////////////////
below is just my speculation in your source code: 
```
sentry_add_breadcrumb(sentry_value_t breadcrumb)
{
    sentry_value_incref(breadcrumb);
    // the `no_flush` will avoid triggering *both* scope-change and
    // breadcrumb-add events.
    SENTRY_WITH_SCOPE_MUT_NO_FLUSH (scope) {
        sentry__value_append_bounded(
            scope->breadcrumbs, breadcrumb, SENTRY_BREADCRUMBS_MAX);
    }

    bool was_added = false;
        if (options->backend && options->backend->add_breadcrumb_func) {
            options->backend->add_breadcrumb_func(options->backend, breadcrumb);
            was_added = true;
        }
    }
    if (!was_added) {
        sentry_value_decref(breadcrumb);
    }
}
```

Looking at the source code, it is a bit strange to me that we have sentry_value_incref and sentry_value_decref with the condition `was_added`. It feels natural to let the options->backend->add_breadcrumb_func's implementation to increase the release counter. `was_added` seems to be reluctant and should be removed. we do not need to call sentry_value_incref  and sentry_value_decref at all, since I guess it should be called inside the implementation of  add_breadcrumb_func. 

I think the  
`sentry_value_incref(breadcrumb); and     bool was_added = false;         sentry_value_decref(breadcrumb);`
should be removed altogether then it should work as expected though it seems to be smart and reasonable, 
as an old Chinese saying, "draw legs on a snake".

Swatinem:
And avoid leaking on some error conditions as well.

Fixes #469


weidonglian:
Really nice work!!! Love it. Expect to be released soon.

src/sentry_backend.h
```diff
@@ -20,6 +20,7 @@ typedef struct sentry_backend_s {
     void (*except_func)(
         struct sentry_backend_s *, const struct sentry_ucontext_s *);
     void (*flush_scope_func)(struct sentry_backend_s *);
+    // NOTE: the breadcrumb is borrowed and does not need to be `decref`-d.
```
flub, line 23: Not going to lie, this is ambiguous to me.  "is borrowed" i understand that the caller retains ownership, so if the caller owns are they not responsible for decref-ing?

src/sentry_core.c
```diff
@@ -469,23 +469,18 @@ sentry_remove_user(void)
...
     }
-    if (!was_added) {
-        sentry_value_decref(breadcrumb);
+
+    // the `no_flush` will avoid triggering *both* scope-change and
+    // breadcrumb-add events.
+    SENTRY_WITH_SCOPE_MUT_NO_FLUSH (scope) {
+        sentry__value_append_bounded(
+            scope->breadcrumbs, breadcrumb, SENTRY_BREADCRUMBS_MAX);
     }
```
flub, line 484: IIUC `sentry__value_append_bounded()` now takes ownership, because it DECREFs on error.  So this means this whole function takes ownership?

tests/unit/test_value.c
```diff
@@ -390,3 +390,27 @@ SENTRY_TEST(value_wrong_type)
...
+    sentry_value_append(list, sentry_value_new_object());
+    sentry_value_append(list, sentry_value_new_object());
+    sentry_value_append(list, sentry_value_new_object());
+
+    // decref the existing values correctly on bounded append
+    sentry__value_append_bounded(list, sentry_value_new_object(), 2);
+    sentry__value_append_bounded(list, sentry_value_new_object(), 2);
+
+    sentry_value_decref(list);
+}
```
flub, line 416: i assume this test only asserts things when valgrind is running?

src/sentry_core.c
flub: "takes ownership" in rust lingo, or "steals the reference" in CPython lingo i guess

tests/unit/test_value.c
weidonglian: Actually, the test could assert the ref count in simplicity if possible.

src/sentry_core.c
Swatinem: "moves", in C lingo :-D But yes, this function always worked that way.

tests/unit/test_value.c
Swatinem: 👍 updated the test to check for refcount explicitly.

flub (approved):

PR #471, opened 2021-02-02
feat: Add ability to control the max_breadcrumbs

### LINKED ISSUE #438: How to control the number of breadcrumbs attached to the captured event?
**How to control the number of breadcrumbs attached to the captured event?**

I only found the SENTRY_BREADCRUMBS_MAX macro and it feels bad to change it directly. There ought to be an API or an option? 

Thanks

Swatinem:
Followup on #470 
fixes #438

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## Unreleased
+
+**Features**:
+
+- Events will automatically get an `os` context with OS version information.
+- Added a new `max_breadcrumbs` option.
+
+**Fixes**:
+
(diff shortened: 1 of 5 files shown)


jan-auer (commented):
There should be an absolute maximum that's allowed to be set. I believe this should be the same as the default 100.

Swatinem:
> There should be an absolute maximum that's allowed to be set.

Discussed this with the rest of the SDK devs, consensus was to not enforce a hard limit. There is a hard limit though in relay (event size).

flub (approved):

PR #476, opened 2021-02-09
ref: Update Android Unwinder

Swatinem:
This updates the `libunwindstack` unwinder that is used on android.

Based on https://github.com/getsentry/libunwindstack-ndk/pull/3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,7 @@ if(NOT CMAKE_C_STANDARD)
 endif()
 
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD 17)
 endif()
 
 include(GNUInstallDirs)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ if(NOT CMAKE_C_STANDARD)
 endif()
 
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD 17)
```
marandaneto, line 33: would this cause problems like https://github.com/getsentry/sentry-native/pull/463?

Swatinem: I don’t think so, c++17 std is supported in CMake 3.10. My concern is rather the NDK version that needs an appropriate compiler.

This is just to get started, we probably want to only do this on android, since we better control the compiler versions that we want to support there.

jan-auer (approved):
LGTM, under the assumption that the unwinder does not use full debug information, as this would be slow.

Swatinem:
> does not use full debug information

I at least removed the gnu debug data handling as it was before.

PR #480, opened 2021-02-15
feat: Support Breakpad on Android

Swatinem:
Actually the changes required to the cmake files was minimal, hopefully CI passes as well.


tests/conditions.py
```diff
@@ -14,6 +14,8 @@
 has_breakpad = (
     not is_valgrind
     and not is_kcov
+    # XXX: we support building breakpad, and it captures minidumps when run through sentry-android,
+    # however running it from an `adb shell` does not work correctly :-(
```
Swatinem, line 18: This is a bit unfortunate :-(

flub (approved):

PR #492, opened 2021-03-03
fix: Resolve some ApplicationVerified Issues on Windows

### LINKED ISSUE #491: ApplicationVerifier: VERIFIER_STOP in function sentry__bgworker_start because of a handle leak
**Description**

When testing sentry-native with Microsoft AppVerifier flag "Leak" there is a HANDLE leak detected.

- [ ] During build
- [x] During run-time

The function `sentry__bgworker_start` seems to leak a handle on this line `if (sentry__thread_spawn(&bgw->thread_id, &worker_thread, bgw) != 0) {`

The `sentry__thread_spawn` returns a handle that is just ignored. It should be closed with [CloseHandle](https://docs.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-closehandle).

```cpp
int
sentry__bgworker_start(sentry_bgworker_t *bgw)
{
    SENTRY_TRACE("starting background worker thread");
    sentry__atomic_fetch_and_add(&bgw->running, 1);
    // this incref moves the reference into the background thread
    sentry__bgworker_incref(bgw);
    if (sentry__thread_spawn(&bgw->thread_id, &worker_thread, bgw) != 0) {    // <= sentry__thread_spawn returns a HANDLE that is then leaked
        sentry__atomic_fetch_and_add(&bgw->running, -1);
        sentry__bgworker_decref(bgw);
        return 1;
    }
    return 0;
}
```

**Environment**

- OS: [Windows 7 and above]

**Steps To Reproduce**

1. Build a little test app and link sentry 0.4.7
2. Download the Windwos Sdk installer from [here](https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk/)
3. Install AppVerifier with `winsdksetup.exe /ceip off /features OptionId.AvrfExternal /quiet`
4. Configure AppVerifier with `C:\Windows\System32\appverif.exe -enable DangerousAPIs DirtyStacks Exceptions Handles Heaps Leak Locks SRWLock Memory TLS Threadpool -for <TEST_APP.exe>`
5. Run the app with debugger attached

Swatinem:
- Correctly initialize non-static Mutex
- Correctly Close Thread Handle

fixes #491

### DIFF
--- a/src/sentry_sync.h
+++ b/src/sentry_sync.h
@@ -157,11 +157,19 @@ typedef struct sentry__winmutex_s sentry_mutex_t;
 #    define sentry__mutex_free(Lock)                                           \
         DeleteCriticalSection(&(Lock)->critical_section)
 
+#    define sentry__thread_init(ThreadId) *ThreadId = INVALID_HANDLE_VALUE
 #    define sentry__thread_spawn(ThreadId, Func, Data)                         \
         (*ThreadId = CreateThread(NULL, 0, Func, Data, 0, NULL),               \
             *ThreadId == INVALID_HANDLE_VALUE ? 1 : 0)
 #    define sentry__thread_join(ThreadId)                                      \
         WaitForSingleObject(ThreadId, INFINITE)
+#    define sentry__thread_free(ThreadId)                                      \
+        do {                                                                   \
+            if (*ThreadId != INVALID_HANDLE_VALUE) {                           \
+                CloseHandle(*ThreadId);                                        \
+            }                                                                  \
+            *ThreadId = INVALID_HANDLE_VALUE;                                  \
+        } while (0)
 
 #    if _WIN32_WINNT < 0x0600
 typedef CONDITION_VARIABLE_PREVISTA sentry_cond_t;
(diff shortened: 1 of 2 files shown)


src/sentry_sync.h
```diff
@@ -143,28 +150,37 @@ typedef struct sentry__winmutex_s sentry_mutex_t;
...
             INIT_ONCE_STATIC_INIT, { 0 }                                       \
         }
+#    define sentry__mutex_init(Lock) sentry__winmutex_init(Lock)
 #    define sentry__mutex_lock(Lock) sentry__winmutex_lock(Lock)
 #    define sentry__mutex_unlock(Lock)                                         \
         LeaveCriticalSection(&(Lock)->critical_section)
 #    define sentry__mutex_free(Lock)                                           \
         DeleteCriticalSection(&(Lock)->critical_section)
-
+g
```
Mixaill, line 159: ?

jan-auer (approved):
Under the assumption that `sentry__thread_free` doesn't have to be thread-safe, this LGTM.

Swatinem:
> Under the assumption that `sentry__thread_free` doesn't have to be thread-safe

So far its only called in the decref, which should be.

PR #500, opened 2021-03-23
ref: Optimize jsonwriter to allocate less in some cases

Swatinem:
Refactors the jsonwriter a bit, taking an existing stringwrite where possible to avoid intermediate allocations.

### DIFF
--- a/src/sentry_envelope.c
+++ b/src/sentry_envelope.c
@@ -212,11 +212,17 @@ sentry__envelope_add_event(sentry_envelope_t *envelope, sentry_value_t event)
         return NULL;
     }
 
+    sentry_jsonwriter_t *jw = sentry__jsonwriter_new(NULL);
+    if (!jw) {
+        return NULL;
+    }
+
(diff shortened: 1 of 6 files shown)


src/sentry_envelope.c
```diff
@@ -212,11 +212,17 @@ sentry__envelope_add_event(sentry_envelope_t *envelope, sentry_value_t event)
...
+    if (!jw) {
+        return NULL;
+    }
+
     sentry_value_t event_id = sentry__ensure_event_id(event, NULL);
 
     item->event = event;
-    item->payload = sentry_value_to_json(event);
-    item->payload_len = strlen(item->payload);
+    sentry__value_write_into_jsonwriter(jw, event);
```
flub, line 223: This name is confusing together with the `sentry__jsonwriter_into_string()`.  In particular the `into` part.  Can this not be called `sentry__jsonwriter_write_value()`?

flub (approved):

PR #509, opened 2021-04-08
Add support for Qt 6

torarnv:


torarnv:
@bruno-garcia

bruno-garcia (commented):
LGTM but @Swatinem might have thoughts

torarnv:
Just a sec, let me push an update

torarnv:
Ok, ready @Swatinem  :)

Swatinem (approved):
So this means if you have both development packages installed, you will always the 6 rather than 5. Is there a way to control that?

torarnv:
> So this means if you have both development packages installed, you will always the 6 rather than 5. Is there a way to control that?

Correct, that's the current behavior. There is [`QT_DEFAULT_MAJOR_VERSION` ](https://doc.qt.io/qt-6/cmake-variable-reference.html#qt-default-major-version), but I need to check with our CMake experts whether that is applicable here. If not I could add custom logic to take it into account in a followup perhaps?

torarnv:
Fixed. You can now choose the version with ` QT_DEFAULT_MAJOR_VERSION`

bruno-garcia:
@Swatinem Docs are landing too: https://github.com/getsentry/sentry-docs/pull/3370
Since it's approved by you, anything blocking to merge? 🥳

Swatinem (approved):

PR #510, opened 2021-04-08
fix: Validate tm put into strftime

Swatinem:


flub (approved):
LGTM but CI is less convinced

Swatinem:
> CI is less convinced

It was #511 basically a CI update started creating an error.

PR #517, opened 2021-04-16
feat: Add more Event Payload convenience methods

### LINKED ISSUE #235: Adding exception details appears to mess up the report on Sentry
I'm not clear if this should be added here or in another location. Please move the issue as necessary.

We're manually reporting exceptions in our app by adding in the `exception` value when sending an event, as described at https://docs.sentry.io/platforms/native/#exceptions:

```
    exc = sentry_value_new_object();
    sentry_value_set_by_key(exc, "type", sentry_value_new_string(ExceptionClassName));
    sentry_value_set_by_key(exc, "value", sentry_value_new_string(Message));
    sentry_value_set_by_key(event, "exception", exc);
```

As soon as we do this, the report on Sentry ends up formatted wrongly, losing visibility of the call stack:

image

Using sentry-native 0.2.3. This issue arises both on sentry.io and our on premise sentry instance.

Originally raised at https://forum.sentry.io/t/corrupted-display-when-exception-data-is-set-using-native-sdk/9167/3

### LINKED ISSUE #376: Strange stacktrace for inline functions (Mac)
**Description**

For some reports I have seen incorrect stacktraces. Seems to happen only for inlined functions,
and they are clearly wrong. It shows the same address, of the inlining function, for all those frames,
maybe some postprocessing done to the stacktrace?

Here's one example issue (hope this is not public?)
https://sentry.io/organizations/reason-studios/issues/1879135326

We use this to capture the stack trace:
sentry_event_value_add_stacktrace(event, nullptr, 0);

It's hard to know where things go wrong; in Sentry native or processing on the server or if our
symbols are wrong somehow, but I thought I'd report it in case you recognise the problem.

I have added our own stacktracing as attribute to our events to see if there's a difference.

- [ ] During build
- [? ] During run-time
- [? ] When capturing a hard crash

**Environment**

- OS: Only macOS so far (10.14, 10.15)
- Compiler: Xcode 11
- CMake version and config: [?, SENTRY_BACKEND=none]

**Steps To Reproduce**

**Log output**

### LINKED ISSUE #503: How to fill in Stack trace when manually capturing an exception?  (Just like what happens when auto captured)
**Description**

Hi thanks for the lib! I wonder how can I fill in the stacktrace? Currently when I report an error manually, it only has exception messages. But as we know, stack traces are very important to debug!

In addition, I do see automatic stack traces when an unexpected exception happens. Thus I guess sentry should have an api like `get_the_current_stack_trace_as_sentry_data_structures()` ;)

I have tried to use the following:

```
            sentry_value_set_by_key(event, "exception", exception_obj);
            sentry_event_value_add_stacktrace(event, nullptr, 0);
            sentry_capture_event(event);
```

but then error occurs:

```
2021-03-26 09:47:05.010 6468-6733/com.cjy.yplusplus E/Sentry: An exception occurred while processing event by processor: io.sentry.android.core.DefaultAndroidEventProcessor
    java.lang.NullPointerException: Attempt to invoke virtual method 'long java.lang.Long.longValue()' on a null object reference
        at io.sentry.android.core.util.MainThreadChecker.isMainThread(MainThreadChecker.java:40)
        at io.sentry.android.core.DefaultAndroidEventProcessor.processNonCachedEvent(DefaultAndroidEventProcessor.java:213)
        at io.sentry.android.core.DefaultAndroidEventProcessor.process(DefaultAndroidEventProcessor.java:143)
        at io.sentry.SentryClient.processEvent(SentryClient.java:185)
        at io.sentry.SentryClient.captureEvent(SentryClient.java:76)
        at io.sentry.Hub.captureEvent(Hub.java:82)
        at io.sentry.Sentry.captureEvent(Sentry.java:254)
        at io.sentry.HubAdapter.captureEvent(HubAdapter.java:27)
        at io.sentry.OutboxSender.processEnvelope(OutboxSender.java:141)
        at io.sentry.OutboxSender.processFile(OutboxSender.java:68)
        at io.sentry.OutboxSender.processEnvelopeFile(OutboxSender.java:101)
        at io.sentry.android.core.EnvelopeFileObserver.onEvent(EnvelopeFileObserver.java:59)
        at android.os.FileObserver$ObserverThread.onEvent(FileObserver.java:163)
        at android.os.FileObserver$ObserverThread.observe(Native Method)
        at android.os.FileObserver$ObserverThread.run(FileObserver.java:113)
```

Ok here is my hack:

```
    sentry_value_t threads = sentry_value_get_by_key(event, "threads");
    sentry_value_t values = sentry_value_get_by_key(threads, "values");
    sentry_value_t thread = sentry_value_get_by_index(values, 0);
    sentry_value_set_by_key(thread, "id", sentry_value_new_int32(-1));
```

but it still does not work. now no errors, but no stacktraces.

the output of `sentry_value_to_json(event)`:

```
{"event_id":"9aabc52f-2ece-439d-f46e-3206346ba765","timestamp":"2021-03-26T02:28:09.814Z","exception":{"values":[{"type":"Exception","value":"cv_error_callback see error status=-215 func_name=debug_throw_exception err_msg=1 == 0 file_name=/Users/tom/QAPMain/Development/rzzsdxx_frontend/vision_utils/ios/Classes/basic_vision_utils.cpp line=190"}]},"threads":{"values":[{"stacktrace":{"frames":[{"instruction_addr":"0x71e6806698"},{"instruction_addr":"0x71aaf8e0a8"},{"instruction_addr":"0x71ab1369cc"},{"instruction_addr":"0x71ab136f90"},{"instruction_addr":"0x71aaf8d980"},{"instruction_addr":"0x71aaf8d44c"},{"instruction_addr":"0x71c5415cc4"},{"instruction_addr":"0x71c5418548"}]},"id":999}]}
```

but the website:

image

Therefore, what I want: **Can we fill in Stack trace when manually capturing an exception? (Just like what happens when auto captured)?** Now there is only instruction addresses - no things like function names, etc!

---

- [ ] During build
- [x] During run-time

**Environment**

- OS: [e.g. Windows 10, 64-bit] android
- Compiler: [e.g. MSVC 19] android ndk compiler
- CMake version and config: [e.g. 3.17.2, SENTRY_BACKEND=inproc]

**Steps To Reproduce**

**Log output**
N/A

Swatinem:
Adds:
* `sentry_value_new_exception`
* `sentry_value_new_thread`
* `sentry_value_new_stacktrace`
* `sentry_event_value_add_exception`
* `sentry_event_value_add_thread`

Deprecates `sentry_event_value_add_stacktrace`

Fixes #235
Fixes #376
Fixes #503

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -340,7 +340,9 @@ typedef enum sentry_level_e {
 /**
  * Creates a new empty Event value.
  *
- * See https://develop.sentry.dev/sdk/event-payloads/
+ * See https://docs.sentry.io/platforms/native/enriching-events/ for how to
+ * further work with events, and https://develop.sentry.dev/sdk/event-payloads/
+ * for a detailed overview of the possible properties of an Event.
  */
 SENTRY_API sentry_value_t sentry_value_new_event(void);
 
@@ -366,6 +368,10 @@ SENTRY_API sentry_value_t sentry_value_new_breadcrumb(
 
 /**
  * Creates a new Exception value.
+ *
+ * This is intended for capturing language-level exception, such as from a
+ * try-catch block. `type` and `value` here refer to the exception class and
+ * a possible description.
  *
  * See https://develop.sentry.dev/sdk/event-payloads/exception/
  *
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -338,12 +338,16 @@ typedef enum sentry_level_e {
 } sentry_level_t;
 
 /**
- * Creates a new empty event value.
+ * Creates a new empty Event value.
+ *
+ * See https://develop.sentry.dev/sdk/event-payloads/
  */
 SENTRY_API sentry_value_t sentry_value_new_event(void);
 
 /**
- * Creates a new message event value.
+ * Creates a new Message Event value.
+ *
+ * See https://develop.sentry.dev/sdk/event-payloads/message/
  *
  * `logger` can be NULL to omit the logger value.
  */
@@ -351,12 +355,68 @@ SENTRY_API sentry_value_t sentry_value_new_message_event(
     sentry_level_t level, const char *logger, const char *text);
 
 /**
- * Creates a new breadcrumb with a specific type and message.
+ * Creates a new Breadcrumb with a specific type and message.
+ *
+ * See https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
  *
  * Either parameter can be NULL in which case no such attributes is created.
  */
 SENTRY_API sentry_value_t sentry_value_new_breadcrumb(
     const char *type, const char *message);
+
+/**
+ * Creates a new Exception value.
+ *
+ * See https://develop.sentry.dev/sdk/event-payloads/exception/
+ *
+ * The returned value needs to be attached to an event via
+ * `sentry_event_value_add_exception`.
+ */
+SENTRY_EXPERIMENTAL_API sentry_value_t sentry_value_new_exception(
+    const char *type, const char *value);
+
+/**
+ * Creates a new Thread value.
+ *
+ * See https://develop.sentry.dev/sdk/event-payloads/threads/
+ *
+ * The returned value needs to be attached to an event via
+ * `sentry_event_value_add_thread`.
+ *
+ * `name` can be NULL.
+ */
+SENTRY_EXPERIMENTAL_API sentry_value_t sentry_value_new_thread(
+    uint64_t id, const char *name);
+
+/**
+ * Creates a new Stack Trace conforming to the Stack Trace Interface.
+ *
+ * See https://develop.sentry.dev/sdk/event-payloads/stacktrace/
+ *
+ * The returned object needs to be attached to either an exception
+ * event, or a thread object.
+ *
+ * If `ips` is NULL the current stacktrace is captured, otherwise `len`
+ * stacktrace instruction pointers are attached to the event.
+ */
+SENTRY_EXPERIMENTAL_API sentry_value_t sentry_value_new_stacktrace(
+    void **ips, size_t len);
+
+/**
+ * Adds an Exception to an Event value.
+ *
+ * This takes ownership of the `exception`.
+ */
+SENTRY_EXPERIMENTAL_API void sentry_event_value_add_exception(
+    sentry_value_t event, sentry_value_t exception);
+
+/**
+ * Adds a Thread to an Event value.
+ *
+ * This takes ownership of the `thread`.
+ */
+SENTRY_EXPERIMENTAL_API void sentry_event_value_add_thread(
+    sentry_value_t event, sentry_value_t thread);
 
 /* -- Experimental APIs -- */
 
@@ -372,6 +432,11 @@ SENTRY_EXPERIMENTAL_API char *sentry_value_to_msgpack(
 
 /**
  * Adds a stacktrace to an event.
+ *
+ * The stacktrace is added as part of a new thread object.
+ * This function is **deprecated** in favor of using
+ * `sentry_value_new_stacktrace` in combination with `sentry_value_new_thread`
+ * and `sentry_event_value_add_thread`.
  *
  * If `ips` is NULL the current stacktrace is captured, otherwise `len`
  * stacktrace instruction pointers are attached to the event.
@@ -398,7 +463,7 @@ typedef struct sentry_ucontext_s {
  *
  * If the address is given in `addr` the stack is unwound form there.
  * Otherwise (NULL is passed) the current instruction pointer is used as
- * start address. The stacktrace is written to `stacktrace_out` with upt o
+ * start address. The stacktrace is written to `stacktrace_out` with up to
  * `max_len` frames being written.  The actual number of unwound stackframes
  * is returned.
  */
@@ -408,7 +473,7 @@ SENTRY_EXPERIMENTAL_API size_t sentry_unwind_stack(
 /**
  * Unwinds the stack from the given context.
  *
- * The stacktrace is written to `stacktrace_out` with upt o `max_len` frames
+ * The stacktrace is written to `stacktrace_out` with up to `max_len` frames
  * being written.  The actual number of unwound stackframes is returned.
  */
 SENTRY_EXPERIMENTAL_API size_t sentry_unwind_stack_from_ucontext(
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -370,7 +370,7 @@ SENTRY_API sentry_value_t sentry_value_new_breadcrumb(
  * See https://develop.sentry.dev/sdk/event-payloads/exception/
  *
  * The returned value needs to be attached to an event via
- * `sentry_event_value_add_exception`.
+ * `sentry_event_add_exception`.
  */
 SENTRY_EXPERIMENTAL_API sentry_value_t sentry_value_new_exception(
     const char *type, const char *value);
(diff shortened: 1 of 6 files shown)


include/sentry.h
```diff
@@ -338,26 +338,86 @@ typedef enum sentry_level_e {
 } sentry_level_t;
 
 /**
- * Creates a new empty event value.
+ * Creates a new empty Event value.
+ *
+ * See https://develop.sentry.dev/sdk/event-payloads/
```
jan-auer, line 343: Thank you for linking into documentation. `develop.sentry.dev` is mainly targeted at SDK engineers. Should we rather link to relevant sections in `docs.sentry.io` that explain these concepts for users?

For the native SDK, it definitely makes sense to keep a secondary link to develop, since we do not have a typed out event schema.

cc @bruno-garcia @HazAT

include/sentry.h
```diff
@@ -338,26 +338,86 @@ typedef enum sentry_level_e {
...
+ */
+SENTRY_EXPERIMENTAL_API sentry_value_t sentry_value_new_stacktrace(
+    void **ips, size_t len);
+
+/**
+ * Adds an Exception to an Event value.
+ *
+ * This takes ownership of the `exception`.
+ */
+SENTRY_EXPERIMENTAL_API void sentry_event_value_add_exception(
```
jan-auer, line 410: _nit_: Could we just call this `sentry_event_add_exception`?

include/sentry.h
```diff
@@ -373,6 +433,11 @@ SENTRY_EXPERIMENTAL_API char *sentry_value_to_msgpack(
 /**
  * Adds a stacktrace to an event.
  *
+ * The stacktrace is added as part of a new thread object.
```
jan-auer, line 436:
```suggestion
 * The stack trace is added as part of a new thread object.
```

include/sentry.h
```diff
@@ -338,26 +338,86 @@ typedef enum sentry_level_e {
...
+/**
+ * Creates a new Exception value.
+ *
+ * See https://develop.sentry.dev/sdk/event-payloads/exception/
+ *
+ * The returned value needs to be attached to an event via
+ * `sentry_event_value_add_exception`.
+ */
+SENTRY_EXPERIMENTAL_API sentry_value_t sentry_value_new_exception(
+    const char *type, const char *value);
```
jan-auer, line 376: What are the nullability-constraints on this API? Please add this to the doc comment.

The implementation uses `sentry_value_new_string`, which returns `sentry_value_new_null()` if the pointer is NULL, which is good behavior in my opinion. In `sentry_value_new_thread` you even make an explicit check for name. 

Note that _one of_ `type` or `value` are required for valid exceptions by the server.

src/sentry_value.c
```diff
@@ -1036,8 +1038,38 @@ sentry_value_new_breadcrumb(const char *type, const char *message)
...
+    return exc;
+}
+
+sentry_value_t
+sentry_value_new_thread(uint64_t id, const char *name)
+{
+    sentry_value_t thread = sentry_value_new_object();
+
+    // TODO: would be nice to be able to actually use a `u64` as value.
+    char buf[100];
```
jan-auer, line 1056: The maximum length is 20 characters plus the zero byte.

```suggestion
    char buf[20 + 1];
```

src/sentry_value.c
```diff
@@ -1036,8 +1038,38 @@ sentry_value_new_breadcrumb(const char *type, const char *message)
...
+    sentry_value_set_by_key(exc, "value", sentry_value_new_string(value));
+    return exc;
+}
+
+sentry_value_t
+sentry_value_new_thread(uint64_t id, const char *name)
+{
+    sentry_value_t thread = sentry_value_new_object();
+
+    // TODO: would be nice to be able to actually use a `u64` as value.
```
jan-auer, line 1055: Unfortunately, this is not really a TODO. You have no choice.

While we [do support integers](https://github.com/getsentry/relay/blob/a1e81e59dfcbc362187dd0d5c04b1be66386d8e3/relay-general/src/protocol/thread.rs#L8-L17), it is not safe to use numbers larger than `2^53` in JSON. Hence, you'll have to stick with the string.

This is something we'll eventually even type out in Relay.

src/sentry_value.c
```diff
@@ -1036,8 +1038,38 @@ sentry_value_new_breadcrumb(const char *type, const char *message)
...
+    char buf[100];
+    size_t written
+        = (size_t)snprintf(buf, sizeof(buf), "%llu", (unsigned long long)id);
+    if (written < sizeof(buf)) {
+        buf[written] = '\0';
```
jan-auer, line 1060: This has multiple issues:
- The return value of `snprintf` is not how many characters were written, but how many would have been written. The name is misleading.
- `snprintf` writes the NUL byte for you in all cases according to the standard.
- if `written == sizeof(buf)`, this overflow and writes past the buffer.

Please audit remaining code for similar issues.

src/sentry_value.c
```diff
@@ -1036,8 +1038,38 @@ sentry_value_new_breadcrumb(const char *type, const char *message)
     return rv;
 }
 
-void
-sentry_event_value_add_stacktrace(sentry_value_t event, void **ips, size_t len)
+sentry_value_t
+sentry_value_new_exception(const char *type, const char *value)
+{
+    sentry_value_t exc = sentry_value_new_object();
+    sentry_value_set_by_key(exc, "type", sentry_value_new_string(type));
```
jan-auer, line 1045: You could check `if (type)` to avoid creating an entry with `sentry_value_new_null()`.

src/sentry_value.c
```diff
@@ -1058,14 +1090,54 @@ sentry_event_value_add_stacktrace(sentry_value_t event, void **ips, size_t len)
     sentry_value_t stacktrace = sentry_value_new_object();
     sentry_value_set_by_key(stacktrace, "frames", frames);
 
-    sentry_value_t thread = sentry_value_new_object();
-    sentry_value_set_by_key(thread, "stacktrace", stacktrace);
+    return stacktrace;
+}
+
+static sentry_value_t
+sentry__get_values_list(sentry_value_t parent, const char *key)
```
jan-auer, line 1097: This is an upsert since it creates the value. Consider calling it `__get_or_insert_values_list` or `__ensure_values_list`.

src/sentry_value.c
Swatinem: Its the same code as in https://github.com/getsentry/sentry-native/pull/508 that I landed recently.
`written` will never be `== sizeof(buf)` per the if. And yes, this is written very defensively, as I’m not sure which parts of printf we can actually trust.

include/sentry.h
Swatinem: hm, https://develop.sentry.dev/sdk/event-payloads/exception/#attributes does not really talk about "*one of* them being required". I read it that both are, and that is what the API reflects.

include/sentry.h
Swatinem: Since developers need to construct event objects themselves, I think its best to link to the page that defines the possible properties.

jan-auer: Agreed, that's why I'm proposing to link _both_, with product documentation being the primary link. In other SDKs, we would skip the develop link.

include/sentry.h
jan-auer: Thanks, we need to update that. The actual behavior is:

https://github.com/getsentry/relay/blob/10f2c93efb8b5a200168a8fb961c3b136c1e5f18/relay-general/src/store/normalize.rs#L618-L623

src/sentry_value.c
jan-auer: My mistake, you're right 👍

include/sentry.h
jan-auer: Update: In the long run we will need two functions, one for creating synthetic errors and one for language-level exceptions. Since this function is for language-level exceptions, it's better to require both the type and the value.

jan-auer (approved):

PR #518, opened 2021-04-16
feat: Introduce `sentry_close`

### LINKED ISSUE #474: Add `close` function to disable completely the SDK
In hybrid SDKs such as React Native, we want to be able to close the SDK in JS and have that `close` call flow through to all inner SDKs.

For that we need make sure `close` is implemented in all layers and works as intended.
The [spec defined a close method on the client](https://develop.sentry.dev/sdk/unified-api/#client) but we're looking for a close on the static API, which will be added to the spec.

The expected behavior is that if a crash happens after close is called, an envelope with the crash data isn't generated.

Swatinem:
This replaces the former `sentry_shutdown`, which is being forwarded.

Fixes #474

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ sentry_init(options);
 
 // your application code …
 
-sentry_shutdown();
+sentry_close();
 ```
 
 Other important configuration options include:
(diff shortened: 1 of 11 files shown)


jan-auer (approved):
Please update all documentation and examples in `sentry-docs`.

PR #519, opened 2021-04-16
feat: Invoke before_send hook when using Crashpad

### LINKED ISSUE #250: Setting a callback with sentry_options_set_before_send does not fire under Windows
I'm currently using the before-send callback under Linux to do my own crash post-processing, but this `before_send` callback does not fire under Windows.

Swatinem:
Fixes #250

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ Other important configuration options include:
 
 ## Known Limitations
 
-- The crashpad backend currently has no support for notifying the crashing
+- The crashpad backend on macOS currently has no support for notifying the crashing
   process, and can thus not properly terminate sessions or call the registered
   `before_send` hook. It will also lose any events that have been queued for
   sending at time of crash.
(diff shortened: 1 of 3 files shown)


jan-auer (commented):
The doc string on `sentry_event_function_t` mentions signal safety and links to Linux docs. Now that it can also be called on Windows, please also add a reference to relevant Windows docs.

Regarding signal safety, we should also mention that we enable our own allocator, which makes all `sentry_value_new` methods safe even though `malloc` isn't.

With this change, this does fire under Windows and Linux, but not under macOS. Is there a chance to support macOS, too?

Swatinem:
I added a link to the Windows docs.

The allocator mechanism is already documented:
https://github.com/getsentry/sentry-native/blob/5fcb1dc4d1c8b85fc5ab225d29c7394b375fd1e9/include/sentry.h#L88-L97

On macOS, the crashes are handled completely outside the process, and crashpad does not offer any API to call back to the process.

jan-auer (approved):
Thanks!

PR #521, opened 2021-04-20
ref: Pass options to scope apply directly

Swatinem:
Extracted from #514

### DIFF
--- a/src/sentry_scope.h
+++ b/src/sentry_scope.h
@@ -67,7 +67,8 @@ void sentry__scope_flush_unlock(const sentry_scope_t *scope);
  * attached.
  */
 void sentry__scope_apply_to_event(const sentry_scope_t *scope,
-    sentry_value_t event, sentry_scope_mode_t mode);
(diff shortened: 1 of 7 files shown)


src/sentry_scope.h
```diff
@@ -67,7 +67,8 @@ void sentry__scope_flush_unlock(const sentry_scope_t *scope);
  * attached.
  */
 void sentry__scope_apply_to_event(const sentry_scope_t *scope,
-    sentry_value_t event, sentry_scope_mode_t mode);
+    const sentry_options_t *options, sentry_value_t event,
+    sentry_scope_mode_t mode);
```
flub, line 71: this formatting offends my rustfmt-trained aesthetics 😄  
but if CI says it's fine I'm not arguing

flub (approved):

src/sentry_scope.h
Swatinem: Yes, its apparently the default clang-format style :-( Either way, at least its automatic, and being checked :-D

PR #528, opened 2021-04-23
Update build docs for windows

kristjanvalur:
Some windows build tips gathered by trial and error.
Probably needs formatting and entering to whatever source editor was used to create the Readme :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,27 @@ install
 
 Please refer to the CMake Manual for more details.
 
+**Windows**:
+Different configurations have different run-time requirements.  The `Debug`
+configuration uses the windows debug C runtime, whereas the `RelWithDebInfo`
+does not.  This may determine which version you can use with your executable.
+
+Some special notes regarding `cmake`:
+- cmake doesn't need the configuration int the `-B` step since it creates
+  all the `*.vcxproj` files at once.
+- for `cmake --build`, **do no** specify `--parallel`, but **do** specify your
+  `--config RelWithDebInfo` or whichever configuration is appropriate.
+
+In short, do something like
+```
+> cmake -B build
+> cmake --build build --config RelWithDebInfo
+> cmake --install build --prefix install --config RelWithDebInfo
+```
+
+In addition to the optional `winhttp.lib`, windows builds require linking
+with `version.lib` and `dbghelp.lib`
+
 **Android**:
 
 The CMake project can also be configured to correctly work with the Android NDK,


README.md
```diff
@@ -112,6 +112,27 @@ install
...
 Please refer to the CMake Manual for more details.
 
+**Windows**:
+Different configurations have different run-time requirements.  The `Debug`
+configuration uses the windows debug C runtime, whereas the `RelWithDebInfo`
+does not.  This may determine which version you can use with your executable.
+
+Some special notes regarding `cmake`:
+- cmake doesn't need the configuration int the `-B` step since it creates
+  all the `*.vcxproj` files at once.
```
Swatinem, line 122: What do you mean by that? The `-B` flag specifies the output directory of the VC project, otherwise it would just clutter the working directory.

README.md
```diff
@@ -112,6 +112,27 @@ install
...
+- for `cmake --build`, **do no** specify `--parallel`, but **do** specify your
+  `--config RelWithDebInfo` or whichever configuration is appropriate.
```
Swatinem, line 124: Its unfortunate that `--parallel` is just ignored on Windows. I found that you actually need to set a cflag to have parallel compilation like so:

https://github.com/getsentry/sentry-native/blob/06d436096653925e5d7a493e413c9981ce7c5f1c/tests/cmake.py#L138-L141

Though I’m not really sure if there might be a better way here. Documentation is quite bad :-(

Swatinem (commented):
left some comments, looks good otherwise.

README.md
kristjanvalur: Sorry, that was unclear.  You don't need the `-DCMAKE_BUILD_TYPE=RelWithDebInfo` setting there, although it is harmless.  All the configurations are embedded in the resulting .vcxproj files.

README.md
kristjanvalur: Right but the point I was making was that if you **do** add --parallel, then the build step will **only** create the `Debug` configuration and ignores `--config`.  Took me half an hour of experimentation last night to figure out what was going on.  If you leave out the `--parallel` flag, then it will build whatever configuration you specify with `--config`

Swatinem: ohhh, I think in that case it might be a different manifestation of https://github.com/getsentry/sentry-native/pull/512, which means that cmakes option parsing is broken

README.md
Mixaill: You do not need it in case of MSVS generator, but you can use `-G Ninja` for example

README.md
```diff
@@ -112,6 +112,27 @@ install
...
+
+In short, do something like
+```
+> cmake -B build
+> cmake --build build --config RelWithDebInfo
+> cmake --install build --prefix install --config RelWithDebInfo
+```
+
+In addition to the optional `winhttp.lib`, windows builds require linking
+with `version.lib` and `dbghelp.lib`
```
Mixaill, line 134: Already handled in https://github.com/getsentry/sentry-native/blob/master/CMakeLists.txt#L343-L366

kristjanvalur: Right, but it probably ought to be pointed out to users of the `sentry.lib` library, who otherwise need to hunt through their linker errors to figure out which windows libraries to include.  It is already mentioned that `winhttp.lib` is a dependency but the others aren't mentioned.

README.md
kristjanvalur: could be.  Don't know what the best solution is.  Only that I am not an avid cmake user myself and always struggle a bit when using it.  Wanted to help save people from the problems I had :)

README.md
kristjanvalur: Right you are.  So it may well be left in place though.  It is harmless enough :)

kristjanvalur:
So, what do we do with this?  Looking at Readme.md, it looks to me as it comes from some external editor.  This PR was not really meant for direct pulling, rather as a suggestion for updates.

Swatinem:
Its just how my vscode (or actually prettier) autoformatted that, so no problem. I can also just copy-paste this in.

kristjanvalur:
Right, but you probably do want to edit it for style.  So feel free to just
grab whatever from it and close the PR :)

þri., 27. apr. 2021 kl. 09:02 skrifaði Arpad Borsos <
***@***.***>:

supervacuus:
The raised topics in this PR concerning the CMake config/build parameters when using the MSBuild generator are all valid points. But I think we covered them in the docs a while back by providing more contrasting build examples: https://docs.sentry.io/platforms/native/#install

PR #545, opened 2021-06-15
fix: possible race conditions in init and close

MikeRumplerSentry:
fixes possible race conditions when init is called multiple times in fast succession
fixes possible race conditions when other threads access options during a longer run of close

WARNING: issues an ERROR now and returns 1 if init is called multiple times.


src/sentry_core.c
```diff
@@ -85,6 +94,7 @@ sentry_init(sentry_options_t *options)
         SENTRY_WARN("failed to create database directory or there is no write "
                     "access to this directory");
         sentry_options_free(options);
```
Swatinem, line 96: maybe `goto fail`?

src/sentry_core.c
```diff
@@ -72,7 +73,15 @@ sentry__should_skip_upload(void)
...
 {
-    sentry_close();
+    // this function is to be called only once, so we do not allow more than
+    // one call to be active. If the function is called twice, create an error
+    // and return fail (1)
+    sentry__mutex_lock(&g_init_lock);
+    if (g_options != NULL) {
+        SENTRY_ERROR("sentry_init() may only be called once");
+        sentry__mutex_unlock(&g_init_lock);
+        return 1;
```
Swatinem, line 83: you leak `options` here. better to `goto fail` here as well.

src/sentry_core.c
```diff
@@ -25,6 +25,7 @@
 
 static sentry_options_t *g_options = NULL;
 static sentry_mutex_t g_options_lock = SENTRY__MUTEX_INIT;
+static sentry_mutex_t g_init_lock = SENTRY__MUTEX_INIT;
```
Swatinem, line 28: I think we should re-use the `g_options_lock` here, so no other thread can touch `g_options` in the middle.

src/sentry_core.c
```diff
@@ -183,9 +195,10 @@ sentry_close(void)
 {
     sentry_end_session();
 
+    // keep the global options until end of function if other calls need the
```
Swatinem, line 198: but isn’t that part of the problem? Suppose 2 concurrent threads call `close`, both of them end up with a non-null `options` and both of them de-init/decref, which means we get use-after-free and double-free.

Swatinem (commented):
I think the `close` is actually worse now than before.

I’m also a bit surprised we don’t have any unit tests for re-init right now. I can’t remember if we communicated if that should be possible or not, but I also can’t remember otherwise.

We already had some problems with users complaining about slight behavior changes (like `sentry_init` returning an error vs just silently not doing anything).

I think we can have the options lock for the complete length of both functions, keep the close call in init (maybe in check that `g_options` was indeed initialized, maybe logging a debug message in that case)
`close` should probably move the `end_session` into its `if (options)` block to avoid doing unnecessary work.

Keep in mind that both these functions *should* be used only once in the program on the main thread, so larger lock regions are perfectly fine, and the locks are re-entrant anyway.

src/sentry_core.c
MikeRumplerSentry: ACK, good option here, crosses no important code doing this.

src/sentry_core.c
MikeRumplerSentry: If the function is responsible to remove the given object, yes. When jumping to fail this is also tackled with this.

src/sentry_core.c
MikeRumplerSentry: Just for readability I would rather keep the additional mutex to see that it guards more than only the access to the data object.

src/sentry_core.c
MikeRumplerSentry: The question here is is it safe to call close twice. If not, I would just do the same as with init, guard the whole function and detect multiple calls. The problem arises when during the time the close function is sending remaining items another thread accesses the global options which are then NULL. With no proper error handling in place on every place global options is used, this may result in a crash, either.

MikeRumplerSentry: Will guard the whole close function as it is done with init, too.

src/sentry_core.c
```diff
@@ -72,7 +73,18 @@ sentry__should_skip_upload(void)
...
-    sentry_close();
+    // this function is to be called only once, so we do not allow more than
+    // one call to be active. If the function is called twice, create an error
+    // and return fail (1)
+    sentry__mutex_lock(&g_initclose_lock);
+
+    // pre-init here, so we can consistently use bailing out to :fail
+    sentry_transport_t *transport = NULL;
+
+    if (g_options != NULL) {
```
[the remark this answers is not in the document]
MikeRumplerSentry: This is a read-only access and this is by definition atomic. So no mutex needed here.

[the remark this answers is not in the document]
andrei-mu: > This is a read-only access and this is by definition atomic. So no mutex needed here.

This is not true

Swatinem (approved):

PR #547, opened 2021-06-18
fix: Restore default Linux signal handlers when shutting down crashpad backend

GenuineAster:
This fixes (on Linux) a bug we were experiencing after calling
`sentry_reinstall_backend` while using the crashpad backend. All
crashes were getting reported in an infinite loop until our server
orchestration killed the process, consuming all of our event budget!

Let me know how I can add a regression test for this.


Swatinem:
I cherry-picked this into #548

PR #551, opened 2021-06-22
fix: Shorten/Split Locked sections to avoid deadlock

Swatinem:
We have received a report that the `sentry_get_modules_list` on mac can deadlock
when other code concurrently does a `dlopen` and thus invokes the `add_image`
callback from a different thread.

We shorten/split the locked blocks in order to avoid holding a lock in the
`get_modules` function whenever the `add_image` function is being invoked possibly
from other threads.

### DIFF
--- a/src/modulefinder/sentry_modulefinder_apple.c
+++ b/src/modulefinder/sentry_modulefinder_apple.c
@@ -43,11 +43,8 @@ add_image(const struct mach_header *mh, intptr_t UNUSED(vmaddr_slide))
         return;
     }
 
-    sentry__mutex_lock(&g_mutex);
-
-    sentry_value_t modules = g_modules;
-    sentry_value_t new_modules = sentry__value_clone(modules);
     sentry_value_t module = sentry_value_new_object();
+    sentry_value_set_by_key(module, "type", sentry_value_new_string("macho"));
     sentry_value_set_by_key(
         module, "code_file", sentry_value_new_string(info.dli_fname));
     sentry_value_set_by_key(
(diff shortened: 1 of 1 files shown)


src/modulefinder/sentry_modulefinder_apple.c
```diff
@@ -130,19 +130,39 @@ remove_image(const struct mach_header *mh, intptr_t UNUSED(vmaddr_slide))
...
         _dyld_register_func_for_add_image(add_image);
         _dyld_register_func_for_remove_image(remove_image);
-        g_initialized = true;
     }
-    sentry_value_t modules = g_modules;
+
+    sentry__mutex_lock(&g_mutex);
+    // now that is strange, having this declaration after the goto label gives a
+    // `expected expression` error.
+    sentry_value_t modules;
```
jan-auer, line 162: You should move that even before the `goto` statement, rather than just the label. Just put this at the very top of the function.

src/modulefinder/sentry_modulefinder_apple.c
```diff
@@ -130,19 +130,39 @@ remove_image(const struct mach_header *mh, intptr_t UNUSED(vmaddr_slide))
...
-    if (!g_initialized) {
+    if (g_initialized) {
+        // avoid an unlock/lock cycle in this case
+        goto is_initialized;
+    } else {
         g_modules = sentry_value_new_list();
+        g_initialized = true;
+        should_initialize = true;
+    }
+    sentry__mutex_unlock(&g_mutex);
```
jan-auer, line 147: You could just keep the initial structure of the function and put unlock+lock in the branch you had there. In pseudo-code:

```c
lock();
if (!g_initialized) {
  g_initialized = true;
  g_modules = new_list();

  unlock();
  //expensive stuff here
  lock();
}
incref()
return 
```

jan-auer (approved):

src/modulefinder/sentry_modulefinder_apple.c
Swatinem: thats a very good idea, makes things more readable and explicit!

PR #566, opened 2021-06-30
fix: Make Crashpad Backend respect max_breadcrumbs setting

### LINKED ISSUE #565: Value supplied in sentry_options_set_max_breadcrumbs not honoured
**Description**

Setting the number of breadcrumbs to a number other that the default does not lead to that number being used 

- [ ] During build
- [x] During run-time

**Environment**

- OS: [Windows/Mac/Linux]
- Compiler: [GCC 10/MSVC 19.16.27030.1]
- CMake version and config: [e.g. 3.18.1, SENTRY_BACKEND=inproc]

**Steps To Reproduce**

During application startup run the line of code

sentry_options_set_max_breadcrumbs(options, 256);

Log more than 256 events, then crash

observe that only 100 (the default) will be logged

**Suggested Resolution**

It appears that sentry__crashpad_backend_add_breadcrumb uses SENTRY_BREADCRUMBS_MAX without looking to the 
value set in options. an alteration such as below may be suitable, though consideration may need to be given to resizing buffers and any aspects of threading that I have not looked into.

static void
sentry__crashpad_backend_add_breadcrumb(
    sentry_backend_t *backend, sentry_value_t breadcrumb)
{
    crashpad_state_t *data = (crashpad_state_t *)backend->data;

    size_t max_breadcrumbs = options->max_breadcrumbs 
		? options->max_breadcrumbs 
		: SENTRY_BREADCRUMBS_MAX;

    bool first_breadcrumb = data->num_breadcrumbs % max_breadcrumbs == 0;

    const sentry_path_t *breadcrumb_file = data->num_breadcrumbs % (max_breadcrumbs * 2) 
	< max_breadcrumbs
        ? data->breadcrumb1_path
        : data->breadcrumb2_path;
    data->num_breadcrumbs++;
    if (!breadcrumb_file) {
        return;
    }

    size_t mpack_size;
    char *mpack = sentry_value_to_msgpack(breadcrumb, &mpack_size);
    if (!mpack) {
        return;
    }

    int rv = first_breadcrumb
        ? sentry__path_write_buffer(breadcrumb_file, mpack, mpack_size)
        : sentry__path_append_buffer(breadcrumb_file, mpack, mpack_size);
    sentry_free(mpack);

    if (rv != 0) {
        SENTRY_DEBUG("flushing breadcrumb to msgpack failed");
    }
}

Swatinem:
fixes #565

### DIFF
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -324,16 +324,17 @@ sentry__crashpad_backend_shutdown(sentry_backend_t *backend)
 }
 
 static void
-sentry__crashpad_backend_add_breadcrumb(
-    sentry_backend_t *backend, sentry_value_t breadcrumb)
+sentry__crashpad_backend_add_breadcrumb(sentry_backend_t *backend,
+    sentry_value_t breadcrumb, const sentry_options_t *options)
 {
     crashpad_state_t *data = (crashpad_state_t *)backend->data;
 
-    bool first_breadcrumb = data->num_breadcrumbs % SENTRY_BREADCRUMBS_MAX == 0;
+    size_t max_breadcrumbs = options->max_breadcrumbs;
+
+    bool first_breadcrumb = data->num_breadcrumbs % max_breadcrumbs == 0;
 
     const sentry_path_t *breadcrumb_file
-        = data->num_breadcrumbs % (SENTRY_BREADCRUMBS_MAX * 2)
-            < SENTRY_BREADCRUMBS_MAX
+        = data->num_breadcrumbs % (max_breadcrumbs * 2) < max_breadcrumbs
         ? data->breadcrumb1_path
         : data->breadcrumb2_path;
     data->num_breadcrumbs++;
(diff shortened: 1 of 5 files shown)


src/backends/sentry_backend_crashpad.cpp
```diff
@@ -324,16 +324,17 @@ sentry__crashpad_backend_shutdown(sentry_backend_t *backend)
...
-    sentry_backend_t *backend, sentry_value_t breadcrumb)
+sentry__crashpad_backend_add_breadcrumb(sentry_backend_t *backend,
+    sentry_value_t breadcrumb, const sentry_options_t *options)
 {
     crashpad_state_t *data = (crashpad_state_t *)backend->data;
 
-    bool first_breadcrumb = data->num_breadcrumbs % SENTRY_BREADCRUMBS_MAX == 0;
+    size_t max_breadcrumbs = options->max_breadcrumbs;
+
+    bool first_breadcrumb = data->num_breadcrumbs % max_breadcrumbs == 0;
```
jan-auer, line 334: To confirm: Can `max_breadcrumbs ever be `0`?

jan-auer (approved):

src/backends/sentry_backend_crashpad.cpp
Swatinem: good point. Code in other places was failing with `max_breadcrumbs = 0`, I fixed that up and its now correctly supported.

PR #568, opened 2021-06-30
Use WinHttpCloseHandle from another thread to cancel the http request.

smibe:

### DIFF
--- a/src/transports/sentry_transport_winhttp.c
+++ b/src/transports/sentry_transport_winhttp.c
@@ -113,17 +113,24 @@ sentry__winhttp_transport_shutdown(uint64_t timeout, void *transport_state)
     sentry_bgworker_t *bgworker = (sentry_bgworker_t *)transport_state;
     winhttp_bgworker_state_t *state = sentry__bgworker_get_state(bgworker);
 
-    if (state->connect) {
-        WinHttpCloseHandle(state->connect);
-        state->connect = NULL;
-    }
-    if (state->session) {
-        WinHttpCloseHandle(state->session);
-        state->session = NULL;
-    }
-    if (state->request) {
-        WinHttpCloseHandle(state->request);
-        state->request = NULL;
+    if (sentry__bgworker_shutdown(bgworker, timeout) != 0) {
+        // Seems like some requests are taking too long/hanging
+        // Just close them to make sure the background thread is exiting.
+        if (state->connect) {
+            WinHttpCloseHandle(state->connect);
+            state->connect = NULL;
+        }
+
+        // NOTE: We need to close the session before closing the request.
+        // This will cancel all other requests which might be queued as well.
+        if (state->session) {
+            WinHttpCloseHandle(state->session);
+            state->session = NULL;
+        }
+        if (state->request) {
+            WinHttpCloseHandle(state->request);
+            state->request = NULL;
+        }
     }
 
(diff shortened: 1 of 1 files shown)


src/transports/sentry_transport_winhttp.c
```diff
@@ -110,6 +111,21 @@ static int
 sentry__winhttp_transport_shutdown(uint64_t timeout, void *transport_state)
 {
     sentry_bgworker_t *bgworker = (sentry_bgworker_t *)transport_state;
+    winhttp_bgworker_state_t *state = sentry__bgworker_get_state(bgworker);
+
+    if (state->connect) {
+        WinHttpCloseHandle(state->connect);
```
Swatinem, line 117: These are also being closed in `state_free`.

src/transports/sentry_transport_winhttp.c
```diff
@@ -110,6 +111,21 @@ static int
...
+        state->connect = NULL;
+    }
+    if (state->session) {
+        WinHttpCloseHandle(state->session);
+        state->session = NULL;
+    }
+    if (state->request) {
+        WinHttpCloseHandle(state->request);
+        state->request = 0;
```
Swatinem, line 126: Just a question: Shouldn’t these be closed in a particular order, in this case request first? Also, while it does not really make a difference, why is this reset to `0` while the others use `NULL`?

Also, since this can race with the background thread, might it be a good idea to exchange `state->request` first before closing it? So the racing background thread does not try to double-close it?

Swatinem (approved):
lgtm!
Since you mention a shutdown hang, will closing the handle from the main thread just lead to an error on the background thread?

src/transports/sentry_transport_winhttp.c
smibe: Yes if everything goes well. In case some WinHttp functions are taking longer than the timeout the background thread will still be alive. Closing the handle will cancel the  WinhttpReceiveResponse call and unblocks the background thred.

src/transports/sentry_transport_winhttp.c
smibe: Yes the order is important. Closing the session first will cancel all other requests in the queue as well.
When the request handle is closed, the background blocking function is unblocked and the loop continues. But because of no session handle all WinHttp requests will fail and the queue will get cleaned up.
This is not a nice solution, but still the best I could com up with.

src/transports/sentry_transport_winhttp.c
```diff
@@ -110,6 +111,21 @@ static int
...
+    if (state->session) {
+        WinHttpCloseHandle(state->session);
+        state->session = NULL;
+    }
+    if (state->request) {
+        WinHttpCloseHandle(state->request);
+        state->request = NULL;
+    }
+
     return sentry__bgworker_shutdown(bgworker, timeout);
```
Swatinem, line 129: Thanks for the explanation why the ordering is important. It might be a good idea to put that as a comment in the code.

I also saw that all the tests are failing, I think in that case also the ordering is important.
The main thread will ask the background thread to shut down here according to the timeout. I think you have to move all your close calls *after* this call here, because you *do* want to wait for the given timeout, instead of just throwing away all the requests that are currently in flight.

Swatinem (changes requested):

PR #576, opened 2021-07-13
build+install breakpad's dump_syms executable

madebr:
Part of the build process when using the breakpad backend, is dumping all symbols from an executable (that contains debug symbols).
This pr adds support to the cmake script for building it.

- [ ] This pr needs https://github.com/getsentry/breakpad/pull/17 because there are some breakpad compile errors
- [ ] The sentry-native unit tests do not trigger building `dump_syms` because it only builds the bare minimum
- [ ] Right now, it will build `dump_syms` unconditionally when using the `breakpad` backend.
I did not add exceptions for e.g. Android/iOS/... .
- [ ] Untested on mingw

Some feedback would be appreciated.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -1,4 +1,26 @@
 # The list of files is drived from: breakpad/Makefile.am
+cmake_minimum_required(VERSION 3.12)
+project(sentry_breakpad)
+
+set(CMAKE_CXX_STANDARD 11)
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    set(LINUX TRUE)
+endif()
+
+# set static runtime, if enabled
+if(SENTRY_BUILD_RUNTIMESTATIC AND MSVC)
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+endif()
+
+if(LINUX)
+    option(BREAKPAD_PIC "Build breakpad as position independent code" ON)
+else()
+    set(BREAKPAD_PIC ON)
+endif()
+if(BREAKPAD_PIC)
+    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+endif()
 
 set(BREAKPAD_SOURCES_COMMON
     breakpad/src/common/convert_UTF.cc
@@ -7,6 +29,38 @@ set(BREAKPAD_SOURCES_COMMON
     breakpad/src/common/md5.h
     breakpad/src/common/string_conversion.cc
     breakpad/src/common/string_conversion.h
+)
+
+set(BREAKPAD_SOURCES_COMMON_EXTRA
+)
+
+set(BREAKPAD_SOURCES_COMMON_EXTRA_UNIX
+    breakpad/src/common/dwarf/bytereader.cc
+    breakpad/src/common/dwarf/bytereader.h
+    breakpad/src/common/dwarf/dwarf2diehandler.cc
+    breakpad/src/common/dwarf/dwarf2diehandler.h
+    breakpad/src/common/dwarf/dwarf2reader.cc
+    breakpad/src/common/dwarf/dwarf2reader.h
+    breakpad/src/common/dwarf/elf_reader.cc
+    breakpad/src/common/dwarf/elf_reader.h
+    breakpad/src/common/dwarf_cfi_to_module.cc
+    breakpad/src/common/dwarf_cfi_to_module.h
+    breakpad/src/common/dwarf_cu_to_module.cc
+    breakpad/src/common/dwarf_cu_to_module.h
+    breakpad/src/common/dwarf_line_to_module.cc
+    breakpad/src/common/dwarf_line_to_module.h
+    breakpad/src/common/dwarf_range_list_handler.cc
+    breakpad/src/common/dwarf_range_list_handler.h
+    breakpad/src/common/path_helper.cc
+    breakpad/src/common/path_helper.h
+    breakpad/src/common/language.cc
+    breakpad/src/common/language.h
+    breakpad/src/common/module.cc
+    breakpad/src/common/module.h
+    breakpad/src/common/stabs_to_module.cc
+    breakpad/src/common/stabs_to_module.h
+    breakpad/src/common/stabs_reader.cc
+    breakpad/src/common/stabs_reader.h
 )
 
 set(BREAKPAD_SOURCES_COMMON_LINUX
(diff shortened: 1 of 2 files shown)


external/CMakeLists.txt
```diff
@@ -1,4 +1,26 @@
...
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+endif()
+
+if(LINUX)
+    option(BREAKPAD_PIC "Build breakpad as position independent code" ON)
+else()
+    set(BREAKPAD_PIC ON)
+endif()
+if(BREAKPAD_PIC)
+    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
madebr, line 22:
```suggestion
    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
else()
    set(CMAKE_POSITION_INDEPENDENT_CODE OFF)
```

external/CMakeLists.txt
```diff
@@ -115,60 +202,117 @@ set(BREAKPAD_SOURCES_CLIENT_IOS
...
-        ${BREAKPAD_SOURCES_COMMON_APPLE}
-        ${BREAKPAD_SOURCES_CLIENT_APPLE})
+    target_sources(breakpad_common PRIVATE ${BREAKPAD_SOURCES_COMMON_APPLE})
+    target_sources(breakpad_common_extra PRIVATE ${BREAKPAD_SOURCES_COMMON_EXTRA_UNIX} ${BREAKPAD_SOURCES_COMMON_EXTRA_APPLE})
+    target_sources(breakpad_client PRIVATE ${BREAKPAD_SOURCES_CLIENT_APPLE})
+    target_sources(dump_syms PRIVATE ${BREAKPAD_SOURCES_DUMPSYMS_MAC})
+    add_executable(macho_dump ${BREAKPAD_SOURCES_MACHODUMP_MAC})
+    target_link_libraries(macho_dump PRIVATE breakpad_common breakpad_common_extra)
+        target_compile_definitions(breakpad_common PRIVATE HAVE_MACH_O_NLIST_H)
+        target_compile_definitions(breakpad_common_extra PRIVATE HAVE_MACH_O_NLIST_H)
```
madebr, line 264:
```suggestion
    target_compile_definitions(breakpad_common PRIVATE HAVE_MACH_O_NLIST_H)
    target_compile_definitions(breakpad_common_extra PRIVATE HAVE_MACH_O_NLIST_H)
```

external/CMakeLists.txt
```diff
@@ -115,60 +202,117 @@ set(BREAKPAD_SOURCES_CLIENT_IOS
...
+            find_package(Threads REQUIRED)
+            target_link_libraries(breakpad_common_extra PRIVATE Threads::Threads)
+        endif()
 
     include(CheckFunctionExists)
     check_function_exists(getcontext HAVE_GETCONTEXT)
     if(HAVE_GETCONTEXT)
-        target_compile_definitions(breakpad_client PRIVATE HAVE_GETCONTEXT)
-    else()
-        target_sources(breakpad_client PRIVATE ${BREAKPAD_SOURCES_COMMON_LINUX_GETCONTEXT})
```
Swatinem, line 134: why exactly was this changed? I thought this file was only needed as a polyfill if `getcontext` is not available on the system?

Swatinem (commented):
lgtm, except one thing to clarify!

we should make this opt-in for sure. philosophically, we should not advocate for using `dump_syms`, as we can use the object files directly, which give us much better results, and can also improve in quality over time as our tools evolve, whereas breakpad files can’t.

external/CMakeLists.txt
Mixaill: source can be included in the project unconditionally (it adds `breakpad_getcontext` function which will be stripped by linker) but L251 is an error, it makes `check_function_exists` call useless and breaks build for some libc implementations like musl.

external/CMakeLists.txt
Mixaill: `set(CMAKE_POSITION_INDEPENDENT_CODE ${BREAKPAD_PIC})`

madebr: It's a trade off between setting `CMAKE_POSITION_INDEPENDENT_CODE` once,
or setting the `POSITION_INDEPENDENT_CODE` property once for each target.

I thought the first was less error prone.

external/CMakeLists.txt
madebr: I misunderstood `HAVE_GETCONTEXT`. I thought it was a user option but apparently it is the result of `check_function_exists`.
This needs to be reverted.

external/CMakeLists.txt
```diff
@@ -115,60 +202,117 @@ set(BREAKPAD_SOURCES_CLIENT_IOS
...
+    target_sources(breakpad_client PRIVATE ${BREAKPAD_SOURCES_CLIENT_WINDOWS})
+    target_sources(dump_syms PRIVATE ${BREAKPAD_SOURCES_DUMPSYMS_WINDOWS})
+
+    target_compile_definitions(breakpad_common PUBLIC _UNICODE UNICODE)
 
-    # set static runtime if enabled
-    if(SENTRY_BUILD_RUNTIMESTATIC AND MSVC)
-        set_property(TARGET breakpad_client PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+    target_compile_options(breakpad_common_extra PUBLIC
+        "-I$(VsInstallDir)/DIA SDK/include"
```
madebr, line 284: With MSVC, the *DIA SDK* library `diaguids` can be found in an awkward location.
The usage of `$(VSInstallDir)` is MSBuild only, I think.
Perhaps this deserves a find_package cmake module.

Also, I haven't tested this yet with the `NMake Makefiles` generator or with mingw.

jan-auer:
@madebr before going ahead with this, can you clarify on this:

> Part of the build process when using the breakpad backend, is dumping all symbols from an executable (that contains debug symbols).

With Sentry, this is not required at all. Like @Swatinem mentioned earlier, you should upload the .exe/dll and .pdb files to Sentry. This allows us to extract higher fidelity debugging information that what the Breakpad symbol format offers. We will also be phasing out support for Breakpad in the SDK pretty soon.

madebr:
> @madebr before going ahead with this, can you clarify on this:
> 
> > Part of the build process when using the breakpad backend, is dumping all symbols from an executable (that contains debug symbols).

My use case is for an open source project such that users can easily create good crash reports.
Because open source projects generally generate a lot of variants of binaries, storing debug symbols centrally and/or using sentry's service is not practical.
That's why I was thinking about using breakpad for users to generate reports locally.
Hiding/obfuscating source code is not needed here.

I'm currently investigating [`backward-cpp`](https://github.com/bombela/backward-cpp).
So perhaps, I won't need crashpad at all.

> With Sentry, this is not required at all. Like @Swatinem mentioned earlier, you should upload the .exe/dll and .pdb files to Sentry. This allows us to extract higher fidelity debugging information that what the Breakpad symbol format offers. We will also be phasing out support for Breakpad in the SDK pretty soon.

If so, then you will also probably stop maintaining the `sentry/breakpad` repo.
Then it won't make sense to push forward with these changes.

Thanks for your time!

PR #577, opened 2021-07-14
made shutdown timer customizable

andrei-mu:

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -1013,6 +1013,12 @@ SENTRY_API void sentry_options_set_system_crash_reporter_enabled(
 SENTRY_API void sentry_options_set_shutdown_timeout(
     sentry_options_t *opts, uint64_t shutdown_timeout);
 
+/**
+ * Gets the maximum time to wait for the asynchronous tasks to end on shutdown,
+ * before attempting a forced termination.
+ */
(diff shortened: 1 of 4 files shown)


include/sentry.h
```diff
@@ -1006,6 +1006,13 @@ SENTRY_API void sentry_options_set_database_pathw(
 SENTRY_API void sentry_options_set_system_crash_reporter_enabled(
     sentry_options_t *opts, int enabled);
 
+/**
+ * Sets the maximum time to wait for the asynchronous tasks to end on shutdown,
+ * before attempting a forced termination.
+ */
+SENTRY_API void sentry_options_set_shutdown_timeout(
```
Swatinem, line 1013: a corresponding `get` would be nice as well

Swatinem (approved):

Swatinem (approved):
Sorry for the turnaround here, forgot to mention this in the last review: It would be nice to document that all the times are in milliseconds.

include/sentry.h
andrei-mu: I've added the getter as well, please review.

andrei-mu:
> Sorry for the turnaround here, forgot to mention this in the last review: It would be nice to document that all the times are in milliseconds.

this is done as well.

Swatinem (approved):

PR #615, opened 2021-12-07
ci(codechecker): Workaround for code checker not building due to node issues

relaxolotl:
It looks like codechecker's using a node feature (node-fibers) that is no longer supported on the version of node on our ubuntu image. I've used the workaround suggested in https://github.com/Ericsson/codechecker/issues/3353. It doesn't look like it should have a negative impact on our runs in CI as I would assume that it shouldn't care for UI builds by design, but I could be misinterpreting the flag.


flub (approved):

Swatinem (approved):
the less unnecessary stuff we build the better ;-)

Issue #601, opened 2021-10-15
Performance Monitoring / Tracing integration with sentry-native


kunaljaydesai:
I noticed JS and Python SDK supports the Sentry Performance Monitoring / Tracing integration. Will sentry-native support this soon too?

Swatinem:
We will be looking into it in the future, yes.

MartinDelille:
I suppose it will be available in the next release ? Do you have a date ?

flub:
> I suppose it will be available in the next release ? Do you have a date ?

currently expected to be released around end of Jan or earlyish Feb (as experimental api, but many parts of the native sdk are such).  but that's not an official promise :)

parsley72:
It's been 8 months, any update?

flub:
This has been released for a while (according to the changelog since 0.4.16 on 21 April), see https://github.com/getsentry/sentry-native/blob/5962300ea77bc546d293f15c7b091051a8abace6/include/sentry.h#L1415 for a starting point in the API.  From there you can add info to transactions, add child spans etc.

parsley72:
The documentation for native on https://docs.sentry.io/platforms/native/performance/ still says:
> The platform or SDK you've selected either does not support this functionality, or it is missing from documentation.

PR #616, opened 2021-12-08
feat: Add internal UUID types

relaxolotl:
This adds in support for internal UUIDs needed by tracing. The major difference between this and the "standard" uuid is that the hyphens are stripped during serialization. sentry appears to not consider the hyphenated representations of these UUIDs to be valid.

This will be used in transactions and spans for their internal `trace_id` and `span_id` values whose specs can be found here:

Transaction: https://develop.sentry.dev/sdk/event-payloads/transaction/#anatomy
Span: https://develop.sentry.dev/sdk/event-payloads/span/#examples

Span IDs are notably missing any sort of explicit specification about their structure; based on examples and other SDKs it appears that they are simply UUID4s truncated to 8 bytes.

This also adds in a convenience make target and some minor doc updates.

relates to #601


Makefile
```diff
@@ -18,6 +18,10 @@ build/sentry_test_unit: build
 test: update-test-discovery test-integration
 .PHONY: test
 
+test-unit: update-test-discovery build/sentry_test_unit
+	./build/sentry_test_unit
```
relaxolotl, line 22: on my machine this still spits out an absurd number of errors that look vaguely like

```
[71087:13339984:20211208,163001.841338:WARNING crash_report_database_mac.mm:767] Failed to read report metadata for .../sentry-native/.sentry-native/completed/49b5cb3a-64f0-4d5f-86fb-ed731f0981df.meta
```

i've tried negative grepping and awking this away to no luck; i would imagine in the worst case scenario this would require some sort of custom build that changes the logging level on crashpad to error at minimum. any help here would be much appreciated as the unit test results are unnecessarily difficult to parse through by default thanks to my poor shell skills.

src/sentry_uuid.h
```diff
@@ -3,11 +3,22 @@
...
+ * Converts a sentry UUID to a string representation used for span IDs.
+ */
+void sentry__span_uuid_as_string(const sentry_uuid_t *uuid, char str[17]);
+
 #ifdef SENTRY_PLATFORM_WINDOWS
 /**
  * Create a new UUID from the windows-native GUID type.
  */
-sentry_uuid_t sentry__uuid_from_native(GUID *guid);
+sentry_uuid_t sentry__uuid_from_native(const GUID *guid);
```
relaxolotl, line 21: looks like including this in `sentry_uuid.c` revealed a mismatch between this signature and the one in `sentry_uuid.c`, causing the windows builds to fail.

unsure if the new include subtly breaks anything; hoping it doesn't.

Makefile
Swatinem: You can ignore this. Just `rm -rf .sentry-native`, as that directory is safe to prune.
However now I think that we might want to have a solution for this, as I bet customers are hitting that as well sometimes and its not as simple to tell them "just remove that directory".

src/sentry_uuid.c
```diff
@@ -101,6 +102,27 @@ sentry_uuid_as_string(const sentry_uuid_t *uuid, char str[37])
 #undef B
 }
 
+void
+sentry__internal_uuid_as_string(const sentry_uuid_t *uuid, char str[37])
+{
+#define B(X) (unsigned char)uuid->bytes[X]
+    snprintf(str, 37,
```
Swatinem, line 109:
```suggestion
    snprintf(str, 33,
```
Since you are only serializing 16 hex bytes, and a zero byte.

src/sentry_uuid.h
Swatinem: This shouldn’t cause any problems.

src/sentry_value.c
```diff
@@ -973,6 +974,30 @@ sentry__value_new_hexstring(const uint8_t *bytes, size_t len)
...
+    }
+    sentry__span_uuid_as_string(uuid, buf);
+    buf[16] = '\0';
+    return sentry__value_new_string_owned(buf);
+}
+
+sentry_value_t
+sentry__value_new_internal_uuid(const sentry_uuid_t *uuid)
+{
+    char *buf = sentry_malloc(37);
```
Swatinem, line 992:
```suggestion
    char *buf = sentry_malloc(33);
```

src/sentry_value.c
```diff
@@ -973,6 +974,30 @@ sentry__value_new_hexstring(const uint8_t *bytes, size_t len)
...
+    if (!buf) {
+        return sentry_value_new_null();
+    }
+    sentry__internal_uuid_as_string(uuid, buf);
+    buf[36] = '\0';
```
Swatinem, line 997:
```suggestion
    buf[32] = '\0';
```

Swatinem (approved):

Makefile
relaxolotl: i'll leave this as is for the time being. if a contributor ends up bumping into this problem we can loop back and look into addressing this then. the solutions i've explored so far to deal with this are all fairly ugly, so i'd rather take the hit and just manually prune the reports stored in my local DB.

PR #617, opened 2021-12-09
feat(tracing): Groundwork to add tracing context to all events

relaxolotl:
This adds the appropriate stubs and fields to start storing spans on the (universal) scope. No actual logic has been added to actually support setting spans on the scope itself in this PR.

The focus of this PR is to begin including tracing info in the context on all events _if_ there is a transaction set on the scope. It does this fairly naively right now as the tooling to merge `sentry_value_t`'s right now are basically nonexistent.

relates to #601


relaxolotl:
i'll fix the unused parameter tomorrow

src/sentry_scope.c
```diff
@@ -269,7 +278,17 @@ sentry__scope_apply_to_event(const sentry_scope_t *scope,
...
     PLACE_CLONED_VALUE("extra", scope->extra);
-    PLACE_CLONED_VALUE("contexts", scope->contexts);
+
+    // TODO: better, more thorough deep merging
+    sentry_value_t contexts = sentry__value_clone(scope->contexts);
+    sentry_value_t trace = sentry__span_get_trace_context(scope->span);
+    if (!sentry_value_is_null(trace)) {
+        sentry_value_set_by_key(contexts, "trace", trace);
+    } else {
+        sentry_value_decref(trace);
```
Swatinem, line 288: you don’t need to decref a `null` value. though it doesn’t hurt either ;-)

src/sentry_scope.h
```diff
@@ -19,6 +19,12 @@ typedef struct sentry_scope_s {
     sentry_value_t breadcrumbs;
     sentry_level_t level;
     sentry_value_t client_sdk;
+    // Not to be confused with transaction, which is a legacy value. This is
+    // also known as a transaction, but to maintain consistency with other SDKs
+    // and to avoid a conflict with the existing transaction field this is named
+    // span. Whenever possible, `transaction` should pull its value from the
+    // `name` property nested in this field.
+    sentry_value_t span;
```
Swatinem, line 27: The description is very nice! Maybe name this `root_span` or something? Not sure what would make this less confusing.
Also not sure if you want to save the innermost span somewhere as well?

src/sentry_tracing.c
```diff
@@ -0,0 +1,31 @@
...
+{
+    if (sentry_value_is_null(span)
+        || sentry_value_is_null(sentry_value_get_by_key(span, "trace_id"))
+        || sentry_value_is_null(sentry_value_get_by_key(span, "span_id"))) {
+        return sentry_value_new_null();
+    }
+
+    sentry_value_t trace_context = sentry_value_new_object();
+
+#define PLACE_VALUE(Key, Source)                                               \
```
Swatinem, line 14: Can you also add a corresponding `#undef PLACE_VALUE` after the method please.
It looks like this is also missing in `sentry_scope.c`, must have been an oversight.

Swatinem (approved):

src/sentry_scope.h
relaxolotl: truth be told i'm actually not the biggest fan of calling these spans as there are subtle differences between a transaction and a span. the nice thing about using `span` here is mostly that it's consistent with the naming scheme used by other SDKs (see https://github.com/getsentry/sentry-python/blob/dd0efc08414ee2ef1a5f22d2cc4e243b54a1b455/sentry_sdk/scope.py#L81-L83). 

given that, i'm reluctant to use `root_span` because in theory, this "span" should be a transaction which means it actually should contain some sort of pointer to a span tree(s). there have been suggestions to call this something like `transaction_object`, which makes it less of a misnomer at the cost of giving up consistency. thoughts?

i think the question you have about the innermost span might be indirectly answered by the second paragraph: the stored "span", aka the transaction should have enough info to find any given span that branches off of it if needed. i'll admit however that i don't have a very good vision of how span tree construction and management will look like yet, and i'm leaving that as a challenge to be solved a little later.

flub (approved):

src/sentry_scope.h
relaxolotl: there's also a setter on scope that the spec explicitly requires to be named `set_span`: https://develop.sentry.dev/sdk/performance/#scope-changes

given that this setter is meant to be internal and this suggested change's intent is to make things easier for us internally, i'd like to do this: i'm going to merge this as-is while it still uses the terminology as required by the spec and as used by all of the other SDKs. a follow-up PR can be opened up which introduces this rename of `span` to `transaction_object` in the scope in addition to the introduction of a `set_transaction_object()` helper to mirror the rename. we can then assess whether the rename works or makes sense or not in there.

PR #623, opened 2021-12-10
fix: Read Windows Version from Registry

Swatinem:
We were previously relying on the `kernel32.dll` version, as does
Crashpad. However that version can lag behind OS releases, which can
manifest itself as failing integration tests, as the version does not
match the one returned from python `platform.version()`.

CC @Mixaill as you suggested this back in https://github.com/getsentry/sentry-native/pull/467#pullrequestreview-578269776

### DIFF
--- a/src/sentry_os.c
+++ b/src/sentry_os.c
@@ -11,13 +11,13 @@ void *
 sentry__try_file_version(LPCWSTR filename)
 {
 
-    DWORD size = GetFileVersionInfoSizeW(L"ntoskrnl.exe", NULL);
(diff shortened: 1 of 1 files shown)


Swatinem:
grafik

^ for reference how this shows up in sentry now with a Windows 10 21H2.

loewenheim (approved):

Mixaill:
@Swatinem For the kernel version I propose to look on `ntoskrnl.exe` (which an actual kernel executable) instead

image

Mixaill:
looks nice

PR #629, opened 2021-12-16
feat(tracing): Restrict `sentry_capture_event` so it only sends non-transaction events

relaxolotl:
Transactions will want to piggyback on most of what the original `sentry_capture_event` already does, but this means that it'll be possible to pass in transactions in to that function. This deliberately prevents users from being able to inadvertently finish transactions with this function.

relates to #601

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -1103,7 +1103,8 @@ SENTRY_API void sentry_user_consent_reset(void);
 SENTRY_API sentry_user_consent_t sentry_user_consent_get(void);
 
 /**
- * Sends a sentry event.
+ * Sends a sentry event. Returns a nil UUID if the event being passed in is a
+ * transaction; `sentry_transaction_finish` should be used to send transactions.
  */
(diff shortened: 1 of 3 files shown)


loewenheim (commented):
It looks like `sentry_capture_event` does nothing but check whether the argument is a transaction and call `sentry__capture_event` otherwise. Then why does `sentry__capture_event` do the same check again?

relaxolotl:
apologies, i should have left a TODO there. the `if` in the internal implementation was left as a stub to loosely indicate that something more meaningful will be filled in there, but i think it just caused more confusion.

https://github.com/getsentry/sentry-native/pull/631/files#diff-45ef05f9e59b750d3d615585ae87121db85ca379ca226f3b26ad809f85703783R387 does exactly this and demonstrates why the `if` is needed - `transaction_finish` uses `sentry__capture_event` and directly feeds a transaction in, which then needs to be handled appropriately. hopefully this answers your question, but let me know if it just caused more confusion.

flub (approved):

Swatinem (approved):

PR #630, opened 2021-12-16
feat(tracing): Add in basic Envelope support for Transactions

relaxolotl:
This makes it possible to add transactions to envelopes, assuming they are well-formed. Most of this follows the envelope spec as defined in https://develop.sentry.dev/sdk/envelopes/#transaction.

There is no checking to ensure the two constraints on Event and Transaction items as the current assumption is that Envelopes are created as-needed for every individual event or transaction that is sent over to sentry. This assumption is also held in other SDKs, so this isn't more lenient than other SDKs in this regard.

relates to #601 

picks up off of #629


src/sentry_envelope.c
```diff
@@ -234,6 +255,45 @@ sentry__envelope_add_event(sentry_envelope_t *envelope, sentry_value_t event)
     return item;
 }
 
+sentry_envelope_item_t *
+sentry__envelope_add_transaction(
```
relaxolotl, line 259: a majority of this is just copypasted from `sentry__envelope_add_event`, with a few differences:
- this sets a different item header value
- this sets a `sent_at` header value

this is duplicated to minimize impact on the event implementation, and to mostly follow the pattern established by other SDKs (they also have separate functions to add different types of events to an envelope). 

of course, this misses out on some of the nice perks that the other SDKs gain from having this pattern, but i think minimizing the impact of this change is enough to justify this for now. open to changing my opinion on this, though. if merged into `sentry__envelope_add_event`, this would only introduce 1 or 2 additional if statements (for now).

loewenheim (approved):

src/sentry_envelope.c
Swatinem: I think the `sent_at` value is applicable to all envelopes? Should we rather add this really at "send" time, as opposed to "adding item to envelope" time?

Also, maybe split this into a `sentry__envelope_add_value(sentry_value_t value, const char *type)` that rather takes the `type` header value explicitly?

Swatinem (approved):
lgtm, although maybe generalizing this further is worth it?

src/sentry_envelope.c
```diff
@@ -234,6 +255,45 @@ sentry__envelope_add_event(sentry_envelope_t *envelope, sentry_value_t event)
...
+
+    sentry_jsonwriter_t *jw = sentry__jsonwriter_new(NULL);
+    if (!jw) {
+        return NULL;
+    }
+
+    sentry_value_t event_id = sentry__ensure_event_id(transaction, NULL);
+
+    item->event = transaction;
+    sentry__jsonwriter_write_value(jw, transaction);
```
flub, line 275: this isn't really wrong because the value is still alive.   but rust wouldn't let you do this since `sentry__jsonwriter_write_value` borrows a value which you just passed ownership away from on the previous line.

flub (approved):

src/sentry_envelope.c
relaxolotl: > I think the sent_at value is applicable to all envelopes? Should we rather add this really at "send" time, as opposed to "adding item to envelope" time?

i wouldn't be opposed to this, although it looks like all of the code paths that use this (and `sentry__envelope_add_event`) send the envelope immediately afterwards. this means that any inaccuracies or timing offsets in`sent_at` would be caused by things like adding/ending the current session, or the amount of time it takes to push it completely through the transport.

what do you think of the idea of documenting this somewhere, and if somebody notices any issues with the `sent_at` timestamps we can set the header a little bit closer to when it's actually sent? something to consider is that events being sent by the native SDK right now (_not_ transactions) do not have the `sent_at` header. i'm not sure if it's because the header precedes this implementation, or if there's some sort of subtle timing issue that lead to this omission.

> maybe split this into a sentry__envelope_add_value(sentry_value_t value, const char *type) that rather takes the type header value explicitly?

this is a good idea and i'll give this a try in a follow-up PR, if that's alright with you. it should be a simple change, hopefully.

Swatinem: Sure, go for it in a followup. And the suggestion for `sent_at` seems to be moot as well. I don’t think we need that much precision after all :-D

PR #631, opened 2021-12-16
feat(tracing): Allow manual creation and sending of spanless Transactions

relaxolotl:
The proverbial "rest of the owl" PR. This hooks up everything so that it is possible to create and send a transaction, provided that the user is capable of manually finishing that transaction by threading it through everything it is meant to measure. 

The main purpose of this is to get some sort of transaction showing up on sentry. Things not covered in this PR:

- storing the current transaction in the scope and all transaction management and manipulation related to the scope (covered in https://github.com/getsentry/sentry-native/pull/632)
- child span creation (covered in #634)
- rate limiting
- resuming a transaction from headers

This is being submitted as a larger PR (in comparison to my previous ones) because much of the code relies on understanding a transaction's inner structure. I was hesitant to expose `transaction_start` nor `transaction_finish` in an incomplete state, but those two are important in providing context on what some of the functions here, such as `prepare_transaction` were doing.

relates to #601 

picks up off of #630


loewenheim (approved):

include/sentry.h
```diff
@@ -1284,6 +1284,20 @@ SENTRY_EXPERIMENTAL_API void sentry_transaction_set_sampled(
 SENTRY_EXPERIMENTAL_API void sentry_transaction_remove_sampled(
     sentry_value_t transaction);
 
+/**
+ * Starts a new transaction based on the provided context, restored
+ * from an external integration (i.e. a span from a different SDK)
+ * or manually constructed by a user.
+ */
+SENTRY_EXPERIMENTAL_API sentry_value_t sentry_transaction_start(
+    sentry_value_t transaction);
```
Swatinem, line 1293:
```suggestion
    sentry_value_t transaction_context);
```

I hate the this `…_context` naming btw, but it is what other SDKs also use, and it makes the distinction a bit clearer.

src/backends/sentry_backend_inproc.c
```diff
@@ -260,6 +260,7 @@ handle_ucontext(const sentry_ucontext_t *uctx)
 
         sentry_envelope_t *envelope
             = sentry__prepare_event(options, event, NULL);
+        // TODO: flush transaction sitting in scope?
```
Swatinem, line 263: related to a decision we still have to make: what happens in the case of a hard crash?
Also note that we have the breakpad and crashpad backends as well that should also execute on whatever we decide.

examples/example.c
```diff
@@ -93,9 +93,15 @@ main(int argc, char **argv)
...
 
+    if (has_arg(argc, argv, "capture-transaction")) {
+        sentry_options_set_traces_sample_rate(options, 1.0);
+    }
+
     sentry_init(options);
 
-    if (!has_arg(argc, argv, "no-setup")) {
+    if (!has_arg(argc, argv, "no-setup")
+        && !has_arg(argc, argv, "capture-transaction")) {
```
Swatinem, line 103: why this change? We should still add all the scope to the transaction (and assert that via tests)

src/sentry_core.c
```diff
@@ -511,6 +500,60 @@ sentry__prepare_event(const sentry_options_t *options, sentry_value_t event,
...
+        sentry__scope_apply_to_event(scope, options, transaction, mode);
+    }
+
+    sentry__ensure_event_id(transaction, event_id);
+    envelope = sentry__envelope_new();
+    if (!envelope || !sentry__envelope_add_transaction(envelope, transaction)) {
+        goto fail;
+    }
+
+    SENTRY_TRACE("adding attachments to envelope");
```
Swatinem, line 532: also related to a decision we still have to make: should we send attachments along with transactions? Probably not for the time being, and maybe turn that on later?

src/sentry_core.c
```diff
@@ -692,3 +735,71 @@ sentry_set_level(sentry_level_t level)
         scope->level = level;
     }
 }
+
+sentry_value_t
+sentry_transaction_start(sentry_value_t tx_cxt)
+{
+    sentry_value_t tx = sentry_value_new_event();
+
+    // TODO: stuff transaction into the scope
```
Swatinem, line 744: shouldn’t API users do that themselves rather?

src/sentry_core.c
```diff
@@ -692,3 +735,71 @@ sentry_set_level(sentry_level_t level)
...
+    sentry_value_set_by_key(
+        tx, "sampled", sentry_value_new_bool(should_sample));
+
+    sentry_value_set_by_key(
+        tx, "transaction", sentry_value_get_by_key_owned(tx_cxt, "name"));
+    sentry_value_set_by_key(tx, "start_timestamp",
+        sentry__value_new_string_owned(
+            sentry__msec_time_to_iso8601(sentry__msec_time())));
+
+    sentry_uuid_t trace_id = sentry_uuid_new_v4();
```
Swatinem, line 755: haven’t looked at the followup PRs yet, but the `trace_id`/`span_id`/`parent_span_id` should probably come from the context (and be defaulted there).

src/sentry_core.c
```diff
@@ -692,3 +735,71 @@ sentry_set_level(sentry_level_t level)
...
+
+    return tx;
+}
+
+void
+sentry_transaction_finish(sentry_value_t tx)
+{
+    sentry_value_t sampled = sentry_value_get_by_key(tx, "sampled");
+    if (!sentry_value_is_null(sampled) && !sentry_value_is_true(sampled)) {
+        sentry_value_decref(sampled);
```
Swatinem, line 773: `sentry_value_get_by_key` returns a borrowed ref, vs `sentry_value_get_by_key_owned`. So you don’t need to decref here.
I think this hides a latent bug, since `sampled` is most likely a bool, which is treated as a primitive type and is thus not refcounted at all.

src/sentry_core.c
```diff
@@ -692,3 +735,71 @@ sentry_set_level(sentry_level_t level)
...
+        sentry__value_new_string_owned(
+            sentry__msec_time_to_iso8601(sentry__msec_time())));
+    sentry_value_set_by_key(tx, "level", sentry_value_new_string("info"));
+
+    // TODO: add tracestate
+    // set up trace context so it mirrors the final json value
+    sentry_value_set_by_key(tx, "status", sentry_value_new_string("ok"));
+
+    sentry_value_t trace_context = sentry__span_get_trace_context(tx);
+    sentry_value_t contexts = sentry_value_new_object();
```
Swatinem, line 791: are we absolutely sure there is no `contexts` yet on the transaction value?

src/sentry_core.c
```diff
@@ -692,3 +735,71 @@ sentry_set_level(sentry_level_t level)
...
+    // clean up trace context fields
+    sentry_value_remove_by_key(tx, "trace_id");
+    sentry_value_remove_by_key(tx, "span_id");
+    sentry_value_remove_by_key(tx, "parent_span_id");
+    sentry_value_remove_by_key(tx, "op");
+    sentry_value_remove_by_key(tx, "description");
+    sentry_value_remove_by_key(tx, "status");
+
+    // This decrefs for us, generates an event ID, merges scope
+    sentry__capture_event(tx);
```
Swatinem, line 804: maybe clear this up: should we rather return the sent `event_id` in case this is being captured? (and a nil uuid in case it is discarded?)

Swatinem (changes requested):
looking good, though the decref of the primitive value technically hides a use after free ;-)
and I noted down some questions that need clearing up maybe.

include/sentry.h
flub, line 1293: after todays discussion are we going roll this function into `sentry_value_new_transaction`?

src/sentry_core.c
```diff
@@ -511,6 +500,60 @@ sentry__prepare_event(const sentry_options_t *options, sentry_value_t event,
...
+sentry_envelope_t *
+sentry__prepare_transaction(const sentry_options_t *options,
+    sentry_value_t transaction, sentry_uuid_t *event_id)
+{
+    sentry_envelope_t *envelope = NULL;
+
+    bool should_skip = !sentry_value_is_true(
+        sentry_value_get_by_key(transaction, "sampled"));
+    if (should_skip) {
+        SENTRY_DEBUG("throwing away transaction due to sample rate");
```
flub, line 512: nitpicking, but it is my understanding that this is not always due to the sample rate as you can explicitly sample-or-not a transaction overriding the sample rate?

src/sentry_core.h
```diff
@@ -56,6 +57,22 @@ sentry_envelope_t *sentry__prepare_event(const sentry_options_t *options,
...
+ * - discard the transaction if it is unsampled
+ * - apply the scope to the transaction
+ * - add the transaction to a new envelope
+ * - add any attachments to the envelope
+ *
+ * The function will ensure the transaction has a UUID and write it into the
+ * `event_id` out-parameter.
+ */
+sentry_envelope_t *sentry__prepare_transaction(const sentry_options_t *options,
+    sentry_value_t event, sentry_uuid_t *event_id);
```
flub, line 74: would be good to clearly document whether this borrows `event` or takes ownership.  basically any time a `sentry_value_t` arg is used i think this is important to write down.

src/sentry_core.c
flub: if we go with today's suggestion 3, then yes i think

src/sentry_core.c
flub: iiuc you can safey incref/decref primitive types so that wrt reference counting you can treat all sentry_value_t's the same?

examples/example.c
relaxolotl: was hoping somebody'd comment on this! looking back at this now, i think this can now be removed - in the past this actually used to break some of my testing while i was still working on getting transactions to work. the level setting, for example used to cause some issues for me. this isn't a problem anymore though, so this can be removed.

include/sentry.h
relaxolotl: i don't think we can/should roll this into `sentry_value_new_transaction`, unfortunately. i do think we may need to re-append `_context` to some of the methods if we're going ahead with this rename: a **transaction context** is at best a subset of what a full transaction contains, and some functions(/methods) on transactions would not work on a transaction context, and vice versa.

some examples:
- transaction sampling: whether a transaction will be sent or not (i.e. its `sampled` field is true/false) needs to be decided upon during transaction start
  - this is an immutable internal field on a transaction
  - this is a mutable field on a transaction context 
- `tag`s and `data`
  - transactions have these
  - transaction contexts do not

src/backends/sentry_backend_inproc.c
relaxolotl: for posterity, i believe we will want to try to flush and send the transaction in case a hard crash does occur, if something like that is possible.

https://github.com/getsentry/develop/issues/443 i believe tracks the work being done across all SDKs to implement this; we'll consider this to be a stretch goal for the time being though. any updates on this not being possible either due to time or technical constraints will be noted in the aforementioned issue.

i'll clean up this TODO so it provides a little bit more context and it's a little easier to find (ie searchable keyword) once we get around to implementing this.

src/sentry_core.c
relaxolotl: true that, i'll add that as a possible cause in here to avoid misleading anybody.

src/sentry_core.c
relaxolotl: i've already got attachments noted down as something to investigate, so it makes sense to remove this for the time being and to investigate this properly later. the docs (spec?) suggest that transactions indeed should have attachments: https://develop.sentry.dev/sdk/envelopes/#attachment

src/sentry_core.c
relaxolotl: no: the transaction which is occasionally referred to as as the root span, and is a superset of a span is managed by the native SDK. transactions tend to contain things that track all of the child spans that spawn off them, and notably have `name`s which spans do not. in addition to that, i'm pretty sure that the native SDK is committing to only ever letting one transaction be active at once, which couples nicely with the way the scope is implemented and managed in this SDK.

spans, which are all children of a transaction will be managed by the end users. 

that was my understanding of the discussion, and is subject to my shoddy short term memory. feel free to correct or comment on this conclusion.

src/sentry_core.c
relaxolotl: seconding @flub here: `sentry_value_decref` appears to "handle" primitives correctly so one could treat all `sentry_value_t`'s the same way.

while we know that this particular instance is likely to be a bool, would it be better to do this here as opposed to omitting it? by ignoring the implementation details it makes the API and its usage consistent across the codebase.

src/sentry_core.c
relaxolotl: by this point, yes. unless a user deliberately tries to insert a `context` properly via the `sentry_value` API instead of using the provided setters for transactions in the public API, the context field should not exist on a transaction yet during this method.

src/sentry_core.c
relaxolotl: good idea, i'll add that in.

src/sentry_core.c
```diff
@@ -389,15 +390,17 @@ sentry__capture_event(sentry_value_t event)
...
+                    // envelope it will be sent onwards.  This means we now
+                    // need to set the init flag to false because we're no
+                    // longer the initial session update.
                     mut_options->session->init = false;
                 }
             }
 
             sentry__capture_envelope(options->transport, envelope);
+        } else {
+            was_captured = false;
```
relaxolotl, line 403: while updating events i noticed this had a subtle bug: `capture_event` returns the event ID even if the event fails to send due to sampling and/or other internal issues. this fix takes care of sampling, however `sentry__capture_envelope` can silently fail due to missing user consent. i'm not sure if this was a deliberate decision, ie we do not want to let SDK users know when user consent has not been provided or if this was just overlooked.

i've made note of this as a TODO in the tests, but any insight on this would be great.

relaxolotl: fixing this comment by re-attaching it to a different line, please hold

src/sentry_core.c
```diff
@@ -377,10 +378,11 @@ sentry__capture_event(sentry_value_t event)
     sentry_envelope_t *envelope = NULL;
 
     bool was_captured = false;
+    bool was_sent = false;
```
relaxolotl, line 381: while updating events i noticed this had a subtle bug: capture_event returned the event ID even if the event failed to send due to sampling and/or other internal issues. this fix takes care of sampling, however `sentry__capture_envelope` can silently fail due to missing user consent. i'm not sure if this was a deliberate decision, ie we do not want to let SDK users know when user consent has not been provided or if this was just overlooked.

i've made note of this as a TODO in the tests, but any insight on this would be great.

relaxolotl:
addressed most comments and left replies to some where a code change wasn't needed, should be good for re-review.

cc @flub @Swatinem

include/sentry.h
Swatinem: > tags and data
> * transactions have these
> * transaction contexts do not

How would you end up adding these to a transaction then? Would they only end up on the transaction based on whatever is set on the scope?

src/sentry_core.c
```diff
@@ -689,3 +710,79 @@ sentry_set_level(sentry_level_t level)
...
+    sentry_value_set_by_key(
+        tx, "sampled", sentry_value_new_bool(should_sample));
+
+    // Avoid having this show up in the payload at all if it doesn't have a
+    // valid value
+    sentry_value_t parent_span
+        = sentry_value_get_by_key_owned(tx_cxt, "parent_span_id");
+    if (sentry_value_get_length(parent_span) > 0) {
+        sentry_value_set_by_key(tx, "parent_span_id", parent_span);
+    }
```
Swatinem, line 730:
```suggestion
    } else {
      sentry_value_decref(parent_span);
    }
```
Since you access it via `sentry_value_get_by_key_owned`.

src/sentry_core.c
Swatinem: I can construct an example that exhibits a double-free:

```
sentry_value_t tx = sentry_value_new_object();
sentry_value_set_by_key(tx, "sampled", sentry_value_new_double(0.0));
```

Yes, this is a very contrived example. But doubles are refcounted (not primitives), and `0.0` is considered `false` and run into this branch.

src/sentry_core.c
```diff
@@ -689,3 +710,79 @@ sentry_set_level(sentry_level_t level)
...
+    // TODO(tracing): stuff transaction into the scope
+    bool should_sample = sentry__should_send_transaction(tx_cxt);
+    sentry_value_set_by_key(
+        tx, "sampled", sentry_value_new_bool(should_sample));
+
+    // Avoid having this show up in the payload at all if it doesn't have a
+    // valid value
+    sentry_value_t parent_span
+        = sentry_value_get_by_key_owned(tx_cxt, "parent_span_id");
+    if (sentry_value_get_length(parent_span) > 0) {
```
flub, line 728: rather minor as the semantics end up being the same, but i was kind of expecting a `!sentry_value_is_null(parent_span)` here

include/sentry.h
```diff
@@ -1251,37 +1251,60 @@ SENTRY_EXPERIMENTAL_API double sentry_options_get_traces_sample_rate(
...
+SENTRY_EXPERIMENTAL_API void sentry_transaction_context_set_name(
     sentry_value_t transaction, const char *name);
 
 /**
- * Sets the `operation` of a Transaction.
+ * Sets the `operation` on a Transaction Context, which will be used in the
+ * Transaction constructed off of the context
  *
  * See https://develop.sentry.dev/sdk/performance/span-operations/ for
  * conventions on `operation`s.
```
flub, line 1269:
```suggestion
 * conventions on `operation`s.
 *
 * Borrows `transaction`.
```

include/sentry.h
```diff
@@ -1251,37 +1251,60 @@ SENTRY_EXPERIMENTAL_API double sentry_options_get_traces_sample_rate(
...
 /**
- * Sets the `sampled` field on a Transaction. When turned on, the Transaction
- * will bypass all sampling options and always be sent to sentry. If this is
- * explicitly turned off in the Transaction, it will never be sent to sentry.
+ * Sets the `sampled` field on a Transaction Context, which will be used in the
+ * Transaction constructed off of the context.
+ *
+ * When passed any value above 0, the Transaction will bypass all sampling
+ * options and always be sent to sentry. If passed 0, this Transaction and its
+ * child spans will never be sent to sentry.
```
flub, line 1280:
```suggestion
 * child spans will never be sent to sentry.
 *
 * Borrows `transaction`.
```

include/sentry.h
```diff
@@ -1251,37 +1251,60 @@ SENTRY_EXPERIMENTAL_API double sentry_options_get_traces_sample_rate(
...
  * actions that can be performed on a Transaction.
  */
-SENTRY_EXPERIMENTAL_API sentry_value_t sentry_value_new_transaction(
+SENTRY_EXPERIMENTAL_API sentry_value_t sentry_value_new_transaction_context(
     const char *name, const char *operation);
 
 /**
- * Sets the `name` of a Transaction.
+ * Sets the `name` on a Transaction Context, which will be used in the
+ * Transaction constructed off of the context.
```
flub, line 1259:
```suggestion
 * Transaction constructed off of the context.
 *
 * Borrows `transaction`.
```

src/sentry_core.c
flub: wait, while what i said is correct.  i misunderstood what was going on and now realise the right thing here is still do handle the reference counting correct because we don't know what type `sampled` actually is.  And because `sentry_value_get_by_key` returns a borrowed value we should not be decrefing it.

src/sentry_core.c
```diff
@@ -689,3 +710,79 @@ sentry_set_level(sentry_level_t level)
...
+
+    // clean up trace context fields
+    sentry_value_remove_by_key(tx, "trace_id");
+    sentry_value_remove_by_key(tx, "span_id");
+    sentry_value_remove_by_key(tx, "parent_span_id");
+    sentry_value_remove_by_key(tx, "op");
+    sentry_value_remove_by_key(tx, "description");
+    sentry_value_remove_by_key(tx, "status");
+
+    // This decrefs for us, generates an event ID, merges scope
```
flub, line 786:
```suggestion
    // This takes ownership of the transaction, generates an event ID, merges scope
```
(pretty minor, i know but i think it's clearer)

include/sentry.h
```diff
@@ -1251,37 +1251,60 @@ SENTRY_EXPERIMENTAL_API double sentry_options_get_traces_sample_rate(
...
+SENTRY_EXPERIMENTAL_API void sentry_transaction_context_set_sampled(
     sentry_value_t transaction, int sampled);
 
 /**
- * Removes the sampled field on a Transaction. The Transaction will use the
- * sampling rate as defined in `sentry_options`.
+ * Removes the sampled field on a Transaction Context, which will be used in the
+ * Transaction constructed off of the context.
+ *
+ * The Transaction will use the sampling rate as defined in `sentry_options`.
```
flub, line 1289:
```suggestion
 * The Transaction will use the sampling rate as defined in `sentry_options`.
 *
 * Borrows `transaction`.
```

include/sentry.h
```diff
@@ -1251,37 +1251,60 @@ SENTRY_EXPERIMENTAL_API double sentry_options_get_traces_sample_rate(
...
+ * external integration (i.e. a span from a different SDK) or manually
+ * constructed by a user.
+ */
+SENTRY_EXPERIMENTAL_API sentry_value_t sentry_transaction_start(
+    sentry_value_t transaction_context);
+
+/**
+ * Finishes and sends a transaction to sentry. The event ID of the transaction
+ * will be returned if this was successful; A nil UUID will be returned
+ * otherwise.
```
flub, line 1305:
```suggestion
 * otherwise.
 *
 * Takes ownership of `transaction`.
```

src/sentry_core.c
flub: my guess would be overlooked or too difficult for a first version (and second version never arrived)

src/sentry_core.h
```diff
@@ -56,6 +57,24 @@ sentry_envelope_t *sentry__prepare_event(const sentry_options_t *options,
...
+ * It will do the following things:
+ * - discard the transaction if it is unsampled
+ * - apply the scope to the transaction
+ * - add the transaction to a new envelope
+ * - add any attachments to the envelope
+ *
+ * The function will ensure the transaction has a UUID and write it into the
+ * `event_id` out-parameter. This takes ownership of the transaction, which
+ * means that the caller no longer needs to call `sentry_value_decref` on the
+ * transaction.
```
flub, line 73:
```suggestion
 * transaction.
 *
 * Takes ownership of `transaction`.
```

flub (approved):

include/sentry.h
relaxolotl: there should be functions to set data and tags on the Transaction itself, i.e. what's returned by `sentry_transaction_start()`. that's implemented in https://github.com/getsentry/sentry-native/pull/626.

`transaction_context`s on the other hand shouldn't have any of that information, nor should they offer the ability to set those. they're only really there so a user can set some values that are immutable after a Transaction has been started. to be fair, the `name` and `operation` on a `transaction_context` are exceptions to that rule, but it otherwise applies to everything else on a `transaction_context`.

src/sentry_core.c
relaxolotl: for some reason github was not able to figure out that the line was successfully changed, possibly due to a diffing algorithm quirk.

src/sentry_core.c
relaxolotl: yeah, i was really relying on the fact that `.._get_length()` handles nulls already here. a lot of the helper functions for `sentry_value_t` do so already, so i'm trying to remove any unnecessary null checks whereever i can as a result.

src/sentry_core.c
```diff
@@ -689,3 +710,79 @@ sentry_set_level(sentry_level_t level)
...
+        = sentry_value_get_by_key_owned(tx_cxt, "parent_span_id");
+    if (sentry_value_get_length(parent_span) > 0) {
+        sentry_value_set_by_key(tx, "parent_span_id", parent_span);
+    }
+    sentry_value_set_by_key(
+        tx, "trace_id", sentry_value_get_by_key_owned(tx_cxt, "trace_id"));
+    sentry_value_set_by_key(
+        tx, "span_id", sentry_value_get_by_key_owned(tx_cxt, "trace_id"));
+    sentry_value_set_by_key(
+        tx, "transaction", sentry_value_get_by_key_owned(tx_cxt, "name"));
```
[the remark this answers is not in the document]
relaxolotl: there's a future PR that validates the contents of these values: https://github.com/getsentry/sentry-native/pull/633

src/sentry_core.h
relaxolotl: L71:
`This takes ownership of the transaction, ...`

relaxolotl:
github is blocked on an explicit approval from arpad and i'd really prefer not to cheat the system, so 🙏 requesting a final OK on the changes here from @Swatinem

i'd like to also note that i've deliberately opted not to call out borrows on functions: i think the default assumption is that any passed in parameters are borrowed, and ownership theft is considered the exception. the documentation on existing public APIs support this in that they explicitly note when ownership is taken from the caller, but they do not go out of their way to mention when something is borrowed. cc @flub

Swatinem (approved):
lets unblock this then ;-)

flub:
> i'd like to also note that i've deliberately opted not to call out borrows on functions: i think the default assumption is that any passed in parameters are borrowed, and ownership theft is considered the exception. the documentation on existing public APIs support this in that they explicitly note when ownership is taken from the caller, but they do not go out of their way to mention when something is borrowed. cc @flub

I'm not going to block on this, but I regret the choice, explicit is often nice. but let's merge this instead of bikeshedding over this.

PR #633, opened 2021-12-18
feat(tracing): Defer some transaction validation and allow creation of internal spans

relaxolotl:
This moves some validation on transaction fields from the setter to `sentry_transaction_finish`. As the validity of these fields aren't crucial to most of the transaction's lifecycle, I figured it was been safe to move it to the end of that very lifecycle. A minor rename was made to the same field.

Since this code was touching the same places already, this also introduces a span creation function which the transaction creator now relies on. This follows the general recommendations made by https://develop.sentry.dev/sdk/performance/ regarding how spans and transactions should be structured and their relationship to one another (transactions should be specialized spans).

relates to #601 

picks up off of #631


loewenheim (approved):

tests/unit/test_tracing.c
```diff
@@ -3,6 +3,8 @@
 #include "sentry_tracing.h"
 #include "sentry_uuid.h"
 
+#define IS_NULL(Src, Field)                                                    \
```
Swatinem, line 6: maybe these helpers make sense in `sentry_testsupport.h`?

Swatinem (approved):

src/sentry_value.c
```diff
@@ -1124,10 +1124,32 @@ sentry_value_new_stacktrace(void **ips, size_t len)
     return stacktrace;
 }
 
+sentry_value_t
+sentry__value_new_span(sentry_value_t parent, const char *operation)
+{
+    sentry_value_t span = sentry_value_new_object();
+
+    sentry_transaction_context_set_operation(span, operation);
```
flub, line 1132: I wonder if this should now be renamed `sentry_span_set_operation(span, operation)`.  Setting the operation on the transaction context is already a very nice API because you might as well pass the right value into the `sentry_value_new_transaction_context()` call.

src/sentry_value.h
```diff
@@ -61,6 +61,13 @@ sentry_value_t sentry__value_new_list_with_size(size_t size);
  */
 sentry_value_t sentry__value_new_object_with_size(size_t size);
 
+/**
+ * Constructs a new Span.
+ *
```
flub, line 66:
```suggestion
 *
 * Borrows `parent`.
```

src/sentry_value.c
```diff
@@ -1124,10 +1124,32 @@ sentry_value_new_stacktrace(void **ips, size_t len)
...
+
+    return span;
+}
+
 sentry_value_t
 sentry_value_new_transaction_context(const char *name, const char *operation)
 {
-    sentry_value_t transaction_context = sentry_value_new_object();
+    sentry_value_t transaction_context
+        = sentry__value_new_span(sentry_value_new_null(), operation);
```
flub, line 1151: On line 1163 (not part of diff so i can't comment there) the operation is set again.  this is now not needed.

tests/unit/test_tracing.c
flub: this is too generic name.  I'd rather have a `sentry_key_is_null()` helper function which would be a lot more readable to me.  (have i mentioned i'm not a fan of macros?)

src/sentry_core.c
```diff
@@ -714,7 +714,13 @@ sentry_set_level(sentry_level_t level)
 void
 sentry_transaction_start(sentry_value_t tx_cxt)
 {
+    // TODO: it would be nice if we could just merge tx_cxt into tx.
+    // `sentry_value_new_transaction_event()` is also an option, but risks
+    // causing more confusion as there's already a
+    // `sentry_value_new_transaction`. The ending timestamp is stripped as well
```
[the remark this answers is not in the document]
relaxolotl: being able to invoke some sort of `sentry_value_merge_objects(tx, tx_cxt)` where `tx_cxt` wins out would be much more preferable to us manually enumerating over every `tx_cxt` field, checking for its existence, and stuffing the value into `tx` if it's present in `tx_cxt`. that's what this means by "just merging tx_cxt into tx". the function just wants a union of those two objects.

src/sentry_value.c
relaxolotl: i think what we can do is define one common implementation for setting operations on spans/transactions/transaction contexts, but i'm thinking that the final API will most likely include a `sentry_span_set_operation(...)`, `sentry_transaction_set_operation(...)`, and `sentry_transaction_context_set_operation(...)`.

i think having a setter regardless can be useful in case one is only able to figure out what a transaction might be after a transaction context is created, or a transaction is started. it's not an immutable field until the transaction is finished.

regarding the naming and reusing `..._span_[operation]` for transactions and transaction contexts, i can't say that i'm too big of a fan of relying on the implicit knowledge that transactions are spans with extra bells and whistles in this SDK, since:
1. there's no such thing as "inheritance" in this to begin with
2. nothing in the types or the codebase hint at the idea that a transaction "is" a span

this might make sense for other SDKs where class/struct inheritance does exist, and transactions do indeed get `set_operation` and other methods for free by virtue of those extending the span implementation. the native SDK doesn't really programmatically support this. i also don't know if the user gains much from knowing that a transaction is loosely equivalent to a span.

src/sentry_value.c
relaxolotl: ah, thanks for catching this

tests/unit/test_tracing.c
relaxolotl: seeing as there's already another helper macro in the tests right below, can i defer any relocations or changes to the macros to https://github.com/getsentry/sentry-native/pull/637?

relaxolotl:
fyi: this has been detached from #632's changes and attached to #631 instead, meaning that this can be merged once the latter lands on master. this is because #632's changes are likely to be contentious or will most likely go through a bit of churn because of the nature of its changes, and this PR doesn't particularly rely on #632's changes. it mostly relies on #631.

src/sentry_value.h
relaxolotl: as noted in https://github.com/getsentry/sentry-native/pull/631#issuecomment-999960844, i think it would be fine to not mention places where parameters are borrowed, as that's the assumed default.

PR #634, opened 2021-12-18
feat(tracing): Basic span support with nesting

relaxolotl:
As noted in the title, this adds support for basic spans (no tags, no extra data). Spans can be created as children of some provided transaction, or as children of other spans.

No checking is made to ensure that a child span being finished belongs to the claimed transaction.

relates to #601 

picks up off of #633


include/sentry.h
```diff
@@ -1295,17 +1295,64 @@ SENTRY_EXPERIMENTAL_API void sentry_transaction_context_remove_sampled(
...
+ */
+SENTRY_EXPERIMENTAL_API sentry_value_t sentry_span_start_child(
+    sentry_value_t parent_span, char *operation, char *description);
+
+/**
+ * Finishes a span.
+ *
+ * Returns a value that should be passed into `sentry_span_finish`. Not
+ * finishing the span means it will be discarded, and will not be sent to
+ * sentry.
```
Swatinem, line 1348: this comment is wrong?

src/sentry_tracing.c
```diff
@@ -11,23 +12,53 @@ sentry__span_get_trace_context(sentry_value_t span)
...
         }                                                                      \
     } while (0)
 
-    PLACE_VALUE("trace_id", span);
-    PLACE_VALUE("span_id", span);
-    PLACE_VALUE("parent_span_id", span);
-    PLACE_VALUE("op", span);
-    PLACE_VALUE("description", span);
-    PLACE_VALUE("status", span);
+    PLACE_CLONED_VALUE("trace_id", span);
```
Swatinem, line 24: I don’t think all of these things should be "cloned". Rather just incref. strings are supposed to be immutable.

Swatinem (commented):
While I think the aggressive discarding has good intentions, IMO it is the wrong approach.

Instead `start` should just create the objects and set the `parent_span_id`, etc.
And then in `finish` you can decide to just discard stuff based on the `max_spans` or sampling.

I think stuff is intentionally threaded through everywhere, so that for example distributed tracing works correctly, with regards to the header etc. You should get a trace_id/parent_span_id/sampled triple no matter in which state you are.
Though we can also discuss this with a broader set of people.

include/sentry.h
```diff
@@ -1295,17 +1295,64 @@ SENTRY_EXPERIMENTAL_API void sentry_transaction_context_remove_sampled(
...
+ *
+ * If `parent_span` is `sentry_value_null`, then the current active Transaction
+ * is used as the parent for the new Span. An active Transaction must be created
+ * via `sentry_transaction_start` in order for the Span to be successfully
+ * created.
+ *
+ * If `parent_span` is another Span, it must belong to the current active
+ * Transaction in order for Span creation to succeed. This will take ownership
+ * of any `parent_span`s that do reference non-existent Spans in the current
+ * active Transaction.
```
flub, line 1326: i can't really parse this paragraph...  I'd expect that the parent span is only borrowed for this function call because the caller will still have to pass the parent to it's own finish call.

flub (commented):
I think in general we want to return an actual span from `sentry_span_start_child()` and not a "span_context".  The span only does get stored once finish is called `sentry_span_finish(sentry_value_t parent, sentry_value_t span)` and that way you can finish spans without needing any global state or being worried about which transaction you belong to.

include/sentry.h
relaxolotl: ownership of `parent_span` is stolen if it logically should not exist anymore: for example, if for some reason the `parent_span` was not finished before its parent Transaction was finished, there is no point letting the user finish the `parent_span` nor is there any point in letting more children be created from it, as all of its descendants will never be reported to sentry. 

the idea was to steal ownership and to discard the `parent_span` to prevent the user from performing more no-ops with an invalid span.

include/sentry.h
relaxolotl: it gets a little too far ahead of itself, yeah; this belongs in https://github.com/getsentry/sentry-native/pull/635.

src/sentry_core.c
```diff
@@ -785,4 +799,170 @@ sentry_transaction_finish(sentry_value_t tx)
...
+    }
+    sentry_value_t spans = sentry_value_get_by_key(parent, "spans");
+    // This only checks that the number of _completed_ spans matches the number
+    // of max spans. This means that the number of in-flight spans can exceed
+    // the max number of spans.
+    if (sentry_value_get_length(spans) >= max_spans) {
+        SENTRY_DEBUG("reached maximum number of spans for transaction, not "
+                     "creating span");
+        goto fail;
+    }
```
relaxolotl, line 871: this is probably going to receive a lot of contention. i don't really see how it makes sense to let users continue to create and populate spans if they've already hit their span limit. this is already short circuiting if the parent span meets certain bad criteria so users are already expected to handle situations where a null is returned during span creation.

relaxolotl:
this has been updated to return a span after a successful `sentry_span_start_child` invocation, and puts the full responsibility of managing spans on the end user.

relaxolotl:
moving to draft for same reasons outlined in https://github.com/getsentry/sentry-native/pull/632#issuecomment-1000555872.

examples/example.c
```diff
@@ -214,14 +218,24 @@ main(int argc, char **argv)
...
 
         if (has_arg(argc, argv, "unsample-tx")) {
             sentry_transaction_context_set_sampled(tx_ctx, 0);
         }
-
         sentry_transaction_start(tx_ctx);
+
+        if (has_arg(argc, argv, "child-spans")) {
+            sentry_value_t child_ctx = sentry_span_start_child(
+                sentry_value_new_null(), "littler.teapot", NULL);
```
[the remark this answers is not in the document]
Swatinem: ☝🏻 same impression, the transaction should be the "parent" here.

include/sentry.h
```diff
@@ -1346,6 +1346,45 @@ SENTRY_EXPERIMENTAL_API sentry_uuid_t sentry_transaction_finish(
  * into this function.
  */
 SENTRY_EXPERIMENTAL_API void sentry_set_span(sentry_value_t transaction);
+
+/**
+ * Starts a new Span.
+ *
+ * Either the return value of `sentry_start_transaction`, `sentry_set_span`, or
```
Swatinem, line 1353:
```suggestion
 * Either the return value of `sentry_start_transaction`, or
```

That fn was changed recently

include/sentry.h
```diff
@@ -1346,6 +1346,45 @@ SENTRY_EXPERIMENTAL_API sentry_uuid_t sentry_transaction_finish(
...
+/**
+ * Finishes a Span.
+ *
+ * Returns a value that should be passed into `sentry_span_finish`. Not
+ * finishing the Span means it will be discarded, and will not be sent to
+ * sentry.
+ *
+ * `root_transaction` is either the parent Transaction of the Span, or
+ * the ancestor Transaction of the Span if the Span is not a direct descendant
+ * of a Transaction.
```
Swatinem, line 1380: Lets land this as is for now to not delay this any further. But IMO this should not be a parameter, but be set during `start_child`.

src/sentry_core.c
Swatinem: IMO the idea should be that you should *always* be able to get the trace_id/span_id to attach to outgoing distributed tracing requests, even if you discard the span internally for whatever reason. So the span needs to "exist" internally.

src/sentry_core.c
```diff
@@ -848,4 +850,108 @@ sentry_set_span(sentry_value_t transaction)
...
+{
+    size_t max_spans = SENTRY_SPANS_MAX;
+    SENTRY_WITH_OPTIONS (options) {
+        max_spans = options->max_spans;
+    }
+
+    if (sentry_value_is_null(parent)) {
+        SENTRY_DEBUG("no transaction available to create a child under");
+        goto fail;
+    }
```
Swatinem, line 866: You were noting in chat that the example stopped working at some point. This is the reason why. In the example your first child span is not linked up to the parent transaction.

Swatinem (approved):
I did have a bunch of comments. Please keep those in mind and address them in a followup. I don’t want to hold this up from progressing right now.

Haven’t looked in detail at the unit tests though, but I trust you ;-)

src/sentry_core.c
relaxolotl: What doesn't line up for me is the fact that existing SDKs _do_ pre-emptively drop spans if they've reached the limit. At the same time, those SDKs also support distributed tracing which suggests to me that one of the two situations is happening:
- existing SDKs do not implement distributed tracing properly
- distributed traces do not continue directly off of spans

If it turns out that we're in the former scenario, the SDK TSC has acknowledged this as a known issue and we're making an active attempt to do better for this specific problem in the native SDK, I would be open to the idea of keeping the span internally in order to avoid breaking distributed tracing. I haven't found any documentation yet that suggests this is the case, however. It is also possible that I may have done a poor job of finding such documentation, so references/links to those would be welcome.

PR #635, opened 2021-12-18
feat(tracing): Drop unfinished spans from a transaction

relaxolotl:
Remove all unfinished spans from transactions as they're being processed and sent over the wire to sentry. This addresses a TODO left behind in #634.

related to #601 
picks up off of #634


loewenheim (approved):

Swatinem (commented):
This ties in with my comment on the previous PR in this chain:
If we only ever record spans in `finish` (that means that `start` only creates the object but otherwise does nothing) this whole PR would become obsolete.

PR #636, opened 2021-12-18
feat(tracing): Don't allow missing orphan spans to be parents or be finished

relaxolotl:
Sometimes it's possible to have orphan spans hanging around due to transactions inadvertently being finished before all of its child spans are complete. This adds in handling if we're able to detect those scenarios, preventing those orphan spans from causing cascading failures. The SDK swallows and deallocs such spans.

related to #601 
picks up off of #635

### DIFF
--- a/src/sentry_core.c
+++ b/src/sentry_core.c
@@ -845,29 +845,41 @@ sentry_span_start_child(
         // There isn't an active transaction. This span has nothing to attach
         // to.
         if (sentry_value_is_null(scope->span)) {
-            return sentry_value_new_null();
+            goto fail;
         }
         // Aggressively discard spans if a transaction is unsampled to avoid
         // wasting memory
         sentry_value_t sampled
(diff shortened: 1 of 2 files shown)


loewenheim (approved):

Swatinem (commented):
This might be a bit more validation than we should be doing in the SDK.
From the data model, the `spans` is just a flat list, and the hierarchy is then reconstructed either on the server, or even the UI. Not sure how much validation happens in later steps, but the SDK might not be the right place for that. Please check with other team members however.

flub:
i also think this might not be needed if we don't receive spans until they are finished as is already suggested in an earlier PR.  Because we'd just drop the orphaned span at the time it is being finished (the transaction ID would not match).

relaxolotl:
full responsibility is now being placed on users to manage spans and to drop them appropriately if spans are being finished after a transaction is finished.

PR #642, opened 2021-12-23
fix: Revert to mmap-ing modules in the modulefinder

Swatinem:
This goes a bit further than #641 and brings back mmap-ing modules in such a way that it still works with libraries loaded from apk on Android.

### DIFF
--- a/src/modulefinder/sentry_modulefinder_linux.c
+++ b/src/modulefinder/sentry_modulefinder_linux.c
@@ -66,6 +66,8 @@ sentry__mmap_file(sentry_mmap_t *rv, const char *path)
         goto fail;
     }
 
+    close(fd);
+
     return true;
 
 fail:
@@ -75,8 +77,6 @@ fail:
     rv->ptr = NULL;
     rv->len = 0;
     return false;
-
-    return rv;
 }
 
 void
--- a/src/modulefinder/sentry_modulefinder_linux.c
+++ b/src/modulefinder/sentry_modulefinder_linux.c
@@ -13,6 +13,8 @@
 #include <elf.h>
 #include <fcntl.h>
 #include <string.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
 #include <sys/syscall.h>
 #include <sys/types.h>
 #include <sys/uio.h>
@@ -40,6 +42,50 @@ static sentry_mutex_t g_mutex = SENTRY__MUTEX_INIT;
 static sentry_value_t g_modules = { 0 };
 
 static sentry_slice_t LINUX_GATE = { "linux-gate.so", 13 };
+
+bool
+sentry__mmap_file(sentry_mmap_t *rv, const char *path)
+{
+    int fd = open(path, O_RDONLY);
+    if (fd < 0) {
+        goto fail;
+    }
+
+    struct stat sb;
+    if (stat(path, &sb) != 0 || !S_ISREG(sb.st_mode)) {
+        goto fail;
+    }
+
+    rv->len = sb.st_size;
+    if (rv->len == 0) {
+        goto fail;
+    }
+
+    rv->ptr = mmap(NULL, rv->len, PROT_READ, MAP_PRIVATE, fd, 0);
+    if (rv->ptr == MAP_FAILED) {
+        goto fail;
+    }
+
+    return true;
+
+fail:
+    if (fd > 0) {
+        close(fd);
+    }
+    rv->ptr = NULL;
+    rv->len = 0;
+    return false;
+
+    return rv;
+}
+
+void
+sentry__mmap_close(sentry_mmap_t *m)
+{
+    munmap(m->ptr, m->len);
+    m->ptr = NULL;
+    m->len = 0;
+}
 
 /**
  * Checks that `start_offset` + `size` is a valid contiguous mapping in the
@@ -111,6 +157,10 @@ sentry__module_read_safely(void *dst, const sentry_module_t *module,
     void *src = sentry__module_get_addr(module, start_offset, size);
     if (!src) {
         return false;
+    }
+    if (module->is_mmapped) {
+        memcpy(dst, src, (size_t)size);
+        return true;
     }
     return read_safely(dst, src, (size_t)size);
 }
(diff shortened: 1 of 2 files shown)


src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -419,13 +475,53 @@ sentry__procmaps_module_to_value(const sentry_module_t *module)
...
+        sentry_mmap_t mm;
+        if (!sentry__mmap_file(&mm, filename)) {
+            sentry_free(filename);
+            sentry_value_decref(mod_val);
+            return sentry_value_new_null();
+        }
+        sentry_free(filename);
+
+        uint64_t needed_mmap_size = module->offset_in_inode + module_size;
+        if (mm.len < needed_mmap_size) {
```
JodiTheTigger, line 507: For me it seems that mappings are larger than the file they are mapping. This results in them failing as this is always true.

For example:

problem file: `libicui18n.so.69.1`

```bash
    # ls -l /usr/lib/libicui18n.so.69.1
    -rwxr-xr-x    1 root     root       2320048 Dec 23 06:07 /usr/lib/libicui18n.so.69.1
```

2,320,048 bytes = 0x23 66B0

```bash
    # cat /proc/688/maps | grep icui
    99255000-9948c000 r--p 00000000 b3:06 1853       /usr/lib/libicui18n.so.69.1 (0x  23 7000 = 2,322,432)
    b2481000-b26af000 r-xp 00000000 b3:06 1853       /usr/lib/libicui18n.so.69.1 (0x  22 E000 = 2,285,568)
    b26af000-b26be000 ---p 0022e000 b3:06 1853       /usr/lib/libicui18n.so.69.1
    b26be000-b26c6000 r--p 0022d000 b3:06 1853       /usr/lib/libicui18n.so.69.1 (0x     8000 =    32,768)
    b26c6000-b26c8000 rw-p 00235000 b3:06 1853       /usr/lib/libicui18n.so.69.1 (0x     2000 =     8,192)
```

the `99255000` mapping is too large for the file size
and the `b26c6000` mapping also overruns the file

needed size is 0x24 7000 but the file size is only 0x23 66b0

I replaced the code with this, and it seems to work for me, however I'm not sure if this is valid or not.

```cpp
        // uint64_t needed_mmap_size = module->offset_in_inode + module_size;
        // if (mm.len < needed_mmap_size) {
        //     sentry__mmap_close(&mm);
        //     sentry_value_decref(mod_val);
        //     return sentry_value_new_null();
        // }

        sentry_module_t mmapped_module;
        memset(&mmapped_module, 0, sizeof(sentry_module_t));
        mmapped_module.mappings[0].size = mm.len; // module_size;
```

Swatinem: 👍🏻 That sounds very good!
I think the mapping can be either bigger because the loader is rounding up to page sizes. Or smaller if the loader just does not map some "unneeded" sections. Using the on-disk size should work there.

Thanks for taking a look!

src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -41,6 +43,53 @@ static sentry_value_t g_modules = { 0 };
 
 static sentry_slice_t LINUX_GATE = { "linux-gate.so", 13 };
 
+bool
+sentry__mmap_file(sentry_mmap_t *rv, const char *path)
+{
+    rv->fd = open(path, O_RDONLY);
```
flub, line 49: you don't need to keep the fd in the sentry_mmap_t struct, part of the point of mmap is that after you called mmap you can close the fd and the mmap stays valid while you don't consume filedescriptors.  So you should store the fd on the stack and close it at the end.

src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -315,15 +371,15 @@ get_code_id_from_text_fallback(const sentry_module_t *module)
...
         for (int i = 0; i < elf.e_shnum; i++) {
             Elf64_Shdr header;
             ENSURE(sentry__module_read_safely(&header, module,
                 elf.e_shoff + elf.e_shentsize * i, sizeof(Elf64_Shdr)));
 
-            const char *name = names + header.sh_name;
+            char name[6];
+            ENSURE(sentry__module_read_safely(name, module,
+                strheader.sh_offset + header.sh_name, sizeof(name)));
+            name[5] = '\0';
```
flub, line 382: to make reviewing easier it is nicer if you set the base to the other branches this started from.  once those branches are merged github will automatically change the base to master and this can then be merged.

src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -419,13 +475,53 @@ sentry__procmaps_module_to_value(const sentry_module_t *module)
...
+
+    // At least on the android API-16, x86 simulator, the linker apparently
+    // does not load the complete file into memory. Or at least, the section
+    // headers which are located at the end of the file are not loaded, and
+    // we would be poking into invalid memory. To be safe, we mmap the
+    // complete file from disk, so we have the on-disk layout, and are
+    // independent of how the runtime linker would load or re-order any
+    // sections. The exception here is the linux-gate, which is not an
+    // actual file on disk, so we actually poke at its memory.
+    if (sentry__slice_eq(module->file, LINUX_GATE)) {
```
flub, line 494: pretty sure its more than just gate, they're all known as "vdso" and they appear under several names, see https://man7.org/linux/man-pages/man7/vdso.7.html which includes a table.

Swatinem: Its still only a single one as the man page says. They just have different names depending on the platform. However we synthesize the name anyway, so I don’t think that matters much.

src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -41,6 +43,50 @@ static sentry_value_t g_modules = { 0 };
...
+    if (rv->len == 0) {
+        goto fail;
+    }
+
+    rv->ptr = mmap(NULL, rv->len, PROT_READ, MAP_PRIVATE, fd, 0);
+    if (rv->ptr == MAP_FAILED) {
+        goto fail;
+    }
+
+    return true;
```
flub, line 69: the fd currently leaks in the success path, you still need to close it yourself.

flub (approved):
if the leaking fd is fixed i think its fine

src/modulefinder/sentry_modulefinder_linux.c
Swatinem: nice catch!

PR #645, opened 2022-01-04
feat(tracing): Move all performance monitoring code behind a compile flag

relaxolotl:
This moves both the external and the internal parts of the API related to performance monitoring (/tracing) behind a compile flag. This is to make it easier for us to iterate on the API, and to potentially break it or leave it in a compilable but incomplete state. 

If an end user runs into any issues with the feature they must have deliberately opted into the feature and/or attempted to use something that isn't publicly documented and is therefore not yet publicly supported by the SDK yet.

relates to #601

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -233,7 +233,7 @@ set_target_properties(sentry PROPERTIES PUBLIC_HEADER "include/sentry.h")
 
 if(DEFINED SENTRY_FOLDER)
 	set_target_properties(sentry PROPERTIES FOLDER ${SENTRY_FOLDER})
-endif()	
+endif()
 
 # check size type
 include(CheckTypeSize)
@@ -457,7 +457,7 @@ if(SENTRY_BACKEND_CRASHPAD)
 			set_target_properties(crashpad_util PROPERTIES FOLDER ${SENTRY_FOLDER})
 			set_target_properties(crashpad_zlib PROPERTIES FOLDER ${SENTRY_FOLDER})
 			set_target_properties(mini_chromium PROPERTIES FOLDER ${SENTRY_FOLDER})
-		endif()	
+		endif()
 
 		target_link_libraries(sentry PRIVATE
 			$<BUILD_INTERFACE:crashpad::client>
@@ -490,8 +490,8 @@ elseif(SENTRY_BACKEND_BREAKPAD)
 
 		if(DEFINED SENTRY_FOLDER)
 			set_target_properties(breakpad_client PROPERTIES FOLDER ${SENTRY_FOLDER})
-		endif()	
-		
+		endif()
+
 		if(NOT SENTRY_BUILD_SHARED_LIBS)
 			sentry_install(TARGETS breakpad_client EXPORT sentry
 				LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
@@ -559,6 +559,9 @@ if(SENTRY_BUILD_EXAMPLES)
 	add_executable(sentry_example examples/example.c)
 	target_link_libraries(sentry_example PRIVATE sentry)
 
+	target_compile_definitions(sentry PRIVATE SENTRY_PERFORMANCE_MONITORING)
+	target_compile_definitions(sentry_example PRIVATE SENTRY_PERFORMANCE_MONITORING)
+
 	if(MSVC)
 		target_compile_options(sentry_example PRIVATE $<BUILD_INTERFACE:/wd5105>)
 	endif()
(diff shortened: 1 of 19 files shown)


CMakeLists.txt
```diff
@@ -559,6 +559,9 @@ if(SENTRY_BUILD_EXAMPLES)
 	add_executable(sentry_example examples/example.c)
 	target_link_libraries(sentry_example PRIVATE sentry)
 
+	target_compile_definitions(sentry PRIVATE SENTRY_PERFORMANCE_MONITORING)
```
Swatinem, line 562: `SENTRY_BUILD_EXAMPLES` is I think the default if you directly compile the project (in contrast to compiling it as a sub-project), so IMO this goes a bit too far (enables it by default).
Better to make it a really opt-in flag, similar to the client-side stack trace feature.

include/sentry.h
```diff
@@ -1110,10 +1112,16 @@ SENTRY_API void sentry_user_consent_reset(void);
  */
 SENTRY_API sentry_user_consent_t sentry_user_consent_get(void);
 
+#ifdef SENTRY_PERFORMANCE_MONITORING
```
Swatinem, line 1115: ifdef-ing doc comments is pretty useless :-D Not even sure what it does.

src/CMakeLists.txt
```diff
@@ -153,3 +153,7 @@ if(SENTRY_INTEGRATION_QT)
 		integrations/sentry_integration_qt.h
 	)
 endif()
+
+if(SENTRY_BUILD_EXAMPLES)
+	target_compile_definitions(sentry PRIVATE SENTRY_PERFORMANCE_MONITORING)
```
Swatinem, line 158: this is redundant with the definition in the main cmake file.

src/sentry_envelope.c
```diff
@@ -200,15 +200,23 @@ sentry_envelope_get_event(const sentry_envelope_t *envelope)
         return sentry_value_new_null();
     }
     for (size_t i = 0; i < envelope->contents.items.item_count; i++) {
+
+#ifdef SENTRY_PERFORMANCE_MONITORING
```
Swatinem, line 204: now that I see this again, maybe we *should* return the transaction here. remember: a transaction *is* an event, just a very special one. It has an id, and can have associated attachments.

src/sentry_tracing.c
```diff
@@ -1,4 +1,5 @@
-#include "sentry_sync.h"
+#ifdef SENTRY_PERFORMANCE_MONITORING
```
Swatinem, line 1: the whole file should ideally only be compiled if the flag is set, otherwise the file is still part of the build, just completely empty.

src/sentry_tracing.h
```diff
@@ -1,14 +1,15 @@
-#ifndef SENTRY_TRACING_H_INCLUDED
-#define SENTRY_TRACING_H_INCLUDED
+#ifdef SENTRY_PERFORMANCE_MONITORING
```
Swatinem, line 1: same here

tests/unit/sentry_testsupport.h
```diff
@@ -7,6 +7,10 @@
 #include <stdio.h>
 #include <string.h>
 
+#ifndef SENTRY_PERFORMANCE_MONITORING
+#    define SENTRY_PERFORMANCE_MONITORING
```
Swatinem, line 11: cmake should set this flag (which it does). otherwise it very much depends on the include order, which you have seen can break things in surprising non-obvious ways.

Swatinem (approved):

include/sentry.h
relaxolotl: "please ignore this block if you are not using this compile flag" 😎

src/sentry_envelope.c
relaxolotl: i'm going to avoid referencing more than one SDK here, but [python](https://github.com/getsentry/sentry-python/blob/49cae6009a4e39c47ef8834b07668f5eb9789ca8/sentry_sdk/envelope.py#L256-L269) for example deliberately has separate getters for an event and a transaction in an envelope, for a counterexample.

CMakeLists.txt
relaxolotl: i don't think i have the knowledge to convert `SENTRY_PERFORMANCE_MONITORING` into a proper opt-in flag considering what i'm trying to prioritize right now:
- anybody building the example via `make build` should get perf monitoring out of the box, targeted towards other devs who may be contributing to perfmon 
- unit tests should have perf monitoring enabled by default

i could create a separate `make` command to build the project with the performance monitoring flag enabled. this just raises other build issues with the unit tests though, as they're missing symbols locked behind the feature flag. 

i'm open to any help to actually make this work, but right now automatically turning the flag on when `SENTRY_BUILD_EXAMPLES` is on appears to be my best bet to getting this working. the long term plan is to either remove the flag entirely or to fully commit to it anyways, so this should be a good enough temporary solution i think.

relaxolotl: i'll leave the PR open until the start of the work day tomorrow just to get some follow-up feedback on this and to see if there are any deal breakers in the current diff. otherwise, i'll be merging this in.

PR #649, opened 2022-01-12
feat(tracing): Introduce structs for performance monitoring constructs

relaxolotl:
This adds in a series of structs that wrap around `sentry_value_t` representations of major concepts in performance monitoring. The changes here are merely an incremental step towards a final goal of having "proper" structs which will fully replace `sentry_value_t` for transactions and spans. Those structs will ideally have their own concurrency and reference counting semantics.

I am currently not entirely convinced that the structs are currently any more meaningful nor useful than `sentry_value_t`s at the moment; There are hints that lead to the idea that the higher level goal as declared in the previous paragraph will mostly amount to just a duplication of the existing design and logic used for `sentry_value_t`.

A bulk of the PR's changes are focused on changing various return values and parameters so they use `sentry_transaction_context_t`/`sentry_transaction_t`/`sentry_span_t` instead of `sentry_value_t`. Other notable changes include:
- moving a majority of the helper functions in `sentry_value.[h|c]` for transactions and spans into `sentry_tracing.[h|c]`
- major amounts of code duplication in tests to get access to the inner `sentry_value_t` struct in transactions/spans for assertion purposes
- refactoring `sentry_span_start_child` by introducing `sentry_transaction_start_child` and then pulling common code between both implementations into an internal helper
- very minor refactoring of the `[set|remove]_[tag|data]` functions

Moving forward, follow up tasks for this should look loosely like the following:
- explore using some union type to represent spans and transactions
- begin removing `sentry_value_t` fields in transaction/span structs and explicitly declare members on them instead
- introduce helpers that convert `sentry_transaction_context_t`/`sentry_transaction_t`/`sentry_span_t` back to a `sentry_value_t`

The latter two goals could be modified to minimize the amount of code required to convert transactions/spans to `sentry_value_t`s. One example of this is to keep the inner `sentry_value_t` on the structs, and to only declare members that are references to other transactions and spans. i.e. `sentry_transaction_t` gets a `spans` field that is an array or linked list of references to `sentry_span_t`s, and `sentry_span_t`gets a `transaction` field that is a reference to an ancestor `sentry_transaction_t`.


src/sentry_core.c
```diff
@@ -829,93 +839,93 @@ sentry_transaction_finish(sentry_value_t tx)
...
-    return child;
-
-fail:
-    return sentry_value_new_null();
+    return sentry__start_child(max_spans, parent, operation, description);
+    // TODO: add pointer to ancestor transaction on child span
 }
 
+// TODO: don't accept the root transaction as a param, the span should have a
+// ref to it already in itself
```
Swatinem, line 902: 👍🏻 that was one of the main reasons to do this change ;-)

tests/unit/test_tracing.c
```diff
@@ -11,67 +11,91 @@
...
 }
 
 SENTRY_TEST(basic_transaction)
 {
-    sentry_value_t tx_cxt = sentry_value_new_transaction_context(NULL, NULL);
+    sentry_transaction_context_t *opaque_tx_cxt
+        = sentry_transaction_context_new(NULL, NULL);
+    TEST_CHECK(opaque_tx_cxt != NULL);
+
+    sentry_value_t tx_cxt = opaque_tx_cxt->inner;
```
Swatinem, line 57: looks like clang-analyzer does not like this:
```
[HIGH] /home/runner/work/sentry-native/sentry-native/tests/unit/test_tracing.c:57:29: Access to field 'inner' results in a dereference of a null pointer (loaded from variable 'opaque_tx_cxt') [core.NullDereference]
    sentry_value_t tx_cxt = opaque_tx_cxt->inner;
```

Its a bit annoying that its flagging test-only code. But sure enough. The assertion is not like in rust in the meaning that it panics and "returns" from the function. The function runs on and would hit a segfault here. In theory.

Swatinem (approved):

Swatinem:
IMO, we could still keep `sentry_transaction_context_t` as a `sentry_value_t`, as it is intentionally just a property bag without any functionality. We just set properties on it and then pass it into `start_transaction`.

include/sentry.h
```diff
@@ -1345,13 +1370,35 @@ SENTRY_EXPERIMENTAL_API sentry_uuid_t sentry_transaction_finish(
  * well as its reference by passing in the same Transaction as the one passed
  * into this function.
  */
-SENTRY_EXPERIMENTAL_API void sentry_set_span(sentry_value_t transaction);
+SENTRY_EXPERIMENTAL_API void sentry_set_span(sentry_transaction_t *tx);
```
Swatinem, line 1373: TBH, I don’t really know what to do with this one here. In Rust, I have this ugly as a mf type `TransactionOrSpan` for usage with `Scope::get/set_span` :-(

loewenheim (approved):

flub:
> IMO, we could still keep `sentry_transaction_context_t` as a `sentry_value_t`, as it is intentionally just a property bag without any functionality. We just set properties on it and then pass it into `start_transaction`.

I think I agree with this.  There is, as far as we are predicting, not going to be a need for this thing to become any smarter so may as well remain a `sentry_value_t`.

flub (approved):

flub:
> Moving forward, follow up tasks for this should look loosely like the following:
> 
> * explore using some union type to represent spans and transactions

I'm not sure there's much to be gained from this currently?

> * begin removing `sentry_value_t` fields in transaction/span structs and explicitly declare members on them instead

let's do this targeted as we need functionality that needs this.

> * introduce helpers that convert `sentry_transaction_context_t`/`sentry_transaction_t`/`sentry_span_t` back to a `sentry_value_t`

IIRC the functionality we wanted to enable was:

- starting a child span of a transaction bound to the scope needs to push span details to the scope (and do the scope locking, so binding a transaction to the scope is expensive).

- finishing a span does not need to pass the transaction

so limiting followup tasks to strictly achieve that is probably good.  I think it's something like this:

- put pointer to scope in each `sentry_span_t‘ (and `sentry_transaction_t`).  Use it to update the span if that pointer is not NULL.

- put pointer to transaction in each scope.  I'm like not 100% sure about this one, because i think it is possible to finish the transaction when there are still spans pointing to it, in that case they'd point to freed memory.  So maybe we shouldn't do this.

include/sentry.h
relaxolotl: scope(s) aren't exposed at all in the public api, hence why it's being omitted from the signature, yes. the existence of scopes just simply do not exist.

src/sentry_core.c
```diff
@@ -829,93 +839,93 @@ sentry_transaction_finish(sentry_value_t tx)
...
-    sentry_value_t spans = sentry_value_get_by_key(parent, "spans");
-    // This only checks that the number of _completed_ spans matches the number
-    // of max spans. This means that the number of in-flight spans can exceed
-    // the max number of spans.
-    if (sentry_value_get_length(spans) >= max_spans) {
-        SENTRY_DEBUG("reached maximum number of spans for transaction, not "
-                     "creating span");
-        goto fail;
+    size_t max_spans = SENTRY_SPANS_MAX;
+    SENTRY_WITH_OPTIONS (options) {
```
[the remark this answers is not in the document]
relaxolotl: i can add a TODO, although our options are likely to be fairly limited in addressing this.

src/sentry_tracing.c
```diff
@@ -1,11 +1,245 @@
...
+    tx->inner = inner;
+
+    return tx;
+}
+
+void
+sentry__transaction_incref(sentry_transaction_t *tx)
+{
+    sentry_value_incref(tx->inner);
+}
```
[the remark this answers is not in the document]
relaxolotl: will probably need some clarification on this; is the idea to remove the ability to create multiple references to a transaction? or is it that the implementation of "reference counting" on the transaction adds an unnecessary layer of abstraction over `sentry_value_t`'s reference counting?

if the intent is to prevent any reference counting at all, we would have to figure out how to let the scope obtain up to date information on a transaction without resorting to refcounting. one idea could be to have the scope contain some separately allocated copy of the transaction's data.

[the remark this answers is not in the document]
flub: my rough thinking was to avoid needing to reference count the transaction entirely.  Instead store a pointer in the transaction to the scope, which can be NULL if the transaction is not bound to the scope.  Then whenever the transaction updates something which needs to be represented on the scope check the pointer and go update the scope.

of course it might be that my rough thinking is just wrong :)

PR #656, opened 2022-01-14
feat(tracing): Spans now carry pointers to the Transactions they belong to

relaxolotl:
This changes the internal structure of spans so they contain a reference to their root Transaction. Finishing a span no longer requires the root Transaction to be explicitly passed in.

This eliminates an entire class of issues and associated validation related to spans being attached to the correct transaction. In return, users are expected to provide certain concurrency-related guarantees for spans and transactions when they are passed in as parameters.


src/sentry_tracing.c
```diff
@@ -161,9 +161,9 @@ sentry__transaction_decref(sentry_transaction_t *tx)
 }
 
 sentry_span_t *
-sentry__span_new(sentry_value_t inner)
+sentry__span_new(sentry_transaction_t *tx, sentry_value_t inner)
```
relaxolotl, line 164: this change to the signature caused a series of cascading changes which are all internal, thankfully.

there should be no reason why a span could or should be started without some root transaction associated with it, so this was added in as a parameter to the "constructor".

setting the transaction after the creation of the span also caused some pointer-related issues, so this also resolves a bug.

src/sentry_core.c
```diff
@@ -864,17 +864,21 @@ sentry_transaction_start_child(
         max_spans = options->max_spans;
     }
 
-    return sentry__start_child(max_spans, parent, operation, description);
-    // TODO: add pointer to transaction on child span
+    sentry_value_t span
+        = sentry__value_span_new(max_spans, parent, operation, description);
+    return sentry__span_new(opaque_parent, span);
```
relaxolotl, line 869: an unfortunate consequence of the change to `sentry__span_new` is that it does create some confusion for contributors(/developers). they'll need to understand the difference between `sentry__value_span_new` and `sentry__span_new`, and why and when one needs to be invoked over the other.

include/sentry.h
```diff
@@ -1408,45 +1447,65 @@ SENTRY_EXPERIMENTAL_API sentry_span_t *sentry_transaction_start_child(
...
- * `root_transaction` is either the parent Transaction of the Span, or
- * the ancestor Transaction of the Span if the Span is not a direct descendant
- * of a Transaction.
- *
  * This takes ownership of `span`, as child Spans must always occur within the
  * total duration of a parent Span and cannot outlive their parent Spans.
+ *
+ * The Span passed in must have an exclusive write lock to be thread-safe. This
+ * will mutate the Span's parent Transaction, and therefore also requires an
+ * exclusive write lock on that Transaction as well.
```
relaxolotl, line 1471: this is actually pretty difficult to communicate. i'm on the fence about this being a positive side-effect of this change; the end user just needs to know which transaction the span belongs to, and to ensure that there's a write lock on that transaction when `sentry_span_finish` is invoked. a lot of this is very implicit.

include/sentry.h
```diff
@@ -1276,13 +1276,22 @@ typedef struct sentry_span_s sentry_span_t;
  * Also see https://develop.sentry.dev/sdk/event-payloads/transaction/#anatomy
  * for an explanation of `operation`, in addition to other properties and
  * actions that can be performed on a Transaction.
+ *
+ * The returned value is not thread-safe. Users are expected to ensure that
+ * appropriate locking mechanisms are implemented over the Transaction Context
+ * if it needs to be mutated across threads. Methods operating on the
+ * Transaction Context will mention what kind of expectations they carry if they
+ * ned to mutate or access the object in a thread-safe way.
```
flub, line 1284:
```suggestion
 * need to mutate or access the object in a thread-safe way.
```

include/sentry.h
```diff
@@ -1313,28 +1328,39 @@ SENTRY_EXPERIMENTAL_API void sentry_transaction_context_set_sampled(
...
+ * it will be discarded, and will not be sent to sentry.
  *
  * To ensure that any Events or Message Events are associated with this
  * Transaction while it is active, invoke and pass in the Transaction returned
  * by this function to `sentry_set_span`. Further documentation on this can be
  * found in `sentry_set_span`'s docstring.
  *
- * Takes ownership of `transaction_context`.
+ * Takes ownership of `transaction_context`. The Transaction Context passed in
+ * must have an exclusive write lock to be thread-safe.
```
flub, line 1357:
```suggestion
 * Takes ownership of `transaction_context` and this should no longer be mutated by user code.
```

I'm not sure this second sentence still makes sense here.  "Taking ownership" means you need to never use the thing again.

include/sentry.h
```diff
@@ -1313,28 +1328,39 @@ SENTRY_EXPERIMENTAL_API void sentry_transaction_context_set_sampled(
...
+ *
+ * The returned value is not thread-safe. Users are expected to ensure that
+ * appropriate locking mechanisms are implemented over the Transaction if it
+ * needs to be mutated across threads. Methods operating on the Transaction will
+ * mention what kind of expectations they carry if they ned to mutate or access
```
flub, line 1362:
```suggestion
 * mention what kind of expectations they carry if they need to mutate or access
```

include/sentry.h
```diff
@@ -1345,8 +1371,10 @@ SENTRY_EXPERIMENTAL_API sentry_transaction_t *sentry_transaction_start(
  * otherwise.
  *
  * Always takes ownership of `transaction`, regardless of whether the operation
- * was successful or not. If `sentry_set_span` was invoked with `transaction`,
- * this will remove the
+ * was successful or not.
+ *
+ * The Transaction passed in must have an exclusive write lock to be
+ * thread-safe.
```
flub, line 1377: again i'm not really sure about this phrasing together with "takes ownership", to me this kind of conflicts.

include/sentry.h
```diff
@@ -1364,6 +1392,8 @@ SENTRY_EXPERIMENTAL_API sentry_uuid_t sentry_transaction_finish(
  * Invoke `sentry_transaction_finish` to remove the Span set by this function as
  * well as its reference by passing in the same Transaction as the one passed
  * into this function.
+ *
+ * The Transaction passed in must have a read lock to be thread-safe.
```
flub, line 1396:
```suggestion
 * This mutates the transaction passed in, which is not thread-safe by itself.  To
 * mutate the transaction from multiple threads a lock must must be used to protect
 * mutations.
```

The "read lock" part is somewhat odd too, sorry if this feels like nitpicking.  The suggestion here is pretty verbose, can probably be made shorter too.

include/sentry.h
```diff
@@ -1386,6 +1416,15 @@ SENTRY_EXPERIMENTAL_API void sentry_set_span(sentry_transaction_t *tx);
  * finishing the Span means it will be discarded, and will not be sent to
  * sentry. `sentry_value_null` will be returned if the child Span could not be
  * created.
+ *
+ * The Transaction passed in must have a read lock to be thread-safe. This
```
flub, line 1420: "read lock" again somewhat odd, since it suggest some kind of RwLock but that's not true since this does mutation.  i do think you mean "mutex" in these cases.

include/sentry.h
flub:
```
Takes ownership of the span, mutates the transaction the span was associated with,
which will require locking if that transaction is used from multiple threads.
```

I agree that it is now harder for users to know which transaction needs the locking.  We haven't made things worse though: with the old API they **had** to keep track of this, now they only have to keep track of this if they want to use a single transaction in multiple threads.

relaxolotl:
thanks greatly for the feedback; i've updated the phrasing so could you take a second look @flub?

flub (approved):

Swatinem (approved):
IMO, calling out the not-threadsafe nature of all the methods explicitly is a bit excessive, but ok ;-)

PR #663, opened 2022-01-20
ref: Make API forward-compatible to Sampling Context (NATIVE-457)

Swatinem:
This also renames `sentry_start_transaction` to use the naming that is defined here: https://develop.sentry.dev/sdk/performance/#static-api-changes

### DIFF
--- a/examples/example.c
+++ b/examples/example.c
@@ -227,7 +227,8 @@ main(int argc, char **argv)
         if (has_arg(argc, argv, "unsample-tx")) {
             sentry_transaction_context_set_sampled(tx_ctx, 0);
         }
-        sentry_transaction_t *tx = sentry_transaction_start(tx_ctx);
+        sentry_transaction_t *tx
+            = sentry_start_transaction(tx_ctx, sentry_value_new_null());
 
         if (has_arg(argc, argv, "error-status")) {
             sentry_transaction_set_status(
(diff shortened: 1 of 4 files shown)


loewenheim (approved):

relaxolotl:
FYI the original name was the way it was (`sentry_transaction_start`) because it aligned with the naming scheme of the rest of the performance monitoring API, which "namespaces" functions in the names themselves. The idea was to make it more compatible with autofill, and chances are users probably had muscle memory from using several `sentry_transaction_...` functions anyways. It was also concluded that users implementing performance monitoring in the native SDK were unlikely to have done so in several other SDKs, cutting down on the odds that prior experience with other SDKs would make users more likely to look for `sentry_start_transaction` instead of `sentry_transaction_start`.

This was also discussed with @rhcarvalho in Slack, and deviating slightly from the "unified" API to better fit the naming schemes of an SDK was accepted as a reasonable justification for the rename.

This change has already been approved and merged, and the rename has been bundled with a change that we had already agreed upon, so I'm not going to ask that the rename gets reverted. I understand why one would look at the name and find it strange that it is different from the documentation, but quite frankly I am a little frustrated to see this considering that there was a considerable amount of thought put behind the original, non-conforming name. I've only discovered this because I was nosy and decided to look at the diff, and I'm pretty sure that any protests on my end would be swept away and labeled as bikeshedding.

PR #671, opened 2022-01-26
test: Add integration test for envelope with transaction

flub:
This adds an integration test which inspects the envelope of a
transaction sent by the native SDK.

-------

### DIFF
--- a/tests/test_integration_http.py
+++ b/tests/test_integration_http.py
@@ -1,9 +1,11 @@
+import time
 import pytest
 import subprocess
 import sys
 import os
 import time
 import itertools
+import uuid
 import json
 from . import make_dsn, check_output, run, Envelope
 from .conditions import has_http, has_breakpad, has_files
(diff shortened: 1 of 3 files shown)


tests/test_integration_http.py
```diff
@@ -413,3 +415,60 @@ def delayed(req):
...
+    # assert trace_context['public_key']
+
+    assert trace_context["span_id"]
+    assert trace_context["status"] == "ok"
+
+    RFC3339_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
+    start_timestamp = time.strptime(json["start_timestamp"], RFC3339_FORMAT)
+    assert start_timestamp
+    timestamp = time.strptime(json["timestamp"], RFC3339_FORMAT)
+    assert timestamp >= start_timestamp
```
Swatinem, line 474: I think this should be fine, as the example runs all these on the main thread, however we did run into the issue previously that our time source is not guaranteed to be fully monotonic in multithreaded scenarios :-(

Swatinem (approved):
❤️ This looks so much better now that someone familiar with pytest does these things ;-)

tests/test_integration_http.py
flub: hum, yes i did worry about this.  if it turns out to be flaky i'll not hesitate the assert though

flub:
> ❤️ This looks so much better now that someone familiar with pytest does these things ;-)

I did try and keep changes to some minimum to just make me debug things.  it could be a lot better but gets the job done.

tests/test_integration_http.py
flub: huh "to remove the assert"

PR #674, opened 2022-01-31
ref: Optimize JSON formatting

Swatinem:
- Inline some `stringbuilder` functions (and remove an unused one)


flub (commented):
I'm a bit confused with the `sentry_string.c` -> `sentry_string.h` changes.  was this intentional?

Swatinem:
Yes, that essentially inlines all those stringbuilder methods (esp the builds check, avoiding the method call to reserve). The `sentry__string_clone(n)` was maybe overkill since I doubt that its particularly hot, but doesn’t hurt either.

flub (approved):

PR #678, opened 2022-02-11
feat: Include performance monitoring by default in experimental API, remove compile flag for feature

relaxolotl:
Does what it says in the title. Removes all instances of `SENTRY_PERFORMANCE_MONITORING` so you don't need to include an additional compile time flag to get access to performance monitoring on the SDK if you've already opted into experimental features.

This touches several test files because they import `sentry.h` and `sentry_testsupport.h`, which is causing a double import of perfmon enums. I believe @Swatinem mentioned in a previous PR that the fix for something like this was to remove the `sentry.h` import, hence the diffs on those files. `sentry.h` has been removed as an import from other headers as well (eg `sentry_scope.h` if `sentry_tracing.h` is already imported due to the same issue.

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -568,7 +568,6 @@ SENTRY_API sentry_value_t sentry_envelope_get_event(
  */
 SENTRY_EXPERIMENTAL_API sentry_value_t sentry_envelope_get_transaction(
     const sentry_envelope_t *envelope);
-#endif
 
 /**
  * Serializes the envelope.
@@ -1708,4 +1707,5 @@ SENTRY_EXPERIMENTAL_API void sentry_transaction_iter_headers(
 
 #ifdef __cplusplus
 }
+#endif
 #endif
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -561,7 +561,6 @@ SENTRY_API void sentry_envelope_free(sentry_envelope_t *envelope);
 SENTRY_API sentry_value_t sentry_envelope_get_event(
     const sentry_envelope_t *envelope);
 
-#ifdef SENTRY_PERFORMANCE_MONITORING
 /**
  * Given an Envelope, returns the embedded Transaction if there is one.
  *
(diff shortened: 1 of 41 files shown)


flub (approved):
I think sentry.h is imported by sentry_boot.h so you don't need to import both those.  But if removing all these imports is fine for the tests i don't see why not

include/sentry.h
```diff
@@ -561,7 +561,6 @@ SENTRY_API void sentry_envelope_free(sentry_envelope_t *envelope);
 SENTRY_API sentry_value_t sentry_envelope_get_event(
     const sentry_envelope_t *envelope);
 
-#ifdef SENTRY_PERFORMANCE_MONITORING
```
Swatinem, line 564: I think this one is now unbalanced?

include/sentry.h
```diff
@@ -1708,9 +1706,6 @@ SENTRY_EXPERIMENTAL_API void sentry_transaction_iter_headers(
     sentry_transaction_t *tx, sentry_iter_headers_function_t callback,
     void *userdata);
 
-#endif
-
 #ifdef __cplusplus
 }
 #endif
-#endif
```
Swatinem, line 1716: I’m also confused by this here?

tests/unit/test_uuid.c
```diff
@@ -1,5 +1,5 @@
 #include "sentry_testsupport.h"
-#include <sentry.h>
+
 #include <sentry_uuid.h>
```
Swatinem, line 3:
```suggestion
#include "sentry_uuid.h"
```

tests/unit/fuzz.c
```diff
@@ -24,8 +24,6 @@ afl-fuzz -i fuzzing-examples -o fuzzing-results -- fuzzing/sentry_fuzz_json @@
 #    define _CRT_SECURE_NO_WARNINGS
```
Swatinem, line 24: I’m wondering if we can remove this whole block, since that is essentially what you need so a direct include of `sentry.h` does not break the build.

Swatinem (commented):
Can you double-check the ifdefs in `sentry.h`, I think they might be unbalanced now.
Other than that, nice work cleaning up the includes! 👍🏻 
As a reminder: The subtle difference between `#include <foo.h>` which is used for system libraries, and `#include "foo.h"` which is used for headers that are part of the project.

tests/unit/fuzz.c
relaxolotl: I'll give this a spin to see what sort of effect this has on our checks. If we fail on compiler warnings then this might not be an option.

include/sentry.h
relaxolotl: thanks for catching this! looks like i didn't remove this one's partner `#endif`, fixed this.

tests/unit/fuzz.c
relaxolotl: So removing this doesn't really break anything as far as the CI is concerned, but after looking into the other directives in this block it looks like some of them help with compilation times, which seems like a pretty nice thing to have. I don't know if these fuzzer tests are being run on Windows not to mention the impact removing these would have on Windows builds, so I'm going to leave this in for the time being just to be safe.

Swatinem (approved):

PR #690, opened 2022-03-29
ref: Switch the internal representation of project IDs from ints over to char*s

relaxolotl:
Much like https://github.com/getsentry/sentry-rust/pull/452, this does what it says in the title. This tries to future-proof project IDs as they aren't guaranteed to be numbers forever. This changes the internal representation, but it still keeps all of the validation prior to the switch until most of sentry treats these values as proper strings.


src/sentry_utils.c
```diff
@@ -259,12 +260,16 @@ sentry__dsn_new(const char *raw_dsn)
     if (!tmp) {
         goto exit;
     }
-
-    dsn->project_id = (uint64_t)strtoll(tmp + 1, &end, 10);
+    // Validate that the project ID is still a valid number until sentry fully
+    // commits to pure string project IDs
+    project_id = strtoll(tmp + 1, &end, 10);
```
Swatinem, line 265: Part of the reason for the change is that we shouldn’t validate these IDs.

src/sentry_utils.c
```diff
@@ -259,12 +260,16 @@ sentry__dsn_new(const char *raw_dsn)
...
     if (end != tmp + strlen(tmp)) {
         goto exit;
     }
+
+    dsn->project_id = sentry__string_clone(tmp + 1);
```
Swatinem, line 270: According to leak checkers, this allocation is never being freed.

relaxolotl:
I've rebased the branch off of #692 which is stacked on top of #691 so this can properly run against the CI again. Changes have also been made to reflect the original expectations that served as the underlying motivation for the change; All feedback should be addressed by the new changes.

This PR should be ready for a second look.

Swatinem (approved):

PR #691, opened 2022-03-30
ref: Update black so it stops breaking our CI

relaxolotl:
black is currently causing our CI to fail. A naive bump up to a version that fixes this shouldn't be high risk for us since: a. it's a code formatter and b. we use python for our integration tests.

The tests don't completely pass but I don't believe it's a hard requirement to at least get this change in. Some of the tests are failing for completely unrelated reasons (deprecated windows environments on GitHub) and can/should be fixed in a separate PR just to keep things compartmentalized.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,8 +75,8 @@ jobs:
             ERROR_ON_WARNINGS: 1
             SYSTEM_VERSION_COMPAT: 0
             RUN_ANALYZER: asan,llvm-cov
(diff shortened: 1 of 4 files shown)


relaxolotl:
pulling out of draft to see if the windows checks are blockers

relaxolotl:
heck yeah it isn't, that can be tackled in a different PR

Swatinem (approved):

flub (approved):
no idea why the windows check was cancelled... CI always has its own mind

PR #692, opened 2022-03-30
ci: Bump CI to use win2019

relaxolotl:
The Windows 2016 environment was [deprecated a few weeks ago](https://github.com/actions/virtual-environments/issues/4312), causing our CI to fail. This bumps the check that used to use that environment up to using Windows Server 2019. [`windows-latest` now points to Windows Server 2022](https://github.com/actions/virtual-environments/issues/4856), so this doesn't reduce any redundant CI jobs.

The README has also been updated to reflect the fact that the project will no longer be actively testing against Windows 2016, and 2022 has been added to its supported platforms list instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ The SDK currently supports and is tested on the following OS/Compiler variations
 - 64bit Linux with clang 9
 - 32bit Linux with GCC 7 (cross compiled from 64bit host)
 - 64bit Windows with MSVC 2019
-- 32bit Windows with MSVC 2017
+- 64bit Windows with MSVC 2022
 - macOS Catalina with most recent Compiler toolchain
 - Android API29 built by NDK21 toolchain
 - Android API16 built by NDK19 toolchain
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -59,7 +59,7 @@ The SDK currently supports and is tested on the following OS/Compiler variations
 - 64bit Linux with clang 9
 - 32bit Linux with GCC 7 (cross compiled from 64bit host)
 - 64bit Windows with MSVC 2019
-- 32bit Windows with MSVC 2017
+- 64bit Windows with MSVC 2022
```
Swatinem, line 62: We are running the 32bit version on Windows 2019 though.
Also, is `windows-latest` coming with VS2022? I wasn’t aware, but thats nice.

Swatinem (approved):

flub (approved):

relaxolotl:
I changed the head on #691 to point to the same head as this PR and pushed, and it turns out this registers as a merge on GitHub. It looks like without the changes in this PR, #691's CI won't pass meaning that the goal of splitting these two fixes into two disparate commits was not possible.

I'll address the feedback in #691 and add breadcrumbs reflecting that fact, just to wrap up loose ends.

PR #698, opened 2022-04-13
feat: More aggressively prune the Crashpad database

Swatinem:
It seems that the defaults in the crashpad_handler only trigger after a long 10min delay, and let the db grow to up to 128M by default. We generally do not care about older/completed reports at all, so make sure to aggressively prune them.

### DIFF
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -31,6 +31,7 @@ extern "C" {
 #include "client/crash_report_database.h"
 #include "client/crashpad_client.h"
 #include "client/crashpad_info.h"
+#include "client/prune_crash_reports.h"
 #include "client/settings.h"
 
 #if defined(__GNUC__)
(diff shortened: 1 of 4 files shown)


src/backends/sentry_backend_crashpad.cpp
```diff
@@ -451,6 +468,7 @@ sentry__backend_new(void)
     backend->user_consent_changed_func
         = sentry__crashpad_backend_user_consent_changed;
     backend->get_last_crash_func = sentry__crashpad_backend_last_crash;
+    backend->prune_database_func = sentry__crashpad_backend_prune_database;
```
flub, line 471: are all other backend guaranteed to initialise this to NULL?

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -426,6 +427,22 @@ sentry__crashpad_backend_last_crash(sentry_backend_t *backend)
...
+static void
+sentry__crashpad_backend_prune_database(sentry_backend_t *backend)
+{
+    crashpad_state_t *data = (crashpad_state_t *)backend->data;
+
+    // We want to eagerly clean up reports older than 2 days, and limit the
+    // complete database to a maximum of 8M. That might still be a lot for
+    // an embedded use-case, but minidumps on desktop can sometimes be quite
+    // large.
+    data->db->CleanDatabase(60 * 60 * 24 * 2);
```
flub, line 439: 2 days still seems a long time to me, why not immediate?  Or are we at risk of removing unsent minidumps?

flub (commented):
do these conditions result in pruning the db even if the minidumps are unsent?

src/backends/sentry_backend_crashpad.cpp
Swatinem: yes, we are `memset`-ing the backend struct to 0

src/backends/sentry_backend_crashpad.cpp
Swatinem: crashpad has 3 states essentially: `new`, `pending` and `completed`. The function cleans pending and completed minidumps. I would have to look into what it does exactly on upload (create as `new`, move into `pending` when it starts uploading?)
Keeping the "most recent" minidump might be fair to do. At least the database should not grow seemingly unbounded.

flub: in theory crashpad tries to upload as soon as the crash happens right?  so this will only find new or pending minidumps if that failed somehow.  I guess 2 days is reasonable, it's just really though to come up with something that suits all :)

flub (approved):

src/backends/sentry_backend_crashpad.cpp
ericriff: I'd strongly prefer to wipe the minidumps as soon as they get uploaded. 
What about making the parameters configurable though API calls?

PR #714, opened 2022-05-17
feat: Capture registers with inproc backend (FEEDBACK-1413)

Swatinem:


src/backends/sentry_backend_inproc.c
```diff
@@ -114,7 +114,8 @@ shutdown_inproc_backend(sentry_backend_t *UNUSED(backend))
     reset_signal_handlers();
 }
 
-#elif defined SENTRY_PLATFORM_WINDOWS
+#elif defined(SENTRY_PLATFORM_WINDOWS)
```
loewenheim, line 117: Is this a functional change or just formatting?

src/backends/sentry_backend_inproc.c
```diff
@@ -168,8 +169,77 @@ shutdown_inproc_backend(sentry_backend_t *UNUSED(backend))
...
+sentry__registers_from_uctx(const sentry_ucontext_t *uctx)
+{
+    sentry_value_t registers = sentry_value_new_object();
+
+#ifdef SENTRY_PLATFORM_UNIX
+
+#elif defined(SENTRY_PLATFORM_WINDOWS)
+    PCONTEXT ctx = uctx->exception_ptrs.ContextRecord;
+
+#    define SET_REG(name, prop)                                                \
```
loewenheim, line 185: Is this the same thing as `#define`?

src/backends/sentry_backend_inproc.c
Swatinem: just formatting

src/backends/sentry_backend_inproc.c
Swatinem: Not sure what you mean? Is it the formatting, or rather what this construct does?

loewenheim: Yeah, the formatting was what I meant. I haven't seen that before.

Swatinem:
I tested this only on x64 on all OSs. I would like to have someone give it a spin on 32/64-bit Android as well.

supervacuus:
I can test Android only on the emulator and give aarch64 on Linux a try. But both probably not before Wednesday.

supervacuus:
I tested this on aarch64 under Linux where it works correctly. I also tried to test this in android using the emulator. I can see that an aarch64 register dump is created in the sentry-native database path on the device, but I cannot really stress the register values for comparison because the crash-causing function is compiled using an `armeabi-v7a` toolchain. My insight into the android build setup is still lacking, but I will investigate further.

supervacuus:
android `aarch64` seems to work too. I cannot test `armeabi-v7a`, because the images provided for the emulator are too old. I do not own a device.

registers from the crashed JNI function

Swatinem:
@supervacuus Thanks for the investigation! I think this is ready to land in that case.

loewenheim (approved):

PR #718, opened 2022-05-24
Setup CodeQL

mdtro:
Sets up CodeQL to run on pull requests targeting `master`, merges to `master`, and weekly.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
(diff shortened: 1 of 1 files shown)


Swatinem (commented):
a couple of things I noticed:

* we need recursive submodules on checkout, as we need to pull in breakpad/crashpad
* please take a look at our normal CI job to install the required system dependencies (curl, etc)

Otherwise it does look like this autobuild tool tried to invoke our cmake build system, which is great, I wonder how far it can get on its own.

Either way, we do have a couple of compile-time options, not sure which permutations of those we need to build/test. At least we should use breakpad/crashpad/inproc.

Swatinem:
Another NOTE: we don’t depend really depend on "external dependencies", so it might not be necessary to run this on a weekly cadance.

mdtro:
> a couple of things I noticed:
> 
> * we need recursive submodules on checkout, as we need to pull in breakpad/crashpad
> * please take a look at our normal CI job to install the required system dependencies (curl, etc)
> 
> Otherwise it does look like this autobuild tool tried to invoke our cmake build system, which is great, I wonder how far it can get on its own.
> 
> Either way, we do have a couple of compile-time options, not sure which permutations of those we need to build/test. At least we should use breakpad/crashpad/inproc.

Thanks for the feedback! I'll go through the build docs and adjust the build step as necessary. 😄 

> Another NOTE: we don’t depend really depend on "external dependencies", so it might not be necessary to run this on a weekly cadance.

For here, the weekly cadence is more so used as a way to scan with updated-detections/rules on risky code patterns that may have been published after our last merge.

markushi:
Closing in favor of https://github.com/getsentry/sentry-native/pull/1058

Issue #716, opened 2022-05-18
Investigate Native SDK Backend/Session behaviour on Shutdown on M1


ashwoods:
We would want to verify how the Native SDK behaves in these instances:

Various ways of process termination:

- KILL / TERM
- Logout / Shutdown

Using the following backends:

- crashpad
- breakpad

All that, specifically on macOS on the M1.

STR:

Possibly patch up the native example, increasing this sleep here: https://github.com/getsentry/sentry-native/blob/bd56100922e1359c08f512e84ab975972788dc98/examples/example.c#L177

Then build the example with the various backends, and start it like so:
`sentry_example log start-session sleep`.
… and then kill it in various ways.
Observe if a crash report and which kind of session status is being reported.

supervacuus:
I can do the testing and create further issues if I see any unwanted behavior.

Swatinem:
The specific case we would like to confirm in our testing, which came from a customer:
They report that with **crashpad**, when the process is forcibly killed on shutdown due to taking too long to terminate, they observe a crash log, which in turn flags the session as `crashed` due to filesystem timestamps.

Not quite sure this will necessarily exhibit that case, maybe you would have to install a handler for SIGTERM or whatever is signalling the process to pls shut down; and then run into a timeout there.

The desired outcome here would be: don’t create a `crashed` session if the process is being forcibly killed on shutdown, but rather a `abnormal` session.

supervacuus:
@ashwoods, @Swatinem, since this was coming from support: was this ever raised again? I would like to close this otherwise.

ashwoods:
Hasn't resurfaced since, closing now.

PR #719, opened 2022-05-28
fix: retrieve only completed crash-reports from the crashpad database...

supervacuus:
when recovering the timestamp of the last potential crash on initialization.

This is one attempt discussed with @Swatinem to fix https://github.com/getsentry/sentry-native/issues/716. I create it as a draft though, because I still cannot reproduce the issue, but it still could be useful for testing if the customer has a stable repro.


supervacuus:
Should we leave this for the time being or are you approving to merge this right away?

Swatinem:
We can as well merge it right away ;-)

Issue #721, opened 2022-05-30
Provide high-level interface to add a stack trace to an exception or thread.


supervacuus:
As discussed here: https://github.com/getsentry/sentry-docs/pull/5085 (specifically: https://github.com/getsentry/sentry-docs/pull/5085#discussion_r884258341)

Swatinem:
This was done in #723

PR #723, opened 2022-06-02
feat: API to add a stack trace to an exception or thread

supervacuus:
Proposal for https://github.com/getsentry/sentry-native/issues/721


Swatinem (approved):

include/sentry.h
```diff
@@ -418,6 +418,17 @@ SENTRY_EXPERIMENTAL_API sentry_value_t sentry_value_new_thread(
 SENTRY_EXPERIMENTAL_API sentry_value_t sentry_value_new_stacktrace(
     void **ips, size_t len);
 
+/**
+ * Adds a Stack Trace conforming to the Stack Trace Interface to a value.
+ *
+ * The value needs to be either an exception event, or a thread object.
```
jan-auer, line 424: An exception _event_ object, or an exception object (not event)? The implementation assumes the latter, which also makes more sense.

supervacuus: You are right, it should be 

```The value needs to be either an exception or a thread object.```

supervacuus:
Hi @jan-auer and @Swatinem, here are a few points to consider before I would make this a non-draft PR:

`sentry-native` with this PR has 3 functions in the public interface concerned with creating and attaching stack traces to values (i.e. excluding the actual unwinding- and the option/configuration interfaces): 
- `sentry_event_value_add_stacktrace` (deprecated here: https://github.com/getsentry/sentry-native/pull/517), 
- `sentry_value_new_stacktrace` (added also here  https://github.com/getsentry/sentry-native/pull/517) and
- `sentry_value_add_stacktrace` (from this PR).

Given the similarity of the latter two, I wondered whether there is a customer use case where the call-site will create a `sentry_value_t` from a stack trace and then doesn't immediately attach it to an `exception` or `thread` object? Would you prefer for this API to tend towards completeness (giving customers more freedom to handle values) or to reduce the chance of creating incoherent (for the backend) value structures?

Is the behavior inherited by `sentry_value_set_key` in case a `"stacktrace"` key is already present in the `value` argument acceptable, or should something more involved happen?

Depending a bit on the last question should the function be called `sentry_value_attach_stacktrace` or `sentry_value_set_stacktrace` (or some other name) instead of `sentry_value_add_stacktrace`?

Swatinem:
I don’t really have a strong opinion on this. The API generally gives more freedom (and with great power comes great responsibility)

As for naming, maybe `sentry_value_set_stacktrace` is a better name indeed. If you want to make things a bit more clearer, you could as well make two (internally identical) functions `sentry_exception_set_stacktrace` and `sentry_thread_set_stacktrace`.

supervacuus:
At this point, I am mostly trying to synchronize with the API idea for `sentry-native` that you folks have. Thanks for the feedback.

include/sentry.h
supervacuus: How do you typically proceed with open conversations? I usually wait until the reviewer thinks the comment has been addressed in a follow-up commit, signalling by resolving.

Issue #592, opened 2021-09-11
Pass EXCEPTION_POINTERS to before_send_func for crashpad backend (on_crash candidate)


espkk:
Hi,

I'm wondering why `ExceptionInfo` is not passed to `before_send_func` as a second (`hint`) argument here: https://github.com/getsentry/sentry-native/blob/c3b7b07c99d19f39a4910093378e29bba86fde37/src/backends/sentry_backend_crashpad.cpp#L143
This could be useful for example to choose a cleanup strategy depending on the exception code when `before_send_func` is called.
Is this by design or can be changed?

Swatinem:
As this callback is being invoked for *every* event (not only crashes), that wouldn’t easily be possible.

I think the best solution here would be to implement a dedicated `on_crash` hook, which does get the exception info, and signal info on unix platforms.

espkk:
> 
> 
> As this callback is being invoked for _every_ event (not only crashes), that wouldn’t easily be possible.
As far as I can see `sentry__prepare_event` and `sentry__crashpad_handler` always set `hint` to `NULL`. 
Since crashpad guarantees `ExceptionInfo` to contain exception pointers, testing the `hint` pointer could be used to differentiate the context.
So I'm still thinking of it as the "cheapest" solution as it's just 1 line change.
Also, that parameter can be used for holding some structure representing Unix signal data following the same considerations. The advantage of this approach is also that the context could be deduced with `nullptr` check using the same callback signature regardless of the platform.

Though, I agree that `on_crash` would be probably semantically cleaner.

Anyway, thank you for your answer! I will stick to my patched solution for now and wait for updates.

ashwoods:
We are currently investigating implementing a "before_crash"/"on_crash" hook.

supervacuus:
This was completed with https://github.com/getsentry/sentry-native/releases/tag/0.5.0.

The author of this issue (@espkk) implemented a considerable part of this feature.

PR #724, opened 2022-06-11
feat: invoke on_crash when handling a crash

espkk:
for #592

not sure if it's possible to add some UT for that


Swatinem (commented):
This looks very good!

Supporting the uctx on breakpad linux / macos would be nice.

It would also be interesting to highlight the difference between this and `before_send`.
Currently, `before_send` in some cases gets an empty event that is being discarded, just to satisfy its function signature, and to give users the ability to return a `null` event, in which case no crash report is being sent.
We should do something similar, maybe add a boolean return value to the hook, to indicate if we would capture the crash, or ignore it.

Another interesting aspect would be to maybe phase out `before_send` as currently used in the crash handler, meaning:
1. If `on_crash` is defined, `before_send` is *only* invoked for normal events, but *not* for crashes.
2. If `on_crash` is not defined, we need to maintain the current `before_send` behavior for backwards compatibility.

This would also give users the possibility to opt into the new behavior just by providing an empty noop `on_crash` handler.

src/backends/sentry_backend_breakpad.cpp
```diff
@@ -95,6 +95,16 @@ sentry__breakpad_backend_callback(
     SENTRY_WITH_OPTIONS (options) {
         sentry__write_crash_marker(options);
 
+#ifdef SENTRY_PLATFORM_WINDOWS
```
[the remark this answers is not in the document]
espkk: it's just a closure data. uctx is not passed, but I think it still can be obtained using `getcontext()`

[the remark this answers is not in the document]
espkk: Yeah I got it, but this `context` parameter is used as a closure since you pass it to `ExceptionHandler` constructor. So it has nothing to do with ucontext

I had an impression that breakpad's `getcontext` retrieves the original context, but it seems that it's just a fallback for envs that don't have `getcontext...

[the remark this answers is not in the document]
Swatinem: Oh sorry, that was my misunderstanding of "closure data/context" then. Too bad :-(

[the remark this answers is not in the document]
Swatinem: We should still call the `on_crash` hook for all platforms. We should probably pass `NULL` is the `uctx` parameter for the platforms where we have access to the crash context, and clearly document this.

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -137,27 +136,46 @@ sentry__crashpad_handler(int UNUSED(signum), siginfo_t *UNUSED(info),
...
+
+                // capture the envelope with the disk transport
+                sentry_transport_t *disk_transport
+                    = sentry_new_disk_transport(options->run);
+                sentry__capture_envelope(disk_transport, envelope);
+                sentry__transport_dump_queue(disk_transport, options->run);
+                sentry_transport_free(disk_transport);
+            }
+        } else {
+            SENTRY_TRACE("event was discarded by the `on_crash` hook");
```
Swatinem, line 178: I think in this case you need to `return true` so crashpad considers the signal "handled" and does not create a crash report for it.

But as you noted, we should still call `dump_queue` in that case as well, we still want to send (on restart) any events that were queued before.

src/backends/sentry_backend_breakpad.cpp
```diff
@@ -142,7 +164,9 @@ sentry__breakpad_backend_callback(
 #ifndef SENTRY_PLATFORM_WINDOWS
     sentry__leave_signal_handler();
 #endif
-    return succeeded;
+
+    // further handling can be skipped via on_crash hook
+    return already_handled || succeeded;
```
espkk, line 169: not sure about this one. probably some tests are needed

Swatinem: If I remember correctly, the breakpad docs are a bit misleading here. You should just return the `succeeded` flag unchanged. But I might be wrong, like I said, the docs are a bit unclear.

Swatinem (approved):
One thing to note here, and I think this is very much in line with the recent changes to sessions, more specifically, we shouldn’t touch them when the event is being discarded by hooks. But that also means that sessions would be marked as "abnormal" instead of "crashed", but I think this is expected. CC @adinauer

src/backends/sentry_backend_breakpad.cpp
Swatinem: Well "we" are not capturing the crash report, which is all that matters from our perspective. I would just pass on the `succeeded` parameter unchanged.

Swatinem (approved):

include/sentry.h
```diff
@@ -776,6 +777,32 @@ typedef sentry_value_t (*sentry_event_function_t)(
...
     sentry_options_t *opts, sentry_event_function_t func, void *data);
 
+/**
+ * Type of the `on_crash` callback.
+ *
+ * Does not work with crashpad on macOS.
+ * The callback passes a pointer to sentry_ucontext_s structure when exception
+ * handler is invoked. For breakpad on Linux this pointer is NULL.
+ *
+ * If the callback returns true outgoing crash report will be discarded.
```
Swatinem, line 787: This might just be personal preference, but I think its also more consistent with similar callbacks:
Can you turn this logic around? So that `false` discards the event, and `true` will continue processing as normal?
This is also in line with how `before_send` behaves, as you return a `null` value to discard the event.

Swatinem (commented):
Sorry for this last minute change request ;-)

Also we fixed our CI, so please merge/rebase master.

include/sentry.h
espkk: I already inverted it to be consistent with crashpad and breakpad handler returns
But it makes sense, especially since it's not used for breakpad return value anymore
Though, maybe it also shouldn't be used for crashpad as well there?
https://github.com/getsentry/sentry-native/blob/e8011b66af01851175bb77a31773cbcb6f96f76e/src/backends/sentry_backend_crashpad.cpp#L188-L190

Swatinem: I admit this is a bit confusing. Ideally the SDK users should not care about the underlying handler, and we can return an appropriate value to make the handler behave like it should.

espkk: The problem is I'm not sure what is the desired behaviour.
The docs says 
`If the custom handler returns true, the signal is considered handled and the signal handler returns. Otherwise, the currently installed Crashpad signal handler is run.`
But the actual behaviour should depend on some formal requirements
I reverted the return to `false` for crashpad too. Let me know if you think it should depend on `on_crash` return

Swatinem: > I reverted the return to false for crashpad too. Let me know if you think it should depend on on_crash return

Yes I think it should. Essentially we need to return `!should_handle`, as returning `true` from the crashpad hook tells it to not handle (aka send to sentry).

espkk: alright, should be fine now

PR #729, opened 2022-06-27
ref(session): align processing sequence in sentry__capture_event() with docs

### LINKED ISSUE #707: Align SDK with docs regarding session update for dropped events
We've updated the docs (https://github.com/getsentry/develop/pull/551) to explain how we would like to handle event filter order, updating of sessions for dropped events and sending of session updates. We should check if we're aligned and adapt if necessary. Here's the issue from the Java SDK where all of this started: https://github.com/getsentry/sentry-java/issues/1916

supervacuus:
Fixes #707.

The specified items in the [filter-order](https://develop.sentry.dev/sdk/sessions/#filter-order) apply to sentry-native accordingly:

- Not implemented: `1. Check for ignored exception types (a.k.a ignore_errors)`
- Not implemented: `2. Apply scoped event_processor (a.k.a error_processor)`
- Is equivalent to `sentry__prepare_event()`: `3. Apply global event_processor`
- Happens within `sentry__prepare_event()`: `4. Apply before_send`
- Happens in `sentry__capture_event()`: `5. Update the session if an event made it this far`
- Happens at the beginning of `sentry__prepare_event()`: `6. Apply sampling rate`

This means, in the current implementation, the sole conflict with the proposed filter order is in the application of the sampling at the beginning of the event processing and thus before `before_send` and the session update. The remaining implemented filters are in the order suggested in the docs.

This PR moves the sampling application right before the transport invocation. Since this is - essentially - a breaking change (e.g. `before_send` is no longer affected by sampling and thus called on every event capture) I think we should reflect this in the API docs too. Please advise on where a good place for this might be (beyond the changelog).


Swatinem (approved):

Swatinem:
As you suggested, we can make this behavior more clear in the fn-level docs, and even link to the docs that specify this behavior. We should also call this change out in the changelog.

supervacuus:
@Swatinem, @ashwoods: builds work again, so we only need to change the branch protection rules to merge.

PR #734, opened 2022-07-16
ref(on_crash): follow-up to make on_crash releasable

supervacuus:
* mentioned in the header docs `before_send` vs `on_crash` behavior
* adapted `sentry__prepare_event` to provide a flag for backends on
  whether `before_send` should be invoked
* adapted backends to prevent running `before_send` even if `on_crash`
  returns true
* added `on_crash` vs `before_send` integration tests for `inproc` and
  `breakpad` (`crashpad` will follow)
* removed `stdbool.h` from `sentry.h` and replaced return value `bool`
  with `int` in `sentry_crash_function_t` (since `bool` seemed to be
  explicitly prevented in the header up to now)
* added remaining CHANGELOG entries for upcoming release

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
 
       - name: Expose llvm PATH for Mac
         if: ${{ runner.os == 'macOS' }}
-        run: echo $(brew --prefix llvm@13)/bin >> $GITHUB_PATH
+        run: echo $(brew --prefix llvm@14)/bin >> $GITHUB_PATH
 
       - name: Installing Android SDK Dependencies
         if: ${{ env['ANDROID_API'] }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,12 @@
 
 **Features**
 
-- Provide `on_crash()` callback to allow clients to act on detected crashes
+- Provide `on_crash()` callback to allow clients to act on detected crashes.
+  Users often inquired about distinguishing between crashes and "normal" events in the `before_send()` hook.
+  `on_crash()` can be considered a replacement for `before_send()` for crash events, where the goal is to use 
+  `before_send()` only for normal events, while `on_crash()` is only invoked for crashes. This change is backward
+  compatible for current users of `before_send()` and allows gradual migration to `on_crash()`
+  ([see the docs for details](https://docs.sentry.io/platforms/native/configuration/filtering/)). 
   ([#724](https://github.com/getsentry/sentry-native/pull/724), 
    [#734](https://github.com/getsentry/sentry-native/pull/734))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,24 @@
 
 ## Unreleased
 
+**Features**
+
+- Provide `on_crash()` callback to allow clients to act on detected crashes
+  ([#724](https://github.com/getsentry/sentry-native/pull/724))
+
 **Fixes**
 
+- Make Windows ModuleFinder more resilient to missing Debug Info 
+  ([#732](https://github.com/getsentry/sentry-native/pull/732))
 - Aligned pre-send event processing in `sentry_capture_event()` with the 
   [cross-SDK session filter order](https://develop.sentry.dev/sdk/sessions/#filter-order) 
   ([#729](https://github.com/getsentry/sentry-native/pull/729))
+
+**Thank you**:
+
+Features, fixes and improvements in this release have been contributed by:
+
+- [@espkk](https://github.com/espkk)
 
 ## 0.4.18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 **Features**
 
 - Provide `on_crash()` callback to allow clients to act on detected crashes
-  ([#724](https://github.com/getsentry/sentry-native/pull/724))
+  ([#724](https://github.com/getsentry/sentry-native/pull/724), 
+   [#734](https://github.com/getsentry/sentry-native/pull/734))
 
 **Fixes**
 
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -2,12 +2,25 @@
 
 ## Unreleased
```
supervacuus, line 3: 0.5.0 or 0.4.19? Moving `before_send()` to the end of the pre-processing (in #729) could be considered a breaking change.

.github/workflows/ci.yml
```diff
@@ -136,7 +136,7 @@ jobs:
 
       - name: Expose llvm PATH for Mac
         if: ${{ runner.os == 'macOS' }}
-        run: echo $(brew --prefix llvm@13)/bin >> $GITHUB_PATH
+        run: echo $(brew --prefix llvm@14)/bin >> $GITHUB_PATH
```
supervacuus, line 139: we might need to rename the macOS + clang job again.

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -175,18 +177,52 @@ sentry__crashpad_handler(int signum, siginfo_t *info, ucontext_t *user_context)
...
+    // * we rectified the situation, so crashpads signal-handler can simply
+    //   return, thereby letting the not-rectified signal-cause trigger a
+    //   signal-handler/exception-filter again, which probably leads to us
+    //   (currently returning `true`)
+    //
+    // TODO(supervacuus):
+    // * we need integration tests for more signal/exception types not only
+    //   for unmapped memory access (which is the current crash in example.c).
+    // * we should adapt the SetFirstChanceExceptionHandler interface in
+    // crashpad
```
supervacuus, line 214: Another option is, as mentioned above, to change the state of the global signal configuration (aka reset to default handlers... which will in turn terminate the app). This looks cleaner on the surface, but I am not sure how happy crashpad is if we pull the rug in the first-chance handler. This is different from resetting when shutting down.

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -149,17 +149,19 @@ sentry__crashpad_handler(int signum, siginfo_t *info, ucontext_t *user_context)
...
             SENTRY_TRACE("invoking `on_crash` hook");
-            should_handle
-                = options->on_crash_func(&uctx, options->on_crash_data);
+            should_dump = options->on_crash_func(&uctx, options->on_crash_data);
         } else if (options->before_send_func) {
             sentry_value_t event = sentry_value_new_event();
             SENTRY_TRACE("invoking `before_send` hook");
             event = options->before_send_func(
                 event, nullptr, options->before_send_data);
+            should_dump = !sentry_value_is_null(event);
```
supervacuus, line 158: This was missing up to now. For the crashpad backend, it was irrelevant whether you returned the event or a null value.

The other difference (which I have not changed) to the other backends: we create the event right before the hook and then discard it again. The other backends allow changes made to the event received in `before_send()` to land in the backend.

src/backends/sentry_backend_breakpad.cpp
```diff
@@ -113,8 +113,8 @@ sentry__breakpad_backend_callback(
 
         if (should_handle) {
             sentry_value_t event = sentry_value_new_event();
-            sentry_envelope_t *envelope
-                = sentry__prepare_event(options, event, NULL);
+            sentry_envelope_t *envelope = sentry__prepare_event(
+                options, event, nullptr, !options->on_crash_func);
```
supervacuus, line 117: without informing `sentry__prepare_event()` whether `before_send()` should be invoked, we cannot make sure that a set `on_crash()` prevents the execution of the `before_send()` callback.

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -175,18 +177,52 @@ sentry__crashpad_handler(int signum, siginfo_t *info, ucontext_t *user_context)
...
+    //
+    // * we accept our fate and want a minidump (currently returning `false`)
+    // * we accept our fate and don't want a minidump (currently not available)
+    // * we rectified the situation, so crashpads signal-handler can simply
+    //   return, thereby letting the not-rectified signal-cause trigger a
+    //   signal-handler/exception-filter again, which probably leads to us
+    //   (currently returning `true`)
+    //
+    // TODO(supervacuus):
+    // * we need integration tests for more signal/exception types not only
```
Swatinem, line 211: We have `assert` / `abort` in there, do these cover this?

CHANGELOG.md
```diff
@@ -2,12 +2,26 @@
...
 
+**Features**
+
+- Provide `on_crash()` callback to allow clients to act on detected crashes
```
Swatinem, line 7: maybe also explain that this replaces `before_send` and how they interact.

src/backends/sentry_backend_inproc.c
```diff
@@ -535,8 +535,8 @@ handle_ucontext(const sentry_ucontext_t *uctx)
         }
 
         if (should_handle) {
```
Swatinem, line 537: I just noticed that we should `decref` (aka free) the event in the else branch. Probably this slipped by all the leak checkers since we are, well, crashing.

Swatinem (commented):
well its almost a bit late, but how about we make the `on_crash` take and return a `sentry_value_t`, similar to `before_send`.

At least for the `inproc` backend, we actually use that very event. The other backends discard it anyway and take all the info from the minidump.
But for in-proc, it might still be useful to add custom stuff to the event in case of hard crash.

src/backends/sentry_backend_inproc.c
supervacuus: Good catch!

supervacuus:
> well its almost a bit late, but how about we make the `on_crash` take and return a `sentry_value_t`, similar to `before_send`.
> 
> At least for the `inproc` backend, we actually use that very event. The other backends discard it anyway and take all the info from the minidump. But for in-proc, it might still be useful to add custom stuff to the event in case of hard crash.

I think this would be a much better interface and more future-proof. I think the best time to make such a change is right now before the interface is released (even considering it is experimental).

Swatinem (approved):

Swatinem (approved):

PR #740, opened 2022-07-29
fix: Iterate all debug directory entries

Swatinem:
Turns out a PE file can have more than one debug directory entry.
We now iterate over all of them, trying to find the one that has a valid
CodeView record that we can use to get a debug_id from.

Crashpad does the same here, for reference:
https://chromium.googlesource.com/crashpad/crashpad/+/refs/heads/main/snapshot/win/pe_image_reader.cc#162

### DIFF
--- a/src/modulefinder/sentry_modulefinder_windows.c
+++ b/src/modulefinder/sentry_modulefinder_windows.c
@@ -44,36 +44,48 @@ extract_pdb_info(uintptr_t module_addr, sentry_value_t module)
         nt_headers->OptionalHeader.SizeOfImage);
     sentry_value_set_by_key(module, "code_id", sentry_value_new_string(id_buf));
 
-    uint32_t relative_addr
-        = nt_headers->OptionalHeader.DataDirectory[IMAGE_DIRECTORY_ENTRY_DEBUG]
-              .VirtualAddress;
+    IMAGE_DATA_DIRECTORY debug_entry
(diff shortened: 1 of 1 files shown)


src/modulefinder/sentry_modulefinder_windows.c
```diff
@@ -44,36 +44,48 @@ extract_pdb_info(uintptr_t module_addr, sentry_value_t module)
...
     if (!relative_addr) {
         return;
     }
 
-    PIMAGE_DEBUG_DIRECTORY debug_dict
-        = (PIMAGE_DEBUG_DIRECTORY)(module_addr + relative_addr);
-    if (!debug_dict || debug_dict->Type != IMAGE_DEBUG_TYPE_CODEVIEW) {
+    size_t table_size = (size_t)debug_entry.Size;
+    size_t entry_size = sizeof(IMAGE_DEBUG_DIRECTORY);
+    if (table_size % entry_size != 0) {
```
kamilogorek, line 57: Would it be possible to somehow iterate over some entries and just drop the one that was somehow malformed and has fewer bytes than necessary? Or is it not possible to tell which one would even be broken at this point?

kamilogorek (approved):

src/modulefinder/sentry_modulefinder_windows.c
Swatinem: I think doing the same thing as crashpad here is okay. These structs have a fixed size, and the table should always be a multiple of that, otherwise the compiler did something wrong

Issue #743, opened 2022-08-02
Use Android system unwinder if API >= 28


marandaneto:
Android API 28 ships the system unwinder if API >= 28
https://android.googlesource.com/platform/system/unwinding/+/refs/heads/master/libunwindstack/AndroidVersions.md#android-9-pie_api-level-28

Today we have a stripped-down version of this library to be back compatible with API >= 16
Because of that, we don't benefit from new features or bundle size gains.

We can't get rid of it right now to gain bundle size, again, because of back compatibility.
What we can do tho is to actually use the Android system unwinder at runtime if API >= 28, so at least, we could benefit from likely better stack trace and its frames.

Related issue to bundle size: https://github.com/getsentry/sentry-native/issues/700

supervacuus:
Thanks, @marandaneto. This issue spans four topics:

* updating our stripped-down `libunwindstack` with the upstream changes
* conditional compilation of our `libunwindstack` fork for projects targetting API >= 28 to reduce binary/bundle-size
* dynamic loading of the system unwinder if run on a system with API >= 28 but were built for a lower target (i.e., we require the fallback)
* introduction (again conditional compile) of useful native APIs like `dl_iterate_phdr`(API >= 21) or the libc-like additions to `exec_info.h` (API >= 33), which are only useful for projects targetting respective versions

It is unclear yet if all of them make sense and in which order we'll approach them (except for the upstream changes, which I'll do first). But I tend towards a runtime solution as the one with the broadest value (even if it means keeping the fallback).

marandaneto:
Option 1 is a good first step.
Option 2 wouldn't work because we pre-compile with the min. supported version and ship it as a dependency of the Sentry Android SDK.
Option 3 is my suggestion, as long as we can dynamically load the system unwinder (runtime dependency), we are good, otherwise, we fall back to the pre-compiled and stripped-down (pre-compiled and shipped as a dependency) version of `libunwindstack`.
Option 4 I believe this causes a compilation issue < API 21, so not sure if this would be possible because we pre-compile the library.

I'd focus on Options 1 and 3.
Option 4 would be nice if possible, so Apps using API >= 21 would benefit from it.
Option 2 would have a huge impact on bundle size, even offering it as a separate package would be a problem because I believe the vast majority of Apps are not targeting minAPI 28, actually, the default today is still between 21 and 23.
We do know people using lower than 21 tho, so it's harder to drop support for them.

supervacuus:
[...]

PR #752, opened 2022-09-22
minimal implementation of sdk-version dependent...

supervacuus:
...dynamic loader, but mostly to show that using the generic unwinder is probably a dead end. There was a lot more time invested than is visible, but I wanted to break this down to a minimal implementation to be able to discuss its issues. This contains an approach to tackle #743, but it has many problems:

* `libunwindstack`, in SDK 28+, is still not exposed via any recent NDK and is thus not accessible to App developers or their dependencies but only to platform developers.
* it seems clear that the generic unwinder does not wrap AOSP `libunwindstack` to make it available to app developers indirectly, so while this PR uses it for SDKs >= 28, the unwinder doesn't provide a better stack-trace than our stripped-down version of libunwindstack.
* in addition to the above, the generic unwinder does not provide an interface to unwind relatively to the registers in `ucontext_t` or a passed-in address, which leads to stack traces that start from `sentry-native` frames.

So, at this point, it is relatively obvious that we cannot release this simply because it provides no visible improvements to the users (or rather regressions if you are deploying on targets with SDK >= 28).

It is not all bad, though:

* by packaging our fork of `libunwindstack` in a shared library, which we dynamically load during `sentry_init`, I could get down to 1.5MiB (for the `.so`) from 2MiB (for the `.a`) on `aarch64`.
* solely depending on the NDK unwinder (not that it would help us in the current implementation), produces an unwinder `.so`of only 26KiB, which would allow users who target only SDK >= 28 to drop the heavy legacy `.so` from their deployment packages.

So the question is how we should proceed with this (if at all):

* we can stay with what we currently have and only provide the [upstream updates](https://github.com/getsentry/libunwindstack-ndk/commit/90d80b7073ae2108f7f42e2f576bddf988069ba2).
* we can pull in the [upstream updates](https://github.com/getsentry/libunwindstack-ndk/commit/90d80b7073ae2108f7f42e2f576bddf988069ba2) and change the unwinder on android to be dynamically loaded during `sentry_init`, cash in the package-size reduction, but not provide any improvements for SDK >= 28.
* we can further investigate and try to resolve the C++ symbols from the platform `libunwindstack` dynamically, but it is very unclear how successful this will be. For one, it requires us to provide a minimal header environment to build the unwinder code in `sentry-native` _and_ it requires the platform's builders to provide symbol-name stability for us to find even the small subset of the interface across deployed androids in the wild.
* we can see if the `backtrace()` implementation in API 33 wraps `libunwindstack` and then replace the generic unwinder implementation we have now with `backtrace()`. In the worst case, it would at least provide us with an interface for a relative start into the stack.

Please chime in with further suggestions.


src/unwinder/libunwindstack_os.cpp
```diff
@@ -0,0 +1,42 @@
...
+}
+
+__attribute__((visibility("default"))) size_t
+android_unwind_stack(
+    void *addr, const sentry_ucontext_t *uctx, void **ptrs, size_t max_frames)
+{
+    (void)addr;
+    (void)uctx;
+    trace_state state = { ptrs, ptrs + max_frames };
+    _Unwind_Backtrace(unwind_callback, &state);
```
Swatinem, line 38: I always confuse the `_Unwind_*` and the `unw_*` APIs. So unfortunate that the first one does not allow unwinding from uctx.

Swatinem (commented):
lgtm, although it is a bit complex admittedly.

No matter what we decide specifically, we should update our fork regardless, maybe independently from this PR.

src/unwinder/sentry_unwinder_libunwindstack.c
```diff
@@ -0,0 +1,73 @@
...
+        = (android_unwind_stack_t)dlsym(handle, "android_unwind_stack");
+    if (!android_unwind_stack) {
+        SENTRY_WARN("failed to load unwinder function symbol");
+    }
+}
+
+void
+sentry__unload_unwinder()
+{
+    int result = dlclose(handle);
```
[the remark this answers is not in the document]
supervacuus: We should surely check this either way.

src/unwinder/libunwindstack_os.cpp
supervacuus: The `unw_` prefix is the "namespace" of `libunwind` which would be very convenient library and I would love to use, but AOSP dropped it and upstream doesn't really support Android.

But even if it would provide a `uctx` interface, I am not sure if the added complexity justifies the changes, as long as it doesn't offer any expected/significant improvements in the stack trace (for instance by mapping to `libunwindstack`).

This is especially true since most users use a pre-built package, which doesn't gain anything from the size improvements when using only a slim`mapper.so`, without ("manually") removing the `legacy.so`.

supervacuus: `libunwind` is still [deployed](https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md#unwinding) with NDK as the unwinder for C++ exception-handling and C scope-cleanups but it is not exposed as an app-developer library. 

The hint that NDK versions below r23 re-export the symbols of the previous libgcc unwinder, potentially screwing with client code that uses exception-handling, shows that stability of that "interface" is not a driver for decisions). Nonetheless, i can also try to access this implementation.

supervacuus:
Closing since there is no conclusive net improvement visible for the changes required:

* `libunwindstack` is still a platform library and not exposed to NDK users.
* we cannot load (via `dlopen`/`dlsym`) the deployed `libunwindstack.so`  system-library, even if we were able to provide a header-less mapping to its symbols, because [android doesn't allow loading system libraries](https://android-developers.googleblog.com/2016/06/improving-stability-with-private-cc.html). I think this is for good reason, because an application cannot provide a stable mapping cross-version or cross-OEM. It is hidden for a reason.
* using the generic unwinder does improve the stack trace slightly (it eliminates some frames, which we cannot for which we haven't even a mapping entry), but has no interface that allows us to provide a `ucontext` to start from (making the stack trace less helpful for users).
* `backtrace()` in SDK_VER 33 is also implemented in terms of the generic unwinder: https://android-review.googlesource.com/c/platform/bionic/+/1863333/6/libc/bionic/execinfo.cpp (which we can 
* `libunwinder` is not supported as an NDK library and is currently only packaged in NDK as an unwinder for C++ exception handling.

what we can extract from this:

* we can use an upstream synced version of our stripped down `libunwindstack` (which does not provide any considerable  improvements visible in our limited testing). This will happen in a separate PR after merging https://github.com/getsentry/libunwindstack-ndk/pull/6
* we found out - as a side investigation - that for native crashdumps on android in release R and above we probably should use the tombstones, which provide a much richer crash context than we could ever extract ourselves (without access to system libraries). This should be discussed with in a bigger round, outside the confines of a PR.
* we can try to further reduce the size a bit using our knowledge from this PR after updating our `libunwindstack`

PR #757, opened 2022-09-28
fix: Be more defensive around transactions

Swatinem:
This adds a bunch more NULL-checks for the outer opaque transaction/_ctx and span types. Also removes the `span_free` function which was pretty much duplicated with span_decref.


supervacuus (approved):

supervacuus:
LGTM, danger is still complaining.

PR #763, opened 2022-10-17
fix: Do not shut down winhttp transport on flush

### LINKED ISSUE #762: with winhttp transport, `sentry_flush` call kills background worker
**Description**

It took me a while to figure out why Sentry events on my Windows builds seemed really delayed. It turns out that `sentry_flush` with WinHTTP does significantly more than this says:

```
/**
 * Instructs the transport to flush its send queue.
 *
 * The `timeout` parameter is in milliseconds.
 *
 * Returns 0 on success, or a non-zero return value in case the timeout is hit.
 */
SENTRY_API int sentry_flush(uint64_t timeout);
```

The first call to `sentry_flush` actually kills the background submission worker as well, so any envelopes created after the first flush will not be sent until the application is started a second time.

The problem seems to be a copy-paste-o, and this would fix it:

```diff
diff --git a/src/transports/sentry_transport_winhttp.c b/src/transports/sentry_transport_winhttp.c
index ef9e74b..9d509f2 100644
--- a/src/transports/sentry_transport_winhttp.c
+++ b/src/transports/sentry_transport_winhttp.c
@@ -111,7 +111,7 @@ static int
 sentry__winhttp_transport_flush(uint64_t timeout, void *transport_state)
 {
     sentry_bgworker_t *bgworker = (sentry_bgworker_t *)transport_state;
-    return sentry__bgworker_shutdown(bgworker, timeout);
+    return sentry__bgworker_flush(bgworker, timeout);
 }

 static int
```

- [ ] During build
- [x] During run-time

**Environment**

- OS: Windows 10
- Compiler: MSVC 2019
- CMake version and config: winhttp backend

**Steps To Reproduce**

Send a sentry event, trigger a flush, try to send a second event. The second event will not arrive until the application restarts and the background worker is re-created.

**Log output**

```
<first sentry_flush call>
[39504] flushing transport
[39504] shutting down background worker thread
[39504] submitting task to background worker thread
[39504] executing task on worker thread
[39504] background worker thread shut down
<later, when trying to flush again>
[39504] flushing transport
[39504] trying to shut down non-running thread
```

Swatinem:
This seems to have been a typo all along.

Fixes #762

### DIFF
--- a/src/transports/sentry_transport_winhttp.c
+++ b/src/transports/sentry_transport_winhttp.c
@@ -111,7 +111,7 @@ static int
 sentry__winhttp_transport_flush(uint64_t timeout, void *transport_state)
 {
     sentry_bgworker_t *bgworker = (sentry_bgworker_t *)transport_state;
-    return sentry__bgworker_shutdown(bgworker, timeout);
+    return sentry__bgworker_flush(bgworker, timeout);
 }
 
 static int
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,20 +6,24 @@
 
 - Crashpad on Windows now supports `fast-fail` crashes via a registered Windows Error Reporting (WER) module. ([#735](https://github.com/getsentry/sentry-native/pull/735))
 
+**Fixes**:
+
+- Fix "flush" implementation of winhttp transport. ([#763](https://github.com/getsentry/sentry-native/pull/763))
+
 **Internal**:
 
-- Updated libunwindstack-ndk submodule to 2022-09-16 ([#759](https://github.com/getsentry/sentry-native/pull/759)) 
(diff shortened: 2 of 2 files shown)


src/transports/sentry_transport_winhttp.c
```diff
@@ -111,7 +111,7 @@ static int
 sentry__winhttp_transport_flush(uint64_t timeout, void *transport_state)
 {
     sentry_bgworker_t *bgworker = (sentry_bgworker_t *)transport_state;
-    return sentry__bgworker_shutdown(bgworker, timeout);
+    return sentry__bgworker_flush(bgworker, timeout);
```
supervacuus, line 114: LGTM. I think with this, we should create a release. The Android emulator-start times out again. I will provide a ci pr.

supervacuus (approved):

PR #775, opened 2022-12-06
Search code-id in ".note" ELF sections

supervacuus:
For some reason the Native SDK only searched in the ELF program headers for a code-/build-id entry, after which if unsuccessful, it immediately fell back to generating a code-/build-id on the fly by hashing the first page of ".text" section.

This change also searches for a ".note.gnu.build-id" section, after searching the program headers was unsuccessful.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,18 +61,21 @@ jobs:
             RUN_ANALYZER: code-checker,valgrind
           - name: macOS (xcode llvm)
             os: macOs-latest
-            ERROR_ON_WARNINGS: 1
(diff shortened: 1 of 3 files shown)


supervacuus:
Mac build errors are curious, I am running on macOS 13 and use the XCode 14.1 toolchain, and I only get deprecation warnings for `kIOMasterPortDefault`. I would prefer not to fix those in this PR given that I'll update crashpad next.

supervacuus:
Okay, just saw that we have ERROR_ON_WARNINGS enabled on all these builds. Duh :-)

Swatinem (approved):

Swatinem:
I think its okay to temporarily relax the error on warnings :-)

PR #784, opened 2022-12-26
fix: increase curl headers buffer size to 512

cnicolaescu:
The company that I work for is using an internal proxy with in-house generated public keys, which can be up to 400 characters long. When substituting the Sentry public key with the custom one, the `curl` transport doesn't write the `X-Sentry-Auth` header because the existing buffer 255 is too small.


Swatinem (approved):

supervacuus:
@cnicolaescu, can you add a corresponding entry to the change log? Then we could merge this change. Thx!

PR #785, opened 2022-12-26
fix: avoid race condition when starting bgw thread

cnicolaescu:
Specifically on Windows, the `sentry_thread_spawn` macro is defined as follows:

```
#    define sentry__thread_spawn(ThreadId, Func, Data)                         \
        (*ThreadId = CreateThread(NULL, 0, Func, Data, 0, NULL),               \
            *ThreadId == INVALID_HANDLE_VALUE ? 1 : 0)
```

We can see that `ThreadId` is initialized with the value returned by `CreateThread`, but the new thread might be running before `CreateThread` returns. If the new thread tries to access `bgw->thread_id`, the value might not be initialized yet, leading to undefined behaviour / crash.

In this change a new function `sentry__thread_get_current_threadid` is added, which returns the current thread id in a reliable way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Fixes**:
 
 - Better error messages in `sentry_transport_curl`. ([#777](https://github.com/getsentry/sentry-native/pull/777))
+- Fix sporadic crash on Windows due to race condition when initializing bgw thread id. ([#785](https://github.com/getsentry/sentry-native/pull/785))
 
 **Internal**:
 
--- a/src/sentry_sync.c
+++ b/src/sentry_sync.c
@@ -20,6 +20,12 @@ typedef struct {
 } THREADNAME_INFO;
 #    pragma pack(pop)
 
+sentry_threadid_t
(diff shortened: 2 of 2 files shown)


cnicolaescu:
Alternatively, the new thread could be created as suspended (passing the `CREATE_SUSPENDED` flag to `CreateThread`), then resumed in a 2nd step (using the `ResumeThread` API). This allows to correctly initialize `bgw->thread_id` before the new thread starts running.
This approach involves either:
- enhanching the `sentry__thread_spawn` macro
- creating the suspended thread in `sentry__thread_init`, then resuming it in `sentry__thread_spawn`

Please let me know if you'd prefer one of these 2 alternative approaches.

Swatinem (approved):

Swatinem:
lgtm. If I remember correctly setting the thread name is indeed the only access of the `thread_id` from within the thread itself. We also only create a single transport thread, and creation should be guarded in the `sentry_init` lock.

supervacuus:
@cnicolaescu, good catch; thx for the fix. Can you add an entry to the change log to merge this?

PR #791, opened 2023-01-24
fix: open the file lock on "UNIX" with O_RDRW

supervacuus:
For some reason, the file lock on UNIX was always opened using `O_RDONLY`. This doesn't make sense in combination with `O_TRUNC` since it is a write operation. On some systems (the trigger scenario was AWS EFS), this will lead to an `EBADF` if we try to `flock()` that descriptor. The docs consider this combination undefined.

This change also removes the AIX compile-conditional since this should work on all systems now.

I also added a warning to the run-initialization in case we get an error when trying to lock.

### DIFF
--- a/src/sentry_database.c
+++ b/src/sentry_database.c
@@ -4,6 +4,7 @@
 #include "sentry_json.h"
 #include "sentry_options.h"
 #include "sentry_session.h"
+#include <errno.h>
 #include <string.h>
 
 sentry_run_t *
@@ -49,13 +50,20 @@ sentry__run_new(const sentry_path_t *database_path)
     run->run_path = run_path;
     run->session_path = session_path;
     run->lock = sentry__filelock_new(lock_path);
-    if (!run->lock || !sentry__filelock_try_lock(run->lock)) {
-        sentry__run_free(run);
-        return NULL;
+    if (!run->lock) {
+        goto error;
     }
-
+    if (!sentry__filelock_try_lock(run->lock)) {
+        SENTRY_WARNF("failed to lock file \"%s\" (%s)", lock_path->path,
+            strerror(errno));
+        goto error;
+    }
     sentry__path_create_dir_all(run->run_path);
     return run;
+
+error:
+    sentry__run_free(run);
+    return NULL;
 }
 
 void
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,7 +134,7 @@ jobs:
 
       - name: Expose llvm PATH for Mac
         if: ${{ runner.os == 'macOS' }}
-        run: echo $(brew --prefix llvm@14)/bin >> $GITHUB_PATH
+        run: echo $(brew --prefix llvm@15)/bin >> $GITHUB_PATH
 
       - name: Installing Android SDK Dependencies
         if: ${{ env['ANDROID_API'] }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 **Fixes**:
 
+- Open the database file-lock on "UNIX" with `O_RDRW` ([#791](https://github.com/getsentry/sentry-native/pull/791))
 - Better error messages in `sentry_transport_curl`. ([#777](https://github.com/getsentry/sentry-native/pull/777))
 - Fix sporadic crash on Windows due to race condition when initializing background-worker thread-id. ([#785](https://github.com/getsentry/sentry-native/pull/785))
 - Increased curl headers buffer size to 512 (in `sentry_transport_curl`). ([#784](https://github.com/getsentry/sentry-native/pull/784))
--- a/src/path/sentry_path_unix.c
+++ b/src/path/sentry_path_unix.c
@@ -54,15 +54,7 @@ sentry__filelock_try_lock(sentry_filelock_t *lock)
 {
     lock->is_locked = false;
 
-    const int oflags =
-#ifdef SENTRY_PLATFORM_AIX
-        // Under AIX, O_TRUNC can only be set if it can be written to, and
-        // flock (well, fcntl) will return EBADF if the fd is not read-write.
-        O_RDWR | O_CREAT | O_TRUNC;
-#else
-        O_RDONLY | O_CREAT | O_TRUNC;
-#endif
-    int fd = open(lock->path->path, oflags,
+    int fd = open(lock->path->path, O_RDWR | O_CREAT | O_TRUNC,
         S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH);
     if (fd < 0) {
         return false;


src/sentry_database.c
```diff
@@ -49,13 +50,20 @@ sentry__run_new(const sentry_path_t *database_path)
...
     run->lock = sentry__filelock_new(lock_path);
-    if (!run->lock || !sentry__filelock_try_lock(run->lock)) {
-        sentry__run_free(run);
-        return NULL;
+    if (!run->lock) {
+        goto error;
+    }
+    if (!sentry__filelock_try_lock(run->lock)) {
+        SENTRY_WARNF("failed to lock file \"%s\" (%s)", lock_path->path,
+            strerror(errno));
```
supervacuus, line 58: Should we go the extra mile and use the thread-safe alternatives to `strerror()`?

Swatinem: I don’t think it hurts, though we have a lock around `sentry_init`, and it is supposed to be called from the main thread during initialization, so maybe not worth it if its a large effort?

Swatinem (approved):
nice catch! Though in the specific customer issue involving AWS EFS, the run directory also exists it would seem, which is only created on the line below.

And I believe we should correctly report an init failure if we fail creating and locking the run directory, can you double-check that please?

src/sentry_database.c
supervacuus: It is not a massive effort, but `strerror_r()` behaves differently in different environments, so I would probably add a wrapper with the right `#ifdefs` to `sentry_util`. 

Interestingly while we do access `errno` in a couple of places, I haven't seen any conversions to readable strings for logging (though we also log `write_buffer_with_flags()`). What I am trying to get across here: while it might not be strictly necessary to have a thread-safe `errno` printer, I can imagine this to help with logging in other UNIXy code.

supervacuus:
> nice catch! Though in the specific customer issue involving AWS EFS, the run directory also exists it would seem, which is only created on the line below.

this is very surprising, it would be interesting to see how that was achieved.

supervacuus:
> And I believe we should correctly report an init failure if we fail creating and locking the run directory, can you double-check that please?

This is what would be logged with the proposed change:

```
[sentry] INFO using database path "/mnt/efs/fs1/sentry-native/sentry-native/.sentry-native"
[sentry] WARN failed to lock file "/mnt/efs/fs1/sentry-native/sentry-native/.sentry-native/e93371c9-015e-4a3b-5a0b-813aa496f55d.run.lock" (Bad file descriptor)
[sentry] WARN failed to initialize run directory
[sentry] WARN `sentry_init` failed
[sentry] INFO sentry_close() called, but options was empty
```

Is this too much?

Swatinem:
That log output is perfect and what I would have expected to see.

Customer support (CC @rbisol) quoted the following log output from the customer:

```
[sentry] INFO entering breakpad minidump callback
[sentry] DEBUG read 12 modules from /proc/self/maps
```

I’m wondering how they ever got to that point if the SDK should properly fail initializing (and log that fact). Maybe the output was from a different test run with different configuration?

Dmitry1987:
> That log output is perfect and what I would have expected to see.
> 
> Customer support (CC @rbisol) quoted the following log output from the customer:
> 
> ```
> [sentry] INFO entering breakpad minidump callback
> [sentry] DEBUG merging scope into event
> [sentry] DEBUG trying to read modules from /proc/self/maps
> [sentry] DEBUG read 12 modules from /proc/self/maps
> [sentry] DEBUG adding attachments to envelope
> [sentry] DEBUG sending envelope
> [sentry] DEBUG serializing envelope into buffer
> [sentry] INFO crash has been captured 
> ```
> 
> I’m wondering how they ever got to that point if the SDK should properly fail initializing (and log that fact). Maybe the output was from a different test run with different configuration?

hi guys 😄  the troublesome customer is here!
so, the story of this stuff in few words is it's aws EKS on ec2 spot instances, pod with c++ app that crashes, a manual trigger in the app was made to "crash on demand" in order to troubleshoot this. we point sentry databasePath setting to EFS mount in the pod, in order to preserve crash data across "respawns" of the pod. because if things run well on 1 node and always will restart container on the same node after crash, in a large busy cluster however (especially on spot instances) the app pod might get evicted and respawned on another ec2 machine (and usually will, if it also consumed too much memory before the crash and is now considered "naughty" by the k8s scheduler, it'll reshuffle the cluster workloads across ec2 hosts and availability zones, that's why I don't use EBS disks because they won't be reattached to another ec2 if it'll end up in another AZ). Honestly, while writing I realized that I can constrain the pods to only be respawned in same AZ, but I don't want to lose the redundancy of being running on multiple AZs in the region 🤔 so EFS mount is the best option to preserve data across crashes of containers in a pod in a busy cluster AFAIK.  I might be wrong, but that's k8s specific things, if you are familiar with a better way to manage this, please share 🙏 .

About this log, 
```
[sentry] DEBUG read 12 modules from /proc/self/maps
```
it is reproducible, I can run strace it as long as we filter the system calls that'll be logged because most likely I cannot post anything from internal activities of the app itself (is it something that'll help?). 
When this happens, the folder .sentry-native and the lock files already exist. the init does not fail, but no session.json, last_crash, xxxxx.run dir, .envelope files, are created, when using EFS.  the moment I restart the app (it's all in interactive mode inside k8s pod shell) with another config file which points sentry to a local disk (not efs) the process is same, crash works, all files are written, the log above gets written (and I see the .run folder, envelope file, and on next start of the app the data flies to sentry dashboard and I see it there). 

Please let me know if you would like me to do some steps to test things out or to do some strace or other kind of debugger to show what's happening inside. I'll need to run the outputs through our devs though, in order to filter what I can share and what not, but I'll post the results to the support ticket that is ongoing, so I might be able to share there more than when posting here.  
Thank you for looking into this issue!

Swatinem:
Hi @Dmitry1987 and thanks for reaching out.

> the lock files already exist. the init does not fail, but no session.json, last_crash, xxxxx.run dir, .envelope files, are created, when using EFS.

That is good to know, so the lack of the `.run` dir is an indication that it might be early-exiting before creating that directory.
However it should log that fact and fail init in that case, that is a bit puzzling.

Can you share the logs of the `sentry_init` call? In particular the liner around:

```
[sentry] INFO using database path ...
```

Swatinem:
Also, @Dmitry1987 if it is not too much trouble, can you test-drive this PR and confirm is it helps your situation?

Dmitry1987:
> Hi @Dmitry1987 and thanks for reaching out.
> 
> > the lock files already exist. the init does not fail, but no session.json, last_crash, xxxxx.run dir, .envelope files, are created, when using EFS.
> 
> That is good to know, so the lack of the `.run` dir is an indication that it might be early-exiting before creating that directory. However it should log that fact and fail init in that case, that is a bit puzzling.
> 
> Can you share the logs of the `sentry_init` call? In particular the liner around:
> 
> ```
> [sentry] INFO using database path ...
> ```

I let the app run for a long time (a backend, websockets api) before manually crashing it, about 1-2 minutes (I do some activities interacting with the app in browser and then crash it after seeing that it's active and running correctly) so if the '.run' folder has to be created immediately after the init or something, it is not happening 😳 .

I will run the whole cycle now and will post the beginning of the log and the crash, then will switch to local non-efs disk path, and will repeat with the local disk path.

About testing the PR, I don't know how to incorporate it (I'm a DevOps) but will ask the developers to apply it to my branch so I can test it.

Dmitry1987:
reproducing the logs:

I exec into pod, made 2 config files, one has efs disk path for sentry in it, another has local path. 
checked EFS mount and cleaned up from previous run of the pod (which made a lock file)
image

strange, I haven't seen these errors in the last run with EFS, maybe because I forgot to enable the debug logging? I did see these errors in the past before upgrading to the last SDK version, but after upgrading there was no more init errors. 
image

now I crash the server.
ok, so this time it does not even print the lines it did during the crash. but luckily we have those lines in a ticket so we know I didn't invented them from thin air lol
image

after this crash, contents of the EFS mount:
image

one more lock file was added but no other things. 

Now I'll run with local disk path and not EFS mount. 
this is the empty local folder in the container that will be used 
image

These are starting logs:
image

Logs on crash
image

and contents of folder:
image


So can be considered reproduced, and it's correct what you mention about logging, looks like logs of crash are not shown with the EFS mount in use, but instead just the init failure shown. And with local disk the init failure logs are not displayed but crash logs are shown 😁 I might've mixed up some results while supplying data to the support ticket. Also in some of the runs I might've had debug logs off, because they're off by default in the config and I turn them on each time I do the manual test like this. 

I asked the developer if he can help me to apply this PR to the branch and test, will be great if this fixes the EFS writing issue.

supervacuus:
@Dmitry1987 This means the behavior is exactly as expected: if the database initialization fails

* the expected warnings appear in the log (if the debug logs are enabled)
* a `*.run.lock` is created but no `*.run` directory or `last_crash`-marker (which is also expected)

If both would not have been the case, then we would have to look for further issues, but now there is a good chance that this PR will also fix the issue you are seeing on your EFS mount.

Dmitry1987:
> @Dmitry1987 This means the behavior is exactly as expected: if the database initialization fails
> 
> * the expected warnings appear in the log (if the debug logs are enabled)
> * a `*.run.lock` is created but no `*.run` directory or `last_crash`-marker (which is also expected)
> 
> If both would not have been the case, then we would have to look for further issues, but now there is a good chance that this PR will also fix the issue you are seeing on your EFS mount.

sounds awesome 🤩 , can you cut a release or it will get into a next release? I was unable to get help in merging this PR to our codebase, but if there will be a new 'version' to pull, we can do it.

Dmitry1987:
oh cool, so I just checked out this branch and copied over all files into deeeeeep path inside our repo that I found, the diff (in another SCM we use) showed perfectly the same diff like in this PR so will be testing it now, yay :octocat:

Dmitry1987:
It works correctly now with the EFS disk mounts, thanks so much for fixing it! 🥇  🔥

supervacuus:
@bitsandfoxes and @markushi, I am not sure if you are "watching" the repo, but just keeping you in the loop.

supervacuus:
@Dmitry1987, so happy this fixed it. The corresponding release is on its way; you can subscribe to this issue to get a notification when it is ready: https://github.com/getsentry/publish/issues/1778

PR #794, opened 2023-01-26
Fixes for mingw builds (+ ARM64)

past-due:
This change helps fix mingw builds (together with: https://github.com/getsentry/crashpad/pull/78)

### DIFF
--- a/src/unwinder/sentry_unwinder_dbghelp.c
+++ b/src/unwinder/sentry_unwinder_dbghelp.c
@@ -56,11 +56,11 @@ sentry__unwind_stack_dbghelp(
 #elif defined(_M_ARM64)
     int machine_type = IMAGE_FILE_MACHINE_ARM64;
     stack_frame.AddrPC.Offset = ctx.Pc;
-# if defined (NONAMELESSUNION)
+#    if defined(NONAMELESSUNION)
     stack_frame.AddrFrame.Offset = ctx.DUMMYUNIONNAME.DUMMYSTRUCTNAME.Fp;
-# else
+#    else
     stack_frame.AddrFrame.Offset = ctx.Fp;
-# endif
+#    endif
     stack_frame.AddrStack.Offset = ctx.Sp;
 #elif defined(_M_ARM)
     int machine_type = IMAGE_FILE_MACHINE_ARM;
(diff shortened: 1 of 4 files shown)


supervacuus (changes requested):
LGTM, can you please run `make format` and commit the resulting formatting changes? Thx!

src/unwinder/sentry_unwinder_dbghelp.c
```diff
@@ -43,16 +43,32 @@ sentry__unwind_stack_dbghelp(
...
     stack_frame.AddrFrame.Offset = ctx.Ebp;
     stack_frame.AddrStack.Offset = ctx.Esp;
+#elif defined(_M_ARM64)
+    int machine_type = IMAGE_FILE_MACHINE_ARM64;
+    stack_frame.AddrPC.Offset = ctx.Pc;
+# if defined (NONAMELESSUNION)
+    stack_frame.AddrFrame.Offset = ctx.DUMMYUNIONNAME.DUMMYSTRUCTNAME.Fp;
+# else
+    stack_frame.AddrFrame.Offset = ctx.Fp;
+# endif
```
Swatinem, line 63: this looks weird, is that a mingw thing? Why is it only needed for `Fp` and not all the others?

Swatinem (approved):

src/unwinder/sentry_unwinder_dbghelp.c
past-due: I probably should have added a comment on this - this is a Windows header thing (reference: [with some details](https://devblogs.microsoft.com/oldnewthing/20170907-00/?p=96956)). 

I have no particular insight into the why, but that’s how it is for this particular structure (reference: https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-arm64_nt_context). So this will match the headers, and work regardless of whether `NONAMELESSUNION` is defined.

past-due:
> LGTM, can you please run `make format` and commit the resulting formatting changes? Thx!

@supervacuus Done 👍

supervacuus (approved):

supervacuus:
CC: @bitsandfoxes, @markushi

PR #795, opened 2023-01-26
Add conan badge

MartinDelille:
Since your project is available on the [conan center](https://conan.io/center/sentry-native), what about advertizing it on the README.md ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![Conan Center](https://shields.io/conan/v/sentry-native)](https://conan.io/center/sentry-native)
+
 <p align="center">
   <a href="https://sentry.io/?utm_source=github&utm_medium=logo" target="_blank">
     <picture>
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Conan Center](https://repology.org/badge/version-for-repo/conancenter/sentry-native.svg)](https://conan.io/center/sentry-native) [![nixpkgs unstable](https://repology.org/badge/version-for-repo/nix_unstable/sentry-native.svg)](https://repology.org/project/sentry-native/versions) [![Vcpkg](https://repology.org/badge/version-for-repo/vcpkg/sentry-native.svg)](https://repology.org/project/sentry-native/versions)
+[![Conan Center](https://shields.io/conan/v/sentry-native)](https://conan.io/center/sentry-native) [![nixpkgs unstable](https://repology.org/badge/version-for-repo/nix_unstable/sentry-native.svg)](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/libraries/sentry-native/default.nix) [![vcpkg](https://shields.io/vcpkg/v/sentry-native)](https://vcpkg.link/ports/sentry-native)
 
 <p align="center">
   <a href="https://sentry.io/?utm_source=github&utm_medium=logo" target="_blank">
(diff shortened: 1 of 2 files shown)


supervacuus:
Thx @MartinDelille! I am fine with adding this. I would personally prefer to have all up-to-date package managers referenced, but there isn't even a `shields.io` badge for `vcpkg`.

I looked into meta-package-manager badge-providers once and it seems [repology](https://repology.org/project/sentry-native/versions)` would answer that, but since it is syncing on the repositories of packages, a badge leads you only there and not to the "official" listing of the respective package-manager.

I know that previously there was opposition to advertising packages not maintained by sentry, so I am looping in @Swatinem and @ashwoods (cc @bitsandfoxes, @markushi) to maybe voice any official position.

Swatinem:
I believe it might be worth adding badges, maybe even a repology overview.

However it would also be a good idea to highlight that these are third party distributions not maintained directly by our team. Though not quite sure if any other open source project does that which is being packaged into various package archives that are not maintained by the original developers?

supervacuus:
> Though not quite sure if any other open source project does that which is being packaged into various package archives that are not maintained by the original developers?

That is a good question, indeed. For most package managers that aren't an integral component of the build tooling in a particular ecosystem, the package maintainer != the source maintainer. I cannot judge how much additional (support) effort advertising would mean on our side, but since we aren't advertising right now and still quite a few people are using vcpkg and conan, my guess is, that we can't really limit their use either.

A more positive interpretation would be, that we can consider any packaging as further bounds on the correct build and export of our artifacts and via open-source package repos we at least have insight into how the Native SDK is built and integrated for some of its users. In the worst case we might have to ask a user to file an issue with the respective package maintainers (which could come back to us, making it a good case again ;) ). 

Well, and there are some hard boundaries in which neither breakpad nor crashpad work (any distribution mechanism where we end up in a sandbox that prevents ptrace to run for instance).

I also found that repology has "tiny badges", so we could combine shields.io badges with those... or use the repology badges for all but link to the official package-manager site for those that provide an addressable resource on their website.

README.md
```diff
@@ -1,3 +1,5 @@
+[![Conan Center](https://shields.io/conan/v/sentry-native)](https://conan.io/center/sentry-native)
```
supervacuus, line 1:
```suggestion
[![Conan Center](https://repology.org/badge/version-for-repo/conancenter/sentry-native.svg)](https://conan.io/center/sentry-native) [![nixpkgs unstable](https://repology.org/badge/version-for-repo/nix_unstable/sentry-native.svg)](https://repology.org/project/sentry-native/versions) [![Vcpkg](https://repology.org/badge/version-for-repo/vcpkg/sentry-native.svg)](https://repology.org/project/sentry-native/versions)
```

supervacuus (changes requested):
Since `shields.io` currently doesn't support `vcpkg`, I would replace the badge with a repology badge (but let it link to Conan Center instead of the GitHub repo) and add the other two repositories that currently host a recent version of the Native SDK. Please also add a short entry to the changelog. Thx!

njakob:
Vcpkg badge support was added to shields.io last month through https://github.com/badges/shields/pull/8923 @supervacuus. Therefore, we could imagine having both Conan and Vcpkg versions!

[Vcpkg Version](https://vcpkg.link/ports/sentry-native)

Since you could only directly link the port files in Vcpkg GitHub repository, the example above contains the [detailed page on Vcpkg.link](https://vcpkg.link/ports/sentry-native) which has much more details.

supervacuus (approved):

Issue #806, opened 2023-02-09
Crash on arbitrary signal that is blocked on main thread


richardhozak:
**Description**

In our application we use epoll to manage our main application loop and we also let it handle plethora of linux signals as opposed to using signal handlers. In order to handle linux signal with epoll you need to block the signal and pass it to [`singalfd`](https://man7.org/linux/man-pages/man2/signalfd.2.html)  which is in turn passed to epoll to check if signal is waiting to be handled. The signals need to be blocked in main thread so they do not crash our application as they are not handled the traditional way (using signal handlers).

The problem arises when there is another thread running which does not have the signals blocked, because when linux kernel sends any kind of signal that is blocked to our app, it sees that the main thread has the signal blocked and thus chooses to send the signal to any thread that does not have the signal blocked and if not handled in that specific thread the whole app crashes.

When using the `curl` transport backend provided by sentry-native library, when `sentry_init` is called it creates bgworker thread for sending any events to `sentry.io` with `libcurl` and this thread does not have any signals blocked, meaning the linux kernel decides to send signal to that thread, crashing our app.

**When does the problem happen**

- [ ] During build
- [x] During run-time
- [ ] When capturing a hard crash (not applicable, crash reporting is disabled)

**Environment**

- OS: [e.g. Linux, 64-bit]
- Compiler: [e.g. gcc 12.2]
- CMake version and config: [e.g. 3.25, SENTRY_BACKEND=none, SENTRY_TRANSPORT=curl]

We do not use any crash handing backend (SENTRY_BACKEND=none) yet and just use the event reporting facilities of sentry-native.

**Steps To Reproduce**

As providing the steps to reproduce are quite involved and harder to provide than the actual solution to this problem I think it is best to demonstrate the possible solution that can be implemented in `sentry-native` to mitigate this problem.

This is a possible implementation of `sentry__thread_spawn` that would mitigate the issue described above:

```c
// block all signals, so threads created by sentry_init do not receive any signals,
// this way it does not intervene with signal handling on main thread
struct sigset_t oldset;
struct sigset_t allset;
sigfillset(&allset);
pthread_sigmask(SIG_SETMASK, &allset, &oldset);

pthread_create(...);

// restore signal mask back to 'oldset', so we do not mess
// with the rest of the program running on main thread
pthread_sigmask(SIG_SETMASK, &oldset, nullptr);
```

**Workaround Steps**

[...]

PR #813, opened 2023-03-05
fix: check for libcurl version and feature AsynchDNS

supervacuus:
As explained in https://github.com/getsentry/sentry-native/issues/806#issuecomment-1446862988, we should ensure that the Native SDK explicitly fails (at compile- and runtime) if `libcurl` wasn't compiled with the feature `AsynchDNS` enabled. I also introduced a lower bound for the `libcurl` version, which currently coincides with the version that introduced the `AsynchDNS`feature.

cc @bitsandfoxes, @markushi


src/sentry_utils.c
```diff
@@ -529,3 +529,20 @@ sentry__snprintf_c(char *buf, size_t buf_size, const char *fmt, ...)
     va_end(args);
     return rv;
 }
+
+bool
+sentry__check_min_version(sentry_version_t actual, sentry_version_t expected)
+{
+    if (actual.major < expected.major) {
```
Swatinem, line 536: I guess we are not considering semver breaks here? As we are allowing the major version to be higher than expected.

supervacuus: That's right; I wanted to encode only the minimum version bound in this function. But checking for an upper bound (in another function) would also make sense when checking versions at runtime (also because symmetry is awesome). 

But I wasn't in a hurry to add such a function because we are currently only using it for the `libcurl` check, and it seems that `libcurl` has more of a Linux major-bump idea for 8: https://daniel.haxx.se/blog/2021/09/13/heading-towards-curl-eight/ 

I am open to adding an (open interval) upper-bound check to utils, but I am unsure if we'll use it against `libcurl` under that premise. What do you think?

Swatinem: If they won’t break anything with a new major, then there is no reason to check for that either :-)

PR #816, opened 2023-03-12
Allow setting CRASHPAD_WER_ENABLED when using system crashpad

Cyriuz:
This is a patch I needed to make to add support for the WER module in the conan package: https://github.com/conan-io/conan-center-index/pull/16530

I included the removal of the dependency on crashpad::wer mostly for discussion. With my limited knowledge I think it shouldn't be needed other than maybe for installing? In the conan sentry-crashpad package that target is not exposed and would only be empty I think, since we don't really want the client to link to the lib. Does it make sense to remove it here or should the sentry-crashpad package export an empty target for it?


CMakeLists.txt
```diff
@@ -482,9 +482,6 @@ if(SENTRY_BACKEND_CRASHPAD)
 		endif()
 	endif()
 	add_dependencies(sentry crashpad::handler)
-	if(CRASHPAD_WER_ENABLED)
-		add_dependencies(sentry crashpad::wer)
-	endif()
```
Swatinem, line 487: I’m wondering if this is implied by `crashpad::handler`, or why is it okay to leave this off?

Cyriuz: I'm thinking that we don't really depend on this since we only really have a runtime dependency that the crashpad_wer.dll exists, and from what I could gather it gets installed fine without this, but it might also be something specific to the sentry-crashpad conan package, that's why I wrote above I'm a bit unsure about this. We shouldn't link to it or use any of its headers right?

supervacuus: Hi @Cyriuz, it makes sense to `add_dependencies` on `crashpad::wer` because it makes the dependency explicit (i.e., "make sure to build the `WER` module if you want to build `sentry.dll` with `WER` enabled") precisely because we don't link the target but only require it at runtime (this is especially useful when doing partial rebuilds during development). 

But you are right that adding it as a dependency is unnecessary when using a system crashpad. Can you add it to the build script, but only in the `false` branch of `if(SENTRY_CRASHPAD_SYSTEM)`?

CMakeLists.txt
```diff
@@ -413,6 +413,9 @@ endif()
 
 if(SENTRY_BACKEND_CRASHPAD)
 	option(SENTRY_CRASHPAD_SYSTEM "Use system crashpad" OFF)
+	if(CRASHPAD_WER_ENABLED)
+		add_compile_definitions(CRASHPAD_WER_ENABLED)
+	endif()
```
supervacuus, line 418: Moving this here will break the current setup because `CRASHPAD_WER_ENABLED` is automatically set by the `crashpad` build script (if the VS runtime target supports it) and then propagated to the parent scope. If we query it up here, we bypass this detection, and the compile definition is never set even though `WER` is enabled.

Please move this to the end of the branch of `if(SENTRY_BACKEND_CRASHPAD)`.

CMakeLists.txt
```diff
@@ -482,9 +482,6 @@ if(SENTRY_BACKEND_CRASHPAD)
 		endif()
 	endif()
 	add_dependencies(sentry crashpad::handler)
-	if(CRASHPAD_WER_ENABLED)
-		add_dependencies(sentry crashpad::wer)
```
supervacuus, line 486: You can move `add_compile_definitions(CRASHPAD_WER_ENABLED)` here and move `add_dependencies(sentry crashpad::wer)` right after the `add_subdirectory()` invocation in the `false` branch of  `if(SENTRY_CRASHPAD_SYSTEM)`.

supervacuus (changes requested):
The PR makes sense to me, but the concrete changes must be adapted. Please also add a changelog entry. Thx!

CMakeLists.txt
Cyriuz: That makes sense! I updated as suggested.

supervacuus (approved):

PR #828, opened 2023-04-04
[build] Fix CMake MSVC builds when WITH_ASAN_OPTION is ON

ndesogus:
When building Sentry with the Address Sanitizer enabled via MSVC on Windows the following warning is hit:

    LNK4044: unrecognized option '/fsanitize=address'; ignore

This is because Sentry's CMakeLists.txt calls:

    link_libraries(-fsanitize=address)

which is only needed on Unix platforms.


Swatinem (approved):

CMakeLists.txt
```diff
@@ -167,13 +167,17 @@ endif()
 option(WITH_ASAN_OPTION "Build sentry-native with address sanitizer" OFF)
 if(WITH_ASAN_OPTION)
 	add_compile_options(-g -fsanitize=address -fno-omit-frame-pointer)
-	link_libraries(-fsanitize=address)
+        if(NOT MSVC)
```
supervacuus, line 170: If we open this to `MSVC` users, we must ensure that it only happens if `MSVC` uses the clang-toolchain. How it is currently implemented will produce a warning with `cl.exe`. But there are other issues with our code base when we build with `clang` in `MSVC` since some build-time conditionals depend on `MSVC` not being `clang`.

Can I ask you how you currently build the project before we go further with this? Specifically of interest are:
* which backend do you use?
* do you let CMake/MSBuild build the whole solution using the `clangcl` toolchain, or do you select specific projects in VS?
* do you use `lld-link`?

supervacuus (changes requested):

supervacuus:
No response for more than two months. Feel free to re-open if it still makes sense.

Issue #831, opened 2023-04-11
Missing const qualifiers in sentry tracing


vitekcerny:
**Description**

Following functions missing const qualifiers in their parameters:

sentry.h:
- sentry_transaction_start_child
- sentry_span_start_child

sentry_tracing.h:
- sentry__value_span_new

Because of that, I can't call these functions with string literals, for example:
`sentry_transaction_start_child(parent, "operation", "description");`

I have looked into the sources and there doesn't seem to be a reason to mutate the original char* sequences (and therefore omit the const).
I also created a pull request for this issue: #832

**When does the problem happen**

- [x] During build
- [ ] During run-time
- [ ] When capturing a hard crash

**Environment**

- OS: Windows 10, 64-bit
- Compiler: MSVC 22
- CMake version and config: 3.26.3, SENTRY_BACKEND=inproc

supervacuus:
This was fixed in #827 and will be part of the upcoming release.

PR #832, opened 2023-04-11
fix: missing const qualifiers in sentry tracing

vitekcerny:
As explained in #831 this PR fixes missing const qualifiers in sentry tracing.


supervacuus:
Thank you for the fix, but this is already fixed as part of a bigger PR here: https://github.com/getsentry/sentry-native/pull/827

Issue #821, opened 2023-03-20
Allow changing `SENTRY_SDK_NAME` at runtime


marandaneto:
The build pipeline already considers the `SENTRY_SDK_NAME` config at compile time.
Ideally, the SDK is able to change its name at runtime.
https://github.com/getsentry/sentry-native/blob/903c17ae20888679caab1871cc74de577509452e/include/sentry.h#L26-L32

For example:
`sentry_set_sdk_name(...)` or `sentry_options_set_sdk_name(...)`, some SDKs have it under the `options` while this attribute is internal/private but allow Hybrid SDKs to do it anyway.
Context https://github.com/getsentry/sentry-dart/issues/1333

PR #834, opened 2023-04-21
Allow setting sdk_name at runtime

markushi:
Allows setting a custom `sdk_name`.

This is useful for hybrid SDK which do not compile their own version of sentry_native (E.g. Dart/Flutter re-uses the pre-compiled Android NDK version).

Fixes: https://github.com/getsentry/sentry-native/issues/821


markushi:
> @markushi how can we help to unblock this?

@marandaneto No all good, I was just waiting for https://github.com/getsentry/sentry-native/pull/827 to be merged, as it changes all public API with string args.

src/sentry_scope.c
```diff
@@ -29,8 +29,10 @@ get_client_sdk(void)
 {
     sentry_value_t client_sdk = sentry_value_new_object();
 
-    sentry_value_t name = sentry_value_new_string(SENTRY_SDK_NAME);
-    sentry_value_set_by_key(client_sdk, "name", name);
+    SENTRY_WITH_OPTIONS (options) {
```
markushi, line 32: Not sure if this is the best way to access the options, open for any feedback! I guess it's safe to assume options are set at this point, right?

tests/unit/test_options.c
```diff
@@ -0,0 +1,65 @@
...
+
+SENTRY_TEST(options_sdk_name_invalid)
+{
+    sentry_options_t *options = sentry_options_new();
+
+    // when the sdk name is set to an invalid value
+    const char *sdk_name = NULL;
+    const int result = sentry_options_set_sdk_name(options, sdk_name);
+
+    // then the value should should be ignored
```
supervacuus, line 57: double whammy

supervacuus (changes requested):
Looks good, but a couple of issues regarding options-initialization need to be fixed first.

tests/unit/test_options.c
```diff
@@ -0,0 +1,65 @@
+#include "sentry_options.h"
+#include "sentry_string.h"
+#include "sentry_testsupport.h"
+
+#include "sentry_options.h"
+#include "sentry_string.h"
+#include "sentry_testsupport.h"
```
supervacuus, line 7: Somehow the includes got duplicated here.

src/transports/sentry_transport_winhttp.c
```diff
@@ -15,7 +15,8 @@
 
 typedef struct {
     sentry_dsn_t *dsn;
-    wchar_t *user_agent;
+    char *sdk_user_agent;
+    wchar_t *winhttp_user_agent;
```
supervacuus, line 19: Argh... sorry for spamming, I was commenting in the sole commit view and the PR doesn't pick up.

This is a little too much for my taste. Let's keep the `wchar_t *user_agent` here and convert to `char*` at the only place it is needed: the call site of `sentry__prepare_http_request()`.

Just `sentry__string_from_wstr()` into a variable because it clones, and then make sure to free that string in the `!req` branch right below and again in the `exit` label.

supervacuus (approved):

PR #836, opened 2023-05-04
chore: update breakpad 2023-05-03

supervacuus:


Swatinem (approved):

supervacuus:
So the issue here: https://github.com/getsentry/sentry-native/actions/runs/4884009202/jobs/8716134172#step:11:514 is related to a change that adds new in favor of deprecated APIs: https://chromium-review.googlesource.com/c/breakpad/breakpad/+/4437109/13/src/common/mac/arch_utilities.cc

The problem is that we cannot use `__builtin_available()` here since `arch_utilities.cc` depends on a system header (`mach-o/utils.h`) that isn't available in earlier SDKs (like the one in our CI). If we want to keep the compatibility with previous SDKs (and it seems that was also the intention upstream), then we must guard the include (as well as the code below) with something like this:

```C
#if defined(__APPLE__) && (defined(__MAC_13_0) || defined(__IPHONE_16_0))
#include <mach-o/utils.h>
#endif
```

This would eliminate the warning pragmas if applied to the dependent functions below. While I can easily change this in our `breakpad` fork, I think it might be the right call to upstream this to understand the intention behind the change.

PR #839, opened 2023-05-05
feat: disable PC adjustment in the backend for libunwindstack

### LINKED ISSUE #835: Specify `instruction_addr_adjustment: "none"` when using Android libunwindstack
We should put the `instruction_addr_adjustment` into stack traces according to https://github.com/getsentry/rfcs/blob/main/text/0043-instruction-addr-adjustment.md

It looks like the `libunwindstack` that we use for unwinding on Android already does its adjustment before handing us the PC values:

https://github.com/getsentry/libunwindstack-ndk/blob/90d80b7073ae2108f7f42e2f576bddf988069ba2/Unwinder.cpp#L97

We should thus use `"none"`, to avoid double-adjusting in the symbolication pipeline which could lead to wrong symbolication results.

supervacuus:
Fixes #835 

Screenshot 2023-05-05 at 13 28 23

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 # Unreleased
 
+**Features**:
+
+- disable PC adjustment in the backend for libunwindstack ([#839](https://github.com/getsentry/sentry-native/pull/839))
+
 **Internal**:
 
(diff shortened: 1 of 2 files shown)


Swatinem (approved):
lets give @bitsandfoxes a chance to chime in and test if this solves his usecase in unity

bitsandfoxes (approved):
This does now work as in we get back from `Assembly-CSharp.cpp` to the proper `.cs` script but the line number itself is off.
But this does not seem to be related to this PR.

PR #840, opened 2023-05-09
feat: crashpad backend allows inspecting the event-id in the on_crash/before_send hooks

### LINKED ISSUE #779: Allow inspecting event-id for crashpad events
# Usecase

A customer would want to collect user feedback for a crash. They are trying to use the `on_crash` hook to do that.
However the user feedback API requires an event-id to attach the user feedback to.

# Background / Proposal

Our crashpad backend currently has no way to get the event-id of the to-be-sent event, as that is created on the server when the minidump is received.

According to @jan-auer, relay will use whatever id is provided in the special `__sentry_event` file which currently contains serialized scope information and is sent alongside crashpad minidumps.

It should be possible to generate a uuid ahead of time and serialize that into the `__sentry_event` file. When calling our hooks and provide a dummy event, we can then at least fill out the event id property so that customers can use it.

There might be the problem that the event/minidump needs to be sent before user feedback can be posted? Not entirely sure how the ordering and timing there could be guaranteed?

A possible flow could be this:
- App triggers crash hook, customer extracts event id from dummy event object.
- Customer spawns their own crash feedback gui using that event id.
- Crash handling is continued and crashpad captures the crash and sends a minidump using the event id embedded in the `__sentry_event` file.
- End user submits a feedback form about the crash using the event id.

supervacuus:
fixes #779

### DIFF
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -78,12 +78,13 @@ typedef struct {
     sentry_path_t *breadcrumb1_path;
     sentry_path_t *breadcrumb2_path;
     size_t num_breadcrumbs;
+    sentry_value_t event_id;
 } crashpad_state_t;
 
 static void
-sentry__crashpad_backend_user_consent_changed(sentry_backend_t *backend)
+crashpad_backend_user_consent_changed(sentry_backend_t *backend)
 {
-    crashpad_state_t *data = (crashpad_state_t *)backend->data;
+    auto *data = static_cast<crashpad_state_t *>(backend->data);
     if (!data->db || !data->db->GetSettings()) {
         return;
     }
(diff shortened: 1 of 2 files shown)


src/backends/sentry_backend_crashpad.cpp
```diff
@@ -141,7 +146,9 @@ sentry__crashpad_handler(int signum, siginfo_t *info, ucontext_t *user_context)
     sentry_value_t event = sentry_value_new_event();
 
     SENTRY_WITH_OPTIONS (options) {
-
+        auto *data = static_cast<crashpad_state_t *>(options->backend->data);
+        data->event_id = sentry_value_get_by_key(event, "event_id");
```
markushi, line 150: I guess there should never be the situation where `data->event_id` has already been set / needs to be freed first right?

markushi (approved):
LGTM!

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -78,31 +78,36 @@ typedef struct {
     sentry_path_t *breadcrumb1_path;
     sentry_path_t *breadcrumb2_path;
     size_t num_breadcrumbs;
+    sentry_value_t event_id;
 } crashpad_state_t;
 
 static void
-sentry__crashpad_backend_user_consent_changed(sentry_backend_t *backend)
+crashpad_backend_user_consent_changed(sentry_backend_t *backend)
```
Swatinem, line 85: is there a particular reason for renaming all these functions?

Swatinem (approved):

src/backends/sentry_backend_crashpad.cpp
supervacuus: The main reason is that across the modules, it seemed that the naming scheme is (or was at some point):

* `sentry_*` for all publicly available interfaces via `sentry.h`
* `sentry__*` for all private interfaces which cross module boundaries (i.e., are only internally exposed via module headers)
* no prefix for all static (i.e., module-private) functions, which shouldn't easily conflict in any "namespace"

So one part is trivial consistency, but it also gets easier to navigate if I know from the name if there might be an external contract to consider.

supervacuus:
closing in favor of #843

PR #846, opened 2023-05-19
Fix compiler error with msvc on non-unicode system

xyz1001:
same as https://github.com/getsentry/crashpad/pull/85

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 - Updated Breakpad backend to 2023-05-03. ([#836](https://github.com/getsentry/sentry-native/pull/836), [breakpad#35](https://github.com/getsentry/breakpad/pull/35))
 - Updated Crashpad backend to 2023-05-03. ([#837](https://github.com/getsentry/sentry-native/pull/837), [crashpad#82](https://github.com/getsentry/crashpad/pull/82))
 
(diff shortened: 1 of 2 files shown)


supervacuus:
Thanks again @xyz1001.  The same applies here with global flags vs. setting via `target_compile_options()`. There is an upcoming refactoring I'll have to do for the MSVC build, and I will change that there. Please add a changelog entry in the following form:

```
**Fixes**:

- Fix MSVC compiler error with on non-Unicode systems ([#846](https://github.com/getsentry/sentry-native/pull/846), [crashpad#85](https://github.com/getsentry/crashpad/pull/85))
```

in the `## Unreleased` section. Thx.

supervacuus (changes requested):
Thanks again @xyz1001.  The same applies here with global flags vs. setting via `target_compile_options()`. There is an upcoming refactoring I'll have to do for the MSVC build, and I will change that there. Please add a changelog entry in the following form:

```
**Fixes**:

- Fix MSVC compiler error with on non-Unicode systems ([#846](https://github.com/getsentry/sentry-native/pull/846), [crashpad#85](https://github.com/getsentry/crashpad/pull/85))
```

in the `## Unreleased` section. Thx.

supervacuus (approved):

PR #847, opened 2023-05-22
feat: crashpad handler http-proxy support

### LINKED ISSUE #788: Adapt the crashpad transports to support HTTP proxies
While all transport implementations of the Native SDK currently support simple HTTP forwarding proxies, `crashpad` doesn't configure them in its transports. Users who configure the Native SDK with an HTTP proxy and use crashpad as a backend will receive events, but no crashes when running behind a proxy.

supervacuus:
fixes #788 together with https://github.com/getsentry/crashpad/pull/86

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Disable PC adjustment in the backend for libunwindstack ([#839](https://github.com/getsentry/sentry-native/pull/839))
 - Crashpad backend allows inspection and enrichment of the crash event in the on_crash/before_send hooks ([#843](https://github.com/getsentry/sentry-native/pull/843))
+- Add http-proxy support to the `crashpad_handler` ([#847](https://github.com/getsentry/sentry-native/pull/847), [crashpad#86](https://github.com/getsentry/crashpad/pull/86))
 
 **Internal**:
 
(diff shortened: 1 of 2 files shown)


supervacuus:
The macOS tests failed because ATS doesn't allow (insecure) HTTP requests (which we use in our integration tests) 🙄. ATS was also a recurring issue during testing (in other scenarios), so I changed the transport code in the handler to fall back on `NSURLConnection` if no HTTP proxy is configured, which also means that all current implementations in the wild won't be surprised/affected by ATS restrictions that introduced under the hood.

markushi (approved):
LGTM!

supervacuus:
> LGTM!

just to be sure, does that include the [actual change](https://github.com/getsentry/crashpad/pull/86) in the crashpad fork too, or is this approval for the adaptation of the crashpad-backend in the Native SDK?

markushi:
> > LGTM!
> 
> just to be sure, does that include the [actual change](https://github.com/getsentry/crashpad/pull/86) in the crashpad fork too, or is this approval for the adaptation of the crashpad-backend in the Native SDK?

My bad: This is just the review for the adaption, I just did a review for the crashpad fork too.

PR #852, opened 2023-06-20
fix: Setting the default level for crash events in Crashpad to `fatal`

bitsandfoxes:
So the story goes like this:
An Unreal crash report lands in relay but the `crash_reason` got ignored and asserts and ensures ended up with the severity `FATAL`. 
Relay [now](https://github.com/getsentry/relay/pull/1677/files) reads the reason from the Unreal context and sets the level appropriately.

Unfortunately, Sentry simply overwrite the level set in Relay with `FATAL` anyway. With [this PR](https://github.com/getsentry/sentry/pull/50717/files), previously set levels get respected.

This PR aims to initialize crash events with `FATAL` but still allow users to overwrite the level.

### DIFF
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -108,6 +108,8 @@ crashpad_backend_flush_scope(
     sentry_value_t event = data->crash_event;
     if (sentry_value_is_null(event)) {
         event = sentry_value_new_object();
+        // FIXME: This should be handled in the FirstChanceHandler but that does
+        // not exist for macOS just yet.
         sentry_value_set_by_key(
             event, "level", sentry__value_new_level(SENTRY_LEVEL_FATAL));
     }
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -105,9 +105,12 @@ crashpad_backend_flush_scope(
     // properties that we do not want here. But in case of a crash we use the
     // crash-event filled in the crash-handler and on_crash/before_send
     // respectively.
-    sentry_value_t event = sentry_value_is_null(data->crash_event)
-        ? sentry_value_new_object()
-        : data->crash_event;
+    sentry_value_t event = data->crash_event;
+    if (sentry_value_is_null(event)) {
+        event = sentry_value_new_object();
+        sentry_value_set_by_key(
+            event, "level", sentry__value_new_level(SENTRY_LEVEL_FATAL));
+    }
 
     SENTRY_WITH_SCOPE (scope) {
         // we want the scope without any modules or breadcrumbs
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -149,6 +149,8 @@ sentry__crashpad_handler(int signum, siginfo_t *info, ucontext_t *user_context)
         auto *data = static_cast<crashpad_state_t *>(options->backend->data);
         sentry_value_decref(data->crash_event);
         data->crash_event = sentry_value_new_event();
+        sentry_value_set_by_key(data->crash_event, "level",
+            sentry__value_new_level(SENTRY_LEVEL_FATAL));
 
         if (options->on_crash_func) {
             sentry_ucontext_t uctx;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Fixes**:
+
+- Events captured by Crashpad are initialized with level `FATAL` ([#852](https://github.com/getsentry/sentry-native/pull/852))
+
 **Features**:
 
 - crashpad_handler: log `body` if minidump endpoint response is not `OK` ([#851](https://github.com/getsentry/sentry-native/pull/851), [crashpad#87](https://github.com/getsentry/crashpad/pull/87))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 **Fixes**:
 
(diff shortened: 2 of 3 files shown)


supervacuus (approved):
LGTM, if @Swatinem does agree with my points in the last comment (and yeah, changelog :P ).

supervacuus:
> The call down below to `crashpad_backend_flush_scope` has previously applied the default `"warning"` level based on the scope, which then ended up in the event/scope file that is written to disk and being sent by crashpad.

I am reluctant to this change because I think I still don't understand what happened in the backend that requires that change. I just tested `master` on macOS, and a crash now has `level` `warning` instead of `fatal`. So the (already live) backend change not only affects minidumps processed by the unreal handler but rather all minidumps. That was unclear to me (is this change tracked somewhere @bitsandfoxes?).

Was your suggestion @Swatinem to initialize the level in the scope to `fatal`? What would that break? At least i would cover all platforms.

supervacuus:
Okay, I found it: https://github.com/getsentry/sentry/pull/50717.

Sorry, @bitsandfoxes, I misunderstood the severity of that change in our sync yesterday. There is no quick way to adapt to this cleanly in the Native SDK. Changing it only in the `crashpad` backend is also wrong because all minidumps will be affected, so we must ensure this also works for `breakpad` (which is easier because the handler works on all platforms and we can add the level to the crash-event before the hooks are invoked similar to the change in `crashpad`).

We could change the scope defaults. But this is a breaking change, and I wonder if it is okay from an SDK pov. 

For `crashpad`, there is no way on `macOS` to set the level at the right point in time without implementing a `FirstChanceHandler`, making the `crashpad_handler` aware of the `MessagePack` scope or touching the defaults.

I am open to alternative proposals, though.

Swatinem:
The fundamental problem here is:
- Previously, the backend would just hardcode `"fatal"` for all minidumps.
- We want a way to overwrite that, so the backend respects an existing `level` without overwriting it
- Well, it turns out that the `apply_scope_to_event`, and by extension the messagepack files written to disk for the crashpad integration get the default level from the scope, which is `"warning"`.
- We did not anticipate this, and as a result, with the backend changes, all minidump crashes are not treated as `"warning"` which is wrong. The should *default* to `"fatal"`, but customers who want to should have a way to override that, via `on_crash` or whatever.

supervacuus:
> We did not anticipate this, and as a result, with the backend changes, all minidump crashes are not treated as `"warning"` which is wrong. The should _default_ to `"fatal"`, but customers who want to should have a way to override that, via `on_crash` or whatever.

Yeah, I understood it the same way. The problem - at least from my current understanding - is that changing the default in the scope affects every envelope we send and not only crashes. So if a user doesn't explicitly set the level in either the event or change it in the scope, everything we send will be `FATAL`. That is at least a breaking change (although I cannot say how severe that would be, especially compared to other SDKs).

I think the most sensible approach would be to 
* explicitly set the event level in every SDK crash handler available to `FATAL` (already happening in `inproc`, easily added for all `breakpad` platforms, possible for Windows and Linux in `crashpad`
* leave the SDK defaults alone
* and add a `FIXME(supervacuus)` clause to  `processing.py` that handles crashes coming from `macOS` using `crashpad` as a backend to how it was done previously (i.e., ignore the level coming from the SDK and only check for the `crashed` property). Something like this:

```python
# FIXME(supervacuus): remove this when we implement a FirstChanceHandler in macOS crashpad.
if os == "macOS" and backend == "crashpad" and response.get("crashed") is not None and response["crashed"]:
    # we crashed on macOS using the crashpad backend, so let us ignore the level from the SDK
    data["level"] = "fatal"
else:
    # do what is currently implemented
    if response.get("crashed") is not None and data.get("level") is None:
        data["level"] = "fatal" if response["crashed"] else "info"
```

This way we don't create a dance around shifting defaults, and the problem is solved for all in a somewhat future-proof way. The level coming from the SDK is only ignored for crashpad-on-macOS if it actually crashed... a situation where SDK users cannot override anyway.

Horrible? Please let me know!

Swatinem (approved):

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -105,9 +105,12 @@ crashpad_backend_flush_scope(
...
     // crash-event filled in the crash-handler and on_crash/before_send
     // respectively.
-    sentry_value_t event = sentry_value_is_null(data->crash_event)
-        ? sentry_value_new_object()
-        : data->crash_event;
+    sentry_value_t event = data->crash_event;
+    if (sentry_value_is_null(event)) {
+        event = sentry_value_new_object();
+        sentry_value_set_by_key(
+            event, "level", sentry__value_new_level(SENTRY_LEVEL_FATAL));
```
supervacuus, line 112: Why?

supervacuus:
The failing `breakpad` integration test is due to this: https://github.com/getsentry/sentry-native/blob/6ac1404dc6f46dd6388c3b9f40ca32b585f3ef58/src/backends/sentry_backend_breakpad.cpp#L119-L121
which can be removed.

supervacuus:
We should extend the integration tests with checks for all crashes to be `FATAL`. No need to do this in this PR; I can see we are in a hurry.

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -105,9 +105,14 @@ crashpad_backend_flush_scope(
...
     // crash-event filled in the crash-handler and on_crash/before_send
     // respectively.
-    sentry_value_t event = sentry_value_is_null(data->crash_event)
-        ? sentry_value_new_object()
-        : data->crash_event;
+    sentry_value_t event = data->crash_event;
+    if (sentry_value_is_null(event)) {
+        event = sentry_value_new_object();
+        // FIXME: This should be handled in the FirstChanceHandler but that does
+        // not exist for macOS just yet.
```
supervacuus, line 112: Thanks! ❤️

supervacuus (approved):

PR #857, opened 2023-06-27
fix: deadlock pattern in dynamic sdk-name assignment

supervacuus:
We recently introduced a deadlock in `sentry_init()` when implementing dynamic SDK names. This was the reason for the uptick in build timeouts in the CI (which was almost exclusively due to `TEST(concurrent_init)` being stuck). Sorry, I missed that during the review.

I also increased the concurrency in the test so that there is a higher chance of triggering such cases. The current level should be high enough for CI. I need to up that considerably on my machines to catch the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Fixes**:
+
+- Remove deadlock pattern in dynamic sdk-name assignment ([#857](https://github.com/getsentry/sentry-native/pull/857))
+
 ## 0.6.4
 
 **Fixes**:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -144,6 +144,7 @@ The example currently supports the following commands:
 - `discarding-before-send`: Installs a `before_send()` callback that discards the event.
 - `on-crash`: Installs an `on_crash()` callback that retains the crash event. 
 - `discarding-on-crash`: Installs an `on_crash()` callback that discards the crash event.
+- `override-sdk-name`: Changes the SDK name via the options at runtime.
 
 Only on Windows using crashpad with its WER handler module: 
 
--- a/examples/example.c
+++ b/examples/example.c
@@ -218,6 +218,10 @@ main(int argc, char **argv)
             options, discarding_on_crash_callback, NULL);
     }
 
+    if (has_arg(argc, argv, "override-sdk-name")) {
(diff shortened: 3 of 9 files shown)


markushi (approved):
Nice catch, thanks for getting this fixed. Can you estimate how severe this issue is for mobile SDKs? If this has a bigger impact I'm happy to help you getting this shipped for all relevant mobile SDKs.

supervacuus:
> Nice catch, thanks for getting this fixed. Can you estimate how severe this issue is for mobile SDKs? If this has a bigger impact I'm happy to help you getting this shipped for all relevant mobile SDKs.

It is hard to say, but I don't think this will affect many. 

The `options` lock is a recursive mutex so you can go pretty wild with nested locking solely on `options`. Also, the `options` lock usually is short-lived because it only guards the access to the ref-count to retrieve a read-only reference. The nesting of options -> scope -> options is generally a lousy pattern if one thread acquires the scope lock before another locks options at the start of `sentry_init()` 😓 .

How realistic this is on the Mobile SDKs depends on how they interact with the Native SDK (threads already running that could access scope-related functions before another thread calls `sentry_init()`, for instance... I think we talked about this recently). I don't know if the last ANR was a rare case or if there is an uptick, like in our CI.

In any case, we can release it immediately, but I am unsure if it is urgent.

The other aspect is that with the current change, you must ensure that the global scope hasn't been touched before `sentry_init()` or the initialization defaults to the build-time SDK name.

We can change this by not setting the dynamic SDK name in `get_scope()` at all but rather doing it in the `SENTRY_WITH_SCOPE_MUT` in `sentry_init()`. This way, we could eliminate the `options` path towards `get_scope()` and an optionally set `sdk_name` will always overwrite the scope in `sentry_init()`. WDYT?

supervacuus:
> We can change this by not setting the dynamic SDK name in `get_scope()` at all but rather doing it in the `SENTRY_WITH_SCOPE_MUT` in `sentry_init()`. This way, we could eliminate the `options` path towards `get_scope()` and an optionally set `sdk_name` will always overwrite the scope in `sentry_init()`. WDYT?

Just in case it is unclear: this would mean we would have to keep the `client_sdk` unfrozen in `get_scope()` and freeze it in `sentry_init()` together with setting the `sdk_name`. It boils down to these two options (they are two sides of a coin).

bitsandfoxes:
Superseded by https://github.com/getsentry/sentry-native/pull/858/files

PR #858, opened 2023-06-27
fix: deadlock in dynamic sdk-name scope init...

supervacuus:
...but this time by moving the `client_sdk` freeze from the scope-initialization to `sentry_init()`.

This alternative approach to https://github.com/getsentry/sentry-native/pull/857 allows scope related functions to be called before `sentry_init()`, where we will overwrite the `sdk_name` a last time before we freeze `client_sdk`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 **Fixes**:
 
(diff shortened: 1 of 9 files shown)


supervacuus:
ping anyone?

bitsandfoxes (approved):
Thanks! 🔥

PR #862, opened 2023-07-07
Update README.md remove beta note

HazAT:
#skip-changelog


supervacuus:
#skip-changelog

supervacuus (approved):

supervacuus:
@HazAT, should we reflect this in the version?

kahest:
> @HazAT, should we reflect this in the version?

Not needed, it was an oversight that we didn't remove the beta note a while ago

PR #872, opened 2023-08-08
Update external/CMakeLists.txt to install breakpad headers

MartinDelille:
This is a naive try to remove the following patch in conan-center-index: https://github.com/conan-io/conan-center-index/blob/5b2024b2a/recipes/sentry-breakpad/all/conanfile.py#L111-L152

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -185,3 +185,43 @@ target_include_directories(breakpad_client
     PUBLIC
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>"
 )
+
+target_compile_features(breakpad_client PUBLIC cxx_std_11)
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
+    find_path(LINUX_SYSCALL_INCLUDE_DIR NAMES linux_syscall_support.h)
+    target_include_directories(breakpad_client PRIVATE ${LINUX_SYSCALL_INCLUDE_DIR})
+endif()
+install(TARGETS breakpad_client
+    ARCHIVE DESTINATION lib
+    LIBRARY DESTINATION lib
+    RUNTIME DESTINATION bin
+)
+file(GLOB COMMON_FILES breakpad/src/common/*.h)
+install(FILES ${COMMON_FILES}
+    DESTINATION include/breakpad/common
+)
+set(PLATFORM_FOLDER)
+if(IOS)
+    set(PLATFORM_FOLDER ios)
+elseif(APPLE)
+    set(PLATFORM_FOLDER mac)
+elseif(UNIX)
+    set(PLATFORM_FOLDER linux)
+endif()
+if(WIN32)
+    set(PLATFORM_FOLDER windows)
+endif()
+if(NOT PLATFORM_FOLDER)
+    message(FATAL_ERROR "Unknown os -> don't know how to install headers")
+endif()
+file(GLOB COMMON_PLATFORM_HEADERS breakpad/src/common/${PLATFORM_FOLDER}/*.h)
+install(FILES ${COMMON_PLATFORM_HEADERS}
+    DESTINATION include/breakpad/common/${PLATFORM_FOLDER})
+install(DIRECTORY breakpad/src/client/${PLATFORM_FOLDER}
+    DESTINATION include/breakpad/client
+    FILES_MATCHING PATTERN *.h
+)
+install(DIRECTORY breakpad/src/google_breakpad/common
+    DESTINATION include/breakpad/google_breakpad
+    FILES_MATCHING PATTERN *.h
+)


supervacuus:
Hi @MartinDelille, thanks for the PR! The changes in this PR touch on multiple topics where I don't see an immediate connection (compile features, LSS replacement, installation of all headers). Can you give me an overview of what you are trying to accomplish and how it relates to the Native SDK or our fork of breakpad? Thanks!

MartinDelille:
Hi @supervacuus,

I'm just trying to make conan center index recipe maintenance easier by removing the current for patches: https://github.com/conan-io/conan-center-index/tree/master/recipes/sentry-breakpad/all/patches

supervacuus:
Hi @MartinDelille! 

I understand why you'd want to eliminate these patches from your code, but the particular changes (also in your [getsentry/breakpad PR](https://github.com/getsentry/breakpad/pull/37)) make little sense outside the Conan build environment. And I don't mean mainly in the code but on an abstract level.

This is also an iffy topic because we are incredibly thankful to any package maintainers who help extend users' access to the Native SDK and are willing to accept changes that improve their work. Still, at the same time, we must ensure that PM-specific decisions do not hurt users not using that particular PM (or no PM at all) and that changes align with the intended usage (and the Native SDK already supports so many directions in that regard). We are also open to some discussion, so please don't consider this response as a trivial negation.

The first discrepancy I see in this context is that sentry's fork of `breakpad` (or `crashpad` or any other future backend) is considered a "private" implementation detail of the Native SDK. What does this mean?

* while everything is open-source, and we cannot and don't want to control how people use the Native SDK and its vendored dependencies, we do see the forks for `crashpad`, `breakpad`, and `libunwindstack` as code 100% shaped by the requirements of the Native SDK and its users
* we do not expose the APIs or headers of any backend dependency for that reason
* for `breakpad` in particular, we only use the client portion of the repository and only include the exception handlers of the supported platforms
* while we currently maintain a complete fork of the upstream `breakpad` with a couple of local changes, this is primarily a detail of version control which eases upstream synchronization. The `breakpad` client changes rarely compared to any other part of the repository, and it could make sense to reduce our fork to that minimum. 
* users can choose a particular backend for their use case, and that decision and its dependencies will - by design - leak through (because of different capabilities, support of platforms, architectures, and our APIs), but there should be no external dependents that would break if we decide to modify these otherwise private interfaces and we also do not want to invest into testing this
* `breakpad` is unique in that regard because we also support it as a system installation (i.e., not from our fork), but I can easily imagine useful changes that would break compatibility (like giving the `crash_handler` on Linux access to `siginfo_t` and `ucontext_t`, as a simple example)

From that perspective, installing our fork of `breakpad` and all its headers makes little sense. I recommend that all package maintainers not expose our vendored dependencies as separate packages.

Your replacement of `lss` with an installed version is something that would have to be entirely optional (if we accept this change at all) because we always want to 

1. be aligned with the version used in the upstream repository
2. provide all users access to that particular version (which we cannot guarantee without vendoring).

So, I could imagine we accept an `lss` change like this if the current setup is still fully supported as the primary use case (and not just replaced) and only provide this scenario via compile definitions.

Of course, this raises the question, why the Conan community doesn't propose this change to the upstream `breakpad` since the [corresponding Conan package also applies a patch](https://github.com/conan-io/conan-center-index/blob/master/recipes/breakpad/all/patches/0001-Use_conans_lss.patch) that replaces the vendored `lss` with the global Conan version?

Be also aware that if you'd consider `breakpad` (and, in turn, `lss`) as a private dependency of the Native SDK, there would be no need for this patch.

external/CMakeLists.txt
```diff
@@ -185,3 +185,43 @@ target_include_directories(breakpad_client
     PUBLIC
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>"
 )
+
+target_compile_features(breakpad_client PUBLIC cxx_std_11)
```
supervacuus, line 189: Why was this line in your build necessary (we already define the `CMAKE_CXX_STANDARD` for our C++ dependencies)? AFAIK as I know upstream `breakpad` switched to C++17 (but it might not yet be necessary to enable in the `breakpad` client). I am more than happy to switch to a target-based setting of C++ standard enforcement, but this should happen in another PR and be applied to all targets so we can remove the global setting.

external/CMakeLists.txt
```diff
@@ -185,3 +185,43 @@ target_include_directories(breakpad_client
...
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
+    find_path(LINUX_SYSCALL_INCLUDE_DIR NAMES linux_syscall_support.h)
+    target_include_directories(breakpad_client PRIVATE ${LINUX_SYSCALL_INCLUDE_DIR})
+endif()
```
supervacuus, line 193: Please read my bigger comment for why a simple replacement like this and your `breakpad` PR will not be accepted.

external/CMakeLists.txt
```diff
@@ -185,3 +185,43 @@ target_include_directories(breakpad_client
...
+install(FILES ${COMMON_PLATFORM_HEADERS}
+    DESTINATION include/breakpad/common/${PLATFORM_FOLDER})
+install(DIRECTORY breakpad/src/client/${PLATFORM_FOLDER}
+    DESTINATION include/breakpad/client
+    FILES_MATCHING PATTERN *.h
+)
+install(DIRECTORY breakpad/src/google_breakpad/common
+    DESTINATION include/breakpad/google_breakpad
+    FILES_MATCHING PATTERN *.h
+)
```
supervacuus, line 227: This part is problematic because it exposes an implementation detail of the Native SDK in a potentially global place where the only scenario it is used in is the `sentry-native` Conan package. Does this patch really entail so much maintenance burden on your side (given you stick with a separate package)? Given the rare changes to the `breakpad_client` this patch should never have to change, or does it?

MartinDelille:
Hi @supervacuus ! Thank you so much for the detailled answer!

If I understand correctly, you would recommand to remove both sentry-breakpad and sentry-crashpad recipes to remove the burden of maintaining three recipes instead of one ?

supervacuus:
> If I understand correctly, you would recommand to remove both sentry-breakpad and sentry-crashpad recipes to remove the burden of maintaining three recipes instead of one ?

Without knowing much about specific Conan packaging guidelines that would provide counter arguments or other technical reasons in that context that would prevent it, yes, I would recommend this very much.

MartinDelille:
Just for the update, I changed the sentry-native recipe for conan center index according to your recommandation: https://github.com/conan-io/conan-center-index/pull/20762

supervacuus:
Thanks for the update @MartinDelille!

PR #883, opened 2023-09-12
docs: clarify behavior of sentry_flush + sentry_close

sapphonie:
This PR
- clarifies that sentry_shutdown does not remove any crash handlers

Copying over a conversation from the Sentry discord;

> supervacuus — Today at 4:25 AM
@sappho
 The crashpad client does not provide a mechanism to uninstall its in-process crash-handler nor shutdown the crashpad_handler executable. So the handler is still installed after sentry_close().

> On Linux we brutally restore signal handlers without making crashpad aware of this. Maybe we should do the same on Windows?

> If you want to disable the crash-handler you can call  SetUnhandledExceptionFilter(NULL) which will disable the local handler installed by crashpad. This can happen in your Shutdown() method right after invoking sentry_close(). 

More TODO involving

- [x] ~~documenting if sentry_flush() is async or not~~ done, it is not async, it blocks
- [x] ~~documenting if sentry_close() is async or not~~ done, see above
- [ ] documenting in example.c that crash handling will still be called after sentry_shutdown()

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+**Thank you**:
+
+Features, fixes and improvements in this release have been contributed by:
+
+- [@sapphonie](https://github.com/sapphonie)
+
 ## 0.6.5
 
 **Fixes**:
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -1275,7 +1275,7 @@ SENTRY_API int sentry_flush(uint64_t timeout);
  *
  * Returns 0 on success.
  *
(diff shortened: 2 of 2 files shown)


supervacuus:
Thanks @sapphonie! Could you let me know if you're done with this PR? Is it okay if I take over and handle a few technicalities?

#skip-changelog

sapphonie:
> Thanks @sapphonie! Could you let me know if you're done with this PR? Is it okay if I take over and handle a few technicalities?
> 
> #skip-changelog

Sure, go ahead

supervacuus (approved):

Issue #882, opened 2023-09-08
Unable to compile crashpad due to missing <stdint.h> #includes (Linux) - fixed in upstream mini_chromium


past-due:
**Description**

Failure to compile crashpad due to missing `<stdint.h>` `#includes`

It looks like mini_chromium _just_ fixed this upstream:
https://chromium.googlesource.com/chromium/mini_chromium/+/eb418ec05f1142fd03cead1ef6201e6bafee7c9a

So updating the mini_chromium submodule is probably desirable: https://github.com/getsentry/crashpad/tree/getsentry/third_party/mini_chromium

> **Note**
> I just happened to notice recent mini_chromium commits moving macOS / iOS files:
> https://chromium.googlesource.com/chromium/mini_chromium/+/f5370228f40bbb8c453e17f6af6c6742858c45d1
> https://chromium.googlesource.com/chromium/mini_chromium/+/e35fc73aa87bb27e10306900b15a18b0e9c7ca42
> 
> So it looks like [the CMakeLists.txt](https://github.com/getsentry/crashpad/blob/getsentry/third_party/mini_chromium/CMakeLists.txt) also needs updating.

**When does the problem happen**

Trying to compile Sentry w/ crashpad on the very latest Freedesktop 23.08 Runtime.

- [x] During build
- [ ] During run-time
- [ ] When capturing a hard crash

supervacuus:
Thanks for the report @past-due! I will update `crashpad` with upstream soon. I hope the referenced `mini_chromium` was also updated in `crashpad` (we track the DEPS from upstream), otherwise I can test it with the particular commit.

supervacuus:
Okay, it seems that [DEPS](https://chromium.googlesource.com/crashpad/crashpad/+/refs/heads/main/DEPS#48) currently points to [10f39a97650a0fe0b305415c15434443c0690a20](https://chromium.googlesource.com/chromium/mini_chromium/+/10f39a97650a0fe0b305415c15434443c0690a20) which is the parent of the commit you are referring to. I will add this with the next update from upstream.

past-due:
@supervacuus Looks great! 🥳 Any chance a new sentry-native release can be pushed this week?

supervacuus:
https://github.com/getsentry/publish/issues/2788

PR #891, opened 2023-09-29
chore: update crashpad 2023 09 28

supervacuus:
This will also fix: https://github.com/getsentry/sentry-native/issues/882

But I am mostly trying to get a cross-check from the CI because I have some weird behavior when reinstalling the crashpad backend on Windows (this is not ready for review).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Fixes**:
+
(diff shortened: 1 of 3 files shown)


supervacuus:
Okay, the failing LLVM-Mingw build is actually due to `ninja` being removed from the Windows runner image: https://github.com/actions/runner-images/issues/8343

supervacuus:
> Okay, the failing LLVM-Mingw build is actually due to `ninja` being removed from the Windows runner image: [actions/runner-images#8343](https://github.com/actions/runner-images/issues/8343)

But the `LLVM-Mingw` build downloads and installs `ninja` in this PowerShell script: https://github.com/getsentry/sentry-native/blob/master/scripts/install-llvm-mingw.ps1#L33-L54

So, this build always used the installed `ninja` vs. the explicitly installed one.

Swatinem (approved):

PR #901, opened 2023-11-10
fix: disable sigaltstack on Android

supervacuus:
I am creating this as a tracking branch for a pre-release. Disabling "our" `sigaltstack` on Android is probably not the solution that will end up in a release, but it should provide a sound basis for broader testing in the downstream SDKs.

### DIFF
--- a/src/backends/sentry_backend_inproc.c
+++ b/src/backends/sentry_backend_inproc.c
@@ -13,6 +13,31 @@
 #include "transports/sentry_disk_transport.h"
 #include <string.h>
 
+/**
+ * Android's bionic libc seems to allocate alternate signal handler stacks for
+ * every thread and also references them from their internal maintenance
+ * structs.
+ *
+ * The way we currently set up our sigaltstack seems to interfere with this
+ * setup and causes crashes whenever an ART signal handler touches the thread
+ * that called `sentry_init()`.
+ *
+ * In addition to this problem, it also means there is no need for our own
+ * sigaltstack on Android since our signal handler will always be running on
+ * an alternate stack managed by bionic.
+ *
+ * Note: In bionic the sigaltstacks for 32-bit devices have a size of 16KiB and
+ * on 64-bit devices they have 32KiB. The size of our own was set to 64KiB
+ * independent of the device. If this is a problem, we need figure out
+ * together with Google if there is a way in which our configs can coexist.
+ *
+ * Both breakpad and crashpad are way more defensive in the setup of their
+ * signal stacks and take existing stacks into account (or reuse them).
+ */
+#ifndef SENTRY_PLATFORM_ANDROID
+#    define SETUP_SIGALTSTACK
+#endif
+
 #define SIGNAL_DEF(Sig, Desc)                                                  \
     {                                                                          \
         Sig, #Sig, Desc                                                        \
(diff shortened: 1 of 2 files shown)


kahest (approved):
🔥

src/backends/sentry_backend_inproc.c
```diff
@@ -87,14 +113,15 @@ startup_inproc_backend(
...
         return 1;
     }
     g_signal_stack.ss_size = SIGNAL_STACK_SIZE;
     g_signal_stack.ss_flags = 0;
     sigaltstack(&g_signal_stack, 0);
-
+#    endif
     sigemptyset(&g_sigaction.sa_mask);
     g_sigaction.sa_sigaction = handle_signal;
     g_sigaction.sa_flags = SA_SIGINFO | SA_ONSTACK;
```
markushi, line 127: Just double checking: We still should set the `SA_ONSTACK` flag, even if we don't allocate the altstack ourselves right?

markushi (approved):
LGTM!

src/backends/sentry_backend_inproc.c
supervacuus: Yeah, `SA_ONSTACK` tells the signal handler to use the alt stack or falls back to the process stack. Since the assumption is that every thread has an alt stack, we should be good.

etnlGD:
Hello, our project has recently collected a lot of crashes related to Android  signal handling. We happened to see this bug fix, which seems to be highly related to the crashes we encountered. However, I didn't find how this fix would cause a crash.

> Android's bionic libc seems to allocate alternate signal handler stacks forevery thread and also references them from their internal maintenance structs.

As far as I know, the stacks allocated by bionic are only referenced for deallocating which should be harmless.


> The way we currently set up our sigaltstack seems to interfere with this setup and causes crashes whenever an ART signal handler touches the thread that called `sentry_init()`.

Could you please provide more details about these crashes? I haven't found any evidence to suggest that setting up stacks repeatedly would lead to a crash.

supervacuus:
Hi @etnlGD,

> Hello, our project has recently collected a lot of crashes related to Android signal handling. We happened to see this bug fix, which seems to be highly related to the crashes we encountered. However, I didn't find how this fix would cause a crash.

The most significant change to our previous implementation is that the default `sigaltstack` from the bionic thread implementation has a reduced stack size in the signal handler, which can be an issue if you allocate something big in either the `before_send` or `on_crash` handler. We generally recommend not to do that, but we do know that some frameworks expose very large crash-context structures that should be heap allocated.

> > Android's bionic libc seems to allocate alternate signal handler stacks forevery thread and also references them from their internal maintenance structs.
> 
> As far as I know, the stacks allocated by bionic are only referenced for deallocating which should be harmless.

Yeah, the reference is not the reason for crashes, but rather the way the `sigaltstack` is set up (separate mmap vs heap allocation, ignoring previously installed sigaltstacks, not creating a stack guard page, maybe even using incompatible flags). The previous `sigaltstack` configuration was not made with Android in mind, and since it was only configured for the thread on which `sentry_init()` was called, it was not a general solution anyway.

> > The way we currently set up our sigaltstack seems to interfere with this setup and causes crashes whenever an ART signal handler touches the thread that called `sentry_init()`.
> 
> Could you please provide more details about these crashes? I haven't found any evidence to suggest that setting up stacks repeatedly would lead to a crash.

The ART team urged us not to set up a `sigaltstack` for any threads after we identified them as a potential culprit. We saw many crashes in the sigchain starting with the August 2023 Google Play System update on Android 12, 13, and 14. Repros tested against our previous `sigaltstack` setup were all fixed using the default Android `sigaltstack` configuration.

This doesn't mean we won't touch the topic again, but it would have to be a non-default behavior where users have more control over which threads should be configured with that stack.

PR #902, opened 2023-11-10
fix: stuck crashpad_client on windows

supervacuus:
This also breaks CI and thus prevents the release of 0.6.7 which should contain https://github.com/getsentry/sentry-native/pull/901

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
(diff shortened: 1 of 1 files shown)


kahest (approved):

supervacuus:
I merged this after the 3rd restart of the Android emulator job... because this change has nothing to do with android. The android build is failing because of GitHub infra.

Issue #900, opened 2023-11-03
 Crashpad Handler Unintended Behaviors after installation of Sentry


JLuse:
**Description**

A customer has reported the following issues only after updating one of their products (the host) to use Sentry.

Issue 1: When the crashpad_handler process is terminated (e.g., using kill -TERM <pid>), the client repetitively attempts to restart the handler, leading to consistent failures.

Issue 2: On teardown of the plugin's crashpad_handler instance, it inadvertently terminates the host process's crashpad_handler. This behavior is observed even when manually terminating the plugin's handler (which is confirmed to be a separate process with a distinct PID); soon after, the host's crashpad handler is also terminated.

Prior to this, when both the customer indicated the host and the plugin were directly using Crashpad for crash reporting, these issues were not observed. 

- [ ] During build
- [X] During run-time
- [ ] When capturing a hard crash

**Steps To Reproduce & Log output**

Customer has provided two repro apps as well as accompanying logs within the support ticket:
https://sentry.zendesk.com/agent/tickets/105906

supervacuus:
Hi @JLuse, I need help understanding the issue (without access to Zendesk for context).

> Issue 1: When the crashpad_handler process is terminated (e.g., using kill -TERM ), the client repetitively attempts to restart the handler, leading to consistent failures.

Why is it necessary to manually kill the `crashpad_handler`? Does the customer expect things to work when the `crashpad_handler` gets killed from the shell? Because I am unaware of support for this.

> Issue 2: On teardown of the plugin's crashpad_handler instance, it inadvertently terminates the host process's crashpad_handler. This behavior is observed even when manually terminating the plugin's handler (which is confirmed to be a separate process with a distinct PID); soon after, the host's crashpad handler is also terminated.

This one needs to be clarified, too: which plugin are you referring to? Is this happening in the context of Unity or Unreal? The words "plugin" and "host" do not refer to anything specific in sentry-native terminology, so this is either usage from the customer domain or from one of the downstream SDKs.

CC: @kahest, @bitsandfoxes

willmbaker:
Hi @supervacuus, I may be able to help you with some of those specifics ...

> Why is it necessary to manually kill the crashpad_handler? Does the customer expect things to work when the crashpad_handler gets killed from the shell? Because I am unaware of support for this.

[...]

PR #910, opened 2023-11-27
fix: Maintain client in crashpad state

supervacuus:
This PR is related to this issue: https://github.com/getsentry/sentry-native/issues/900#issuecomment-1825457223.

The crashpad client is stateful (on macOS, it contains the send right to the exception port), and it might be a good idea to keep the client instance for the life cycle of the Native SDK. This also fixes part of the issue (terminating one `crashpad_handler` kills the other). I don't see any downside to this, although having a second client will almost always overwrite a previously initialized one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,17 @@
 # Changelog
 
+## Unreleased
+
+**Fixes**:
+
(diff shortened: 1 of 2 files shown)


markushi (approved):
LGTM!

Swatinem (approved):

PR #922, opened 2023-12-07
Remove options memory leak during consent setting

stima:
Do not do early exit loop early, that leads to a incorrect reference counting of options during consent setting.


stima:
Not sure if it worth adding some testing around that.

supervacuus (changes requested):
Thanks, @stima, nice catch!

By the way, the correct approach to exit these synchronization scopes early would be to use `continue` instead of `break`. I looked through the code to see if there was another occurrence of such an early exit, but there was none.  From my side, it is okay to add the nesting in this function.

> Not sure if it worth adding some testing around that.

It is 💯 worth it, given that triggering this bug in a test is trivial (you only need to consent twice in a row), and our unit tests are run through `ASAN` in CI, which would have revealed this earlier if the branch had been covered. Not adding this case as a regression test would be careless.

Please give another consent here
https://github.com/getsentry/sentry-native/blob/a51ccbbdb00cd3042022f7df6d4e352e9ae6c0c3/tests/unit/test_consent.c#L30

and add a short comment that explains why it was done with a reference to this PR (you can also write a separate test if you want, but that should be enough). Please also add the fix to the changelog. Thx!

supervacuus (approved):

supervacuus:
Can you please fix the formatting of your comment https://github.com/getsentry/sentry-native/actions/runs/7165776324/job/19508507795?pr=922#step:3:90 accordingly? Thx!

Issue #908, opened 2023-11-23
Backend selection for static lib


vaind:
I'm trying to integrate sentry-native crash handling to sentry-dotnet SDK when compiled to NativeAOT. I'm integrating as a static library to avoid having the end consumer application to need to bundle an additional library. Bundling and compilation all work fine without any backend and transport (originally, I've just used sentry-native to get the list of loaded debug images to enable server-side symbolication; that all worked fine).

However, I'm trying to figure out which backend to use for desktop platforms (Windows, Linux, macOS) in this scenario, specifically for a static library. Since crashpad is an external process (i.e. an external binary which I've tried to avoid with the static lib), I was thinking `breakpad` or `inproc`. The thing is, I could not find any documentation what are the main differences, functionally, between the various backends, aside from platform support. It would be nice to have a doc that I could consult. I've already had the same issue when implementing sentry-native to the Unity and Unreal SDKs and because All these use-cases/platforms are a bit different, I'm having trouble making the decision each time, other than "pick the default" (then new questions arise, like: is the default the default because it's the best one or because it has been the default years ago and is just kept that way to not break users, etc)

supervacuus:
> The thing is, I could not find any documentation what are the main differences, functionally, between the various backends, aside from platform support. It would be nice to have a doc that I could consult. 

You are right. We don't have a document to help users decide on a specific backend. After 1.5 years of being exposed to questions surrounding the backend, I can short-cut it to always using Crashpad, except when you shouldn't.

The reasons for "shouldn't" by order of occurrence are:

1. can't distribute/deploy an external handler artifact/process (for which there are uncountable, but often artificial reasons)
2. want to use a crash handler hook on macOS
3. clients running on a version of the OS that Crashpad doesn't support

If you touch one of those, you should always default to `breakpad` first. It immediately eliminates 1.) and 2.) above while still providing decent coverage of platforms, and it, like crashpad, sends `minidumps`, which provide more context than what the `inproc` backend offers.

You'll need to use `inproc` only if you have tested `breakpad` and `crashpad` and are 100% sure they don't fit your use case.

> I've already had the same issue when implementing sentry-native to the Unity and Unreal SDKs and because All these use-cases/platforms are a bit different, I'm having trouble making the decision each time, other than "pick the default" (then new questions arise, like: is the default the default because it's the best one or because it has been the default years ago and is just kept that way to not break users, etc)

[...]

PR #927, opened 2023-12-19
build: make crashpad the default backend on Linux

supervacuus:
We've been discussing this for a year now, and it is a recurring and understandable mental burden for newcomers (as well as developers of integrations within Sentry). From my POV, there is no reason to have `breakpad` over `crashpad` as the default backend for Linux besides the worry that we might break builds that rely on that default. Most recent interactions with new users show that people choose `crashpad` in the end anyway.

I would make this a breaking change toward 0.7.0.

This is also related to the context touched by https://github.com/getsentry/sentry-native/issues/908.

I am open to counter-arguments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Breaking changes**:
+
+- Make `crashpad` the default backend for Linux. ([#927](https://github.com/getsentry/sentry-native/pull/927))
+
 **Fixes**:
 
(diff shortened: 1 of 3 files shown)


anonrig (approved):

Swatinem (approved):

bsergean:
fyi when I tried crashpad, it was requiring some extra permissions which aren't always available in kubernetes (like ability to ptrace a process), which is why we are still using breakpad.

but I understand that's it's a more reasonable default, maybe this constraint should be documented somewhere.

Issue #934, opened 2024-01-04
[windows] including only wtypes in sentry.h makes Swift fail to import sentry.h


hyp:
**Description**

https://github.com/thebrowsercompany/swift-sentry is a Swift wrapper for the native SDK for Sentry. It uses Swift's embedded clang to import `sentry.h`, by importing it as a Clang module. Clang modules rely on modularized headers, i.e. the header must be self contained, and it must import all header dependencies. Unfortunately `sentry.h`'s import `wtypes` is problematic for two reasons:
- it's incomplete, it uses APIs from other headers like `winnt.h`
- the `wtypes` header is not modularized as a standalone module in the Windows SDK.

This causes a Swift build error with Swift-5.10 or Swift main snapshot. The error is shown below.

**When does the problem happen**

- [x] During build

**Environment**

- OS: Windows 11, 64-bit
- Compiler: Swift 5.10 snapshot
- SDK: Windows 10 SDK (10.0.20348.0)

**Steps To Reproduce**

Build swift-sentry.

**Log output**

```
......\build\debug\_deps\swiftsentry-src\Sources\sentry\include/SentryApp.h:6:10: note: in file included from ......\build\debug\_deps\swiftsentry-src\Sources\sentry\include/SentryApp.h:6:
#include <sentry.h>
         ^
......\build\debug\NugetPackages\io.sentry.sentry-native.crashpad.windows.amd64.0.0.0-ede193b\include\sentry.h:93:14: note: in file included from ......\build\debug\NugetPackages\io.sentry.sentry-native.crashpad.windows.amd64.0.0.0-ede193b\include\sentry.h:93:
#    include <wtypes.h>
             ^
C:\Program Files (x86)\Windows Kits\10\\include\10.0.20348.0\\shared\wtypes.h:323:15: error: missing '#include <windef.h>'; 'RECTL' must be declared before it is used
typedef const RECTL *LPCRECTL;
              ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.20348.0\shared\windef.h:170:3: note: declaration here is not visible
} RECTL, *PRECTL, *LPRECTL;
  ^
```

hyp:
This can be resolved by including `#windows.h` instead.

PR #935, opened 2024-01-04
include <windows.h> to ensure that sentry.h is modularized correctly

hyp:
This makes it possible to use this header in Swift

Fixes https://github.com/getsentry/sentry-native/issues/934


hyp:
@Swatinem What do you think about this change? I see you previously dropped the use of `windows.h` in favor of `wtypes.h`, can we bring it back instead?

Swatinem:
The problem is that `windows.h` includes the whole world, and you need a bunch of other defines to contain the blast radius a bit. We previously had some problems related to including `windows.h`, thats why it was removed.

@supervacuus hopefully knows more about this, and can also guide us to a solution here. Lets wait a bunch more days for people to return from winter break :-)

compnerd:
@Swatinem totally understandable that Windows.h is pretty broad

I think that the documentation does clearly state that you should use `Windows.h` _most_ of the time (there are a few headers which can be imported without `Windows.h`).  Adding a wrapping header to define all the control macros locally is an option that you might consider. The other possible alternative might be to start pulling in random headers in addition to this particular header (in some aspects re-creating a `Windows.h` as if you had defined the other macros). Unfortunately, many of the headers are not standalone nor do they include what they use.

supervacuus (changes requested):
A short check with `/showIncludes` reveals that we indirectly include `windows.h` via `rpc.h` anyway, so I see no reason to block this. This change is not only relevant to making the Native SDK work with clang modules but also makes explicit that we depend on `windows.h`. 

There is currently no way around this without hiding the platform-specific context types in the header via some abstraction, and this is not a priority given that we still need to expose and track machine- and platform-dependent interfaces somewhere (but it might be the right choice down the road).

So, LGTM, but please add a changelog entry. Thanks!

hyp:
Thanks, I updated the release notes!

PR #948, opened 2024-02-08
Not all modules were added to the modules list resulting in <unknown> stack traces

sc-aki:
Not all modules were added to the modules list resulting in <unknown> stack traces.

The problem was that in order to a module to get added to the list, the next module needed to have a valid elf header, which is not always the case. The fixed code checks if the module names are different.

### DIFF
--- a/src/modulefinder/sentry_modulefinder_linux.c
+++ b/src/modulefinder/sentry_modulefinder_linux.c
@@ -602,17 +602,6 @@ get_linux_vdso(void)
     return 0;
 }
 
-static bool
-is_valid_elf_header(void *start)
-{
-    unsigned char e_ident[EI_NIDENT];
-    if (!read_safely(e_ident, start, EI_NIDENT)) {
-        return false;
-    }
-    return e_ident[EI_MAG0] == ELFMAG0 && e_ident[EI_MAG1] == ELFMAG1
-        && e_ident[EI_MAG2] == ELFMAG2 && e_ident[EI_MAG3] == ELFMAG3;
-}
-
 static void
 load_modules(sentry_value_t modules)
 {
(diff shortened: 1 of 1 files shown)


src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -687,7 +677,14 @@ load_modules(sentry_value_t modules)
...
 
-        if (is_valid_elf_header((void *)(size_t)module.start)) {
+        // Module is appended if next module has different file name
+        if ((!last_module.file.len
+                && (!module.file.len || last_module.file.len != module.file.len
+                    || memcmp(last_module.file.ptr, module.file.ptr,
+                        module.file.len)))
+            || (!module.file.len || last_module.file.len != module.file.len
+                || memcmp(
+                    last_module.file.ptr, module.file.ptr, module.file.len))) {
```
Swatinem, line 687: you should use `!sentry__slice_eq(module.file, last_module.file)` here to make this more readable.

Swatinem (commented):
I wonder if checking for the executable flag is enough here, as I would be hesitant to remove the ELF header check. Can you elaborate your issue a bit more (maybe in the private chat we shared) why exactly the previous check failed on your executable?

src/modulefinder/sentry_modulefinder_linux.c
sc-aki: The key here is that the ELF check must be removed. Not all memory areas are valid ELF binaries. It doesn't make sense to search for ELF header of the next memory area as a condition to add a module to the list. Quite often there's anonymous memory areas between loaded libraries, and in such case the original code doesn't work.

supervacuus: Hi @sc-aki. Can you provide an example of such a memory map (or just a few entries that visualize the problem)?

> Not all memory areas are valid ELF binaries.

But do they still contain valid ELF code referenced from the stack trace?

> Quite often there's anonymous memory areas between loaded libraries, and in such case the original code doesn't work.

Yes, in these cases, we would ignore that map entry. Are you saying that you have anonymous entries that contain code that your stack trace is referencing? How do you ensure the code is readable in the backend symbolication? 

Wouldn't this ignore anonymous map entries as well: https://github.com/getsentry/sentry-native/blob/master/src/modulefinder/sentry_modulefinder_linux.c#L677-L681

supervacuus: > How do you ensure the code is readable in the backend symbolication?

Or is this change only meant for Android (where symbolication happens in the client)?

PR #954, opened 2024-02-22
feat: sentry add gzipped with compression

Strive-Sun:
Hi, I noticed that the http request from sentry is missing content-encoding, so I added this parameter. `DeflateInit2` references header file in `crashpad`, and I am not sure if this reference is appropriate. if further modifications are needed, please let me know at any time. Thx in advance.


src/sentry_transport.h
```diff
@@ -89,4 +89,10 @@ sentry_prepared_http_request_t *sentry__prepare_http_request(
  */
 void sentry__prepared_http_request_free(sentry_prepared_http_request_t *req);
 
+/**
+ * Gzip
+ */
+bool sentry_gzipped_with_compression(const char *body, const size_t body_len,
+    char **compressed_body, size_t *compressed_body_len);
+
```
Swatinem, line 97: this seems to be local to the transport, so not really necessary to expose it via a header.
Also the naming convention uses two underscores (`sentry__XXX`) for internal functions.

src/sentry_transport.c
```diff
@@ -165,9 +167,24 @@ sentry__prepare_http_request(sentry_envelope_t *envelope,
...
+    size_t compressed_body_len = 0;
+    bool compressed = sentry_gzipped_with_compression(
+        body, body_len, &compressed_body, &compressed_body_len);
+    if (compressed) {
+        if (body_owned) {
+            sentry_free(body);
+            body = NULL;
+            body_owned = false;
+        }
+    }
```
Swatinem, line 180: Once you compress the body, you do not need the original `body` anymore.
Therefore, I would just reuse the `body/_owned` variables so you don’t have to duplicate all the code down below.

Swatinem (commented):
What is missing here is building / linking the zlib code

src/sentry_transport.c
```diff
@@ -5,9 +5,11 @@
 #include "sentry_ratelimiter.h"
 #include "sentry_string.h"
 
+#include "external/crashpad/third_party/zlib/zlib/zlib.h"
```
supervacuus, line 8: Yeah, @Strive-Sun, as you mentioned, that is problematic. The transport implementation is entirely independent of any backend being built, so a direct dependency on a particular backend doesn't work.

If you want to continue this PR, I would recommend to:

* search for an installed `zlib` in the top-level `CMakeLists.txt` (but do not make it a required dependency)
* if `zlib` was found, use a compile definition that enables your compression code in the transport and uses the system-include instead of the `crashpad` one
* make sure that the `crashpad` build uses the same `zlib` implementation (because it will be linked with `libsentry` via `crashpad-util`) by setting the `CRASHPAD_ZLIB_SYSTEM` option in the top-level `CMakeLists.txt` before `add_subdirectory()` for `crashpad`.

This way, the transport compression will be enabled whenever a `zlib` system implementation is available, and we have no clash in case `crashpad` is used as a backend.

An alternative path to the first point in my recommendations would be to introduce a build-option `SENTRY_TRANSPORT_COMPRESSION`, where you can make the dependency required after checking for that option. The remaining points would be the same. The difference would be that people can opt-in/-out from compression and don't get that feature simply by an existing dependency on their build-system.

Strive-Sun: @supervacuus Thank you for your detailed suggestions. I added the option `SENTRY_TRANSPORT_COMPRESSION` to `CMakeLists.txt`, which make the dependency required after checking for that option. But now there are two questions.

- whether `zlib` needs to add the `cmake < 3.12` support like the libcurl?
- The default installation method for third-party libraries in Windows is to use the `vcpkg` tool, which is a universal approach. Dependency cannot find the library installed by `vcpkg`, how can find these libs?
image

supervacuus: > whether zlib needs to add the cmake < 3.12 support like the libcurl?

No. We no longer support CMake 3.12 (except on Android, where no transport is used). Let's keep it slim.

> The default installation method for third-party libraries in Windows is to use the vcpkg tool, which is a universal approach. Dependency cannot find the library installed by vcpkg, how can find these libs?

`SENTRY_TRANSPORT=curl` on Windows is not supported. Please also consider that the dependency was found, but the `AsynchDNS` feature wasn't.

Strive-Sun: @supervacuus I updated this pr, add `zlib.h` in `sentry_transport.c` and add `SENTRY_TRANSPORT_COMPRESSION`  in CMakeLists. It should be noted that, for compilation under Windows, setting `CMAKE_TOOLCHAIN_FILE` is required if the `zlib` library installed via vcpkg is needed.

src/sentry_transport.h
Strive-Sun: Yes, I has updated `sentry_gzipped_with_compression` to `sentry__gzipped_with_compression.`

src/sentry_transport.c
Strive-Sun: Yes, if enabled compress,  original body need free.

src/sentry_transport.c
```diff
@@ -165,9 +167,24 @@ sentry__prepare_http_request(sentry_envelope_t *envelope,
...
+    char *compressed_body = NULL;
+    size_t compressed_body_len = 0;
+    bool compressed = sentry__gzipped_with_compression(
+        body, body_len, &compressed_body, &compressed_body_len);
+    if (compressed) {
+        if (body_owned) {
+            sentry_free(body);
+            body = NULL;
+            body_owned = false;
+        }
```
Swatinem, line 179: I rather mean this:
```suggestion
        if (body_owned) {
            sentry_free(body);
        }
        body = compressed_body;
        body_len = compressed_body_len;
        body_owned = true;
```

That way, you do not need to duplicate all the code below depending on `if (compressed)`, as everything will just work the same way with `body`.

Strive-Sun:
There is a new question, if diable `SENTRY_TRANSPORT_COMPRESSION`, `sentry_transport. c` should not complie failed on due to missing `zlib.lib`. Is there a macro that does not compile `sentry_gzipped_with_compression` function.

supervacuus:
> There is a new question, if diable `SENTRY_TRANSPORT_COMPRESSION`, `sentry_transport. c` will complie failed on due to missing `zlib.lib`. Is there a macro that does not compile `sentry_gzipped_with_compression` function and `zlib.h`.

It would be best if you created a compile definition in the `CMakeLists.txt` and `#ifdef` that definition in the transport module. Specifically, you must include the `zlib` header and your entire compression code only when that definition exists. Otherwise, users without `zlib` developer packages will have failed builds.

Given that `sentry__gzipped_with_compression` is only called from within `sentry_transport. c`, there is no need to expose it in a header. Please,

* remove the header declaration
* remove the `sentry__` prefix from the function name and
* move the function before its first use and
* make it static

Thx!

src/sentry_transport.c
Swatinem: It does, as the `compressed_body` is definitely owned and needs to be freed once you are re-assigning `body` here.

Strive-Sun:
> > There is a new question, if diable `SENTRY_TRANSPORT_COMPRESSION`, `sentry_transport. c` will complie failed on due to missing `zlib.lib`. Is there a macro that does not compile `sentry_gzipped_with_compression` function and `zlib.h`.
> 
> It would be best if you created a compile definition in the `CMakeLists.txt` and `#ifdef` that definition in the transport module. Specifically, you must include the `zlib` header and your entire compression code only when that definition exists. Otherwise, users without `zlib` developer packages will have failed builds.
> 
> Given that `sentry__gzipped_with_compression` is only called from within `sentry_transport.c`, exposing it in a header is unnecessary. Please,
> 
> * remove the header declaration,
> * remove the `sentry__` prefix from the function name,
> * move the function before its first use and
> * make the function static
> 
> Thx!

Great idea, please help me review again. Thx a lot.

src/sentry_transport.c
Strive-Sun: Yes. I had modified. Thx

supervacuus:
This is a quick heads-up, @Strive-Sun: the error in the MinGW build is unrelated to this PR (the fact that it is `zlib` is a coincidence). I'll find out what is going on and continue my review tomorrow. Thx.

supervacuus:
The CMake failure to find zlib in the MinGW build is due to a change in the GHA runner image. In previous builds, zlib was found in the StrawberryPerl installation 😅 . 

I will provide a fix for this, which the open PRs can rebase on, next week.

CMakeLists.txt
```diff
@@ -291,6 +293,15 @@ if(SENTRY_TRANSPORT_CURL)
 	endif()
 endif()
 
+if(SENTRY_TRANSPORT_COMPRESSION)
+	if(NOT ZLIB_FOUND)
+		find_package(ZLIB REQUIRED)
+	endif()
+
```
supervacuus, line 300: As suggested here: https://github.com/getsentry/sentry-native/pull/954#discussion_r1502525718, you will need to inform the crashpad build to use the system zlib, or else we'll have two different zlib versions linked.

Please add:

```
set(CRASHPAD_ZLIB_SYSTEM ON CACHE BOOL "Force CRASHPAD_ZLIB_SYSTEM when enabling transport compression" FORCE)
```

This will further require a change in the crashpad build where the `find_package(ZLIB REQUIRED)` will also have to wrapped with a check for `ZLIB_FOUND`.

supervacuus (changes requested):
A few minor changes are still necessary. Thanks for your work so far.

Integration tests that validate the setting on supported platforms would be great, but I can add those after merging this PR.

CMakeLists.txt
Strive-Sun: Thank you for your careful review and helpful suggestions. I fixed them.

src/sentry_transport.c
```diff
@@ -5,9 +5,18 @@
 #include "sentry_ratelimiter.h"
 #include "sentry_string.h"
 
-#define ENVELOPE_MIME "application/x-sentry-envelope"
```
supervacuus, line 8: I think this removal was unintended.

Strive-Sun: fixed. `:sad:`

Strive-Sun:
Hi, I have rebased #964. Please help review again, thx a lot

src/sentry_transport.c
```diff
@@ -5,9 +5,14 @@
...
+#ifdef SENTRY_GZIPPED_COMPRESSION
+#  include "zlib.h"
+#endif
+
 #define ENVELOPE_MIME "application/x-sentry-envelope"
-// The headers we use are: `x-sentry-auth`, `content-type`, `content-length`
-#define MAX_HTTP_HEADERS 3
+// The headers we use are: `x-sentry-auth`, `content-type`, `content-encoding`,
+// `content-length`
+#define MAX_HTTP_HEADERS 4
```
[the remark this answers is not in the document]
supervacuus: Was there a specific reason not to apply this change (I am also not 100% sure, but I think it makes sense to limit the header count according to actual use).

CMakeLists.txt
```diff
@@ -291,6 +293,19 @@ if(SENTRY_TRANSPORT_CURL)
...
+	if(SENTRY_BACKEND_CRASHPAD)
+		set(CRASHPAD_ZLIB_SYSTEM ON CACHE BOOL "Force CRASHPAD_ZLIB_SYSTEM when enabling transport compression" FORCE)
```
supervacuus, line 302: If we set this here, we need to ensure that `find_package()` in the crashpad `CMakeLists.txt` is only invoked if `NOT ZLIB_FOUND` as we do above:

https://github.com/getsentry/crashpad/blob/58d7207c3c5fac96bbbca8e760824714ec3f8a29/CMakeLists.txt#L25

supervacuus: If you are unsure how to apply this change via submodules and separate PR to the crashpad fork, I can also do this.

supervacuus (changes requested):
We can soon merge this. Only minor additions, please

* add a change-log 
* wrap `find_package()` in the crashpad build
* apply the suggestions with `MAX_HTTP_HEADERS` (or tell me why that is a bad idea 😸)
* please rebase once more because yet another GHA runner image update introduced an ASAN issue (#965).

Thx!

I will take care of an integration test in a follow-up PR, because it might also require a few changes to the CI config.

src/sentry_transport.c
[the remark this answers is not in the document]
Strive-Sun: Sorry, I don't understand what you meant. In fact, I have modified it to determine whether to `#define SENTRY_TRANSPORT_COMPRESSION` to use `MAX_HTTP_HEADERS 4` or `MAX_HTTP_HEADERS 3`.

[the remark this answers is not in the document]
supervacuus: Sorry, my fault. I must have been looking at an old diff.

CMakeLists.txt
Strive-Sun: Yes, I am not familiar with the zip dependency relationship between `crashpad` submodules and `sentry-native`, so I removed this line(`set(CRASHPAD_ZLIB_SYSTEM...`)

supervacuus: Okay, no worries, I can take care of it.

supervacuus (approved):

supervacuus:
Thanks a lot for your contribution @Strive-Sun!

Strive-Sun:
> Thanks a lot for your contribution @Strive-Sun!

My pleasure. Without your suggestions and help, I wouldn't have been able to complete it smoothly.:smile:

PR #963, opened 2024-03-07
feat: add linux distributions to os context

### LINKED ISSUE #943: Add Linux distro meta-data to OS context.
### Discussed in https://github.com/getsentry/sentry-native/discussions/942

<div type='discussions-op-text'>

<sup>Originally posted by **tkatsoulas** February  1, 2024</sup>
Hello guys, 

The current implementation to understand the host OS https://github.com/getsentry/sentry-native/blob/master/src/sentry_os.c provides poor information for any distro. Are there any plans to update them? For instance `/etc/os-releases` file is there for most* of the Linux distros and can provide detailed information about a host. The current mechanisms can be used as a fallback. 

Thanks in advance, 
Tasos.</div>

CC: @kahest would be interesting if other SDKs already have a tag reserved for this information.

supervacuus:
Fixes #943.

This adds Linux distribution meta-data to the os context as follows:

Screenshot 2024-03-07 at 14 55 00

Only `name` and `version` should be indexed.

Further steps required before we merge/release this:
- [ ] update [event payload docs for OS context](https://develop.sentry.dev/sdk/event-payloads/contexts/#os-context)
- [ ] add definitions in relay: https://github.com/getsentry/relay/blob/master/relay-event-schema/src/protocol/contexts/os.rs
- [ ] add indexing support for `name` and `version` in [sentry-frontend](https://github.com/getsentry/sentry/pull/44157/files) and [-backend](https://github.com/getsentry/sentry/pull/43964/files)


tests/fixtures/os_releases/CODE_OF_CONDUCT.md
supervacuus, line 1: not sure if we should keep this

tests/fixtures/os_releases/README.md
supervacuus, line 1: Should I add a line to where I got this from?

src/sentry_os.c
```diff
@@ -195,8 +195,105 @@ sentry__get_os_context(void)
...
+
+        // Handle any partial line left at the end of the buffer
+        if (line < buffer + bytes_read) {
+            buffer_rest = buffer + bytes_read - line;
+            memmove(buffer, line, buffer_rest);
+            line = buffer;
+        }
+    }
+
+    if (bytes_read == -1) {
```
supervacuus, line 286: I would love to cover this, and I implemented it locally with a minimal FUSE setup that allows opening a file but returns an error on any read. I just wonder whether this PR is where we'll add it to the test setup.

supervacuus:
The failing Mingw test is unrelated to the change (a change in the GHA runner image requires us to install `zlib` for MinGW).

supervacuus:
@markushi, @Swatinem, can one of you give this a thorough look?

src/sentry_os.c
```diff
@@ -195,8 +195,105 @@ sentry__get_os_context(void)
...
+    key[key_length] = 0;
+
+    unsigned long value_length = strlen(equals + 1);
+    strncpy(value, equals + 1, value_length);
+    value[value_length] = 0;
+
+    // some values are wrapped in double quotes
+    if (value[0] == '\"') {
+        value[value_length - 1] = 0;
+        memmove(value, value + 1, value_length);
```
markushi, line 224: Shouldn't this be `value_length-2`? As we want to ignore the quotation marks at beginning + end?

tests/unit/test_os_release.c
```diff
@@ -0,0 +1,194 @@
...
+    { .name = "fedora", .version = "36" },
+    { .name = "endeavouros", .version = "" },
+    { .name = "virtuozzo", .version = "7" },
+    { .name = "opensuse", .version = "42.3" },
+    { .name = "fedora", .version = "31" },
+    { .name = "pop", .version = "22.04" },
+    { .name = "sled", .version = "12.3" },
+};
+
+const int num_test_dists = sizeof(test_dists) / sizeof(struct distro);
```
markushi, line 106: nit: this could be of type `size_t`

markushi (approved):
Left a few minor nits, looking good!

src/sentry_os.c
Swatinem: instead of operating on `value` and using a `memmove`, it might be a good idea to detect this before calling into `strncpy`, and adjust the slice range then.

src/sentry_os.c
```diff
@@ -195,8 +195,105 @@ sentry__get_os_context(void)
 }
 #elif defined(SENTRY_PLATFORM_UNIX)
 
+#    include <fcntl.h>
 #    include <sys/utsname.h>
 
+#    if defined(SENTRY_PLATFORM_LINUX)
+#        define OS_RELEASE_MAX_LINE_SIZE 128
+#        define OS_RELEASE_MAX_KEY_SIZE 64
```
Swatinem, line 203: I don’t think you are ever bounds checking for this?

tests/fixtures/os_releases/README.md
markushi: Yes, just a line on the top of the file should be enough. We have some guidelines over [here](https://github.com/getsentry/team-ospo/issues/120#issuecomment-1479900824), [example here](https://github.com/getsentry/sentry/blob/master/src/sentry/db/models/fields/jsonfield.py#L2).

src/sentry_os.c
supervacuus: We could probably change the interface to separate `open()` from `read()`, pass an `fd` instead of a path, and modify the `fd` in between. But that would move the goalposts in non-public interfaces. 

In C unit tests, error handling is a significant part of the uncovered paths, and introducing a heavier approach should help cover most of the related test cases in the codebase. So, whatever approach is taken, this will happen in a different PR 😅.

supervacuus:
again, failing test here is due to: https://github.com/getsentry/sentry-native/pull/968

supervacuus:
PR is on hold until the context format is clarified. Nested context formats with three layers are unacceptable for the backend to search, so we either change the format in the SDK or map it locally in the sentry backend. 

Since the format was at the core of how to add this feature from the start, it is best to change the format in the SDK (and with it, in the develop-docs, relay, and sentry backend).

This is currently deprioritized.

JoshuaMoelans (approved):
LGTM 🚀

PR #964, opened 2024-03-11
fix: build zlib for mingw ci

supervacuus:
The `mingw` CI currently breaks, because it relied on a `zlib` packaged with `StrawberryPerl` which can no longer be found by `CMake`. It would generally be a good idea that our builds are independent of GHA image packages (except if there is a `zlib` package, of course).

#skip-changelog


supervacuus:
This currently blocks two other PRs, #963 and #954, which can be rebased after we merge this.

markushi (approved):
LGTM!

PR #965, opened 2024-03-15
fix: failing clang-asan/llvm-cov tests

supervacuus:
After fixing the failing mingw build in #964, we have another failing build caused by an update to the runner image. We can't connect to the pytest-httpserver anymore, but seemingly only in the clang-asan/llvm-cov build.

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
             CC: clang-15
             CXX: clang++-15
             ERROR_ON_WARNINGS: 1
-            RUN_ANALYZER: asan
+            RUN_ANALYZER: llvm-cov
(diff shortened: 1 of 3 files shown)


Swatinem (approved):

supervacuus:
Hmm... it seems we get "connection reset by peer" when trying to receive via the Unix socket when starting the `crashpad_handler`,  not the connection to the test HTTP server. Maybe the `crashpad_handler` from the previous run still runs in some cases?

supervacuus:
So, disabling the `AddressSanitizer` in the build prevents test failures...

supervacuus:
Okay, that made things much easier to find: https://github.com/actions/runner-images/issues/9491

Swatinem (approved):

Issue #885, opened 2023-09-14
Provide user-feedback capability to the Native SDK


supervacuus:
The wish to provide some access to this capability was recently raised. This is not necessarily meant as an enhancement of the code but could also be documentation that describes how to use existing Native SDK API and sentry endpoints to accomplish this on the client side.

A user was [reaching out](https://github.com/getsentry/sentry-native/issues/375) a couple of years ago but closed the issue, and we have yet to provide a solution.

Following the [SDK development docs](https://develop.sentry.dev/sdk/features/#user-feedback) and considering the level of application-level support the Native SDK can provide without introducing a lot of dependencies, we could provide something like an `onCrashedLastRun` hook that provides enough context so that users create their own feedback UI and send a request to the [user feedback endpoint](https://docs.sentry.io/api/projects/submit-user-feedback/).

cc: @kahest

sapphonie:
+1 to this, really want to avoid having to make janky json POST calls in a crash handler 😄

bruno-garcia:
Also note that recently @michellewzhang added in-product onboarding wizard for all platforms we had support at that time.

For Native, we're showing this:
<img width="585" alt="image" src="https://github.com/getsentry/sentry-native/assets/1633368/afe8a42a-a844-44a9-8bbb-e905f0e44871">

[Here are some example PRs](https://github.com/getsentry/sentry/pulls?q=is%3Apr+sort%3Aupdated-desc+feedback+onboarding+is%3Aclosed+author%3Amichellewzhang) from Michelle if you folks want to add it to Sentry

kahest:
Note that https://github.com/getsentry/sentry-native/pull/966 covers an API to prepare and send User Feedback. It does not, however, provide functionality and hooks to retrieve details about previous crashed runs, like `onCrashedLastRun` suggested above.

supervacuus:
While #966 covers some ground, we shouldn't close this issue yet.

bruno-garcia:
Also, note that this PR only added the ability to capture a user feedback that's tied to an event:
* https://github.com/getsentry/sentry-native/pull/966/

It relies on our 'old' user feedback format, that is, by itself not an event. It requires an `event_id` and that event must make it into Sentry for the feedback to show up. Otherwise it's just lost:

https://github.com/getsentry/sentry-native/blob/b555a978986f882abb3161bd1c848339f9a2cc1a/src/sentry_envelope.c#L318

[...]

PR #966, opened 2024-03-19
Add user feedback capability to the Native SDK

tustanivsky:
This PR introduces a new API allowing to create/capture user feedback for specific events (similar functionality is offered by native [Android](https://docs.sentry.io/platforms/android/user-feedback/) and [Apple](https://docs.sentry.io/platforms/apple/user-feedback/) SDKs).

It will be later exposed to the [Unreal Engine plugin](https://github.com/getsentry/sentry-unreal) where user feedback capabilities have been requested recently. While the existing [user feedback endpoint](https://docs.sentry.io/product/user-feedback/) provides an alternative it requires deviating from our current approach of utilizing the native SDKs for client app communication with Sentry there.

Related to #885


supervacuus (changes requested):
This looks great! Thanks a lot! One minor change.

I think this should also find its way into the docs.

tustanivsky:
@supervacuus Thanks for the prompt feedback! I've updated the integration tests as you suggested and opened a corresponding PR for the docs.

supervacuus (approved):

PR #969, opened 2024-03-23
test: integration-tests for transport compression

supervacuus:
This is a follow-up to #954.

* ~there might still be failing tests related to the fix here: https://github.com/getsentry/sentry-native/pull/968~
* ~I think we can merge the matrix entries for Linux and macOS because the non-gzip-encoded path is already covered, but for Windows it might make sense to have separate ones.~
* ~I only tried it with the `minGW` build first, because there is no `zlib` available on Windows by default (we need to either build it, like we do for the `minGW` build or we use `vcpkg`).~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,13 @@ jobs:
             CXX: clang++-15
             ERROR_ON_WARNINGS: 1
             RUN_ANALYZER: asan,llvm-cov
+          - name: Linux (clang + transport-compression + asan + llvm-cov)
+            os: ubuntu-22.04
+            CC: clang-15
+            CXX: clang++-15
+            ERROR_ON_WARNINGS: 1
+            TRANSPORT_COMPRESSION: 1
+            RUN_ANALYZER: asan,llvm-cov
           - name: Linux (clang + kcov)
             os: ubuntu-22.04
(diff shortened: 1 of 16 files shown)


Swatinem (commented):
instead of having separate CI tasks for this, why not add this as variants within pytest?

supervacuus:
> instead of having separate CI tasks for this, why not add this as variants within pytest?

Because it is solely a build parameter and parameterizing every cmake invocation in every test and each test itself would be a hassle. If `pytest` has a sensible abstraction to automate the parameterization, I am more than happy to integrate it. 

In the current configuration, I can just enable it for all tests, which are all oblivious to the encoding (as they should be). Using `@pytest.mark.parametrize` (or something similar) would allow us to use the same test run, but we'd need to enable it per test even though the test doesn't change otherwise.

Swatinem:
I don’t think we really want to parameterize *all* the tests, it should suffice to just do it for one or two, and maybe even explicitly check that the payloads really *are* compressed.

Depending on whether we would want to enable this by default in the future (we probably should), we can just enable this flag everywhere, and maybe have explicit tests (the unit test suite is a good candidate) to disable this flag and test without compression.

Swatinem (approved):

PR #979, opened 2024-04-11
Add ability to disable page allocator

azat:
This is required to avoid leaks, since free() with page allocator is noop.


azat:
Mislooked, it will not leak, only after SIGSEGV/TRAP/ILL/ABRT/BUS/FPE, and this is OK

supervacuus:
Yes, @azat, the page allocator is intentionally leaky because, at the point it is used, the program is already in a crashed state. This allows us to use functionality in the SDK that requires heap allocation but which can't use system `malloc()`, which isn't async-signal safe. You might already understand this, but I keep this here for others who reach this PR.

PR #982, opened 2024-05-02
fix: correctly support stack overflows across platforms

### LINKED ISSUE #977: Stack overflow not sent to backend
**Description**
We have discovered that crashes due to stack overflow are not uploaded to Sentry. The application silently crash without any additional details reported.

- [ ] During build
- [x] During run-time

**Environment**
- OS: Windows 10 and later, 32-bit
- Compiler: MSVC2022
- vcpkg, 0.7.2, also applies to older versions

**Steps To Reproduce**
Generate stack overflow by allocating all available stack and then calling into any function.

**Log output**
None.

**Mitigations**
After discovering this issue we applied several mitigations:
- Increased stack size to reduce probability
- Reserve one extra stack page for recovery `SetThreadStackGuarantee`
- Inject SEH handler that calls `_resetstkoflw` to avoid subsequent crash in error handling

However, I would very much prefer that this situations was handled also with "naive" integration.

### LINKED ISSUE #987: Only set up sigaltstack if there is none (Linux + Android)
Fixing this was a side-effect of https://github.com/getsentry/sentry-native/pull/982

supervacuus:
Currently, stack overflows on Windows die somewhere in the handler because we run out of stack space. This is the first attempt to fix the situation by reserving stack space before it is exhausted. This is an analog to `sigaltstack()` on Linux, but Windows only provides a "reserved amount" interface, and with that, you have fewer options to shoot yourself in the foot.

Fixes #987.
Fixes #977.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 **Fixes**:
 
 - Allow `crashpad` to run under [Epic's Anti-Cheat Client](https://dev.epicgames.com/docs/game-services/anti-cheat/using-anti-cheat#external-crash-dumpers) by deferring the full `crashpad_handler` access rights to the client application until a crash occurred. ([#980](https://github.com/getsentry/sentry-native/pull/980), [crashpad#99](https://github.com/getsentry/crashpad/pull/99))
-- Reserve enough stack-space on Windows for our handler to run when stack is exhausted from stack-overflow. ([#982](https://github.com/getsentry/sentry-native/pull/982))
+- Reserve enough stack space on Windows for our handler to run when the stack is exhausted from stack-overflow. ([#982](https://github.com/getsentry/sentry-native/pull/982))
+- Only configure a `sigaltstack` in `inproc` if no previous configuration exists on Linux and Android. ([#982](https://github.com/getsentry/sentry-native/pull/982))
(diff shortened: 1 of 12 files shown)


supervacuus:
> The macOS tests fail because AppleClang 13.0.0.13000029 seemingly ignores -fno-optimize-sibling-calls. However, locally with 15.0.0.15000309, the tests run successfully. I will add a corresponding @pytest.mark.skipif until we update our macOS runner images (which is a bigger change than I am willing to introduce here).

After rebasing `master` to fix the GitHub-runner issue, the same problem is apparent with `clang` from the older NDK.

supervacuus:
This has been quite a revealing exercise. That a runtime approach to the `sigaltstack` fixed the "Old NDK"-build means that the bionic-libc implementations differ between the NDK versions regarding whether a `sigaltstack` was installed for each thread. This doesn't necessarily mean this will also be true when these threads are started through the ART. Still, it is noteworthy, especially given the recent influx of issues with older devices.

This started as a Windows fix, but since I tried to make the stack overflow tests run on all builds, we also have improvements on Linux and Android. So, renaming the PR and additional change-log entries might be in order.

cc: @kahest

supervacuus:
Fixes #987 and #977.

supervacuus:
Fixes #987 and #977.

Swatinem (approved):
nice! 🎉

Issue #983, opened 2024-05-02
sentry_transaction_set_data() - Discarded unknown attribute


supervacuus:
### Discussed in https://github.com/getsentry/sentry-native/discussions/981

<div type='discussions-op-text'>

<sup>Originally posted by **AntonTenstar** April 25, 2024</sup>
Hello trying to send custom data to with a transaction, but guess I'm doing something wrong. Anyone got any ideas?

Code:
```C
sentry_transaction_context_t* tx_ctx = sentry_transaction_context_new("test transaction", "performance");
sentry_transaction_t* tx = sentry_transaction_start(tx_ctx, sentry_value_new_null());
sentry_transaction_set_data(tx, "performance", sentry_value_new_int32(1337));
sentry_transaction_finish(tx);
```

Result:
![image](https://github.com/getsentry/sentry-native/assets/90307454/39f87f95-e98d-4b45-a8b7-c4947e22f0c3)

</div>

PR #986, opened 2024-05-07
fix: store transaction `data` in event `extra`

supervacuus:
fixes https://github.com/getsentry/sentry-native/issues/983

This aligns with the current implementations in sentry-java, sentry-dotnet, and sentry-cocoa (and probably others). The [RFC 0101](https://github.com/getsentry/rfcs/blob/main/text/0101-revamping-the-sdk-performance-api.md) mentions the issue as one of its drivers. In the meantime we should allow transactions to serialize `data` where it doesn't get discarded.

### DIFF
--- a/src/sentry_tracing.c
+++ b/src/sentry_tracing.c
@@ -533,14 +533,16 @@ sentry_transaction_set_data(
     }
 }
 
+static const char txn_data_key[] = "extra";
+static const size_t txn_data_key_len = sizeof(txn_data_key) - 1;
+
 void
 sentry_transaction_set_data_n(sentry_transaction_t *tx, const char *key,
     size_t key_len, sentry_value_t value)
 {
     if (tx) {
-        const char data_key[] = "extra";
-        const size_t data_key_len = sizeof(data_key) - 1;
-        set_data(tx->inner, data_key, data_key_len, key, key_len, value);
+        set_data(
+            tx->inner, txn_data_key, txn_data_key_len, key, key_len, value);
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/sentry_tracing.c
```diff
@@ -538,50 +528,51 @@ void
 sentry_transaction_set_data(
     sentry_transaction_t *tx, const char *key, sentry_value_t value)
 {
-    if (tx) {
-        set_data(tx->inner, key, value);
+    if (key) {
```
kahest, line 531: nit: took me a bit to understand why we check `key` here and `tx`/`span` below. in the end I wondered why we don't just check both in `set_data`

markushi (approved):
LGTM!

src/sentry_tracing.c
supervacuus: I keep all checks close to where the dereferences happen. This way, the contract between a check and dereference doesn't have to move with the usage of a particular function down the road.

For instance, many outer API functions pass on `NULL` pointers because they can be valid downstream. If I move the `txn`/`span` check to the outer layer, then any function that uses `set_data()` must know that it also needs to check `txn`/`span`, because `set_data()` would otherwise crash. There is no way to express this in C besides writing documentation, but then I duplicate the check in the outer layers and the comments.

If you mean why we don't move the `key` check into `set_data()`: because `strlen()` in the outer layer would crash if we passed it a `NULL` pointer.

If you ask why I chose to move the ptr/len interface towards the outer layer, the answer is that the previous implementation duplicated the code between C-strings and ptr/len interfaces (instead of implementing one in terms of the other), which would lead to duplication of the configurable keys for the `data' storage.

kahest: gotcha, thanks for the explanation!

PR #996, opened 2024-05-28
Use static STL, hide it from prefab

markushi:
With the JNI move we switched from using the static stdlib to `c++_shared`.  This had the side effect of `libc++_shared.so` being integrated into the build artifact. As the version of `libc++_shared.so` is tied to the version of the NDK, this could cause troubles / undefined behavior if customer is using a different NDK version than us.

image


supervacuus (changes requested):
LGTM, but please, with the changelog. Thx!

supervacuus (approved):

Issue #1002, opened 2024-06-04
Request: Merge getsentry/crashpad PR #101 to fix llvm-mingw arm64 crashpad builds


past-due:
**Description**

llvm-mingw builds targeting arm64 did not succeed if crashpad was the selected backend

**Solution**

Crashpad fix PR: https://github.com/getsentry/crashpad/pull/101

(Once merged, the crashpad submodule in this repo will need to be updated to point the latest commits.)
```[tasklist]
### Tasks
- [x] Merge https://github.com/getsentry/crashpad/pull/101
- [ ] Merge https://github.com/getsentry/sentry-native/pull/1003 (to update submodule)
```

past-due:
Thanks @supervacuus!

Would it be possible to issue a new sentry-native release with this included?

supervacuus:
> Would it be possible to issue a new sentry-native release with this included?

I will release a new version in the first half of next week.

supervacuus:
@past-due, [0.7.6](https://github.com/getsentry/sentry-native/releases/tag/0.7.6) is out.

PR #1003, opened 2024-06-04
fix: crashpad build for Windows ARM64 via LLVM-MINGW

past-due:
Fixes: #1002.

Actual change: [getsentry/crashpad#/101](https://github.com/getsentry/crashpad/pull/101).


supervacuus (approved):
LGTM, thanks!

PR #1025, opened 2024-08-01
ci: introduce codecov token

### LINKED ISSUE #1023: codecov requires a token
I recently saw that Codecov no longer produces coverage reports in our PRs.

It looks like our token is no longer valid (or just missing?):

https://github.com/getsentry/sentry-native/actions/runs/10166682391/job/28117412853#step:14:47

Since Codecov often did not show a coverage report for reasons, I wasn't immediately aware that this was an authentication issue. I don't know how to fix this since I have no idea how the token setup works (and also have no access because that is probably a secret).

@kahest, @Swatinem, do you have any idea what can be done? 
Were other repos affected by this? 
Is the token a repo secret, or do we simply not have one (because I can't see a reference to a secret)? 
If we never had one, could it be that we didn't get a report when Codecov throttled (429) our uploads previously, and now they actually need one every time?
Can we get one?

supervacuus:
Additionally, bump `codecov-action` to 4.5.0

Fixes #1023.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,9 +191,12 @@ jobs:
 
       - name: "Upload to codecov.io"
         if: ${{ contains(env['RUN_ANALYZER'], 'cov') }}
-        uses: codecov/codecov-action@125fc84a9a348dbcf27191600683ec096ec9021c # pin@v4.4.1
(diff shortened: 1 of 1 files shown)


Swatinem (approved):

Swatinem:
Nice, looks like clang might report branch coverage by default now 🎉

supervacuus:
> Nice, looks like clang might report branch coverage by default now 🎉

Sometimes it can be so easy.

PR #1028, opened 2024-08-05
feat: support 16kb page size on Android 15

### LINKED ISSUE #989: [Android] Support dynamic page size for Android 15+ devices
See https://developer.android.com/about/versions/15/behavior-changes-all#16-kb for more details.

> Beginning with Android 15, Android supports devices that are configured to use a page size of 16 KB (16 KB devices). If your app uses any NDK libraries, either directly or indirectly through an SDK, then you will need to rebuild your app for it to work on these 16 KB devices.

supervacuus:
This fixes #989.

I tried to reach the most minimal version I could successfully deploy to a 16kb page-size image on the emulator.

Important here:
* NDK 27 is required because the toolchain dependencies must also be compiled for 16kb page sizes. That's why I explicitly configured this
* appended new linker settings to all CMake library targets
* enable legacy packaging because we want to stay compatible with `AGP <= 8.2`: https://developer.android.com/guide/practices/page-sizes#agp_version_82_or_lower

Open topics:

- [x]  This should be tested with the NDK sample app by at least another Android developer in the team
- [x] Manual check if there are any remaining hardcoded 4096 page-size assumptions
- [x] How (and when) to integrate this in `sentry-android` since it will also have to be added there for version < 8.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Features**:
 
 - Let the envelope serialization stream directly to the file. ([#1021](https://github.com/getsentry/sentry-native/pull/1021))
+- Support 16kb page sizes on Android 15 ([#1028](https://github.com/getsentry/sentry-native/pull/1028))
 
 ## 0.7.7
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -596,4 +596,9 @@ endif()
 if(SENTRY_BUILD_SHARED_LIBS)
 	target_link_libraries(sentry PRIVATE
 			"$<$<OR:$<PLATFORM_ID:Linux>,$<PLATFORM_ID:Android>>:-Wl,--build-id=sha1,--version-script=${PROJECT_SOURCE_DIR}/src/exports.map>")
+
+	# Support 16KB page sizes
+	target_link_libraries(sentry PRIVATE
+	    "$<$<PLATFORM_ID:Android>:-Wl,-z,max-page-size=16384>"
+	)
 endif()
--- a/ndk/lib/CMakeLists.txt
+++ b/ndk/lib/CMakeLists.txt
@@ -12,6 +12,7 @@ set(SENTRY_BUILD_SHARED_LIBS ON)
 add_subdirectory(${SENTRY_NATIVE_SRC} sentry_build)
 
 # Link to sentry-native
-target_link_libraries(sentry-android PRIVATE
-    $<BUILD_INTERFACE:sentry::sentry>
-)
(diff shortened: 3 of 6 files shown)


markushi:
@supervacuus thanks for taking a look into this!


> * This should be tested with the NDK sample app by at least another Android developer in the team
> * How (and when) to integrate this in `sentry-android` since it will also have to be added there for version < 8.

1) I manually tested the changes on an "Android 15 16KB page size" emulator, looking good, no more obvious crashes 
2) I manually tested the changes on non 16KB devices, just to ensure the build flags don't cause a regression, looking good as well!
3) The required changes for `sentry-java` 7.x are here: https://github.com/getsentry/sentry-java/pull/3620 I did some manual testing here as well, looks promising.

supervacuus:
I have also tested with a separate app, only initializing sentry-native on a 16kb image but using the local maven AAR instead of the sample APK. 

I think it should be enough to specify the NDK version in the sample build script because our packaged .so does not have 16kb-built runtime dependencies.

markushi (approved):
LGTM!

supervacuus:
> Manual check if there are any remaining hardcoded 4096 page-size assumptions

Context: https://developer.android.com/guide/practices/page-sizes#check-code

* libunwindstack-ndk uses `getpagesize()` throughout: 
  * https://github.com/getsentry/libunwindstack-ndk/blob/master/Memory.cpp#L105-L106
  * https://github.com/getsentry/libunwindstack-ndk/blob/master/Memory.cpp#L299-L300
* `mpack` hardcodes the page-size to `4096`, but we don't use it in the Android build and the way it is used, doesn't really lead to misalignment re page-sizes (but it could mean inefficient node layout).
* we [`mmap()` in the page allocator](https://github.com/getsentry/sentry-native/blob/master/src/sentry_unix_pageallocator.c#L57-L58), but don't use any page-aligned parameters and the page size is [retrieved from `getpagesize()`](https://github.com/getsentry/sentry-native/blob/master/src/sentry_unix_pageallocator.c#L45) for the internal allocation increments (not that it matters).
* the [module-loader mmaps](https://github.com/getsentry/sentry-native/blob/master/src/modulefinder/sentry_modulefinder_linux.c#L64) [each module](https://github.com/getsentry/sentry-native/blob/master/src/modulefinder/sentry_modulefinder_linux.c#L534), but also without using any of the parameters that must be page-aligned.
* there is no use of `PAGE_SIZE` or any other hard-coded `4096` page-size in the remaining code base

So, from my side, we're good if everything works on the `sentry-android` side.

PR #1044, opened 2024-09-27
Decrease SetThreadStackGuarantee value

stima:
Decrease guaranteed stack size to smaller value to be able execute chromium embedded task.

### DIFF
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -375,7 +375,7 @@ crashpad_backend_startup(
         }
     }
 
-#ifdef SENTRY_PLATFORM_WINDOWS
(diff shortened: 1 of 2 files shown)


supervacuus:
Hi @stima! Thanks for the PR. I understand what you want to do here, but I wonder if this is the right approach. How about we introduce a similar contract as with the `sigaltstack` for UNIX signal handling and only set an expected stack size if the client didn't specify one for the thread up to that point?

We can also provide a compile-time option to turn off the convenience per-thread setting via `DllMain` so that you can fully control your configuration without ours interfering. Lowering the stack size to 4 pages will likely break use cases with more involved crash hooks, which are often necessary for downstream SDKs.

stima:
I wanted to keep it simple (although increasing to 32k might be sufficient for CEF). However, my best guess is that only the client knows the size they are willing to provide, so autosetting disabling is best option.

Issue #1012, opened 2024-06-19
Ensure unsafe functions are properly guarded


markushi:
**Description**

We recently received a security report about unsafe usages. Whilst there's not always a safe alternative available (especially cross-platform) we can still have a check and ensure we have best-effort implementations for:
* input validation
* buffer management
* return values are in proper format (e.g. null terminated strings)

Detailed report:
### Format String Vulnerabilities
```
libsentry-android.so
  snprintf: 00001c92
libsentry.so
  fprintf: 000d4d98
  snprintf: 00074252
  vfprintf: 000d102a
  vprintf: 0008c4da
  vsnprintf: 0007914a
```

### Buffer Overflow Vulnerabilities
```
libsentry.so
  strncpy: 000d34ea
  memcpy: 000d40b0
```

### Buffer Overflow Danger Functions
```
libsentry.so
  realpath: 0007cd4e
```

markushi:
We should also consider integrating CodeQL into CI to highlight potential issues earlier.

JoshuaMoelans:
Tracking progress on [this Notion page](https://www.notion.so/sentry/1012-Ensure-unsafe-functions-are-properly-guarded-1048b10e4b5d80b8b5d8f073fd0d6f96?pvs=4) (+ subpages)

JoshuaMoelans:
Investigation led to 2 minor fixes (https://github.com/getsentry/sentry-native/pull/1046 and https://github.com/getsentry/sentry-native/pull/1047). All other occurrences of 'unsafe' functions were already properly guarded, or handled static data that made their uses safe.

PR #1047, opened 2024-10-04
fix: adjusted memcpy size param to match remaining destination size

JoshuaMoelans:
As part of an investigation into unsafe functions (https://github.com/getsentry/sentry-native/issues/1012) a slight mismatch between buffer size and copy amount was found for a call to `memcpy`. 

This PR fixes this by making the copied amount match the to-copy parameter's size.

#skip-changelog


src/path/sentry_path_windows.c
```diff
@@ -164,14 +164,14 @@ sentry__path_join_wstr(const sentry_path_t *base, const wchar_t *other)
...
-            sentry_path_t *rv = path_with_len(len);
+            size_t other_len = wcslen(other);
+            sentry_path_t *rv = path_with_len(other_len + 3);
             if (!rv) {
                 return NULL;
             }
             rv->path[0] = base->path[0];
             rv->path[1] = L':';
-            memcpy(rv->path + 2, other, sizeof(wchar_t) * len);
+            memcpy(rv->path + 2, other, sizeof(wchar_t) * other_len);
```
supervacuus, line 174: The problem is that you copy only `other_len` `wchat_t` over to `rv->path`, and those do not include the terminating `\0`.

Let's assume the following example:

```
other = "/Users"
base = "C:/"
```

In memory, if we assume each cell to be `wchar_t` wide, those (including the expected result) would look like this:

| Variable  | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 
| -------- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| `other`  | `/` | `U` | `s` | `e` | `r` | `s` | `\0` | ☠ | ☠ |
| `base`  | `C` | `:` | `/` | `\0` | ☠  | ☠ | ☠ | ☠ | ☠ |
| `rv->path`  | `C` | `:` | `/` | `U` | `s` | `e` | `r` | `s` | `\0` |

We cannot assume (from within `sentry__path_join_wstr`) that `other` or `base` have allocated memory beyond their `\0` (and yes, that assumes that we received a valid null-terminated string).

In the original code, `len = 6 + 3`, which is the correct size for `rv->path`, but `memcpy` would now access 2 `wchar_t` beyond the null termination of `other`. One immediate fix would be to subtract `2` from `len` in the `memcpy` invocation:

```C
memcpy(rv->path + 2, other, sizeof(wchar_t) * (len - 2));
```

In your fix, `other_len = 6`, `rv->path` is correctly pre-allocated to `other_len + 3`, but now you only copy the contents of the string but not the terminating `\0` from `other`. A fix here would be to add `1` to `other_len` in the `memcpy` invocation or to explicitly set `rv->path[8] = L'\0';` after the copy.

JoshuaMoelans: Thanks for the visualization! Updated to manually set the null terminator in https://github.com/getsentry/sentry-native/pull/1047/commits/ea858f6872d0c4bd46665c0a2bb04b70164be639 .

supervacuus (approved):

PR #1049, opened 2024-10-08
feat: build-parameter for handler stack size

supervacuus:
This allows users to change the currently hard-coded handler stack size during the build.

Summary:

* The default value is 64 KiB, equivalent to our current hard-coded value. This means for most users, everything stays the same.
* For Windows, the following additional changes were introduced:
  * similar to the `sigaltstack` setup on the signal-handler side, we only configure the thread stack if no one else has done that before us.
  * `sentry__reserve_thread_stack()` is only called from the backend initialization for a static build. Dynamic libraries install a `DllMain` that sets the thread stack for each thread starting after our library is loaded.

I could imagine providing two more build parameters:
* a CMake option to turn on/off the check for a previous stack configuration.
* a CMake option to turn on/off the automatic configuration of each thread's crash stack (i.e., the `DllMain`).

For the sake of simplicity, I would like to defer this decision until we have a case where someone needs to configure the stack size and can not use these mechanisms.

Beyond the configurable stack size, most users prefer not to configure any of these subtle configurations themselves. Introducing any of these build parameters later can be done non-breakingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,15 +2,23 @@
 
 ## Unreleased
 
+**Features**:
+
+- Provide a build-parameter `SENTRY_HANDLER_STACK_SIZE` that allows to change the targeted stack-size reserved for the crash handler on Linux and Windows. ([#1049](https://github.com/getsentry/sentry-native/pull/1049))
+
 **Fixes**:
 
-- Reject invalid trace- and span-ids in context update from header ([#1046](https://github.com/getsentry/sentry-native/pull/1046))
+- Reject invalid trace- and span-ids in context update from header. ([#1046](https://github.com/getsentry/sentry-native/pull/1046))
 
 ## 0.7.10
 
 **Fixes**:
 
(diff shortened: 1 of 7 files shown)


supervacuus:
Hi @stima, would this be enough for you as an alternative to #1044? I thought about this again and would like to defer adding additional build parameters. With this approach, you can decrease the stack size without affecting other usage. If it works for you, I will include it in the next release.

stima:
@supervacuus I don't think it's feasible to find a one-size-fits-all solution from an SDK perspective. Having the option to configure is better than not having it at all. However, in our specific case, we aren't aware of the stack size requirements for the third-party libraries we use, so we'll completely disable auto-attaching and set a guaranteed stack size for the threads we manage. Probably adding one more compilation flag (autoset: on/off) and some wrapper API around stack-size setting will provide more flexability.

supervacuus:
closing in favor of #1196

PR #1051, opened 2024-10-11
fix: look up `GetSystemTime()` implementations at runtime.

supervacuus:
This is a fix for the issue raised here: https://github.com/getsentry/sentry-native/pull/1039#issuecomment-2398218262

The problem is that the previous build-time condition only checks if the SDK supports the new API, but the application could later run on a Windows without support.

We look up  `GetSystemTimePreciseAsFileTime()` and runtime (during `sentry_init`, at which point it is cached) and fall back to `GetSystemTimeAsFileTime()` if we can't find it.

@dacap and @ShawnCZek, please look to see if that works for you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Fixes**:
 
 - Reject invalid trace- and span-ids in context update from header ([#1046](https://github.com/getsentry/sentry-native/pull/1046))
+- Lookup `GetSystemTimePreciseAsFileTime()` at runtime and fall back to `GetSystemTimeAsFileTime()` to allow running on Windows < 8. ([#1051](https://github.com/getsentry/sentry-native/pull/1051))
 
 ## 0.7.10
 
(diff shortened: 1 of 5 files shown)


ShawnCZek (approved):
This works perfectly fine for me. Thank you for the quick action! I am just surprised there is an incentive to support such an old OS, especially when it is done via a slightly non-standard way of using a newer target Windows SDK version.

supervacuus:
> This works perfectly fine for me. Thank you for the quick action! 

Great!

> I am just surprised there is an incentive to support such an old OS

I don't particularly appreciate breaking changes (and Sentry agrees). If the effort of implementing and maintaining a fallback is within acceptable bounds and the 99 percentile users are not affected by it, then there is no reason to break.

dacap:
@supervacuus thanks for this patch, I've just tested 331f8fc91004a716af7a9136314420d8ffc8ea18 on Windows 7 and it worked 👍

JoshuaMoelans (approved):

Issue #1050, opened 2024-10-09
Allow capturing minidumps programatically


bruno-garcia:
A customer who has their own crash handler mechanism would like to upload minidumps they capture by themselves.
One of the solutions proposed was using HTTP and [uploading them to our `minidump` endpoint](https://docs.sentry.io/platforms/native/guides/minidumps/).

Since that means having to deal with retries, managing files on disk etc. I suggesting using one of our SDKs to manage that.

Our envelope format allows attachment, including type `minidump`. Which should trigger symbolication.

While we might not have a `capture_minidump` yet, the suggestion is to create this function so the SDK takes over the job of establishing the connection to Sentry. And deleting the file upon completion.

kahest:
@bruno-garcia you mention

> Since that means having to deal with retries, managing files on disk etc.

`sentry-native` currently doesn't do this, and there's no synchronous behaviour between client and transport. so if that's the expectation it would require additional effort than what is suggested here

PlasmaDev5:
Wrote up some initial notes on this
[HandleCustomMinidumps.pdf](https://github.com/user-attachments/files/17335867/HandleCustomMinidumps.pdf)

vaind:
> [<img alt="" width="16" height="16" src="https://avatars.githubusercontent.com/u/1633368?s=64&amp;u=856a9e52d4cdc92eac2ff9e1f97900df39ae5e4d&amp;v=4">@bruno-garcia](https://github.com/bruno-garcia) you mention
> 
> > Since that means having to deal with retries, managing files on disk etc.
> 
> `sentry-native` currently doesn't do this, and there's no synchronous behaviour between client and transport. so if that's the expectation it would require additional effort than what is suggested here

I don't think we need that - IMO, it would be fine to just process all minidumps at once, which moves them over to the envelope cache dir and transport handles retries. Also, we can probably set any large timeout to flush() so it runs until the server accepts everything.

vaind:
> A customer who has their own crash handler mechanism would like to upload minidumps they capture by themselves.

do they want to deal with sentry-native API, building an executable to run this? IDK but build a simple script that runs HTTP requests to our backend sounds a lot simpler.

PlasmaDev5:
> do they want to deal with sentry-native API, building an executable to run this? IDK but build a simple script that runs HTTP requests to our backend sounds a lot simpler.

[...]

PR #1052, opened 2024-10-11
feat: programatic minidump capture

PlasmaDev5:
Resolves: https://github.com/getsentry/sentry-native/issues/1050

### Summary
Provides a new function that allows sentry to capture user created minidumps.


include/sentry.h
```diff
@@ -1362,6 +1362,14 @@ SENTRY_API sentry_user_consent_t sentry_user_consent_get(void);
  */
 SENTRY_API sentry_uuid_t sentry_capture_event(sentry_value_t event);
 
+struct sentry_path_s;
+typedef struct sentry_path_s sentry_path_t;
```
vaind, line 1366: If we must expose this (can/should we just use string instead?), users would also need a way to create this opaque struct.

include/sentry.h
```diff
@@ -1362,6 +1362,14 @@ SENTRY_API sentry_user_consent_t sentry_user_consent_get(void);
...
+
+/**
+ *  This function allows for independently created minidumps to be captured.
```
vaind, line 1369: This could be simplified, e.g.
```suggestion
 *  Allows capturing independently created minidumps.
```

vaind (changes requested):
Thanks for the change. looks straightforard so far. 

Is this PR done and ready for review? because it's marked as such, I'll asume so (if not, you ca use draft PRs in the future so that reviewers don't get notifications yet).

CI is currently failing so that will need to be fixed. Also, the newly added API should have a new test case.

PlasmaDev5:
> Thanks for the change. looks straightforard so far.
> 
> Is this PR done and ready for review? because it's marked as such, I'll asume so (if not, you ca use draft PRs in the future so that reviewers don't get notifications yet).
> 
> CI is currently failing so that will need to be fixed. Also, the newly added API should have a new test case.

i guess somewhere between draft and final. I for sure not indented to be be merged yet but at the same time i need feedback and review before moving further.

include/sentry.h
```diff
@@ -1362,6 +1362,14 @@ SENTRY_API sentry_user_consent_t sentry_user_consent_get(void);
...
+ */
+SENTRY_API void sentry_capture_minidump(sentry_path_t *dump_path, sentry_value_t event, const sentry_options_t *options);
```
ShawnCZek, line 1371: The usage of `sentry_options_t` in the public interface after the backend's initialization seems unusual. For instance, the SDK for Unreal Engine does not maintain an instance of its own. This appears to align with `sentry_init()`, which takes ownership of the object.

PlasmaDev5: Changed all that on my next commit. Essentially im trying to mirror what is in the breakpad backend without platform specific API stuff. Mainly relating to ```on_crash_func``` where in the breakpad backend it uses ```EXCEPTION_POINTERS *exinfo``` but only on windows. I am hoping to have the breakpad backend call into the new function to minimize duplicate code.

include/sentry.h
PlasmaDev5: Its more an issue of lack of API knowledge. I found ```sentry__path_from_str``` so going to pass in ```const char*``` and convert within the function.

supervacuus: Yeah, we don't currently expose the path and we should take it as a string here. However, we do provide `_n` interfaces that take ptr/len strings in addition to null-terminated strings (because people increasingly use `strings_views` and other things that cannot guarantee null-termination), where typically the null-terminated version is implemented in terms of the ptr/len interface.

include/sentry.h
supervacuus: The critical part is that options are sunk into the SDK when `sentry_init()` is called. Access to it is unsafe after `sentry_init()`, and all functions requiring access to the global state should do so internally via the `SENTRY_WITH_OPTIONS` scope macro, which handles synchronized access to the ref-count.

supervacuus: > Essentially im trying to mirror what is in the breakpad backend without platform specific API stuff. Mainly relating to on_crash_func where in the breakpad backend it uses EXCEPTION_POINTERS *exinfo but only on windows.

These shouldn't be necessary to touch for the function at all. Only the stuff inside the `should_dump` conditional (and the `SENTRY_WITH_OPTIONS` access to options) is essential.

supervacuus: > I am hoping to have the breakpad backend call into the new function to minimize duplicate code.

Let's leave deduplication open until the PR is done. There is a good chance that these implementations will diverge down the road, so there is no immediate need to keep one bound to the other.

src/sentry_core.c
```diff
@@ -1169,3 +1169,29 @@ sentry_clear_crashed_last_run(void)
...
 }
+
+void sentry_capture_minidump(sentry_path_t *dump_path, sentry_value_t event,
+    const sentry_options_t *options)
+{
+    sentry_envelope_t *envelope = sentry__prepare_event(
+        options, event, NULL, !options->on_crash_func);
+    sentry_session_t *session = sentry__end_current_session_with_status(
+        SENTRY_SESSION_STATUS_CRASHED);
+    sentry__envelope_add_session(envelope, session);
```
supervacuus, line 1180: The question is whether these two should run unconditionally or at all.

* does it make sense for this use case (or any other that essentially bypasses the SDK) to have a session or scope applied, given that this interface probably transfers an event from an entirely different session/scope?
* if we use `sentry__prepare_event()` here, its fourth parameter, `invoke_before_send`, should probably not depend on the `options->on_crash_func` being set (another reason not to extract the function and instead have two different implementations): even if we invoke a hook for an event that is entirely in the user's hands, it should always be `before_send` and not `on_crash` (which should strictly be called from a crash handler).
* if we don't apply scope or session here, then there are still a few things that `sentry__prepare_event()` does that might be useful:
  * checking for a valid `event_id` (`sentry__ensure_event_id()`)
  * creating an envelope to hold the event and minidump attachment
  * adding attachment items for the attachments configured via options (which could make sense?).
* as mentioned in https://github.com/getsentry/sentry-native/issues/1050#issuecomment-2404728084, while I think an event parameter would make this more generically useful, the question is what this parameter is going to be used for vs. auto-generating an event (which is what I had in mind when the argument is a null object).

supervacuus (changes requested):
Thanks @PlasmaDev5!

There are still quite a few open design decisions to consider. In this case, I tend to leave the duplication because there are already visible diverging needs.

I know that you haven't finished this yet, but the function needs 
* clear header documentation of tbd design decisions mentioned in my comments
* an integration test and an `example.c` entry

PlasmaDev5:
Feel i addressed feedback provided.
Unit test and mentioned ```example.c``` still need doing tomorrow. but wanted to get a little more feedback

src/sentry_core.c
```diff
@@ -1169,3 +1169,69 @@ sentry_clear_crashed_last_run(void)
...
+                sentry_value_new_string(
+#endif
+                    sentry__path_filename(sentry_dump_path)));
+        }
+
+        sentry__capture_envelope(options->transport, envelope);
+
+        // now that the envelope was written, we can remove the temporary
+        // minidump file
+        sentry__path_remove(sentry_dump_path);
```
supervacuus, line 1227: While this makes sense in the crash handler (including the comment), removing the dump file is probably not expected in the general case. I would provide a parameter `remove_dump_on_send` (or something similar) to make this deletion conditional on the client's request.

src/sentry_core.c
```diff
@@ -1169,3 +1169,69 @@ sentry_clear_crashed_last_run(void)
...
+        sentry__path_free(sentry_dump_path);
+
+        // after capturing the crash event, try to dump all the in-flight
+        // data of the previous transports
+        sentry__transport_dump_queue(options->transport, options->run);
+        // and restore the old transport
```
supervacuus, line 1233: This doesn't make sense here since we only do this in the crash handler because the process crashed, and we want to exit next (and the user won't be able to trigger a queue dump via `sentry_close()` themselves).

Here, we want to wait for as little as possible so that users can pile up send requests in the background. The forced queue dump will happen when they call `sentry_close()` at the end of the application, where they can specify a sufficiently large shutdown timeout via `sentry_options_set_shutdown_timeout()` as @vaind suggested.

supervacuus (changes requested):
Thanks, things are getting in shape.

PlasmaDev5:
Addressed review feedback.
Note: feedback to split declaration and initialization of ```envelope``` creates a lint error.

I also have work towards the test case and will post a draft version in a comment on here once i start work tomorrow for feedback as i am a lot more unsure with how we want to handle this step.

src/sentry_core.c
```diff
@@ -1169,3 +1169,69 @@ sentry_clear_crashed_last_run(void)
...
+        SENTRY_WARN(
+            "'sentry_capture_minidump' Failed due to null path to minidump");
+        return;
+    }
+
+    SENTRY_WITH_OPTIONS (options) {
+        sentry_envelope_t *envelope
+            = NULL;
+        sentry__ensure_event_id(event, NULL);
+        envelope = sentry__envelope_new();
```
supervacuus, line 1190: I was probably not clear about this, but I meant not to split but to have declaration and initialization in one:
```suggestion
        sentry__ensure_event_id(event, NULL);
        sentry_envelope_t *envelope = sentry__envelope_new();
```

src/sentry_core.c
```diff
@@ -1169,3 +1169,69 @@ sentry_clear_crashed_last_run(void)
     sentry__options_unlock();
     return success ? 0 : 1;
 }
+
+void
+sentry_capture_minidump(
+    const char *dump_path, sentry_value_t event, int removeDumpOnSend)
```
supervacuus, line 1175: We use `snake_case` throughout the SDK; please apply it here too.

supervacuus (changes requested):
If the format linter is bugging you, the Native SDK has a `.clang-format` spec that you can use locally.

If you have any questions about integration testing, please let me know.

PlasmaDev5:
To give a rundown on what i  have planned for the tests

I am referencing the attachment tests.
It essentially comes down to using ```sentry__path_write_buffer```, from what i can tell it should be able to create a .dmp file and just add some unused data. I can then use the new API to send capture this file.
Im not sure if there is a better solution as im still getting to grips with the available API.

PlasmaDev5:
@supervacuus Any feedback on the test case stuff? and hopefully all review feedback is addressed.

src/sentry_core.c
```diff
@@ -1227,8 +1225,8 @@ sentry_capture_minidump(
 
         sentry__capture_envelope(options->transport, envelope);
 
-        bool removeDumpOnSendBool = removeDumpOnSend;
-        if (removeDumpOnSendBool) {
+        bool remove_dump_on_send_bool = remove_dump_on_send;
+        if (remove_dump_on_send_bool) {
```
supervacuus, line 1229: It is okay to use the parameter directly here.

supervacuus (changes requested):
> I am referencing the attachment tests.
> It essentially comes down to using `sentry__path_write_buffer`, from what i can tell it should be able to create a .dmp file and just add some unused data. I can then use the new API to send capture this file.
> Im not sure if there is a better solution as im still getting to grips with the available API.

Given that the inner parts of `sentry_capture_minidump()` are already covered by unit tests and its side effects primarily affect the network and the file system, I think a basic integration test should suffice for starters.

What I would do is the following:

* Build and run the `sentry_example` with `log` and `crash` arguments to produce a minidump (which you can find in the database directory after the crash) and store it in `tests/fixtures`. At this point, you can also upload the debug info files, which will be useful if you want to verify what the function uploaded in the WebUI.
* Add a new command (`capture-minidump`) to `example.c` that passes that fixture as an (initially hard-coded absolute system-) path and an empty event. This also lets you experiment manually with the function to see what the WebUI renders for the incoming event.
* Alternatively, you can also send _any file_ as a first step since we don't check whether the path pointed to contains a valid minidump
* Get an integration test running that invokes the example using your new `capture-minidump` command (take inspiration from `test_capture_http` in `test_integration_http.py`, removing anything relating to transport-compression, but adding an `assert_minidump(envelope)` at the end)
* You can run integration tests locally via `make tests`, `scripts\run_tests.ps1` if you are on Windows or directly via `pytest`. You can check out the Makefile or the Powershell script to see the venv setup if you want to run this manually.
* to allow the test to run in CI, you must pass the fixture path from the test to the example executable (either via command-line arg or via the environment) or let the test or build script copy the fixture to a known location relative to the example executable (which might be helpful to test the removal case).
* if the above works, 
  * add a test case for (other) attachments (using the command `attachment` + adding `assert_attachment(envelope)` to the case) 
  * add a test case for `capture-minidump-remove` verifying that the

include/sentry.h
vaind: I've added the _n variant

src/sentry_core.c
```diff
@@ -1174,3 +1174,76 @@ sentry_clear_crashed_last_run(void)
...
+                sentry_value_new_string(
+#endif
+                    sentry__path_filename(sentry_dump_path)));
+        }
+
+        sentry__capture_envelope(options->transport, envelope);
+
+        if (remove_dump_on_send) {
+            sentry__path_remove(sentry_dump_path);
+        }
```
vaind, line 1245: Because this parameter doesn't actualy do what it says in the name (and it cannot at the moment), I'd remove it completely and let users move/remove the file after this function returns.

vaind: Note: I could of course remove it myself since I'm taking over this PR, just raised this as a discussion in the PR because I know the original issue had a request to remove processed files. It just doesn't fit with the rest of the API in my opinion.

bruno-garcia:
Closing in favor of:

* https://github.com/getsentry/sentry-native/pull/1067/

PR #1058, opened 2024-10-24
Setup CodeQL

markushi:
This came up via the `sentry-java` [code scanning alerts over here](https://github.com/getsentry/sentry-java/security/code-scanning).

As we've moved the NDK into sentry-native, this PR now enables CodeQL for both the C(++) parts, as well for the Java parts within `sentry-native`.  

Relevant `sentry-java` 8.x PR: https://github.com/getsentry/sentry-java/pull/3819

#skip-changelog

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,62 @@
+name: 'CodeQL'
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+  schedule:
+    - cron: '17 23 * * 3'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: ['cpp', 'java']
+
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@c36620d31ac7c881962c3d9dd939c40ec9434f2b # pin@v2
+        with:
+          languages: ${{ matrix.language }}
+
+      - if: matrix.language == 'java'
+        name: Setup Java Version
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+
+      - if: matrix.language == 'java'
+        name: Setup Gradle
+        uses: gradle/actions/setup-gradle@bb0c460cbf5354b0cddd15bacdf0d6aaa3e5a32b # pin@v3
+        with:
+          gradle-home-cache-cleanup: true
+
+      - if: matrix.language == 'java'
+        name: Build for Android NDK
+        working-directory: ./ndk
+        run: |
+          ./gradlew compileJava
+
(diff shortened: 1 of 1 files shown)


adinauer (approved):

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,67 @@
...
+      - if: matrix.language == 'java'
+        name: Build for Android NDK
+        working-directory: ./ndk
+        run: |
+          ./gradlew compileJava
+
+      - if: matrix.language == 'cpp'
+        name: Build sentry-native
+        run: |
+          cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo && cmake --build build --parallel
```
supervacuus, line 64: As someone who was only exposed to CodeQL using its CLI to write queries, how do we parametrize the analysis? I mean, yes, RTFM, but do you know about any defaults? I can remember multiple build modes. Will this configuration

* scan the source only? And why, then, do we build? Where does the scan stop (will vendored items be considered)
* scan the source via the build directory? We would have to create a build matrix to cover our build-time configurations.
* log its choices/defaults somewhere? I see no workflow output, probably because it uploads stuff to GitHub. However, I only see three "security" items for this branch, not the configuration used or the source files considered. The workflow log is also silent regarding the source files or build mode.

Do you know if the idea here is to parameterize the analysis from the GitHub scanner configuration web-ui or to do this from the Yaml? Are they aware of each other or entirely independent?

supervacuus (approved):
Thanks, @markushi. LGTM as a start, but I have many questions 😅 

There is nothing that prevents merging and having this running. Still, I wonder how much of the CLI capability is accessible and what the right path is to configure it for an SDK with so many build parameters that it essentially becomes an SDK to build SDKs.

.github/workflows/codeql.yml
supervacuus: I began by answering some of these questions by merging to master. Then, the relationship between the workflow and the security scanning ui became more apparent.

If you have more input, I am still all ears.

markushi: Yeah as of now many defaults are chosen, but in general both the source and the compiled code get analysed. If you haven't seen by now, here's the output of a [recent run](https://github.com/getsentry/sentry-native/actions/runs/11502771788/job/32018507800), the docs around [the default queries](https://docs.github.com/en/code-security/code-scanning/managing-your-code-scanning-configuration/c-cpp-built-in-queries) can be found here. On top of that the security page allows you to download the final list of rules being used.
image

Having said that, I fully agree - this is just some basic vanilla setup, if we want to have the most out of it, we should have a look at the more advanced configuration and using the `security-extended` [query suite](https://docs.github.com/en/code-security/code-scanning/managing-your-code-scanning-configuration/codeql-query-suites).

supervacuus: Thanks! My question is less about queries used and more about source files analyzed. 

In local usage, when you point CodeQL to a "manual" build (which native typically requires), only the source files used in the build will be considered for analysis. 

So, my immediate question would be: In the current configuration, does CodeQL only consider files used in the Linux build? Sadly, the workflow run log doesn't provide any insights into this question, nor do any rules-related outputs.

But I saw that the top-level CodeQL tool status lists scanned files (not on Friday, though, so maybe these are updated with a delay), and it provides a CSV for details, which both validate my assumption:

Screenshot 2024-10-28 at 09 18 26

Curiously, no Java/Kotlin files are listed even though this is a top-level tool status (i.e., not language-specific output), and even though you build the JNI project, none of the JNI-specific C/C++ files aren't included either.

Issue #1056, opened 2024-10-16
Wrong documentation for 'dsn' option in Native SDK


VolodymyrKorniichuk-GeoComply:
### SDK

Native SDK (Windows, macOS)

### Description

On the page https://docs.sentry.io/platforms/native/configuration/options/ documentation says:

> [dsn](https://docs.sentry.io/platforms/native/configuration/options/#dsn)
The DSN tells the SDK where to send the events. If this value is not provided, the SDK will try to read it from the SENTRY_DSN environment variable. If that variable also does not exist, the SDK will just not send any events.

This was true in some older versions of SDK (for example, when I used Native SDK 0.6.2). Now, I switched 0.7.9 and in case I don't set DSN (in debug builds, for example), `sentry_init` returns error 1. So, the description "If that variable also does not exist, the SDK will just not send any events" is not complete. SDK will also return 1 from `sentry_init` and not start crash handler at all.

PR #1059, opened 2024-10-24
fix: don't fail `crashpad` startup on invalid `DSN`

JoshuaMoelans:
fixes https://github.com/getsentry/sentry-native/issues/1056

When leaving the DSN empty, the crashpad handler wouldn't start. This PR fixes this by warning the user about the empty DSN and not failing the initialization anymore.

### DIFF
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -443,7 +443,7 @@ crashpad_backend_startup(
             "failed to construct minidump URL (check DSN or user-agent)");
     }
     bool success = data->client->StartHandler(handler, database, database, 
-        minidump_url ? mindump_url : "",
+        minidump_url ? minidump_url : "",
         options->http_proxy ? options->http_proxy : "", annotations, arguments,
         /* restartable */ true,
         /* asynchronous_start */ false, attachments);
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -442,13 +442,12 @@ crashpad_backend_startup(
         SENTRY_WARN(
             "failed to construct minidump URL (check DSN or user-agent)");
     }
-    const std::string url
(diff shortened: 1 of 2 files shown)


src/backends/sentry_backend_crashpad.cpp
```diff
@@ -434,23 +434,21 @@ crashpad_backend_startup(
...
-            annotations, arguments,
-            /* restartable */ true,
-            /* asynchronous_start */ false, attachments);
-        sentry_free(minidump_url);
     } else {
         SENTRY_WARN(
             "failed to construct minidump URL (check DSN or user-agent)");
-        crashpad_state_dtor(data);
-        return 1;
     }
```
supervacuus, line 444: We can drop the "else" here, given that the `DSN`-validation will complain before that.

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -434,23 +434,21 @@ crashpad_backend_startup(
...
+    const std::string url
+        = minidump_url ? std::string(minidump_url) : std::string();
+    sentry_free(minidump_url);
+    bool success = data->client->StartHandler(handler, database, database, url,
+        options->http_proxy ? options->http_proxy : "", annotations, arguments,
+        /* restartable */ true,
+        /* asynchronous_start */ false, attachments);
```
supervacuus, line 451: Why don't we do it like we do it for `http_proxy` without explicitly wrapping it in `std::string`? A wrap will happen when passing the argument anyway.

```suggestion
    bool success = data->client->StartHandler(handler, database, database, 
        minidump_url ? mindump_url : "",
        options->http_proxy ? options->http_proxy : "", annotations, arguments,
        /* restartable */ true,
        /* asynchronous_start */ false, attachments);
    sentry_free(minidump_url);
```

supervacuus (changes requested):
LGTM.

supervacuus:
Btw, this should end up in the change log as a fix.

supervacuus (approved):

Issue #1006, opened 2024-06-07
Breadcrumbs should use more efficient ringbuffer


Swatinem:
Adding a breadcrumb right now is using `sentry__value_append_bounded` to limit the number of breadcrumbs according to settings:

https://github.com/getsentry/sentry-native/blob/a05b260359246c75ccb594e86110a1fd326d04b3/src/sentry_core.c#L650-L651

However that internally does a costly `memmove`:

https://github.com/getsentry/sentry-native/blob/a05b260359246c75ccb594e86110a1fd326d04b3/src/sentry_value.c#L658

It would be nice to find a way to optimize this, for example by using a ringbuffer / linked list, or some other method to avoid doing a `O(n)` `memmove` every time when adding breadcrumbs that are at their maximum.

PR #1060, opened 2024-10-24
feat: add breadcrumb ringbuffer to avoid O(n) memmove

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/1006

(in progress)

- [x] Creating ringbuffer
- [x] Storing items to ringbuffer
- [x] Rethink `decref` test -> add a test that checks reference counts of objects added/overwritten in ringbuffer
- [x] Extracting items from ringbuffer (by converting back to regular list)
- [ ] Updating docs
- [x] Updating changelog


src/sentry_value.c
```diff
@@ -639,30 +640,16 @@ sentry__value_append_bounded(sentry_value_t value, sentry_value_t v, size_t max)
 
     list_t *l = thing->payload._ptr;
 
-    if (l->len < max) {
+    if (l->len <= max) {
```
JoshuaMoelans, line 643: Since we now store [start_idx, breadcrumb 1, bc 2, ...] the full list will be max+1 long

tests/unit/test_value.c
```diff
@@ -509,59 +510,6 @@ SENTRY_TEST(value_wrong_type)
     TEST_CHECK(sentry_value_get_length(val) == 0);
 }
 
-SENTRY_TEST(value_collections_leak)
```
JoshuaMoelans, line 512: Removed this test since it no longer makes sense with the new ringbuffer append. Maybe this should be replaced by a similar test (e.g. one where we add objects to the ringbuffer & make sure they get refcount reduced when they get overwritten)

supervacuus: Yes, that makes sense. I am unsure about a _new_ test as long as we test breadcrumb overflow _somewhere_ because Valgrind or ASAN would scream if the ring buffer append didn't correctly ref-count values.

src/sentry_value.c
supervacuus: That is true, but can we also express it as you described it in the code?
```suggestion
    if (l->len < max + 1) {
```
This makes the intention much more evident.

src/sentry_scope.c
```diff
@@ -337,7 +338,8 @@ sentry__scope_apply_to_event(const sentry_scope_t *scope,
     sentry_value_decref(contexts);
 
     if (mode & SENTRY_SCOPE_BREADCRUMBS) {
-        PLACE_CLONED_VALUE("breadcrumbs", scope->breadcrumbs);
+        PLACE_CLONED_VALUE("breadcrumbs",
+            sentry__value_ring_buffer_to_list(scope->breadcrumbs));
```
JoshuaMoelans, line 342: Do we need to clear the list after cloning it? Currently tests are failing, which might be because the old `scope->breadcrumbs` got cleared out anyway, but we now allocate a new list to clone instead.

tests/unit/test_value.c
JoshuaMoelans: Added a new test `SENTRY_TEST(value_ringbuffer)` which aims to check exactly the behaviour I described above. If you agree on the test structure/contents I can mark this point as resolved ^

src/sentry_scope.c
```diff
@@ -338,8 +338,10 @@ sentry__scope_apply_to_event(const sentry_scope_t *scope,
     sentry_value_decref(contexts);
 
     if (mode & SENTRY_SCOPE_BREADCRUMBS) {
-        PLACE_CLONED_VALUE("breadcrumbs",
-            sentry__value_ring_buffer_to_list(scope->breadcrumbs));
+        sentry_value_t l
+            = sentry__value_ring_buffer_to_list(scope->breadcrumbs);
+        PLACE_CLONED_VALUE("breadcrumbs", l);
+        sentry_value_decref(l);
```
supervacuus, line 344: This is actually double cloning, right? You clone the contents of `scope->breadcrumbs` once via `sentry__value_ring_buffer_to_list` and then again in `PLACE_CLONED_VALUE`, the latter is no longer necessary, because `l` already contains the cloned values. You must set the `breadcrumbs` by key, but without a clone (i.e., `PLACE_VALUE`).

src/sentry_value.c
```diff
@@ -840,6 +827,25 @@ sentry_value_as_string(sentry_value_t value)
...
+    const thing_t *thing = value_as_thing(rb);
+    if (!thing || thing_get_type(thing) != THING_TYPE_LIST) {
+        return sentry_value_new_null();
+    }
+    const list_t *rb_list = thing->payload._ptr;
+    const int32_t start_idx = sentry_value_as_int32(rb_list->items[0]);
+
+    sentry_value_t rv = sentry_value_new_list();
+    for (size_t i = 0; i < rb_list->len - 1; i++) {
+        int32_t idx = (start_idx - 1 + i) % (rb_list->len - 1) + 1;
```
supervacuus, line 842: Due to API constraints, we must _store_ an `int32` value (or the integral component of a 64-bit float), but we should use it always as an unsigned type here. I think this what the VC compiler complains about.

src/sentry_value.h
```diff
@@ -81,14 +81,25 @@ sentry_value_t sentry__value_clone(sentry_value_t value);
...
- * It will remove the first value of the list, is case the total number if items
- * would exceed `max`.
+ * To make this work properly as a ring buffer, the value list needs to have
+ * the ring buffer start index as the first element
+ * (e.g, 1 until max is exceeded, then it will update for each added item)
+ *
+ * It will remove the oldest value in the list, in case the total number of
+ * items would exceed `max`.
+ *
+ * The list is of size `max + 1` to store the start index.
```
supervacuus, line 91: This is an exemplary implementation description and should land above the function definition of the C file because that is also where another implementer would be scratching their head regarding the layout.

Here, I would expect a high-level description that explains to callers what we expect from them and what they expect from us:
* append will exponentially grow (reallocate) the list until we reach `max`
* after reaching `max`, it will act as a ring buffer (no further list allocations or mem moves).
* what happens when `max` changes over multiple invocations, or at least what we expect call sites to do

The last point is particularly relevant because it is extraordinary that an argument should stay fixed over multiple invocations. This leaky abstraction for a particular _internal_ usage is okay, but we should highlight it.

src/sentry_scope.c
```diff
@@ -74,6 +74,7 @@ get_scope(void)
     g_scope.contexts = sentry_value_new_object();
     sentry_value_set_by_key(g_scope.contexts, "os", sentry__get_os_context());
     g_scope.breadcrumbs = sentry_value_new_list();
+    sentry_value_append(g_scope.breadcrumbs, sentry_value_new_int32(1));
```
supervacuus, line 77: I think it would be much better to move this as a special case (`l->len == 0`) into `sentry__value_append_ringbuffer()`, so that we do not spread the invariants over multiple locations.

supervacuus (changes requested):
This looks great; we can merge it with minor finishing touches.

src/sentry_scope.c
JoshuaMoelans: Makes sense, I adapted the append code & the conversion to list (since we can now have a ringbuffer without items if the first value is never added) 
https://github.com/getsentry/sentry-native/pull/1060/commits/190a4a5d1dc195d5b038634f83902f79c8e84b0b#diff-c47c40069b4dfa1e5cc367dcd84f2c8845913ad9dff9ff82686139981f95c5abR858-R860

src/sentry_value.h
JoshuaMoelans: I've updated the description on both sides; let me know if you still feel something's missing/vague:

https://github.com/getsentry/sentry-native/pull/1060/files#diff-c47c40069b4dfa1e5cc367dcd84f2c8845913ad9dff9ff82686139981f95c5abR632-R644

https://github.com/getsentry/sentry-native/pull/1060/files#diff-975c006356fcc80792308ad2d771a20ca88700679287e4eabcfc94346c703e21R84-R89

supervacuus (approved):

PR #1062, opened 2024-10-28
fix: remove a potential overflow before conversion

supervacuus:
This is in response to CodeQL security scan alerts 1-3.

`Elf[32|64]_Ehdr[.e_phnum|.e_phentsize|.e_shnum|.e_shentsize]` are all `uint16_t`. This means the loop-var `i` is bounded by `uint16_t` and should fit in a `uint32_t` (to prevent unsigned overflow in the loop). A switch to unsigned still makes sense because we reduce the future chance of unnecessary signed overflow (=UB) in the loop body.

All program/section-header table entry sizes are cast to `uin64_t` even though the multiplication is bound to `uint32_t` by both factors bound by `uint16_t`. This fixes the potential overflow before conversion to the bigger type.

#skip-changelog

### DIFF
--- a/src/modulefinder/sentry_modulefinder_linux.c
+++ b/src/modulefinder/sentry_modulefinder_linux.c
@@ -362,7 +362,7 @@ fail:
                                                                                \
         Elf64_Shdr strheader;                                                  \
         ENSURE(sentry__module_read_safely(&strheader, module,                  \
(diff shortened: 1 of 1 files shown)


JoshuaMoelans (approved):
lgtm!

Issue #1030, opened 2024-08-14
Need support for socks5proxy for sentry-native


saivivek1987:
### Problem Statement

Can we get support for socks5proxy also. I know we can use httpproxy can we also get support for socks5proxy in [sentry-native](https://github.com/getsentry/sentry-native). Currently using [sentry-native sdk](https://docs.sentry.io/platforms/native/). It has support for using httpproxy but need support for socks5proxy also.

### Product Area

APIs

JoshuaMoelans:
Hi @saivivek1987 , thanks for this feature request! Could you share with us which platform you'd want socks5 proxy support for?

PR #1063, opened 2024-10-29
feat: add support for SOCKS5 proxy

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/1030

Todo

- [x] Add SOCKS5 proxy option
- [x] Handle transport when SOCKS5 proxy is given
    -> currently checked for MacOS using Curl transport
- [x] also check for Linux transport
- [x] Add Proxy integration tests
    -> using `mitmproxy`
- [x] Add CI step to install `mitmproxy`
    -> replaced by adding it to [`requirements.txt`](https://github.com/getsentry/sentry-native/pull/1063/files#diff-27c5febcc9ce93e8e5341f263b7d68962733b8a2a86c583723d025a1db39d7a3R8)
- [x] Merge [Crashpad PR](https://github.com/getsentry/crashpad/pull/113)
- [x] Update changelog
- [x] Update [docs](https://docs.sentry.io/platforms/native/configuration/transports/)
    -> especially about [socks5 being circumvented ](https://forums.developer.apple.com/forums/thread/705504?answerId=712418022#712418022)if proxy is unavailable on mac
    -> similar story on [Windows with Autoproxy](https://learn.microsoft.com/en-us/windows/win32/winhttp/winhttp-autoproxy-support) circumventing the proxy if a connection is possible
        -> ❗️ this only manifested in the `capture` test, NOT in the `crash` test -> ToDo: reason about why crashpad doesn't autoproxy?
    -> document which transports/platforms don't support it (e.g. [winhttp](https://www.vbforums.com/showthread.php?889332-Vb6-winhttp-proxy-socks5-support))


src/transports/sentry_transport_curl.c
```diff
@@ -218,6 +221,9 @@ sentry__curl_send_task(void *_envelope, void *_state)
     if (state->http_proxy) {
         curl_easy_setopt(curl, CURLOPT_PROXY, state->http_proxy);
     }
+    if (state->socks5_proxy) {
```
JoshuaMoelans, line 224: Should we disallow having both http and socks5 proxies (e.g. by having one take priority over the other if both are given), or can these work in tandem?

supervacuus: No, you are right. We should prevent double proxy configs at the options interface (i.e., reset an HTTP proxy when setting a SOCKS5 proxy and vice versa). 

The HTTP proxy option results in a `CONNECT` method proxy, a persistent tunnel, and thus a stateful connection. We can't have two persistent tunnels (that would require two clients).

Please also consider providing a single proxy option (where we keep but deprecate the current HTTP-only proxy option or reuse it with sufficient documentation). In this option, we decide, based on the URI, which proxy configuration we attempt when configuring our HTTP clients. 

This is more generic and extensible and provides the convenience of not carrying around two (or, in the future, more) mutually exclusive config items.

JoshuaMoelans: @supervacuus I've added a single proxy option & kept the old API methods (which now just call to the new `sentry_options_set/get_proxy` function). I think that should resolve this comment?

src/transports/sentry_transport_curl.c
```diff
@@ -215,8 +215,8 @@ sentry__curl_send_task(void *_envelope, void *_state)
     curl_easy_setopt(curl, CURLOPT_HEADERDATA, (void *)&info);
     curl_easy_setopt(curl, CURLOPT_HEADERFUNCTION, header_callback);
 
-    if (state->http_proxy) {
-        curl_easy_setopt(curl, CURLOPT_PROXY, state->http_proxy);
+    if (state->proxy) { // TODO do we need to process the proxy here?
```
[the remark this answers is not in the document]
supervacuus: Please be aware that we have a runtime and a compile-time check for `libcurl`.

tests/test_integration_http.py
```diff
@@ -606,3 +608,64 @@ def test_capture_minidump(cmake, httpserver):
...
+            ["log", "start-session", "capture-event"]
+            + run_args,  # only passes if given proxy is running
+            check=True,
+            env=dict(os.environ, SENTRY_DSN=make_dsn(httpserver)),
+        )
+        if proxy_status == ["on"]:
+            assert len(httpserver.log) == 2
+        elif proxy_status == ["off"]:
+            # Window's Autoproxy will send the request even if the http proxy fails
+            # https://learn.microsoft.com/en-us/windows/win32/winhttp/winhttp-autoproxy-support
```
supervacuus, line 666: Curiously, doesn't this also affect crashpad transport? Are you sure that the PAC auto proxy configuration is the cause here?

supervacuus (changes requested):
This looks great, thanks! Just a few minor changes and questions.

src/transports/sentry_transport_curl.c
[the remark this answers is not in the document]
JoshuaMoelans: I think this was a very early comment (when I was still trying to find out where the proxy implementation needed changing) which I was just using as a marker for myself. It doesn't have any significance anymore, so I'll just remove it.

tests/test_integration_http.py
JoshuaMoelans: I'm not sure, I was also puzzled by the fact that it doesn't affect the crashpad test. This might need some further investigation before merging, as I cannot say for certain that the autoproxy setting is the cause.

JoshuaMoelans: Upon closer investigation, this [code snippet](https://github.com/getsentry/sentry-native/pull/1063/files#diff-8492f5ddad4505921f8eb4d12bdeba3e62a966aafc94fd235dd77e4aba7c0e80R86-R105) (also pasted below) seemed to imply that there is no fallback on the Windows crashpad proxy crash test (whilst there is one on the proxy-capture test); However, changing the assertion at the end to ensure we only expect a log IF the proxy was running (instead of relying on the `Assertionerror` to handle that situation), the test fails on Windows **and** on macOS because it falls back.

The argument for Apple's fallback stands with the blogpost already mentioned in the `Assertionerror` scope, but there is probably a similar method (potentially the Autoproxy) that results in the same behaviour on Windows. This can use some further investigation from my end to make sure we have the root cause documented.

I'll also adapt the test to more clearly indicate the expected return when the proxy is turned off, given that we now know about these fallback methods.

```diff
        try:
            with httpserver.wait(timeout=10) as waiting:
                child = run(
                    tmp_path, "sentry_example", ["log", "crash"] + run_args, env=env
                )
                assert child.returncode  # well, it's a crash after all
        except AssertionError:
            # only want to end up here on (http-proxy, off) tuple (we expect the request to fail)
            if run_args == ["http-proxy"] and proxy_status == ["off"]:
                assert len(httpserver.log) == 0
                return
            elif run_args == ["socks5-proxy"] and proxy_status == ["off"]:
                # Apple's NSURLSession will send the request even if the socks proxy fails
                # https://forums.developer.apple.com/forums/thread/705504?answerId=712418022#712418022
                assert len(httpserver.log) == (1 if (sys.platform == "darwin") else 0)
                return

        assert waiting.result

+       assert len(httpserver.log) == (1 if proxy_status == ["on"] else 0);
```

JoshuaMoelans: I've added some comments to the code explaining the observed proxy behaviour: https://github.com/getsentry/sentry-native/pull/1063/commits/22982f83fb56c62de1043fccdc16d52ece2f3d4d

- We see that on the [`capture_proxy` test](https://github.com/getsentry/sentry-native/pull/1063/files#diff-5254164492b3599852a520cb29d9b5cdc964768ea2dfc6ddbd9714a45ee03f6eR663-R668), only Windows provides a fallback (we still expect `0` when `proxy_status=off` for non-win32 platforms). 
- In the [`crash_proxy` test](https://github.com/getsentry/sentry-native/pull/1063/files#diff-8492f5ddad4505921f8eb4d12bdeba3e62a966aafc94fd235dd77e4aba7c0e80R105-R107) on the other hand, regardless of the platform, if the `httpserver.wait` returns we always get an `httpserver.log` of size `1` even when the `proxy_status=off`.

supervacuus: So, to be clear, the outcome is the following:

* on `macOS`, we have a different bypass behavior (bypasses in `crashpad_handler` if the proxy is unreachable but honors proxy without in our transport), which is expected because we use the `NS*` networking API in the `crashpad_handler` but defer to `libcurl` in our transport.
* on `Linux`, the proxy is honored without bypass in both, which is expected because both use `libcurl`
* on `Windows`, the proxy is by-passed in both, if the proxy is unreachable because both use WinHTTP

Am I summarizing this correctly? If yes, that should be documented in the sentry header's inline docs. Of course, only the observable behavior should be documented, not my justification.

include/sentry.h
```diff
@@ -972,17 +972,43 @@ SENTRY_API void sentry_options_set_dist_n(
 SENTRY_API const char *sentry_options_get_dist(const sentry_options_t *opts);
 
 /**
- * Configures the http proxy.
+ * Configures the proxy.
  *
- * The given proxy has to include the full scheme, eg. `http://some.proxy/`.
+ * The given proxy has to include the full scheme,
+ * eg. `http://some.proxy/`or 'socks5://some.proxy/'.
```
supervacuus, line 978:
```suggestion
 * eg. `http://some.proxy/` or `socks5://some.proxy/`.
```

CONTRIBUTING.md
```diff
@@ -146,6 +146,8 @@ The example currently supports the following commands:
 - `discarding-on-crash`: Installs an `on_crash()` callback that discards the crash event.
 - `override-sdk-name`: Changes the SDK name via the options at runtime.
 - `stack-overflow`: Provokes a stack-overflow.
+- `http-proxy`: Uses a localhost HTTP proxy on port 8080.
+- `socks-proxy`: Uses a localhost SOCKS proxy on port 1080.
```
supervacuus, line 150:
```suggestion
- `http-proxy`: Uses a localhost `HTTP` proxy on port 8080.
- `socks5-proxy`: Uses a localhost `SOCKS5` proxy on port 1080.
```

src/transports/sentry_transport_curl.c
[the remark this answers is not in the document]
supervacuus: Just realized that I dropped the idea of a compile-time check in #813, there is only a runtime check. You already updated that one.

PR #1066, opened 2024-10-31
chore: add clang-format to venv

vaind:
I've run `make format` as instructed by the contributing.md and linter complained in CI. Later, I've found out it's because of a different version of clang-format so I've added it to .venv so that everyone is on the same version.

#skip-changelog


supervacuus (approved):
Looks great, but what is wrong with CI?

PR #1067, opened 2024-10-31
feat: expose function to send minidumps

### LINKED ISSUE #1050: Allow capturing minidumps programatically

vaind:
Adds `sentry_capture_minidump(const char*)` to send an existing minidump. 

This PR builds on top of #1052 but makes significant API changes so I've made a new PR instead of rewriting the old one.

* Closes #1050
* Closes #1052 

Note: CI failures seem to be unrelated, see https://github.com/getsentry/sentry-native/issues/1069


bruno-garcia:
Thank you @vaind !

src/sentry_core.c
```diff
@@ -18,6 +18,7 @@
 #include "sentry_tracing.h"
 #include "sentry_transport.h"
 #include "sentry_value.h"
+#include "transports/sentry_disk_transport.h"
```
supervacuus, line 21: This is no longer necessary, right? I would love to keep direct dependencies from core to concrete implementations at a minimum.

tests/test_capture_minidump.py
```diff
@@ -0,0 +1,52 @@
...
+pytestmark = pytest.mark.skipif(not has_http, reason="tests need http")
+
+# fmt: off
+auth_header = (
+    "Sentry sentry_key=uiaeosnrtdy, sentry_version=7, sentry_client=sentry.native/0.7.11"
+)
+# fmt: on
+
+
+def test_capture_minidump(cmake, httpserver):
```
supervacuus, line 23: Is this function sufficiently different from all other HTTP integration tests to end up in a separate module?

include/sentry.h
```diff
@@ -1362,6 +1362,12 @@ SENTRY_API sentry_user_consent_t sentry_user_consent_get(void);
  */
 SENTRY_API sentry_uuid_t sentry_capture_event(sentry_value_t event);
 
+/**
+ *  Allows capturing independently created minidumps.
```
supervacuus, line 1366: Given that this interface has no peers in the standard SDK interfaces, a quick spec of what happens is in order here, especially that:

* it generates a dummy event with log-level `FATAL` and
* this generated event is processed like any other event, including attachments, scope application, and before-send hook.

supervacuus (changes requested):
Thanks! Only minor things.

include/sentry.h
vaind: OK, updated the doc

PR #1072, opened 2024-11-04
docs: clarify macOS support of the `inproc` backend

supervacuus:
as a review result of: https://github.com/getsentry/sentry-docs/pull/11721

#skip-changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,8 @@ Legend:
 - `**`: `breakpad` on Android and iOS builds and should work according to upstream but is untested.
 - `***`: `curl` as a transport works on Android but isn't used in any supported configuration to reduce the size of our artifacts.
 
+In addition to platform support, the "Advanced Usage" section of the SDK docs now [describes the tradeoffs](https://docs.sentry.io/platforms/native/advanced-usage/backend-tradeoffs/) involved in choosing a suitable backend for a particular use case.
+
 ### Build Targets
 
 - `sentry`: This is the main library and the only default build target.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -270,6 +270,7 @@ Legend:
 - ☑ default
 - ✓ supported
 - unsupported
+- `*`: `inproc` has not produced valid stack traces on macOS since version 13 ("Ventura"). Details: https://github.com/getsentry/sentry-native/issues/906
```
kahest, line 273: should we instead/additionally point to the advanced usage page?

kahest (approved):
thanks for adding, left a question/suggestion

README.md
```diff
@@ -260,7 +260,7 @@ using `cmake -D BUILD_SHARED_LIBS=OFF ..`.
 | - none     | ✓       | ✓     | ✓     | ☑       | ☑   |
 |            |         |       |       |         |     |
 | Backends   |         |       |       |         |     |
-| - inproc   | ✓       | ✓     | ✓     | ☑       |     |
+| - inproc   | ✓       | (✓)*  | ✓     | ☑       |     |
```
JoshuaMoelans, line 263: nit: is there a reason why we keep `inproc` as the first listed item if macOS doesn't really support it? Maybe we should move the default `crashpad` to the top of this list; as a user that would be less confusing imo

JoshuaMoelans (approved):
LGTM (+ small nit)

kahest (approved):
🚀

PR #1075, opened 2024-11-06
feat: add emitting of transaction data inside context trace data

JoshuaMoelans:
fixes https://github.com/getsentry/sentry-native/issues/1074
related to https://github.com/getsentry/team-sdks/issues/95

ToDo:
- [x] emit transaction data through context trace data
- [ ] add proper test for this


src/sentry_core.c
```diff
@@ -934,6 +934,8 @@ sentry_transaction_finish(sentry_transaction_t *opaque_tx)
     sentry_value_t trace_context
         = sentry__value_get_trace_context(opaque_tx->inner);
     sentry_value_t contexts = sentry_value_new_object();
+    sentry_value_set_by_key(
+        contexts, "data", sentry_value_get_by_key(tx, "extra"));
```
JoshuaMoelans, line 938: could this cause issues in tests where we didn't expect the transaction's data to be passed through before (e.g. context data differs because it's expected to be empty?)

tests/unit/test_tracing.c
```diff
@@ -1097,6 +1097,19 @@ SENTRY_TEST(txn_data)
...
+{
+    sentry_transaction_t *txn
+        = sentry__transaction_new(sentry_value_new_object());
+
+    sentry_transaction_set_data(
+        txn, "os.name", sentry_value_new_string("Linux"));
+    check_after_set(txn->inner, "extra", "os.name", "Linux");
+    sentry_value_set_by_key(txn->inner, "sampled", sentry_value_new_bool(true));
+
+    sentry_transaction_finish(txn);
```
JoshuaMoelans, line 1110: We need to add an assertion here (or else there is no test); how can we assert that the context.trace.data actually holds the data from our transaction?

src/sentry_core.c
```diff
@@ -944,6 +946,7 @@ sentry_transaction_finish(sentry_transaction_t *opaque_tx)
     sentry_value_remove_by_key(tx, "op");
     sentry_value_remove_by_key(tx, "description");
     sentry_value_remove_by_key(tx, "status");
+    sentry_value_remove_by_key(tx, "extra");
```
JoshuaMoelans, line 949: the current failing test (in [test_integration_http.py](https://github.com/getsentry/sentry-native/blob/bf14b8533a3b26853e4e6fecf2f955deaa29e2d8/tests/test_integration_http.py#L544-L548)) fails because we remove the "extra" data from the transaction; should we not remove it, or change the test? (also related to the comment above)

src/sentry_core.c
supervacuus: well, I hope it at least misses the "extra" key.

src/sentry_core.c
supervacuus: I'm glad it failed; that is precisely what this test is for. We're changing the contract to the backend. In this case, the backend invited the change, so we must assert whatever the test expected in `extra` to be in `contexts.trace.data`. Changing the test is the right choice here.

tests/unit/test_tracing.c
supervacuus: You can't do it with the current unit tests. At one point, we could introduce a "test transport" where we assert all the things happening in the pre-send pipeline.

In the meantime, we capture these with our integration tests (like the one currently failing). Writing tests for unpacking nested structures is much more convenient to express (and thus readable) in Python than in C (achievable with a little string and macro magic). 

Test execution speed is also a factor that hints towards having unit-test access to the transport pipeline.

src/sentry_core.c
supervacuus, line 938: By the way, when you do a `sentry_value_set_by_key()`, where another ref owns the set value, you need to increase the ref count because `sentry_value_set_by_key()` doesn't do this (it is a ref sink, for convenience).

This is probably why you see ASAN and Valgrind cry: you add it to `contexts.trace`` here but later remove it from `tx`, at this point, it gets freed because there is still only one ref counted. Later, when we create the envelope and need to serialize it, we use that value after freeing it.

tests/test_integration_http.py
```diff
@@ -575,6 +574,8 @@ def test_transaction_only(cmake, httpserver, build_args):
     timestamp = time.strptime(payload["timestamp"], RFC3339_FORMAT)
     assert timestamp >= start_timestamp
 
+    assert trace_context["data"] == {"url": "https://example.com"}
```
JoshuaMoelans, line 577: Moved the expected data from the `transaction_data` check into this `trace_context.data` check; should we now also make sure that the "extra" data in the transaction is cleared up at this point (since we remove the value when we store it in the trace context data)?

supervacuus: Nope, since this was a workaround (adopted from the workarounds of SDKs), it is unnecessary to encode that it is missing. 

But I think it would make sense to rename the field from "extra" back to "data", since naming it "extra" was the workaround. We're not breaking anything by changing this back to "data". Asserting that "data" is not an existing key in the transaction root can make sense.

supervacuus: To be clear, starting from the client interface. We now do the following:

```
sentry_transaction_set_data() -> 
sets tx["extra"] -> 
sentry_transaction_finish() -> 
moves tx["extra"] to tx["contexts.trace.data"]
```

This means that the interface names it "data," and then we name it "extra" until the transaction finishes. At this point, we name it "data" again and move to another child object.

That will be indecipherable at some point. We should change the intermediate "extra" back to "data" so that the change only concerns moving it from the transaction root to "contexts.trace".

JoshuaMoelans: Changed the "extra" key to "data"; would we now want to assert + warn when the "data" key already exists in the [`set_data` function](https://github.com/getsentry/sentry-native/blob/0623fe6224f55e42ed8c107bedeff90b82c69186/src/sentry_tracing.c#L559-L570)

src/sentry_core.c
```diff
@@ -934,6 +934,9 @@ sentry_transaction_finish(sentry_transaction_t *opaque_tx)
     sentry_value_t trace_context
         = sentry__value_get_trace_context(opaque_tx->inner);
     sentry_value_t contexts = sentry_value_new_object();
+    sentry_value_set_by_key(
+        trace_context, "data", sentry_value_get_by_key(tx, "data"));
+    sentry_value_incref(sentry_value_get_by_key(tx, "data"));
```
supervacuus, line 939: There is one missing piece left (added a todo to the PR description): we should apply the currently scoped transaction/span "data" to every event sent inside that transaction/span here:

https://github.com/getsentry/sentry-native/blob/ffdf6ed024ff1f2fe2e55fbe7d86940ca0228207/src/sentry_scope.c#L321-L328

If the `scope_trace` isn't a null object, we can also read the "data" item of that scoped transaction/span and place it in the events trace context. Again, you must use `incref` here because it will be `decref`ed when sending the envelope, meaning it would be freed and no longer accessible to the transaction/span.

Removal is unnecessary since the transaction/span will continue to exist beyond the lifetime of the event.

supervacuus: The data item should only be added to the event if it isn't a null object in the scoped transaction/span. It doesn't make a difference in the backend (null object == non-existent key, most of the time), but there is no need to send data over the wire that signifies non-existent.

JoshuaMoelans: Is this a correct implementation of what you asked for? (i.e. when a transaction object exists inside the scope, we get its "data" key item and put it at the scope_trace "data" key)
```c
sentry_value_t scope_trace = sentry__value_get_trace_context(
    sentry__get_span_or_transaction(scope));
if (!sentry_value_is_null(scope_trace)) {
    if (sentry_value_is_null(contexts)) {
        contexts = sentry_value_new_object();
    }
    if (scope->transaction_object) {
        sentry_value_set_by_key(scope_trace, "data",
            sentry_value_get_by_key(
                scope->transaction_object->inner, "data"));
        sentry_value_incref(sentry_value_get_by_key(
            scope->transaction_object->inner, "data"));
    }
    sentry_value_set_by_key(contexts, "trace", scope_trace);
}
```

supervacuus: You can use either `span` or `transaction` as returned from `sentry__get_span_or_transaction(scope)` and don't have to retrieve the transaction specifically. Both can have `data` items, and they should apply theirs to the event. If, for instance, a span should inherit its `data` from the transaction, then that should happen at span creation.

@markushi, do you also apply `data` when you apply the trace context to events from your scope transaction/spans? If so, do you take it solely from the transaction or also from spans?

supervacuus: To clarify, I would probably do something like this:

```C
sentry_value_t scoped_txn_or_span = sentry__get_span_or_transaction(scope)
sentry_value_t scope_trace = sentry__value_get_trace_context(scoped_txn_or_span);
if (!sentry_value_is_null(scope_trace)) {
    if (sentry_value_is_null(contexts)) {
        contexts = sentry_value_new_object();
    }
    sentry_value_t scoped_txn_or_span_data = sentry_value_get_by_key(scoped_txn_or_span, "data");
    if (!sentry_value_is_null(scoped_txn_or_span_data)) {
        sentry_value_incref(scoped_txn_or_span_data);
        sentry_value_set_by_key(scope_trace, "data", scoped_txn_or_span_data);
    }
    sentry_value_set_by_key(contexts, "trace", scope_trace);
}
```

markushi: @supervacuus @JoshuaMoelans Yes, we do the same but we solely take the context from the transaction:
https://github.com/getsentry/sentry-java/blob/28a11a7925b8e45bec38684f2002192787f13477/sentry/src/main/java/io/sentry/SentryClient.java#L216-L229

It's worth mentioning that at least in `sentry-java` a span doesn't really have it's own trace context, it solely returns the top level txn context (see https://github.com/getsentry/sentry-java/blob/2e90ac7733c6c9b941e7c64175f3d097d1d15a37/sentry/src/main/java/io/sentry/Span.java#L170)

supervacuus (approved):
We can merge this because transaction `data` is now handled as expected.

However this:

> It's worth mentioning that at least in sentry-java a span doesn't really have it's own trace context, it simply returns the top level txn context (see https://github.com/getsentry/sentry-java/blob/2e90ac7733c6c9b941e7c64175f3d097d1d15a37/sentry/src/main/java/io/sentry/Span.java#L170)

is worth further inspection because we deviate quite a bit, and the current implementation of how the trace context is created (and maintained over transaction and span life cycles is insufficient).

I had a deeper look into our implementation and the one in sentry-java. I could imagine anything from only fixing the retrieval function (to choose either the scoped transaction or the transaction linked to from the scoped span) to a more involved rewrite of how we carry around and maintain the trace context. Either way, this won't happen in this PR.

CC: @markushi, @kahest

PR #1077, opened 2024-11-12
fix: crashpad_handler proxy-upload semaphore-timeout

supervacuus:
Fixes an incorrectly set up timeout value when waiting for a `dispatch_semaphore_signal()` from the HTTP-proxy upload task on macOS. See: https://github.com/getsentry/crashpad/pull/111


JoshuaMoelans (approved):
LGTM

PR #1082, opened 2024-11-19
[NDK] Add loadNativeLibraries method to allow pre-loading .so files

markushi:
`sentry-java` `7.x` loads the `.so` libraries using a background thread, see https://github.com/getsentry/sentry-java/pull/3670

Instead of porting this behaviour 1:1 to `sentry-native`, I'm proposing a new `.loadNativeLibraries()` method, allowing you to pre-load the libraries if necessary. It's up to the consumer, like sentry-java `8.x` to use this method, and e.g. call it on a background thread.

This is a change in behaviour, as the `.so` files no longer get loaded during class init.


supervacuus (approved):
We already talked about how little love I have for static init blocks, but for a library, it makes even less sense since different clients have different needs. LGTM.

PR #1083, opened 2024-11-19
feat: thread context from breakpad on macos

supervacuus:
Upstream `breakpad` doesn't provide the thread context to the handler callbacks (it does write it to the minidump, though). This change fills the context of the crashed thread and passes it on to the handler callback (which in turn passes it on to the `on_crash` and `before_send` callbacks).

The change keeps compatibility between upstream and our fork in both directions even though this breaks one of the public interfaces:
* our `breakpad` backend in the Native SDK uses the upstream interface if `SENTRY_BREAKPAD_SYSTEM` is defined
* our `breakpad` fork provides the modified interface only if compiled with the Native SDK (i.e., `SENTRY_BACKEND_BREAKPAD` is defined in the `breakpad_client` target).

### DIFF
--- a/src/backends/sentry_backend_breakpad.cpp
+++ b/src/backends/sentry_backend_breakpad.cpp
@@ -42,7 +42,7 @@ extern "C" {
 
 #ifdef SENTRY_PLATFORM_WINDOWS
 static bool
-sentry__breakpad_backend_callback(const wchar_t *breakpad_dump_path,
+breakpad_backend_callback(const wchar_t *breakpad_dump_path,
     const wchar_t *minidump_id, void *UNUSED(context),
     EXCEPTION_POINTERS *exinfo, MDRawAssertionInfo *UNUSED(assertion),
     bool succeeded)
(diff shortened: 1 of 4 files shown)


src/backends/sentry_backend_breakpad.cpp
```diff
@@ -106,6 +113,12 @@ sentry__breakpad_backend_callback(
         if (options->on_crash_func) {
             sentry_ucontext_t *uctx = nullptr;
 
+#if defined(SENTRY_PLATFORM_DARWIN) && !defined(SENTRY_BREAKPAD_SYSTEM)
```
markushi, line 116: Shouldn't we only fill in the data if SENTRY_BREAKPAD_SYSTEM _is_ defined?
```suggestion
#if defined(SENTRY_PLATFORM_DARWIN) && defined(SENTRY_BREAKPAD_SYSTEM)
```

markushi (commented):
Looks good

src/backends/sentry_backend_breakpad.cpp
supervacuus: No, `SENTRY_BREAKPAD_SYSTEM` means that we are using the system-installed `breakpad`, which is the upstream version and that doesn't provide thread-contexts to the crash handler on macOS.

markushi: Oh right, yeah now it makes sense!

PR #1084, opened 2024-11-20
Don't define sentry_add_version_resource() command when building the static version

dacap:
As crashpad now checks the existence of sentry_add_version_resource() command in:

  https://github.com/getsentry/crashpad/commit/8961424c23c2ba98eaecbee526e0b729aabea1b2

We cannot define the command if the sentry.rc file is not going to be created, i.e. only the static version of Sentry is going to be built.

Without this fix we'll get the following error:
```
CMake Error: File .../sentry.rc.in does not exist. CMake Error at sentry-native/cmake/utils.cmake:14 (configure_file):
  configure_file Problem configuring file
Call Stack (most recent call first):
  sentry-native/external/crashpad/handler/CMakeLists.txt:118 (sentry_add_version_resource)
```

This regression was introduced in #1076

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -369,8 +369,8 @@ if(WIN32)
 	endif()
 endif()
 
-include(cmake/utils.cmake)
 if (WIN32 AND SENTRY_BUILD_SHARED_LIBS)
+	include(cmake/utils.cmake)
 	sentry_add_version_resource(sentry "Client Library")
 endif()
 


supervacuus:
Thanks for the report, @dacap. As you can see by the failing tests, this won't work because the `crashpad` artifacts are no longer versioned with your change.

The placement of the `include` outside the `if` was intentional because we added the `.rc` only to the `crashpad_handler` (an executable) and `crashpad_wer.dll` (a shared library by design), both of which are not affected by a static build configuration.

Even if we build shared, there is never a `crashpad` shared library (because all backends are linked into the main sentry library), so we don't need to differentiate between static and shared builds on the crashpad level (because the versioned artifacts are unaffected by these flags).

> We cannot define the command if the sentry.rc file is not going to be created, i.e. only the static version of Sentry is going to be built.

As you can see in your error message, the problem seems to be that `cmake/utils.cmake` can't find `sentry.rc.in` in the `CMAKE_SOURCE_DIR`. Do you see `sentry.rc.in` in your `CMAKE_SOURCE_DIR`? It would be interesting to understand how that is possible.

We do test our builds in the static configuration: https://github.com/getsentry/sentry-native/actions/runs/11930882865/job/33252573289#step:13:180

And with the given configuration the build seems to work.

supervacuus:
> Do you see sentry.rc.in in your CMAKE_SOURCE_DIR? It would be interesting to understand how that is possible.

Do you add sentry as a subdirectory to your project?

supervacuus:
> > Do you see sentry.rc.in in your CMAKE_SOURCE_DIR? It would be interesting to understand how that is possible.
> 
> Do you add sentry as a subdirectory to your project?

If yes, can you please try #1086?

dacap:
> If yes, can you please try #1086?

That's right, I'm using Sentry as a "subdirectory" (actually importing it from outside with `add_subdirectory(${SENTRY_DIR} sentry)`, [our CMakeLists.txt line](https://github.com/aseprite/aseprite/blob/a7c8d4cb06e344178f9142f5a6a4e61679cb96a5/src/app/CMakeLists.txt#L799))

I'll give a try to your PR and answer there. Thanks!

supervacuus:
closing in favor of #1086

PR #1086, opened 2024-11-20
fix: DLL versioning for SDK usage as CMake subdirectory

supervacuus:
This should fix the issue raised in #1084.

@dacap, can you give this a try in your setup?


JoshuaMoelans (approved):
LGTM!

PR #1088, opened 2024-11-21
Using libunwind for mac, since backtrace do not expect thread context…

saf-e:
… (i.e. working just with current thread)


supervacuus:
Thanks for the PR; introducing `libunwind` for macOS has been on our wish list for some time. This requires a bit more work on our side before we can merge it since ucontext-based stack tracing also affects the signal handler in the `inproc` backend. Further, the unwinder, as currently implemented, will have issues if system functions end up as leaves in the ucontext after being delivered from the signal trampoline.

If you are okay with it, I will take this over, adapt the current unwinder code for usage in `inproc` (and other signal handler users), and extract it into a separate unwinder module. But not before next week.

saf-e:
> If you are okay with it, I will take this over, adapt the current unwinder code for usage in `inproc` (and other signal handler users), and extract it into a separate unwinder module. But not before next week.

Yes, no problem with that, fill free to adapt it to your needs!

Issue #941, opened 2024-01-30
Feature Request: Provide explicit timings to the Performance Monitoring API


leops:
Exposing a lower-level API for manually creating transactions and spans with explicit timing values (instead of automatically reading the system clock when the spans are started / finished) would be useful to capture timing informations at a higher resolution than the default millisecond (which is often not precise enough for native code so microseconds / nanoseconds may be needed), or to build up a transaction from timing data coming from an external source (eg. performance counters from the GPU). I could easily open a PR implementing this, but it does raise the question of what the naming convention and general API design for these explicit transaction management functions should be.

supervacuus:
Hi @leops, as far as I know, microseconds are the current limit in the backend for performance/tracing events. Anything smaller than this would be truncated (or rounded), so opening the API to arbitrary scales doesn't make much sense (without considerable changes in the backend, and that discussion would have to happen in another issue tracker). Changing our timestamping to microseconds might be sensible, though. 

I am unsure whether the tracing API is the proper interface for instrumentation at that scale level because we can gather a maximum of 1000 spans before a transaction must flush. Transaction flushing is a costly operation (the transport happens in a backend thread, but preparing the envelope for sending happens in the calling thread and includes serialization, something you might not want to have in a tight loop).

This sounds like the perfect use case for `[developer metrics](https://develop.sentry.dev/delightful-developer-metrics/)`, where you have client-side aggregation of arbitrary measurements (you can use it for profiling code at that low level) over time. This is implemented only in the backend and in the Python SDK (?) in a very early phase and only rolled out to some customers. However, a discussion regarding the needs of Native SDK users might already make sense.

CC: @kahest.

supervacuus:
Hi @leops. That makes a lot of sense. So, to quickly summarize:

* microseconds would be enough fine-grained resolution for now
* you'd prefer to have hierarchical tracing rather than generic metrics (the latter aren't currently implemented anyway, but even if they were, performance/tracing would still be the better match for your use case)
* you would run the transaction flush asynchronously. I mainly mention this because people use the tracing API synchronously in screen-refresh loops or UI threads, which is different from the scenario for which they were initially implemented.
* if I got this right, you would still prefer to pass explicit timestamps or construct the span brackets yourself rather than defining them explicitly in the code, not only due to different timing resolutions but also because the measurements you want to take are not expressible in the host code. To be clear, you would still need to align these measurements with our wall clock to make these interpretable in traces.

[...]

PR #1093, opened 2024-11-28
feat: extend performance API with explicit timings

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/941

- [x] add tests for new API functions
- [x] add changelog entry
- [x] update docs https://github.com/getsentry/sentry-docs/pull/12041


src/sentry_tracing.c
```diff
@@ -378,7 +378,7 @@ sentry__value_span_new_n(size_t max_spans, sentry_value_t parent,
         sentry_value_new_string_n(description.ptr, description.len));
     sentry_value_set_by_key(child, "start_timestamp",
         sentry__value_new_string_owned(
-            sentry__usec_time_to_iso8601(sentry__usec_time())));
+            sentry__usec_time_to_iso8601(timestamp)));
```
JoshuaMoelans, line 381: Should our API's in-code docs mention what kind of timestamp we 'expect' (e.g. a value relative to the epoch time)?

src/sentry_core.c
```diff
@@ -874,14 +874,29 @@ sentry_transaction_start(
...
+sentry_transaction_t *
+sentry_transaction_start_ts(sentry_transaction_context_t *opaque_tx_cxt,
+    sentry_value_t sampling_ctx, uint64_t timestamp)
+{
+    return sentry__transaction_start(opaque_tx_cxt, sampling_ctx, timestamp);
+}
+
 sentry_uuid_t
-sentry_transaction_finish(sentry_transaction_t *opaque_tx)
+sentry__transaction_finish(sentry_transaction_t *opaque_tx, uint64_t timestamp)
```
supervacuus, line 899: You have two options here: either you implement `sentry_transaction_start()` in terms of `sentry_transaction_start_ts()` (like we did with the `_n`-functions) or you push down the implementation to a shared function that duplicates the interface wrt the `_ts` counterpart. If you do the latter, the pushed-down implementation should be static and lose the `sentry__` prefix.

src/sentry_core.c
```diff
@@ -984,9 +1012,9 @@ sentry_set_span(sentry_span_t *span)
 }
 
 sentry_span_t *
-sentry_transaction_start_child_n(sentry_transaction_t *opaque_parent,
+sentry__transaction_start_child_n(sentry_transaction_t *opaque_parent,
```
supervacuus, line 1015: Sames as above.

src/sentry_core.c
```diff
@@ -1053,8 +1117,26 @@ sentry_span_start_child(sentry_span_t *opaque_parent, const char *operation,
...
+    const char *description, uint64_t timestamp)
+{
+    return sentry_span_start_child_ts_n(opaque_parent, operation,
+        sentry__guarded_strlen(operation), description,
+        sentry__guarded_strlen(description), timestamp);
+}
+
 void
-sentry_span_finish(sentry_span_t *opaque_span)
+sentry__span_finish(sentry_span_t *opaque_span, uint64_t timestamp)
```
supervacuus, line 1139: Sames as above.

tests/unit/test_tracing.c
```diff
@@ -198,36 +199,69 @@ SENTRY_TEST(basic_function_transport_transaction)
...
+            tx->inner, "start_timestamp", "1970-01-01T00:00:00.000003Z");
+    } else {
+        tx = sentry_transaction_start(tx_cxt, sentry_value_new_null());
+    }
     CHECK_STRING_PROPERTY(tx->inner, "transaction", "honk");
     CHECK_STRING_PROPERTY(tx->inner, "op", "beep");
-    event_id = sentry_transaction_finish(tx);
-    TEST_CHECK(!sentry_uuid_is_nil(&event_id));
+    if (timestamped) {
+        sentry_uuid_t event_id = sentry_transaction_finish_ts(tx, 4);
```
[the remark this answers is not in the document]
supervacuus: Yes, this can only be verified with an integration test since `sentry_transaction_finish()` is a capture function.

supervacuus (changes requested):
LGTM, minor changes.

tests/unit/test_tracing.c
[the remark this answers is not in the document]
JoshuaMoelans: (discord discussion) After refactoring we already implicitly test a path with this, hence we don't need the additional integration test.

tests/assertions.py
```diff
@@ -205,8 +206,9 @@ def assert_minidump(envelope):
     assert minidump.payload.bytes.startswith(b"MDMP")
 
 
-def assert_timestamp(ts, now=datetime.now(UTC)):
-    assert ts[:11] == now.isoformat()[:11]
+def assert_timestamp(ts):
+    elapsed_time = datetime.now(timezone.utc) - datetime.fromisoformat(ts)
```
JoshuaMoelans, line 210: Had to add the timezone to avoid `TypeError: can't subtract offset-naive and offset-aware datetimes`. I think the subtraction still works since it [converts both to utc](https://docs.python.org/3/library/datetime.html#:~:text=input%20is%20aware.-,Subtraction,-of%20a%20datetime)

supervacuus: You could have just used the `now` parameter, right?

supervacuus (approved):

tests/assertions.py
JoshuaMoelans: right, removed the unnecessary imports & moved the `now` into the function body to evaluate it for every call

Issue #888, opened 2023-09-22
Add option to output warnings from sentry_init()


triplef:
If `sentry_init()` fails, it seems that currently the only way to figure out what caused it to fail is to enable debug logs with `sentry_options_set_debug()`. It would be helpful to have an option to always output initialization warnings and errors, without also outputting debug output.

This could be achieved by adding a method like this to set the output level:
```cpp
sentry_options_set_output_level(sentry_options_t *opts, sentry_level_t level);
```

In the following example logs (obtained by enabling debug logs), calling the above function with `SENTRY_LEVEL_WARNING` only the logs marked `WARN` should be output:

```
[sentry] INFO using database path "C:\..."
[sentry] DEBUG starting transport
[sentry] DEBUG starting background worker thread
[sentry] DEBUG starting backend
[sentry] WARN unable to start crashpad backend, invalid handler_path
[sentry] WARN failed to initialize backend
[sentry] WARN `sentry_init` failed
[sentry] DEBUG shutting down transport
[sentry] DEBUG shutting down background worker thread
[sentry] DEBUG submitting task to background worker thread
[sentry] DEBUG background worker thread started
[sentry] DEBUG executing task on worker thread
[sentry] DEBUG background worker thread shut down
```

PR #1107, opened 2024-12-18
feat: add option to set debug log level

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/888
Related to https://github.com/getsentry/sentry-cocoa/issues/4467

The main goal is to allow for outputting just the warnings from `sentry_init`. To accommodate this, we introduce log levels as a way to control what level of output the user gets. We decided against splitting up these levels into `init`-only and global levels, since this doesn't seem like a necessary feature as of yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## Unreleased
 
 **Features**:
-- Add `sentry_init` and global logger log levels ([#1107](https://github.com/getsentry/sentry-native/pull/1107))
+
+- Add option to set debug log level. ([#1107](https://github.com/getsentry/sentry-native/pull/1107))
 
 ## 0.7.17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+**Features**:
+- Add `sentry_init` and global logger log levels ([#1107](https://github.com/getsentry/sentry-native/pull/1107))
+
 ## 0.7.17
 
 **Features**:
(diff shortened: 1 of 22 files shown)


src/sentry_options.c
```diff
@@ -396,6 +401,20 @@ sentry_options_set_logger(
     opts->logger.logger_data = userdata;
 }
 
+void
+sentry_options_set_logger_level(sentry_options_t *opts, sentry_level_t level)
+{
+    opts->logger.logger_level = level;
+    opts->debug = true;
```
JoshuaMoelans, line 408: We implicitly expect that the user wants to have debug output when setting the logger level (hence we call this from `sentry_options_set_debug` as well).

supervacuus (changes requested):
I expected the macro name change to highlight a few inconsistencies.

A short sync might also be acceptable regarding the options interface.

include/sentry.h
```diff
@@ -1078,6 +1078,18 @@ SENTRY_API void sentry_options_set_debug(sentry_options_t *opts, int debug);
...
+ */
+SENTRY_API void sentry_options_set_logger_level(
+    sentry_options_t *opts, sentry_level_t level);
+
+/**
+ * Sets the log level for the sentry_init function
+ */
+SENTRY_API void sentry_options_set_init_log_level(
+    sentry_options_t *opts, sentry_level_t level);
+
```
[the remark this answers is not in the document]
JoshuaMoelans: Removed this overcomplication splitting up global and init log levels. Now we just allow setting of the log level globally.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+**Features**:
+- Add `sentry_init` and global logger log levels ([#1107](https://github.com/getsentry/sentry-native/pull/1107))
```
supervacuus, line 6:
```suggestion
**Features**:

- Add option to set debug log level. ([#1107](https://github.com/getsentry/sentry-native/pull/1107))
```

supervacuus (approved):

Issue #864, opened 2023-07-18
Support `tracesSampler`


kahest:
Add support for `tracesSampler`. 

Relevant docs:
- https://develop.sentry.dev/sdk/performance/#tracessampler
- https://docs.sentry.io/platforms/native/performance
- https://docs.sentry.io/platforms/native/configuration/sampling

markushi:
After some more discussion both `tracesSampler` and `beforeSendTransaction` should be low effort. Compared to `on_crash` they are not backend-specific and should be straightforward to implement.

Implementation detail: Let's keep in mind that resetting the `tracesSampler` should restore the original behaviour.

PR #1108, opened 2024-12-20
feat: add `traces_sampler`

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/864

- [x] Add to options
- [x] Add `sentry_traces_sampler_function` typedef (like the [`on_crash`](https://github.com/getsentry/sentry-native/blob/b24b30c1233ff72f173f7090ad628f6d194b2e68/include/sentry.h#L838-L886) and [`before_send`](https://github.com/getsentry/sentry-native/blob/b24b30c1233ff72f173f7090ad628f6d194b2e68/include/sentry.h#L796-L828) ones
- [x] Add `samplingContext`
- [x] Add tests (also for resetting sampler->restoring default tracesSampleRate behaviour)
- [ ] Update docs ([tracing](https://docs.sentry.io/platforms/native/tracing/) and [sampling](https://docs.sentry.io/platforms/native/configuration/sampling/)). Write about callback values access outside their lifetime (cleared after `transaction_start` is done). (docs PR: https://github.com/getsentry/sentry-docs/pull/12333)

[Core changes](https://github.com/getsentry/sentry-native/pull/1108/files/217525187fc0d8e74e89be9f3095162844a808c8) (without `tx_cxt` -> `tx_ctx` refactor)


src/sentry_sampling_context.h
```diff
@@ -0,0 +1,16 @@
+// sentry_sampling_context.h
+#ifndef SENTRY_SAMPLING_CONTEXT_H_INCLUDED
+#define SENTRY_SAMPLING_CONTEXT_H_INCLUDED
+
+#include "sentry_tracing.h"
+
+typedef struct sentry_sampling_context_s {
+    sentry_transaction_context_t *transaction_context;
+    void *custom_sampling_context; // TODO what type should this be?
```
JoshuaMoelans, line 9: Do we want to provide something similar to [the cocoa SDK](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentrySamplingContext.m#L16), or is it up to the users to define what type and how they'll use it in the tracesSampler callback?

e.g.
```c
typedef struct {
    char *key;
    void *value;
} custom_sampling_context_entry_t;

typedef struct {
    custom_sampling_context_entry_t *entries;
    size_t entry_count;
} custom_sampling_context_t;
```

supervacuus: No, this should be a `void*` since that should be able to point to whatever the client code wants to make available for the callback. There is no need to create and explicitly fill yet another data structure. It is also consistent with the context usage in all other callbacks we provide.

src/sentry_sampling_context.h
```diff
@@ -0,0 +1,19 @@
...
+} sentry_sampling_context_t;
+
+// TODO add refcounting for this; users might want to reuse this
```
JoshuaMoelans, line 12: Do these sampling contexts need refcounting?

src/sentry_core.c
```diff
@@ -875,6 +881,8 @@ sentry_transaction_start_ts(sentry_transaction_context_t *opaque_tx_cxt,
     sentry_value_remove_by_key(tx, "timestamp");
 
     sentry__value_merge_objects(tx, tx_cxt);
+    // TODO construct a sampling context based on tx_cxt and
```
JoshuaMoelans, line 884: Do we construct this based on the given `sampling_context` (of type `sentry_value_t`) or should this passed-in value already be of the new `sentry_sampling_context_t` type (which the user must then construct before calling `sentry_transaction_start`?

JoshuaMoelans: In commit https://github.com/getsentry/sentry-native/pull/1108/commits/78f01c93269f45b9da083d8a4153093c6de4d8ed I implemented it as follows:
- The **user constructs a custom sampling context** of type `sentry_value_t` for a transaction, filling it with key-value pairs.
- In `sentry_transaction_start_ts` we **take this custom context and the transaction context, and construct a sampling context** which gets passed into the sampling decision function `sentry__should_send_transaction(tx_cxt, sampling_ctx); `
- There we **call the callback function** as provided in `options->traces_sampler` if it exists, and pass the created `sampling_ctx` to it.
- Now the user can **use their defined callback** like below to fetch data from their custom sampling context to make a sampling decision:
```c
static double
traces_sampler_callback(sentry_sampling_context_t *sampling_ctx)
{
    if (sentry_value_as_int32(sentry_value_get_by_key(
            *sentry_sampling_context_get_custom_context(sampling_ctx), "b"))
        == 42) {
        return 1;
    }
    return 0;
}
```

examples/example.c
```diff
@@ -403,8 +418,15 @@ main(int argc, char **argv)
...
+        sentry_value_t custom_sampling_ctx = sentry_value_new_object();
+        sentry_value_set_by_key(
+            custom_sampling_ctx, "a", sentry_value_new_string("first_value"));
+        sentry_value_set_by_key(
+            custom_sampling_ctx, "b", sentry_value_new_int32(42));
+        sentry_value_incref(custom_sampling_ctx);
         sentry_transaction_t *tx
-            = sentry_transaction_start(tx_ctx, sentry_value_new_null());
+            = sentry_transaction_start(tx_ctx, custom_sampling_ctx);
+        sentry_value_decref(custom_sampling_ctx);
```
JoshuaMoelans, line 429: ASAN complains about this object leaking memory, but I'm not sure if we can expect the users to always provide a `sentry_value_t` custom sampling context (currently that is the required argument type for the `transaction_start` because it was already that way before, but in the new sampling context struct we allow for `void *`.

I think we maybe should change the argument type for the custom context to also be `void *` then.

supervacuus: Maybe I also misunderstood the design idea behind the custom context in the `traces_sampler` scenario. I thought of it more as a context-access parameter for the callback (similar to our `on_crash` and `before_sen` hook) rather than a value used in the context of the remaining performance environment. Maybe it should be a `sentry_value_t`.

Let's not fix the ASAN issue, but I'd like to review the custom context first. I might have given you the wrong design guidance here.

src/sentry_sampling_context.c
```diff
@@ -0,0 +1,28 @@
...
+#include "sentry_sampling_context.h"
+#include <sentry_alloc.h>
+#include <stdlib.h>
+
+sentry_sampling_context_t *
+sentry_sampling_context_new(sentry_transaction_context_t *transaction_context)
+{
+    sentry_sampling_context_t *context = SENTRY_MAKE(sentry_sampling_context_t);
+    if (context) {
+        context->transaction_context = transaction_context; // todo incref?
```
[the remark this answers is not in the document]
JoshuaMoelans: not anymore, since we replaced it by a designated initializer

examples/example.c
JoshuaMoelans: For now we only allow for `sentry_value_t` type custom contexts. If there is every an ask for more flexibility, we can reconsider the `void *` implementation.

examples/example.c
```diff
@@ -403,8 +418,14 @@ main(int argc, char **argv)
...
         }
+        sentry_value_t custom_sampling_ctx = sentry_value_new_object();
+        sentry_value_set_by_key(
+            custom_sampling_ctx, "a", sentry_value_new_string("first_value"));
+        sentry_value_set_by_key(
+            custom_sampling_ctx, "b", sentry_value_new_int32(42));
         sentry_transaction_t *tx
-            = sentry_transaction_start(tx_ctx, sentry_value_new_null());
+            = sentry_transaction_start(tx_ctx, custom_sampling_ctx);
+        sentry_value_decref(custom_sampling_ctx);
```
JoshuaMoelans, line 428: This needs to be documented properly, since it is necessary to `decref` to avoid leaks. We alternatively could `decref` inside the transaction start, just like we do with the `tx_ctx`, but users might want to reuse the one custom context multiple times.

I'm in favor of the first option, providing a clear example in the docs of this necessary `decref` cleanup

examples/example.c
```diff
@@ -28,6 +28,17 @@
 #    define sleep_s(SECONDS) sleep(SECONDS)
 #endif
 
+static double
+traces_sampler_callback(sentry_sampling_context_t *sampling_ctx)
+{
+    if (sentry_value_as_int32(sentry_value_get_by_key(
+            sentry_sampling_context_get_custom_context(sampling_ctx), "b"))
```
JoshuaMoelans, line 35: We are currently passing in this sampling context consisting of custom context (`sentry_value_t`) and a transaction context (`sentry_transaction_context_t`), but the latter is not all that useful as we can't get any data from it.

 We have the following setters `transaction_context_set_name` & `transaction_context_set_operation` but no getter functions (e.g. for these `name`/`operation` values or `trace_id`/`parent_span_id` when updating from header).

I'm not sure if adding access to transaction context data should be part of this PR (or even happen in the first place), but it does feel like something worth considering.

src/sentry_core.c
```diff
@@ -457,9 +459,14 @@ sentry__should_send_transaction(sentry_value_t tx_cxt)
...
     SENTRY_WITH_OPTIONS (options) {
-        send = sentry__roll_dice(options->traces_sample_rate);
-        // TODO(tracing): Run through traces sampler function if rate is
-        // unavailable.
+        if (options->traces_sampler) {
+            double result
+                = ((double (*)(void *))options->traces_sampler)(sampling_ctx);
+            send = sentry__roll_dice(result);
+        } else {
+            // TODO if there is a parent sampling decision, use it
```
JoshuaMoelans, line 467: I think we already do? In `transaction_context_update_from_header_n` we [get the parent sampling decision](https://github.com/getsentry/sentry-native/blob/master/src/sentry_tracing.c#L262) if it exists, and in this `should_send_transaction` we [check if the `tx_cxt` has this key already set ](https://github.com/getsentry/sentry-native/blob/master/src/sentry_core.c#L453-L456)(which, if the parent had a sampling decision, will be the case).

JoshuaMoelans: + following the develop docs section on [precedence](https://docs.sentry.io/platforms/native/configuration/sampling/#precedence) the order should be `passed in decision` (not supported) > `traces_sampler` > parent decision > `traces_sample_rate`, so if I understand the above fetching of the `"sampled"` key correctly, we should change the order.

examples/example.c
supervacuus: It should exist because the transaction context would otherwise be useless inside the callback. The question is how much of it should be exposed and whether we provide specialized accessors for each item or a getter to access the inner.

@markushi, what are typical things people filter on? Does providing an interface to a subset make sense, or should we expose the whole inner value?

I would add this to this PR and would prefer to provide access to the inner value, which we can freeze before returning.

examples/example.c
supervacuus: I think there are two opposing arguments:

* A custom sampling context should exist close to where this "context" is. Thus, it could make sense to pass ownership on to us and let users create a new custom sampling context wherever required.
* A custom sampling context should be created only once and reused throughout the code. This requires client ownership, but the upside is fewer memory (de-)allocations.

With a fresh view of the current flow, the better model is to let us take ownership. This approach doesn't require the client to track usage, and they won't need to maintain access to their context references throughout their code. If code duplication of the context construction is a concern, nothing stops them from extracting that bit into a function.

src/sentry_core.c
supervacuus: Yes to all adaptations. 😄

supervacuus (changes requested):
Great work up to now, but a few places still need to be fixed/clarified before we can merge this. A few more tests would also make sense, especially when considering changing the order of the sampling mechanisms.

examples/example.c
```diff
@@ -31,6 +31,14 @@
...
 traces_sampler_callback(sentry_sampling_context_t *sampling_ctx)
 {
+    if (!sentry_value_is_null(
+            sentry_sampling_context_get_parent_sampled(sampling_ctx))) {
+        if (sentry_value_is_true(
+                sentry_sampling_context_get_parent_sampled(sampling_ctx))) {
+            return 0.8; // high sample rate for children of sampled transactions
+        }
+        return 0; // parent is not sampled
+    }
```
JoshuaMoelans, line 41: I'm not a huge fan of how we expose this, but also couldn't come up with a cleaner solution. We will always have to differentiate between a null decision or a true/false decision, so instead of having the user call `sentry_value_is_null` and `sentry_value_is_true` we could make two functions
- `sentry_sampling_context_has_parent_sampled_decision` -> returning true if there is a value
- `sentry_sampling_context_get_parent_sampled` -> returning the value itself (but what if it doesn't exist?)

Which abstracts away the function compositions, but IMO this could be more confusing due to the naming thereof. Since we already expose `sentry_sampling_context_get_custom_context` which users will have to perform `sentry_value_...` functions on, I think we can expect them to do the same for the parent sampling decision (when provided with an example like this in the docs).

supervacuus: To be clear, the opaque type is more extensible. However, it introduces a lot of public interface for a singular use case. Since we now cover even the parent and the custom context provides users with a more generic way to pass in data, I think the extensibility aspect can be questioned vs the wordy interface it requires.

JoshuaMoelans: Do you mean we should change
```c
typedef double (*sentry_traces_sampler_function)(
    sentry_sampling_context_t *sampling_ctx);
```
to something like
```c
typedef double (*sentry_traces_sampler_function)(
    sentry_transaction_context_t *transaction_ctx,
    sentry_value_t custom_sampling_ctx,
    bool *parent_sampled);
```
where we also change `parent_sampled` to a `bool` pointer, so we can still differentiate between a non-existing parent sampling decision (it pointing to `NULL`) or an existing one (so we point to a location with true/false) instead of using the `sentry_value_t`?

supervacuus: If a missing parent doesn't mean plain `false,` we must find a ternary solution. Passing in a `bool` pointer (ideally, `const`, signifying a non-mutating parameter) is a sensible option. 

Of course, this still requires two checks, but that is the essential complexity of the `traces_sampler` spec, and users can express it using language syntax rather than function calls. Now that the opaque interface is gone, I think querying for `is_null` or `is_true` is less hassle, so a `sentry_value_t` would also be acceptable. I prefer a boolean pointer to a `sentry_value_t` because it encodes the right amount of meaning.

We could also consider passing in the transaction context as a frozen `sentry_value_t` because the opaque type doesn't contain more than `inner` and the context isn't used to accumulate state although fields are added during transaction construction (`parent_span_id`, although I am not sure what you'd want to use this for in sampling). However, I am less sure about this because it exposes internals (although I question whether those will change before we make considerable other changes). Also, `sentry_transaction_context_t` is already exposed as an opaque type, so providing getters should be fine.

JoshuaMoelans: For now I replaced the function typedef to the one I suggested above (taking in a transaction ctx, custom sampling ctx and parent sampling decision pointer). 

I think we can keep the `sentry_transaction_context_t` type, and provide getters for the things we already have setters for (without exposing the `inner` value).

src/sentry_core.c
```diff
@@ -448,18 +449,36 @@ sentry__capture_event(sentry_value_t event)
...
+            if (sampling_ctx->parent_sampled != NULL) {
+                send = *sampling_ctx->parent_sampled;
+            } else {
+                send = sentry__roll_dice(options->traces_sample_rate);
+            }
+        }
+    }
+    if (sampling_ctx->parent_sampled != NULL) {
+        sentry_free(sampling_ctx->parent_sampled);
+        sampling_ctx->parent_sampled = NULL;
```
supervacuus, line 481: You can eliminate the `free` by pointing to the stack variable. Writing `NULL` at the end is still excellent citizenship.

examples/example.c
```diff
@@ -28,6 +28,25 @@
 #    define sleep_s(SECONDS) sleep(SECONDS)
 #endif
 
+static double
+traces_sampler_callback(sentry_transaction_context_t *transaction_ctx,
+    sentry_value_t custom_sampling_ctx, const bool *parent_sampled)
+{
+    (void)transaction_ctx; // unused for now
```
supervacuus, line 35: Primarily because the getters are not exposed, right? Could we start by exposing getters for `name` and `operation` and forget about the external trace inheritance? This way, we could provide an example that does something side-effect free as a first step like
```suggestion
    if (strcmp(sentry_transaction_context_get_name(tx_ctx), "little.teapot")
        != 0) {
        return 0;
    }
```

CONTRIBUTING.md
```diff
@@ -148,6 +148,8 @@ The example currently supports the following commands:
 - `stack-overflow`: Provokes a stack-overflow.
 - `http-proxy`: Uses a localhost `HTTP` proxy on port 8080.
 - `socks5-proxy`: Uses a localhost `SOCKS5` proxy on port 1080.
+- `capture-transaction`: Captures a transaction.
```
supervacuus, line 151: thx 🙏

src/sentry_core.c
```diff
@@ -448,18 +449,36 @@ sentry__capture_event(sentry_value_t event)
...
+                    sampling_ctx->custom_sampling_context,
+                    sampling_ctx->parent_sampled);
+            send = sentry__roll_dice(result);
+        } else {
+            if (sampling_ctx->parent_sampled != NULL) {
+                send = *sampling_ctx->parent_sampled;
+            } else {
+                send = sentry__roll_dice(options->traces_sample_rate);
+            }
+        }
```
supervacuus, line 477: 🥇

supervacuus (changes requested):
Thanks for the test effort ❤️. There are only minor things left.

examples/example.c
JoshuaMoelans: Yes indeed, I think exposing the getters in this PR makes sense (as there's now a reason for them to be exposed).

JoshuaMoelans: Added the getters in https://github.com/getsentry/sentry-native/pull/1108/commits/a9d72219f122893d1f128630cbddabd5bf8adc9b and tested them in the context of the traces sampler in https://github.com/getsentry/sentry-native/pull/1108/commits/370b16264e2476e3dada827829e69ce659e7b880

src/sentry_core.c
```diff
@@ -448,18 +449,35 @@ sentry__capture_event(sentry_value_t event)
...
     SENTRY_WITH_OPTIONS (options) {
-        send = sentry__roll_dice(options->traces_sample_rate);
-        // TODO(tracing): Run through traces sampler function if rate is
-        // unavailable.
+        if (options->traces_sampler) {
+            const double result
+                = ((sentry_traces_sampler_function)options->traces_sampler)(
+                    sampling_ctx->transaction_context,
+                    sampling_ctx->custom_sampling_context,
+                    sampling_ctx->parent_sampled);
```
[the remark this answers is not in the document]
JoshuaMoelans: Casting didn't work for me (seems like going from `bool *` to `int *` just reads more than it should, giving strange int values). Hence I added another place to store the int temporarily (see [sentry_core.c](https://github.com/getsentry/sentry-native/pull/1108/files#diff-45ef05f9e59b750d3d615585ae87121db85ca379ca226f3b26ad809f85703783R461-R464))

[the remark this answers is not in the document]
supervacuus: Yeah, it is much safer to convert explicitly.

src/sentry_core.c
```diff
@@ -448,18 +448,40 @@ sentry__capture_event(sentry_value_t event)
...
-        return sentry_value_is_true(context_setting);
-    }
+    sentry_value_t context_setting = sentry_value_get_by_key(tx_ctx, "sampled");
+    bool sampled = sentry_value_is_null(context_setting)
+        ? false
+        : sentry_value_is_true(context_setting);
+    sampling_ctx->parent_sampled
+        = sentry_value_is_null(context_setting) ? NULL : &sampled;
 
+    int parent_sampled_int; // bool to int conversion
```
markushi, line 461: `nit`: How about being more explicit here? I see the value is only read if has been previously set, so it should be fine. But in case someone makes changes to the code below in future, this could easily slip through.
```suggestion
    int parent_sampled_int = 0; // bool to int conversion
```

markushi (approved):
LGTM, nice!

ShawnCZek:
From what I understand, this part of the documentation is now obsolete:

https://github.com/getsentry/sentry-native/blob/092de88b7369a8d39088003d3773323a7ca66f50/include/sentry.h#L1738-L1741

supervacuus:
> From what I understand, this part of the documentation is now obsolete:

The first sentence is still valid. The second sentence no longer is.

Issue #787, opened 2023-01-12
Honor http_proxy environment variables


supervacuus:
A user of the Native SDK currently must configure an HTTP proxy themselves via `sentry_options_set_http_proxy()`, which also means their application must parse the `http_proxy` (and potentially `https_proxy`, `no_proxy`) variables. Since these are fairly standard in all supported platforms, we should provide the mechanism ootb. I think other SDKs already support this.

PR #1111, opened 2025-01-09
feat: honor `http_proxy` environment variables

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/787

## Overview

This started as an additional option to configure, but testing revealed that (some) transports already honor the variable. This PR is now mainly an extensive testing effort to reason about the boundaries & quirks of `http_proxy`.

One major addition was made to the [crashpad handler](https://github.com/getsentry/sentry-native/blob/joshua/feat/http_proxy_from_env_variables/src/backends/sentry_backend_crashpad.cpp#L442-L453) which now also reads the environment variable if it was set.

## Done
- [x] ~~add option~~ removed it in favor of relying on transports
- [x] add tests
- [x] document what types of proxy strings work and which don't
- [x] add to [docs](https://docs.sentry.io/platforms/native/configuration/transports/) as to which variables we read (and what order they are read) (tracking on https://github.com/getsentry/sentry-docs/pull/12424)
- [x] document how to bypass proxy-from-env (set `options->proxy` to "")
- [x] rework section about 'fallback' https://docs.sentry.io/platforms/native/configuration/transports/
- [x] merge related Crashpad PR https://github.com/getsentry/crashpad/pull/119

### DIFF
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -447,7 +447,7 @@ crashpad_backend_startup(
                                            : "";
 #ifdef SENTRY_PLATFORM_LINUX
     // explicitly set an empty proxy to avoid reading from env. vars. on Linux
-    if (options->proxy == "") {
+    if (strcmp(options->proxy, "") == 0) {
         proxy_url = "<empty>";
     }
 #endif
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -445,6 +445,12 @@ crashpad_backend_startup(
     const char *proxy_url = options->proxy ? options->proxy
         : env_proxy                        ? env_proxy
                                            : "";
+#ifdef SENTRY_PLATFORM_LINUX
+    // explicitly set an empty proxy to avoid reading from env. vars. on Linux
+    if (options->proxy == "") {
+        proxy_url = "<empty>";
+    }
+#endif
     bool success = data->client->StartHandler(handler, database, database,
         minidump_url ? minidump_url : "", proxy_url, annotations, arguments,
         /* restartable */ true,
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -448,7 +448,7 @@ crashpad_backend_startup(
 #ifdef SENTRY_PLATFORM_LINUX
     // explicitly set an empty proxy to avoid reading from env. vars. on Linux
     if (options->proxy == "") {
-        proxy_url = "<empty>"
+        proxy_url = "<empty>";
     }
 #endif
     bool success = data->client->StartHandler(handler, database, database,
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -439,12 +439,11 @@ crashpad_backend_startup(
     if (minidump_url) {
         SENTRY_DEBUGF("using minidump URL \"%s\"", minidump_url);
     }
-    char *proxy_url;
-    ;
+    const char *proxy_url;
 #ifdef SENTRY_PLATFORM_MACOS
     proxy_url = getenv("https_proxy"); // other platforms do this already
 #endif
-    proxy_url = options->proxy ? options->proxy : proxy_url;
+    proxy_url = options->proxy ? options->proxy : proxy_url ? proxy_url : "";
     bool success = data->client->StartHandler(handler, database, database,
         minidump_url ? minidump_url : "", proxy_url, annotations, arguments,
         /* restartable */ true,
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -439,9 +439,14 @@ crashpad_backend_startup(
     if (minidump_url) {
         SENTRY_DEBUGF("using minidump URL \"%s\"", minidump_url);
     }
+    char *proxy_url;
+    ;
+#ifdef SENTRY_PLATFORM_MACOS
+    proxy_url = getenv("https_proxy"); // other platforms do this already
+#endif
+    proxy_url = options->proxy ? options->proxy : proxy_url;
     bool success = data->client->StartHandler(handler, database, database,
-        minidump_url ? minidump_url : "", options->proxy ? options->proxy : "",
-        annotations, arguments,
+        minidump_url ? minidump_url : "", proxy_url, annotations, arguments,
         /* restartable */ true,
         /* asynchronous_start */ false, attachments);
     sentry_free(minidump_url);
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -447,7 +447,7 @@ crashpad_backend_startup(
                                            : "";
 #ifdef SENTRY_PLATFORM_LINUX
     // explicitly set an empty proxy to avoid reading from env. vars. on Linux
-    if (strcmp(options->proxy, "") == 0) {
+    if (options->proxy && strcmp(options->proxy, "") == 0) {
         proxy_url = "<empty>";
     }
 #endif
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -439,15 +439,11 @@ crashpad_backend_startup(
     if (minidump_url) {
         SENTRY_DEBUGF("using minidump URL \"%s\"", minidump_url);
     }
-    auto proxy_url = "";
-    if (options->dsn->is_secure) {
-        proxy_url = getenv("https_proxy");
-    } else {
-        proxy_url = getenv("http_proxy");
-    }
-    proxy_url = options->proxy ? options->proxy
-        : proxy_url != NULL    ? proxy_url
-                               : "";
+    const char *env_proxy
+        = getenv(options->dsn->is_secure ? "https_proxy" : "http_proxy");
+    const char *proxy_url = options->proxy ? options->proxy
+        : env_proxy                        ? env_proxy
+                                           : "";
     bool success = data->client->StartHandler(handler, database, database,
         minidump_url ? minidump_url : "", proxy_url, annotations, arguments,
         /* restartable */ true,
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -445,12 +445,6 @@ crashpad_backend_startup(
     const char *proxy_url = options->proxy ? options->proxy
         : env_proxy                        ? env_proxy
                                            : "";
-#ifdef SENTRY_PLATFORM_LINUX
-    // explicitly set an empty proxy to avoid reading from env. vars. on Linux
-    if (options->proxy == "") {
-        proxy_url = "<empty>";
-    }
-#endif
     bool success = data->client->StartHandler(handler, database, database,
         minidump_url ? minidump_url : "", proxy_url, annotations, arguments,
         /* restartable */ true,
--- a/src/backends/sentry_backend_crashpad.cpp
+++ b/src/backends/sentry_backend_crashpad.cpp
@@ -439,11 +439,13 @@ crashpad_backend_startup(
     if (minidump_url) {
         SENTRY_DEBUGF("using minidump URL \"%s\"", minidump_url);
     }
-    const char *proxy_url;
+    auto proxy_url = "";
 #ifdef SENTRY_PLATFORM_MACOS
     proxy_url = getenv("https_proxy"); // other platforms do this already
 #endif
-    proxy_url = options->proxy ? options->proxy : proxy_url ? proxy_url : "";
+    proxy_url = options->proxy ? options->proxy
+        : proxy_url != NULL    ? proxy_url
+                               : "";
     bool success = data->client->StartHandler(handler, database, database,
         minidump_url ? minidump_url : "", proxy_url, annotations, arguments,
         /* restartable */ true,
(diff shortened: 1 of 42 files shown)


src/sentry_core.c
```diff
@@ -138,6 +138,10 @@ sentry_init(sentry_options_t *options)
             "the provided DSN \"%s\" is not valid", raw_dsn ? raw_dsn : "");
     }
 
+    if (options->read_proxy_from_environment) {
+        sentry__set_proxy_from_environment(options);
+    }
```
JoshuaMoelans, line 143: With this new option, I'm not sure we still need to support `no_proxy`; if users want to use the environment variable value in their own code, they can just not set this flag to true (which is the default anyway).

supervacuus: Yes and no. You always have to consider that this is a feature that the users of our users require from them.

So, for instance, if your application ignores the `http_proxy` env-vars, your users don't care about the topic and accept that they have to manually configure the proxy config for that application (or even have no proxy config at all).

At some point, you might provide a UI to your users that does something like the following:

```
(   ) manually specify HTTP proxy: _________
( * ) read proxy from the environment.
```

You can route this directly to our proposed options interface. However, once your users have configured the app to read from the environment, they will want to stay with it because it allows them to have a centrally managed proxy configuration rather than defining it separately in every application.

This is where `no_proxy` enters the picture. So, yes, overwriting the manual proxy-config of that app with an empty proxy solves the problem, but then the users of our users are essentially back to square 1. Again, this is not a requirement for an initial implementation of that feature but a likely follow-up.

src/sentry_options.c
```diff
@@ -276,6 +276,29 @@ sentry_options_get_http_proxy(const sentry_options_t *opts)
...
+}
+
+void
+sentry__set_proxy_from_environment(sentry_options_t *opts)
+{
+    const char *https_proxy = getenv("https_proxy");
+    if (https_proxy) {
+        sentry_options_set_proxy(opts, https_proxy);
+    } else {
+        sentry_options_set_proxy(opts, getenv("http_proxy"));
```
supervacuus, line 298: What happens if the env contains `http_proxy=http://user:pass@my-proxy:12345/`? Do we support this everywhere? Shouldn't we parse the URL coming from the env?

JoshuaMoelans: Afaict this seems to work, but it makes sense to add tests that add `--proxyauth=user:pass` to the mitmproxy run command to make sure this is the case.

supervacuus: If it works on all configs and thus doesn't require us to clean URLs, we should have tests to back that claim.

src/sentry_options.c
```diff
@@ -276,6 +276,29 @@ sentry_options_get_http_proxy(const sentry_options_t *opts)
...
+int
+sentry_options_get_read_proxy_from_environment(const sentry_options_t *opts)
+{
+    return opts->read_proxy_from_environment;
+}
+
+void
+sentry__set_proxy_from_environment(sentry_options_t *opts)
+{
+    const char *https_proxy = getenv("https_proxy");
```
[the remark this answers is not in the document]
supervacuus: I am fine with only accepting the lower-case versions, but since this is a "thing", we should clearly document the behavior.

src/transports/sentry_transport_winhttp.c
```diff
@@ -71,7 +105,12 @@ sentry__winhttp_transport_start(
     // ensure the proxy starts with `http://`, otherwise ignore it
     if (opts->proxy && strstr(opts->proxy, "http://") == opts->proxy) {
         const char *ptr = opts->proxy + 7;
```
JoshuaMoelans, line 107: Apparently our Windows transport doesn't take https:// proxies. This should be documented, since we now also read the `HTTPS_PROXY` environment variable, which on this platform would get ignored.

src/transports/sentry_transport_winhttp.c
```diff
@@ -51,10 +53,47 @@ sentry__winhttp_bgworker_state_free(void *_state)
     sentry__dsn_decref(state->dsn);
     sentry__rate_limiter_free(state->ratelimiter);
     sentry_free(state->user_agent);
+    sentry_free(state->proxy_username);
+    sentry_free(state->proxy_password);
     sentry_free(state->proxy);
     sentry_free(state);
 }
 
+// Function to extract and set credentials TODO replace with `sentry__url_parse`
```
JoshuaMoelans, line 62: Since `sentry___url_parse` was built to handle DSNs, it won't parse the general `http_proxy` formats (e.g. the one we have in our test, `http://user:password@127.0.0.1:8080`, will fail the parsing as it has no `/path` behind the port). We would either need to add a flag to the parser signaling whether it needs the `/path` section or not, or we can keep this `set_proxy_credentials` function.

supervacuus: I would add the flag as a parameter. We should let setting the proxy fail anyway after the parsing. If a proxy requires authentication, then it won't work on Windows. That should generate a `WARN` log.

supervacuus (changes requested):

supervacuus (approved):
👍

tests/test_integration_http.py
```diff
@@ -650,13 +660,27 @@ def test_capture_proxy(cmake, httpserver, run_args, proxy_status):
...
+                os.environ["http_proxy"] = f"http://{auth}localhost:8080"
+            elif run_args == ["socks5-proxy"]:
+                os.environ["http_proxy"] = f"socks5://{auth}localhost:1080"
 
+        httpserver.expect_request("/api/123456/envelope/").respond_with_data("OK")
+        current_run_arg = run_args[0]
+        if proxy_auth == ["on"]:
+            current_run_arg += "-auth"
+        if proxy_from_env == ["proxy-from-env"]:
+            current_run_arg = "proxy-from-env"  # overwrite args if proxy-from-env is set (e.g. don't manually set)
```
JoshuaMoelans, line 678: In cURL transport we don't even need to call our proxy-from-env setter, since it will attempt to read the environment variable automatically https://everything.curl.dev/usingcurl/proxies/env.html . 
This kind of makes this feature only really useful for Windows.

supervacuus: Is this also true for the library (i.e., `libcurl`)? There are a couple of such features where the library only provides the mechanism, but the reading from env is then implemented in the application (i.e., `curl`).

JoshuaMoelans: Seems like it is https://curl.se/libcurl/c/libcurl-env.html

tests/test_integration_http.py
```diff
@@ -625,7 +625,12 @@ def test_capture_minidump(cmake, httpserver):
     ],
 )
 @pytest.mark.parametrize("proxy_status", [(["off"]), (["on"])])
-def test_capture_proxy(cmake, httpserver, run_args, proxy_status):
+@pytest.mark.parametrize("proxy_auth", [(["off"]), (["on"])])
+@pytest.mark.parametrize("proxy_IPv", [(["4"]), (["6"])])
+@pytest.mark.parametrize("proxy_from_env", [(["proxy-from-env"]), ([""])])
```
JoshuaMoelans, line 630: This test is becoming exponentially larger. It might make sense to split apart some test cases to not cover the full matrix at once. There are still some untested cases:

- [ ] What if `proxy-from-env` but the env. variable is empty?  -> expected is that if there was a manually set proxy, it gets used.
- [ ] Test whether`proxy-from-env` takes precedence over manually set proxy if both are given.

tests/test_integration_crashpad.py
```diff
@@ -39,6 +39,49 @@ def test_crashpad_capture(cmake, httpserver):
...
+                assert child.returncode  # well, it's a crash after all
+        except AssertionError as e:
+            if port_correct:
+                raise e
+            else:
+                assert len(httpserver.log) == 0
+                return
+
+        assert waiting.result
+        assert len(httpserver.log) == 1
```
JoshuaMoelans, line 76: We currently can't know if the event was sent over the proxy, or if it used a fallback on the existing connection. It might make sense to listen to the port the proxy is on & confirm it passes through there.

On the other hand, the `port_correct` kind of already checks this by seeing if the example reads the proxy from the env (which is now 'wrong' and doesn't point to the running proxy server) and therefor no event ends up in the httpserver.

JoshuaMoelans:
Seems like the latests changes broke some tests (these seemed broken locally but I didn't see why). Will investigate 🕵️

tests/test_integration_crashpad.py
supervacuus: After talking with you, I prefer to check the `mitmdump` output and count the logs there. We can do this at the end when `proxy_process` is terminated. This way, we don't have to guess whether another route by-passing the proxy allowed the request to succeed. This is also true for the other proxy tests.

Only thinking out aloud, maybe I missed a detail, but that would be the route I would choose first.

tests/test_integration_http.py
```diff
@@ -611,6 +611,114 @@ def test_capture_minidump(cmake, httpserver):
     assert_minidump(envelope)
 
 
+@pytest.mark.parametrize("port_correct", [True, False])
+def test_proxy_from_env(cmake, httpserver, port_correct):
+    # TODO how can we test this? If it doesn't get read but it's there, we don't know (since it'll just fallback)
+    #     can we listen to mitmdump's port and see if it gets traffic?
```
JoshuaMoelans, line 617: We now have a way to read the stdout after terminating `mitmdump`, and we can check it as such:
```c
proxy_process.terminate()
proxy_process.wait()
stdout, stderr = proxy_process.communicate()
assert "POST" in stdout
```
It makes sense to have this check for all of our proxy tests, as this is a more rigorous check on what goes through/around our proxy server.

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -439,9 +439,13 @@ crashpad_backend_startup(
     if (minidump_url) {
         SENTRY_DEBUGF("using minidump URL \"%s\"", minidump_url);
     }
+    const char *proxy_url;
+#ifdef SENTRY_PLATFORM_MACOS
+    proxy_url = getenv("https_proxy"); // other platforms do this already
+#endif
+    proxy_url = options->proxy ? options->proxy : proxy_url ? proxy_url : "";
```
JoshuaMoelans, line 446: Currently we only look at `https_proxy`, but maybe if this is empty we should check `http_proxy` as well? (although these two should ideally hold the same value). 
Afaict cURL doesn't use `http_proxy` when trying to send to a configured `https://` DSNs, so I would keep our implementation as similar as possible.

supervacuus: Yeah, we talked about this once. I think a "no fallback" if `DSN` has a `https`-scheme use `https_proxy`; otherwise, use `http_proxy` is enough.

JoshuaMoelans:
After the last commit, tests _only_ fail on Windows, so will investigate these failing asserts locally.

JoshuaMoelans:
We still need to find a way to work around [WinINet bypassing the localhost->localhost proxy](https://learn.microsoft.com/en-us/windows/win32/wininet/enabling-internet-functionality?redirectedfrom=MSDN#listing-the-proxy-bypass:~:text=By%20default%2C%20WinINet%20will%20bypass%20the%20proxy%20for%20requests%20that%20use%20the%20host%20names%20%22localhost%22%2C%20%22loopback%22%2C%20%22127.0.0.1%22%2C%20or%20%22%5B%3A%3A1%5D%22.%20This%20behavior%20exists%20because%20a%20remote%20proxy%20server%20typically%20will%20not%20resolve%20these%20addresses%20properly.) in our test environment. Locally this was tested by adding a line to the Windows hosts file

```
127.0.0.1        sentry_native.test 
```

JoshuaMoelans:
Also, now that we test whether the proxy actually gets used when we prompt it to, this crashpad socks5-proxy test fails on macOS (see [runner output](https://github.com/getsentry/sentry-native/actions/runs/13117507632/job/36595294578?pr=1111)). It seems like the proxy isn't used, but when trying this with a non-localhost DSN it seems to work (similar to the remaining Windows failures)).
```
FAILED tests/test_integration_crashpad.py::test_crashpad_crash_proxy[True-run_args1] - AssertionError
```

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -439,9 +439,17 @@ crashpad_backend_startup(
...
     }
+    auto proxy_url = "";
+    if (strncmp(options->dsn->raw, "https", 5) == 0) {
+        proxy_url = getenv("https_proxy");
+    } else if (strncmp(options->dsn->raw, "http", 4) == 0) {
+        proxy_url = getenv("http_proxy");
+    }
+    proxy_url = options->proxy ? options->proxy
+        : proxy_url != NULL    ? proxy_url
+                               : "";
```
JoshuaMoelans, line 450: Do we need a test for this?

JoshuaMoelans (commented):
Almost there, just a few remarks left.

src/backends/sentry_backend_crashpad.cpp
JoshuaMoelans: yes, we want to test the following:
- setting `https_proxy` when the DSN is `http`, we expect the proxy to remain free from requests
- setting the `proxy` option to `""` when the `http(s)_proxy` variables are set, should lead to the proxy to remain free from requests too

tests/leaks.txt
```diff
@@ -11,3 +11,5 @@ leak:SCDynamicStoreCopyProxiesWithOptions
 # This is a known issue in ASAN packaged with llvm15/16 (and below): https://github.com/google/sanitizers/issues/1501
 # I cannot reproduce it with the current brew llvm package (18.1.7). TODO: remove when GHA macOS runner image updates the llvm package.
 leak:realizeClassWithoutSwift
+
+leak:Curl_getaddrinfo_ex
```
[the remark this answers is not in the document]
supervacuus: @JoshuaMoelans, can I ask you to send me this specifically via some channel?

PR #1124, opened 2025-01-22
Auto detect latest GDK and Windows SDK

PlasmaDev5:
This PR modifies the CMake configuration to automatically detect the GDK version and WindowsSDK. This method ensures the selection of the latest installed version of the SDK. Consequently, the minimum required CMake version has been increased to 3.18 to utilize the list compare behavior.


supervacuus (changes requested):
Looks great, only minor things. 

I wondered if removing any explicit settings regarding the Windows SDK wouldn't lead to the same outcome (i.e., `CMake` will select the latest Windows SDK installed by default).

Then, we would only have the `GDK` detection code. However, since we need the SDK paths anyway, it is okay to couple it with the version detection.

supervacuus (approved):

Issue #1123, opened 2025-01-22
sentry-native crashpad build fails on Linux arm64: arm64 does not support xpaclri


aokerholm-cobot:
**Description**

sentry-native crashpad build fails on Linux arm64

```
/tmp/ccxvkBk7.s: Assembler messages:
/tmp/ccxvkBk7.s:27: Error: selected processor does not support `xpaclri'
make[2]: *** [crashpad_build/util/CMakeFiles/crashpad_util.dir/build.make:930: crashpad_build/util/CMakeFiles/crashpad_util.dir/linux/pac_helper.cc.o] Error 1
```

**When does the problem happen**

- [x] During build
- [ ] During run-time
- [ ] When capturing a hard crash

**Environment**

- OS: Linux 5.10.104-tegra aarch64
- Compiler: g++ (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0
- CMake version and config: 3.30.2

**Steps To Reproduce**

cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSENTRY_BACKEND=crashpad
cmake --build build --parallel

**Log output**

```
$ cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSENTRY_BACKEND=crashpad
-- SENTRY_TRANSPORT=curl
-- SENTRY_BACKEND=crashpad
-- SENTRY_LIBRARY_TYPE=SHARED
-- SENTRY_SDK_NAME=
-- Found ZLIB: /usr/lib/aarch64-linux-gnu/libz.so (found version "1.2.11")
-- Configuring done (0.1s)
-- Generating done (0.1s)
-- Build files have been written to: ~/third-party/sentry-native/build
$ cmake --build build --parallel
[  0%] Building CXX object crashpad_build/tools/CMakeFiles/crashpad_tools.dir/tool_support.cc.o
[  0%] Building CXX object crashpad_build/compat/CMakeFiles/crashpad_compat.dir/linux/sys/mman_memfd_create.cc.o
[  0%] Building CXX object crashpad_build/third_party/mini_chromium/CMakeFiles/mini_chromium.dir/mini_chromium/base/files/file_path.cc.o
[  2%] Building CXX object crashpad_build/third_party/mini_chromium/CMakeFiles/mini_chromium.dir/mini_chromium/base/debug/alias.cc.o
[...]

PR #1125, opened 2025-01-24
fix: support crashpad PAC-stripping on older assemblers.

supervacuus:
fixes: #1123
details: https://github.com/getsentry/crashpad/pull/118


JoshuaMoelans (approved):
LGTM

Issue #1121, opened 2025-01-16
Set options.debug = true if we can reliably detect that the app runs in a debug build


kahest:
see https://github.com/getsentry/team-mobile/issues/201

PR #1128, opened 2025-01-27
feat: enable debug-option by default in debug-build.

supervacuus:
fixes: #1121

The implementation allows users to set the environment variable `SENTRY_DEBUG` to `0` to override.


JoshuaMoelans (approved):
LGTM!

Issue #1131, opened 2025-01-28
Value for `max_spans` defaults to `0` in `sentry_options_new`


nwhite-riot:
... despite the documentation stating that SDKs [should implement it as a default of 1000](https://develop.sentry.dev/sdk/telemetry/traces/#maxspans).

supervacuus:
Agreed. The initialization of `max_spans` was bound to setting the static `sample_rate` to a value `> 0` in the initial implementation. While this might have made sense before `parent_sampled` and `sample_tracer` were implemented, it doesn't make sense anymore. Fixing this is on our to-do list.

PR #1132, opened 2025-01-29
fix: set default `max_spans` to 1000

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/1131

- [x] set default for `sentry_options_new`
- [x] think about other occurrences of `SENTRY_SPANS_MAX`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 **Fixes**:
 
 - Allow older toolchains with assemblers that don't support PAC-stripping instructions on `aarch64` to compile `crashpad`. ([#1125](https://github.com/getsentry/sentry-native/pull/1125), [crashpad#118](https://github.com/getsentry/crashpad/pull/118))
+- Set default `max_spans` to 1000. ([#1132](https://github.com/getsentry/sentry-native/pull/1132))
 
 ## 0.7.19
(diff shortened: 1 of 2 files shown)


JoshuaMoelans:
We have three of these snippets for `transaction/span_start_child` and `span_finish`.
Even if we set the `options->max_spans` at creating the options object, I'm not sure we can just remove these?
```c
  size_t max_spans = SENTRY_SPANS_MAX;
      max_spans = options->max_spans;
  }
```

supervacuus (approved):

supervacuus:
> We have three of these snippets for `transaction/span_start_child` and `span_finish`. Even if we set the `options->max_spans` at creating the options object, I'm not sure we can just remove these?

The sole reason why those exist is that the initial implementer didn't want to scope the remaining code in `SENTRY_WITH_OPTIONS`. However, they had to declare it outside so that it was accessible, and that naturally led to potential uninitialized use, which led them to initialize it to the most sensible value. 😅 

I would leave them in there. They don't hurt anyone. If you want to get rid of them, you will have to put the entire dependent code inside the `SENTRY_WITH_OPTIONS`, which doesn't add a lot but might visualize the dependency better.

PR #1134, opened 2025-01-29
fix: allow test discovery to handle one-liner unit-tests

supervacuus:
#1093 recently broke our make/bash/powershell unit-test runner (i.e., the ones not running via `pytest`, which we use in CI) because `clang-format` decided to put the test body on the same line as the test header. Test discovery supports this now 😅 

#skip-changelog

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 all: test
 
 update-test-discovery:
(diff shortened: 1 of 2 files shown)


JoshuaMoelans (approved):
lgtm, thanks!

PR #1137, opened 2025-01-30
feat: Added `sentry_set_trace`

bitsandfoxes:
### Motivation

Looking at it from one of the upstream SDKs (i.e. `Android SDK`, `Unity SDK`) there is currently no way to connect errors and events from their environment with the events sent by `sentry-native`. 
The idea is to provide a way for those SDKs to propagate their trace to be used and applied to events. The trace ID can then be used to link those events across multiple SDKs, languages, and layers.


src/sentry_core.c
```diff
@@ -826,6 +826,19 @@ sentry_remove_fingerprint(void)
     }
 }
 
+void
+sentry_set_trace_id(const char *trace_id)
```
bitsandfoxes, line 830: `char*` is most probably not the appropriate type here. Since this is supposed to mimic the header-parsing functionality from `sentry_tracing` we could pull that out from there and reuse it here?

include/sentry.h
```diff
@@ -1493,6 +1493,12 @@ SENTRY_API void sentry_set_fingerprint_n(
  */
 SENTRY_API void sentry_remove_fingerprint(void);
 
+/**
+ * Sets the trace ID.
+ */
```
supervacuus, line 1498: Since this is now part of the public API and could be used by all clients, not only downstream SDKs, it would be great to have some context here. Just basic information on what it is used for, what clients shouldn't use it for, and a link to conceptual docs (not developer docs), so clients understand the usage scenario.

supervacuus (changes requested):
Besides a changelog, I missed a basic test for the interaction with transactions. Something like this in `test_tracing.c`:

```C
void
apply_scope_and_check_trace_context(
    sentry_options_t *options, const char *trace_id, const char *parent_span_id)
{
    // simulate scope application onto an event
    sentry_value_t event = sentry_value_new_object();
    SENTRY_WITH_SCOPE (scope) {
        sentry__scope_apply_to_event(scope, options, event, SENTRY_SCOPE_NONE);
    }

    // check that the event has a trace context
    sentry_value_t event_contexts = sentry_value_get_by_key(event, "contexts");
    TEST_CHECK(!sentry_value_is_null(event_contexts));
    TEST_CHECK(
        sentry_value_get_type(event_contexts) == SENTRY_VALUE_TYPE_OBJECT);

    sentry_value_t event_trace_context
        = sentry_value_get_by_key(event_contexts, "trace");
    TEST_CHECK(!sentry_value_is_null(event_trace_context));
    TEST_CHECK(
        sentry_value_get_type(event_trace_context) == SENTRY_VALUE_TYPE_OBJECT);

    // check trace context content
    const char *event_trace_id = sentry_value_as_string(
        sentry_value_get_by_key(event_trace_context, "trace_id"));
    TEST_CHECK_STRING_EQUAL(event_trace_id, trace_id);

    const char *event_trace_parent_span_id = sentry_value_as_string(
        sentry_value_get_by_key(event_trace_context, "parent_span_id"));
    TEST_CHECK_STRING_EQUAL(event_trace_parent_span_id, parent_span_id);

    sentry_uuid_t event_trace_span_id = sentry__value_as_uuid(
        sentry_value_get_by_key(event_trace_context, "span_id"));
    TEST_CHECK(!sentry_uuid_is_nil(&event_trace_span_id));

    sentry_value_decref(event);
}

SENTRY_TEST(set_trace_id_with_txn)
{
    // initialize SDK so we have a scope
    sentry_options_set_traces_sample_rate(options, 1.0);
    sentry_options_set_sample_rate(options, 1.0);
    sentry_init(options);

    // inject a trace via trace-header into a transaction
    const char *trace_header
        = "2674eb52d5874b13b560236d6c79ce8a-a0f9fdf04f1a63df-1";
    const char *txn_trace_id = "2674eb52d5874b13b560236d6c79ce8a";
    const char *txn_parent_span_id = "a0f9fdf04f1a63df";
    sentry_transaction_context_t *tx_ctx
        = sentry_transaction_context_new("wow!", NULL);
    sentry_transaction_context_update_from_header(
        tx_ctx, "sentry-trace", trace_header);
    sentry_transaction_t *tx
        = sentry_transaction_start(tx_ctx, sentry_value_new_null());

    // set the direct trace first
    const char *direct_trace_id = "aaaabbbbccccddddeeeeffff00001111";
    const char *direct_parent_span_id = "f0f0f0f0f0f0f0f0";
    sentry_set_trace(direct_trace_id, direct_parent_span_id);

    // events should get that trace applied
    apply_scope_and_check_trace_context(
        options, direct_trace_id, direct_parent_span_id);

    // now set a scoped transaction
    sentry_set_transaction_object(tx);

    // events should get the transaction's trace applied
    apply_scope_and_check_trace_context(
        options, txn_trace_id, txn_parent_span_id);

    sentry_transaction_finish(tx);

    // after finishing the transaction, the direct trace should hit again
    apply_scope_and_check_trace_context(
        options, direct_trace_id, direct_parent_span_id);

    sentry_close();
}
```

tests/unit/tests.inc
```diff
@@ -117,7 +119,9 @@ XX(update_from_header_no_sampled_flag)
 XX(update_from_header_null_ctx)
 XX(url_parsing_complete)
 XX(url_parsing_invalid)
+XX(url_parsing_no_path)
 XX(url_parsing_partial)
```
bitsandfoxes, line 123: I'm not sure where those are coming from.

markushi (approved):
Looks good to me, especially the Java part 🤓 
You probably need to apply the code formatter + .api file generator using 
`./gradlew spotlessApply` + `./gradlew :sentry-native-ndk:apiDump` to make CI happier.

tests/unit/tests.inc
supervacuus: Joshua probably didn't run the test discovery before merging. It is not a biggie, but those files were not yet collected.

supervacuus (approved):
If downstream, y'all are also happy, let's do this.

zsd4yr:
Question on usage -- where do the trace and parent span ids originate from?

zsd4yr:
Like are you just meant to make one? with `sentry_uuid_new_v4` or something? that does not seem right

supervacuus:
As explained in the inline docs and changelog, this function should primarily be used by downstream SDKs. Do you plan to write one?

zsd4yr:
Hey @supervacuus yes, I am following that -- but where does the originating trace and parent span id come from? Like what API creates them? 

Let's say I own an upstream SDK. How do I create the necessary parameters to this method?

zsd4yr:
For example, imagine I have 2 applications, one is my crash reporter and other is the application it is observing. My hope is to use this new API to "share a trace" between the two apps so that I can create breadcrumbs in the primary application which get married to the crash event when a crash is reported by the other application 

For that to work, I have to have some means of _creating the traceid and parent span id_ from the primary application

zsd4yr:
Ah perhaps sentry_add_breadcrumb won't work for this at all as it needs a backend. Ours is null as we generate the dump ourselves by calling breakpad directly. :(

zsd4yr:
sorry for the noise, going to just start my own dedicated issue on this topic https://github.com/getsentry/sentry-native/issues/1197

Issue #1135, opened 2025-01-29
Request: API Change / Addition to sentry_capture_minidump


zsd4yr:
RE: https://github.com/getsentry/sentry-native/pull/1067/files

Would it be possible to change / add new API here to make the result more manageable / informational? 
Some ideas
* sentry_capture_minidump should have a return value indicating whether or not it succeeded 
* perhaps the report event id would be good? Some information that can be used to lookup the actual dump 

These even ids appear to be coming back when observed in fiddler, we just need some way to get ahold of them :)

zsd4yr:
I was able to make this change and validate it locally through our sentry usage -- for whatever reason I am getting denied trying to push my changes into a branch in this repo

The changes are very simple if someone had any advise on how I can create a branch to PR them

I'll admit I'm not usually a fan of changing API surface area, but an additive change like this should be fine. If you're worried about back compatibility, we could always just make a second version of the methods too, `sentry_capture_minidump_with_return` or something
 
`sentry.h`
```c
	/**
	 * Allows capturing independently created minidumps.
	 *
	 * This generates a fatal error event, includes the scope and attachments.
	 * If the event isn't dropped by a before-send hook, the minidump is attached
	 * and the event is sent.
	 *
	 * If returns a nil UUID, the minidump was not successfully captured. Otherwise,
	 * will return the UUID of the event processed with the minidump attachment.
	 */
	SENTRY_API sentry_uuid_t sentry_capture_minidump(const char* path);
	SENTRY_API sentry_uuid_t sentry_capture_minidump_n(const char* path, size_t path_len);
```

`sentry_core.c`
```c
sentry_uuid_t
sentry_capture_minidump(const char *path)
{
    return sentry_capture_minidump_n(path, sentry__guarded_strlen(path));
}

sentry_uuid_t
sentry_capture_minidump_n(const char *path, size_t path_len)
{
    sentry_uuid_t eventId = sentry_uuid_nil();

    sentry_path_t *dump_path = sentry__path_from_str_n(path, path_len);

    if (!dump_path) {
        SENTRY_WARN(
            "sentry_capture_minidump() failed due to null path to minidump");
        return eventId;
    }

    SENTRY_DEBUGF(
        "Capturing minidump \"%" SENTRY_PATH_PRI "\"", dump_path->path);

    sentry_value_t event = sentry_value_new_event();
    sentry_value_set_by_key(
        event, "level", sentry__value_new_level(SENTRY_LEVEL_FATAL));

[...]

PR #1138, opened 2025-01-31
API Change / Addition to sentry_capture_minidump

zsd4yr:
See https://github.com/getsentry/sentry-native/issues/1135


include/sentry.h
```diff
@@ -2220,4 +2224,4 @@ SENTRY_EXPERIMENTAL_API const char *sentry_sdk_user_agent(void);
 #ifdef __cplusplus
 }
 #endif
-#endif
+#endif
```
supervacuus, line 2227: This breaks some of the builds. Please use `clang-format`.

src/sentry_core.c
```diff
@@ -1257,21 +1255,23 @@ sentry_clear_crashed_last_run(void)
...
 {
-    sentry_capture_minidump_n(path, sentry__guarded_strlen(path));
+    return sentry_capture_minidump_n(path, sentry__guarded_strlen(path));
 }
 
-void
+sentry_uuid_t
 sentry_capture_minidump_n(const char *path, size_t path_len)
 {
+    sentry_uuid_t eventId = sentry_uuid_nil();
```
supervacuus, line 1267: we use snake-case throughout the code-base.
```suggestion
    sentry_uuid_t event_id = sentry_uuid_nil();
```

src/sentry_core.c
```diff
@@ -1257,21 +1255,23 @@ sentry_clear_crashed_last_run(void)
...
+
     sentry_path_t *dump_path = sentry__path_from_str_n(path, path_len);
 
     if (!dump_path) {
         SENTRY_WARN(
             "sentry_capture_minidump() failed due to null path to minidump");
-        return;
+        return eventId;
```
supervacuus, line 1274:
```suggestion
        return event_id;
```

src/sentry_core.c
```diff
@@ -1304,11 +1304,14 @@ sentry_capture_minidump_n(const char *path, size_t path_len)
             }
 
             sentry__capture_envelope(options->transport, envelope);
+            eventId = sentry__envelope_get_event_id(envelope);
```
supervacuus, line 1307:
```suggestion
            event_id = sentry__envelope_get_event_id(envelope);
```

src/sentry_core.c
```diff
@@ -1304,11 +1304,14 @@ sentry_capture_minidump_n(const char *path, size_t path_len)
...
 
             SENTRY_INFOF("Minidump has been captured: \"%" SENTRY_PATH_PRI "\"",
                 dump_path->path);
         }
     }
 
     sentry__path_free(dump_path);
+
+    return eventId;
```
supervacuus, line 1316:
```suggestion
    return event_id;
```

include/sentry.h
```diff
@@ -1409,9 +1409,13 @@ SENTRY_API sentry_uuid_t sentry_capture_event(sentry_value_t event);
  * This generates a fatal error event, includes the scope and attachments.
  * If the event isn't dropped by a before-send hook, the minidump is attached
  * and the event is sent.
+ * 
+ * If returns a nil UUID, the minidump was not successfully captured and sent to transport (not backend yet).
+ * Otherwise, will return the UUID of the event sent to transport with the minidump attachment.
+ * This can be used look up the upload later.
```
supervacuus, line 1415:
```suggestion
 * Returns a nil `UUID` if capturing the minidump failed and the event-id
 * otherwise. Uploads can fail because capturing is asynchronous, so a non-nil
 * `UUID` is not a delivery guarantee. However, if the minidump is successfully
 * delivered, the ID is guaranteed to be the same as the event in the Sentry UI.```

**Review by supervacuus — CHANGES_REQUESTED** [review]
Please add a changelog entry using the header `##Unreleased` if it is not there yet. Then, add a section `**Breaking changes**:` to list the changes.

File: include/sentry.h (hunk 1 above)
Comment (zsd4yr) [reply include/sentry.h:2228]: Hey @supervacuus ty for the help -- in our codebase clang-format runs automatiically on file save so I'm not familar with running it manually -- can you advise me there?

File: src/sentry_core.c [hunk 7]
```diff
@@ -1304,11 +1304,22 @@ sentry_capture_minidump_n(const char *path, size_t path_len)
             }
 
             sentry__capture_envelope(options->transport, envelope);
+            event_id = sentry__envelope_get_event_id(envelope);
 
-            SENTRY_INFOF("Minidump has been captured: \"%" SENTRY_PATH_PRI "\"",
-                dump_path->path);
+            if (event_id == sentry_uuid_nil())
+            {
+                SENTRY_ERRORF("Minidump was not captured: \"%" SENTRY_PATH_PRI "\"",
```
supervacuus, line 1311: Please use `SENTRY_WARNF` instead of `SENTRY_ERRORF` (we can keep that macro in the header). A Sentry _error_ should only be logged for fatal cases (i.e., if the program should immediately terminate).

supervacuus (changes requested):

src/sentry_core.c
zsd4yr: Interesting! Is there some kind of discussion / docs I can read about that distinction. From a customer perspective, this _is_ a fatal error

supervacuus: Sentry considers something "fatal" when a program can't recover (a `SIGSEGV`, an uncaught exception, etc.) and must immediately shut down, meaning "fatal" to the program because it must terminate. We send out a minidump as a fatal error event. Anything else is less than "fatal".

I assumed a failed minidump capture does not require shutting down the application (i.e., you could continue uploading other minidumps or anything else since your program is not in a failed state). 

For instance, the `UUID` could be `nil` because you didn't call `sentry_init()`. That is certainly not a fatal error. However, it could also be `nil` because reading from the random number generator failed, which tends to be more unrecoverable.

In any case, I incorrectly remembered the code base to still convert error to fatal at some point, which is why we don't have any error logs, but this is no longer true. So, feel free to use `SENTRY_ERRORF`, since this is still one below "fatal", and an error it certainly is.

include/sentry.h
supervacuus: You can run `make format`, which will automatically acquire `clang-format` and `black` and format the entire code base.

If you run `make setup`, it will install a pre-commit hook which you can examine here: https://github.com/getsentry/sentry-native/blob/341a64f5c50fa1449aabe6e93b8cdae245022e72/scripts/git-precommit-hook.sh.

This is documented in the [contribution guide](https://github.com/getsentry/sentry-native/blob/master/CONTRIBUTING.md#setting-up-environment).

zsd4yr: Thanks!

When I run `make setup` I hit this issue
```
'ln' is not recognized as an internal or external command, operable program or batch file.
```
It looks like that is related to https://github.com/getsentry/sentry-native/blob/03143a8e01a4fe52a03a87fb566c6d669f036356/Makefile#L61 

It's been a very long time since I fiddled with makefiles, but I think it's trying to create a symbolic link? It seems like in general the scripts struggle to run on Windows

Putting that aside, I was able to run this
```
python .\scripts\check-clang-format.py --clang-format-executable "...MicrosoftVisualStudio\2022\BuildTools\VC\Tools\Llvm\x64\bin\clang-format.exe" -r include/sentry.h
```
```
--- include/sentry.h    (original)
+++ include/sentry.h    (reformatted)
@@ -1409,14 +1409,15 @@
  * This generates a fatal error event, includes the scope and attachments.
  * If the event isn't dropped by a before-send hook, the minidump is attached
  * and the event is sent.
- *
+ *
  * Returns a nil `UUID` if capturing the minidump failed and the event-id
  * otherwise. Uploads can fail because capturing is asynchronous, so a non-nil
  * `UUID` is not a delivery guarantee. However, if the minidump is successfully
  * delivered, the ID is guaranteed to be the same as the event in the Sentry UI.
  */
 SENTRY_API sentry_uuid_t sentry_capture_minidump(const char *path);
-SENTRY_API sentry_uuid_t sentry_capture_minidump_n(const char *path, size_t path_len);
+SENTRY_API sentry_uuid_t sentry_capture_minidump_n(
+    const char *path, size_t path_len);

 /**
  * Captures an exception to be handled by the backend.
```

However, git shows no changes

src/sentry_core.c
zsd4yr: Thanks for the clarification! I think having that separation for Fatal (internal failure) and Error (customer facing failure) is great

include/sentry.h
supervacuus: > It's been a very long time since I fiddled with makefiles, but I think it's trying to create a symbolic link? It seems like in general the scripts struggle to run on Windows

That's true. The scripts are primarily for Unix because, on Windows, most devs use either Visual Studio, Visual Studio Code, or CLion. All of these can be used to build and format (they all support `clang-format`) without the help of a `Makefile`. Another topic is that people on Windows are still split between `cmd` and `PowerShell`; supporting both due to the already well-supported tooling landscape on the Windows side is not warranted.

> However, git shows no changes

This script is only used in CI to check for formatting errors.

If you want to format the header without a lot of manual dependency tracking and without the support of an IDE/editor, I recommend doing the following in Powershell from the repo root:

```powershell
# if no `.venv` directory exists, start from here
python -m venv .venv

# if a `.venv` directory exists, start from here
.\.venv\Scripts\pip.exe install --upgrade --requirement .\tests\requirements.txt
.\.venv\Scripts\clang-format.exe -i .\include\sentry.h
```

With that last line, you can format any source file in-place. However, I recommend that your editor be configured to use clang-format.

zsd4yr: Thank you for your patience! I'll get that hooked up in VS / VSC but for now the CLI is very helpful!
I've run it on all 3 source files I touched

zsd4yr:
And these
```
FAILED tests/test_integration_http.py::test_capture_minidump - Failed: running command failed: valgrind --suppressions=/home/runner/work/sentry-native/sentry-native/tests/valgrind.txt --error-exitcode=33 --leak-check=yes ./sentry_example log attachment capture-minidump
FAILED tests/test_unit.py::test_unit[capture_minidump_basic] - Failed: running command failed: valgrind --suppressions=/home/runner/work/sentry-native/sentry-native/tests/valgrind.txt --error-exitcode=33 --leak-check=yes ./sentry_test_unit --no-summary capture_minidump_basic
FAILED tests/test_unit.py::test_unit[capture_minidump_without_sentry_init] - Failed: running command failed: valgrind --suppressions=/home/runner/work/sentry-native/sentry-native/tests/valgrind.txt --error-exitcode=33 --leak-check=yes ./sentry_test_unit --no-summary capture_minidump_without_sentry_init
FAILED tests/test_unit.py::test_unit_transport[capture_minidump_basic] - Failed: running command failed: valgrind --suppressions=/home/runner/work/sentry-native/sentry-native/tests/valgrind.txt --error-exitcode=33 --leak-check=yes ./sentry_test_unit --no-summary capture_minidump_basic
FAILED tests/test_unit.py::test_unit_transport[capture_minidump_without_sentry_init] - Failed: running command failed: valgrind --suppressions=/home/runner/work/sentry-native/sentry-native/tests/valgrind.txt --error-exitcode=33 --leak-check=yes ./sentry_test_unit --no-summary capture_minidump_without_sentry_init
```

supervacuus:
> Looking into this failure in tests

There is no need, @zsd4yr. I had to rewrite the proposed implementation (and parts of the initial implementation). I'll take over from here. Thanks!

supervacuus (approved):

zsd4yr:
@supervacuus Thanks for getting this across the finish line! When will the next drop be available? I'll update our fork with this :)

Issue #1129, opened 2025-01-27
Documentation out of sync with library implementation for Span


nwhite-riot:
The [Sentry documentation for Span](https://develop.sentry.dev/sdk/data-model/event-payloads/span/#attributes) indicates that `parent_span_id` is _optional_, but the [native library](https://github.com/getsentry/sentry-native/blob/2aa4b16fb40a280744a9a5e74c55d82283b159cb/src/sentry_core.c#L1095) requires that a valid parent Span object be passed in, otherwise the Span creation fails and returns `NULL`.

Even if the divergence here is due to library expectations (for example that Span is created as a child of a Transaction only), I think the library should acknowledge and explain the behavior difference.

nwhite-riot:
In principle I agree, but `sentry.h` pushes the [user towards the documentation](https://github.com/getsentry/sentry-native/blob/2aa4b16fb40a280744a9a5e74c55d82283b159cb/include/sentry.h#L1891C4-L1893) for information about the properties, so it may not be as obvious that these things are not bound by the same contract to folks who are less familiar with the library.

supervacuus:
> In principle I agree, but `sentry.h` pushes the [user towards the documentation](https://github.com/getsentry/sentry-native/blob/2aa4b16fb40a280744a9a5e74c55d82283b159cb/include/sentry.h#L1891C4-L1893) for information about the properties, so it may not be as obvious that these things are not bound by the same contract to folks who are less familiar with the library.

That's fair. I will adapt the docs to clarify.

JoshuaMoelans:
Hi @nwhite-riot , thanks for raising this. I [updated the docstrings ](https://github.com/getsentry/sentry-native/pull/1139/files)for the span-creating functions. Is it clearer now why the span interface docs & the native code differ?

nwhite-riot:
> Hi [@nwhite-riot](https://github.com/nwhite-riot) , thanks for raising this. I [updated the docstrings ](https://github.com/getsentry/sentry-native/pull/1139/files)for the span-creating functions. Is it clearer now why the span interface docs & the native code differ?

Yes this is definitely helpful! Thank you :)

PR #1139, opened 2025-01-31
docs: explicitly mention non-null span parent

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/1129

#skip-changelog

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -1829,7 +1829,7 @@ SENTRY_EXPERIMENTAL_API void sentry_set_span(sentry_span_t *span);
  * Starts a new Span.
  *
  * The return value of `sentry_transaction_start` should be passed in as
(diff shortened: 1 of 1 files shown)


JoshuaMoelans:
The original issuer mention [this snippet](https://github.com/getsentry/sentry-native/pull/1139/files#diff-9c77242571e813d699b88663fdc7d7ad7bb5b481a3c1af8d34e1a9e242cbf076R1893-R1895) pointing to the [develop docs](https://develop.sentry.dev/sdk/data-model/event-payloads/span/) and how `parent_span_id` is listed as _optional_ there. 

I feel like with the change in this PR, prefacing that link by stating how spans can't be orphans, this is no longer an issue.

supervacuus (approved):

PR #1142, opened 2025-02-04
feat: Add `sentry_value_get_key(value, index)`

limbonaut:
This PR introduces the `sentry_value_get_key(value, index)` function.

## Rationale

There is currently no way to get the keys of a map object. We need such a function to build a proper representation in upstream SDKs (like `sentry-godot`). We can potentialy improve [existing code in other SDKs](https://github.com/getsentry/sentry-unreal/blob/8e159cec39ef22eeef536ad650a7fe8233e02dfe/plugin-dev/Source/Sentry/Private/GenericPlatform/Infrastructure/GenericPlatformSentryConverters.cpp#L140-L167) if we add this function. It can also be used to find which contexts or tags are added to a sentry event and present such data through `SentryEvent` class.

`sentry_value_get_length(value)` is already set up to return the number of elements in a map. This PR simply adds the last missing piece - the `sentry_value_get_key(value, index)` function.

## How we will use it
WIP upstream SDK code that benefits from these changes:
- https://github.com/getsentry/sentry-godot/blob/2fae26ef6151ef983b45d5c97280f66545c3b64d/src/sentry/native/native_breadcrumb.cpp#L47-L54
- https://github.com/getsentry/sentry-godot/blob/2fae26ef6151ef983b45d5c97280f66545c3b64d/src/sentry/native/native_util.cpp#L85-L95

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Features**:
+
+- Add `sentry_value_get_key(value, index)` function. ([#1142](https://github.com/getsentry/sentry-native/pull/1142))
+
 ## 0.7.20
 
 **Features**:
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -291,6 +291,12 @@ SENTRY_API sentry_value_t sentry_value_get_by_key_owned_n(
     sentry_value_t value, const char *k, size_t k_len);
 
 /**
+ * Returns a key in a map by index.  If missing or value is not a map,
+ * an empty string is returned.
+ */
+SENTRY_API const char *sentry_value_get_key(sentry_value_t value, size_t index);
+
+/**
  * Looks up a value in a list by index.  If missing a null value is returned.
  * The returned value is borrowed.
  */
--- a/src/sentry_value.c
+++ b/src/sentry_value.c
@@ -778,6 +778,18 @@ sentry_value_get_by_key_owned(sentry_value_t value, const char *k)
     return rv;
 }
 
+const char *
+sentry_value_get_key(sentry_value_t value, size_t index) {
+    const thing_t *thing = value_as_thing(value);
+    if (thing && thing_get_type(thing) == THING_TYPE_OBJECT) {
+        obj_t *o = thing->payload._ptr;
+        if (index < o->len) {
+            return o->pairs[index].k;
+        }
+    }
+    return "";
+}
+
 sentry_value_t
 sentry_value_get_by_index(sentry_value_t value, size_t index)
 {
--- a/src/sentry_value.c
+++ b/src/sentry_value.c
@@ -779,7 +779,8 @@ sentry_value_get_by_key_owned(sentry_value_t value, const char *k)
 }
 
 const char *
-sentry_value_get_key(sentry_value_t value, size_t index) {
+sentry_value_get_key(sentry_value_t value, size_t index)
+{
     const thing_t *thing = value_as_thing(value);
     if (thing && thing_get_type(thing) == THING_TYPE_OBJECT) {
         obj_t *o = thing->payload._ptr;
--- a/tests/unit/test_value.c
+++ b/tests/unit/test_value.c
@@ -252,8 +252,10 @@ SENTRY_TEST(value_object)
         sentry_value_t child = sentry_value_get_by_key(val, key);
         if (i < 10) {
             TEST_CHECK(sentry_value_as_int32(child) == (int32_t)i);
+            TEST_CHECK(strcmp(sentry_value_get_key(val, i), key) == 0);
         } else {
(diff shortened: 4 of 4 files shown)


supervacuus:
Hi @limbonaut! Providing a way to iterate the items (key and value) of the object value type is very useful. However, I would prefer not to expose implementation details or keep compatibility with an index-based lookup if we change internal structures in the future since the index is meaningless in the object's context, and it might give clients weird ideas about the stability of such an index.

Since the primary use-case seems to be iteration, using the key to retrieve the value, I suggest having a stateful iterator design instead of an index-based lookup. Something like this:

```C
for (item_iter_t iter = item_iter(obj); is_item_iter_valid(iter); item_iter_next(&iter)) {
    const char* key = item_iter_get_key(iter);
    value_t value = item_iter_get_value(iter); // Borrowed (== ref-count stays the same, the client can always `inc` if they want to hold on)
}
```

This is undoubtedly more upfront effort but a sensible mid- to long-term solution. The iterator can be an opaque pointer hiding any details of the implementation. In the first version, the iterator wouldn't provide guarantees if the object is modified concurrently (i.e., any modification to the object invalidates any iterator, similar to how the index lookup cannot offer any guarantees), which we could later improve.

limbonaut:
Hey @supervacuus! Indeed, the main use case is iteration, and I'm glad this functionality is considered useful. Your suggestions sound so much better than what I proposed. I'll try to present a new PR using the stateful iterator approach in a bit. Thanks for looking into it!

supervacuus:
I quickly hacked up a basic implementation yesterday for you to play with.

`sentry.h`
```C
struct sentry_item_iter_s;
typedef struct sentry_item_iter_s sentry_item_iter_t;

SENTRY_API sentry_item_iter_t *sentry_value_new_item_iter(sentry_value_t value);
SENTRY_API void sentry_value_item_iter_next(sentry_item_iter_t *item_iter);
SENTRY_API int sentry_value_item_iter_valid(sentry_item_iter_t *item_iter);
SENTRY_API const char *sentry_value_item_iter_get_key(
    sentry_item_iter_t *item_iter);
SENTRY_API sentry_value_t sentry_value_item_iter_get_value(
    sentry_item_iter_t *item_iter);
```

`sentry_value.c`
```C
struct sentry_item_iter_s {
    size_t len;
    obj_pair_t *pairs;
    size_t index;
};

sentry_item_iter_t *
sentry_value_new_item_iter(sentry_value_t value)
{
    const thing_t *thing = value_as_thing(value);
    if (thing && thing_get_type(thing) == THING_TYPE_OBJECT) {
        obj_t *o = thing->payload._ptr;
        sentry_item_iter_t *item_iter = SENTRY_MAKE(sentry_item_iter_t);
        item_iter->len = o->len;
        item_iter->pairs = o->pairs;
        item_iter->index = 0;
        return item_iter;
    }
    return NULL;
}

void sentry_value_item_iter_next(sentry_item_iter_t *item_iter)
{
    item_iter->index++;
}

const char *
sentry_value_item_iter_get_key(sentry_item_iter_t *item_iter)
{
    if (item_iter->index >= item_iter->len) {
        return NULL;
    }
    return item_iter->pairs[item_iter->index].k;
}

sentry_value_t
sentry_value_item_iter_get_value(sentry_item_iter_t *item_iter)
{
    if (item_iter->index >= item_iter->len) {
        return sentry_value_new_null();
    }
    return item_iter->pairs[item_iter->index].v;
}

int
sentry_value_item_iter_valid(sentry_item_iter_t *item_iter)
{
    return item_iter->index < item_iter->len && item_iter->pairs != NULL;
}
```

Which can be used like this:

```C
sentry_item_iter_t *item_iter = sentry_value_new_item_iter(val);
for (; sentry_value_item_iter_valid(item_iter);
         sentry_value_item_iter_next(item_iter)) {
    const char *item_key = sentry_value_item_iter_get_key(item_iter);
    sentry_value_t item_value = sentry_value_item_iter_get_value(item_iter);
    // ...
}
sentry_free(item_iter);
```

As you can see, there are still open design/implementation questions. For instance, I was defensive in the `get_key()` and `get_value()` functions. One could require the client to check validity before the accessor; in that case, we could remove the check. In any case, you can pick up or ignore it.

limbonaut:
@supervacuus 
I've tried your code and it works great. I also measured it with and without the defensive guards: for traversing `event.contexts` there was no tangible difference, at least for our use case. Having such safety measures is probably for the best. I've added a branch for the new code, feel free to commit directly to it: #1143

I'm trialing it with my PR branch which wants this functionality over here: https://github.com/getsentry/sentry-godot/pull/83

I'm contemplating adding `sentry_value_iter_item_erase()`. It seems useful for filtering. How do you feel about it?

limbonaut:
Closed in favor of #1143

PR #1143, opened 2025-02-06
feat: Map item iterators

limbonaut:
WIP

Supercedes #1142

#skip-changelog for now

### DIFF
--- a/src/sentry_value.c
+++ b/src/sentry_value.c
@@ -783,7 +783,7 @@ sentry_value_get_by_key_owned(sentry_value_t value, const char *k)
  */
 
 struct sentry_item_iter_s {
-    size_t len;
+    size_t *len; // Pointer to length!
     obj_pair_t *pairs;
     size_t index;
 };
@@ -795,7 +795,7 @@ sentry_value_new_item_iter(sentry_value_t value)
     if (thing && thing_get_type(thing) == THING_TYPE_OBJECT) {
         obj_t *o = thing->payload._ptr;
         sentry_item_iter_t *item_iter = SENTRY_MAKE(sentry_item_iter_t);
-        item_iter->len = o->len;
+        item_iter->len = &o->len;
         item_iter->pairs = o->pairs;
         item_iter->index = 0;
         return item_iter;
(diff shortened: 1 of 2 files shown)


src/sentry_value.c
```diff
@@ -778,6 +778,75 @@ sentry_value_get_by_key_owned(sentry_value_t value, const char *k)
...
+    return item_iter->index < *item_iter->len && item_iter->pairs != NULL;
+}
+
+int
+sentry_value_item_iter_erase(sentry_item_iter_t *item_iter)
+{
+    if (item_iter->frozen || item_iter->index >= *item_iter->len) {
+        return 1;
+    }
+    obj_pair_t *pair = &item_iter->pairs[item_iter->index];
```
supervacuus, line 840: Again, do we assume that `pairs` and `len` are valid pointers in this case? 

If so, why? Because `valid()` does the check? If so, we must decide whether users must check for validity every time either the iterator or the object changed and `erase()`, `get_key()` and `get_value()` assume valid iterators or add the checks in all of those functions. 

Whichever way we choose, we should be consistent regarding the guarantees of calling `valid()`, the accessors, and mutators (since adding new keys will then also be used if iterators allow object mutation).

This demonstrates how easily checks (and thus guarantees) can become inconsistent quickly when an in-place erasure is introduced. If you introduce the guarantees for each getter and mutator (making `valid()` solely a function for a loop-invariant), then you can't change that without breaking code. OTOH, adding those guarantees later won't break any code.

tests/unit/test_value.c
```diff
@@ -290,6 +290,94 @@ SENTRY_TEST(value_object)
...
+        for (; sentry_value_item_iter_valid(it);
+            sentry_value_item_iter_next(it)) {
+            const char *key = sentry_value_item_iter_get_key(it);
+            int32_t key_idx;
+            sscanf(key, "key%d", &key_idx);
+            TEST_CHECK(key_idx % 2 != 0);
+        }
+        sentry_free(it);
+    }
+
```
supervacuus, line 364: If we keep the mutating iterator, please also add a loop that adds keys or documents that this is forbidden.

PR #1148, opened 2025-02-11
Fix Xbox OS name

PlasmaDev5:
This change fixes an issue where the os name on Xbox was being reported as Windows.

### DIFF
--- a/src/sentry_os.c
+++ b/src/sentry_os.c
@@ -117,7 +117,7 @@ sentry__get_os_context(void)
 #    if defined(_GAMING_XBOX_SCARLETT)
 #        pragma warning(push)
 #        pragma warning(disable : 4996)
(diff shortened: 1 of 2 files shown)


supervacuus (approved):

src/sentry_os.c
```diff
@@ -117,7 +117,7 @@ sentry__get_os_context(void)
 #    if defined(_GAMING_XBOX_SCARLETT)
 #        pragma warning(push)
 #        pragma warning(disable : 4996)
-    sentry_value_set_by_key(os, "name", sentry_value_new_string("Windows"));
+    sentry_value_set_by_key(os, "name", sentry_value_new_string("Xbox"));
```
bruno-garcia, line 120: Will the version number always match?

This also doesn't tell us which model (One, Series X or S etc)?

PR #1153, opened 2025-02-20
feat(crashpad): add option to delay shutdown until upload thread completes

JoshuaMoelans:
See https://github.com/getsentry/crashpad/pull/121

- [x] add Sentry option
- [ ] implement option passthrough into Crashpad
- [ ] add wait for all platforms
- [ ] add tests

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -1272,11 +1272,20 @@ SENTRY_API void sentry_options_set_system_crash_reporter_enabled(
     sentry_options_t *opts, int enabled);
 
 /**
- * Sets the maximum time (in milliseconds) to wait for the asynchronous tasks to
- * end on shutdown, before attempting a forced termination.
+ * Enables a wait for the crash report upload to be finished before shutting
+ * down. This is disabled by default.
+ *
+ * This setting only has an effect when using the crashpad backend.
  */
-SENTRY_API void sentry_options_set_shutdown_timeout(
-    sentry_options_t *opts, uint64_t shutdown_timeout);
+SENTRY_API void sentry_options_set_on_crash_wait_for_upload(
+    sentry_options_t *opts, int wait_for_upload)
+
+    /**
+     * Sets the maximum time (in milliseconds) to wait for the asynchronous
+     * tasks to end on shutdown, before attempting a forced termination.
+     */
+    SENTRY_API void sentry_options_set_shutdown_timeout(
+        sentry_options_t *opts, uint64_t shutdown_timeout);
 
 /**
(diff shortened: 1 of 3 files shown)


JoshuaMoelans:
Do we already want to add the dockerfile & docker-compose files to the repo for posterity, or should we just leave them in the PR description as-is (until we add a proper Docker test into our CI testing setup)

CONTRIBUTING.md
```diff
@@ -157,6 +157,10 @@ The example currently supports the following commands:
 - `capture-transaction`: Captures a transaction.
 - `traces-sampler`: Installs a traces sampler callback function when used alongside `capture-transaction`.
 
+ 
+Only on Linux using crashpad:
+- `crashpad-wait-for-upload`: Couples application shutdown to finishing the upload thread.
```
supervacuus, line 162:
```suggestion
- `crashpad-wait-for-upload`: Couples application shutdown to complete the upload in the `crashpad_handler`. 
```

examples/example.c
```diff
@@ -286,6 +286,10 @@ main(int argc, char **argv)
         sentry_options_set_proxy(options, "socks5://127.0.0.1:1080");
     }
 
+    if (has_arg(argc, argv, "crashpad-wait-for-upload")) {
+        sentry_options_set_on_crash_wait_for_upload(options, true);
```
supervacuus, line 290: I would rename the option to the same as the argument string. Not for consistency, but because it better highlights what it is for: `sentry_options_set_crashpad_wait_for_upload()` (this also applies to the option flag itself). It only works and makes sense for `crashpad`, which should be highlighted since it is a highly backend-specific function.

include/sentry.h
```diff
@@ -1288,8 +1288,17 @@ SENTRY_API void sentry_options_set_system_crash_reporter_enabled(
     sentry_options_t *opts, int enabled);
 
 /**
- * Sets the maximum time (in milliseconds) to wait for the asynchronous tasks to
- * end on shutdown, before attempting a forced termination.
+ * Enables a wait for the crash report upload to be finished before shutting
+ * down. This is disabled by default.
+ *
+ * This setting only has an effect when using the crashpad backend.
```
supervacuus, line 1294:
```suggestion
 * This setting only has an effect when using the `crashpad` backend on Linux.
```

supervacuus (changes requested):

examples/example.c
JoshuaMoelans: good catch, thanks!

supervacuus (approved):

PR #1155, opened 2025-02-21
upload debug files for ndk build

bruno-garcia:
While looking at this diff I realized we're no longer uploading debug symbols from the NDK build since we're no longer compiling it: https://github.com/getsentry/sentry-unreal/pull/783/files

But since we're downloading from here. We should upload debug files via [symbol collector](https://github.com/getsentry/symbol-collector/) when making releases.

Note: I didn't test if paths etc match. This is how it was set up before when compiling on the sentry-java repo

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -26,5 +26,9 @@ targets:
       distDirRegex: /^(sentry-native-ndk).*$/
       fileReplaceeRegex: /\d+\.\d+\.\d+(-\w+(\.\d+)?)?(-SNAPSHOT)?/
       fileReplacerStr: release.aar
+  - name: symbol-collector
+    includeNames: /libsentry(-android)?\.so/
+    batchType: android
+    bundleIdPrefix: sentry-unreal-android-ndk-
 requireNames:
   - /^sentry-native.zip$/


.craft.yml
```diff
@@ -26,5 +26,9 @@ targets:
       distDirRegex: /^(sentry-native-ndk).*$/
       fileReplaceeRegex: /\d+\.\d+\.\d+(-\w+(\.\d+)?)?(-SNAPSHOT)?/
       fileReplacerStr: release.aar
+  - name: symbol-collector
+    includeNames: /libsentry(-android)?\.so/
+    batchType: android
+    bundleIdPrefix: sentry-unreal-android-ndk-
```
bruno-garcia, line 32:
```suggestion
    bundleIdPrefix: sentry-native-ndk-
```

markushi (approved):

supervacuus (approved):
This may not be the right place to ask, but are there any further 7.x sentry-android releases planned?

kahest:
> This may not be the right place to ask, but are there any further 7.x sentry-android releases planned?

@supervacuus possibly for hotfixes, especially while v8 adoption is still low

supervacuus:
>> This may not be the right place to ask, but are there any further 7.x sentry-android releases planned?

> @supervacuus possibly for hotfixes, especially while v8 adoption is still low

In this case, we should ensure that debug files for both majors are still uploaded and do not conflict (I am guessing the latter is a non-issue since they are different versions anyway, but still).

PR #1156, opened 2025-02-25
chore: update ubuntu runners

### LINKED ISSUE #1168: Ubuntu-20.04 jobs will break on 2025-04-01
Just a heads-up to update CI beforehand.  See https://github.com/actions/runner-images/issues/11101

Image

JoshuaMoelans:
due to [upcoming deprecation](https://github.com/actions/runner-images/issues/11101#issue-2719265843) we update our runners as such:
`ubuntu-22.04 -> ubuntu-24.04`

`ubuntu-20.04 -> ubuntu-22.04`

- [ ] we should also do this on [crashpad](https://github.com/getsentry/crashpad/blob/getsentry/.github/workflows/build.yml#L24)

fixes #1168.

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,7 @@ jobs:
         if: ${{ runner.os == 'Linux' && matrix.os != 'ubuntu-22.04' && !env['TEST_X86'] }}
         run: |
           sudo apt update
-          sudo apt install cmake clang-14 clang-tools llvm g++-12 valgrind zlib1g-dev libcurl4-openssl-dev
+          sudo apt install cmake clang-15 clang-tools llvm g++-12 valgrind zlib1g-dev libcurl4-openssl-dev
           # Install kcov from source
           sudo apt-get install binutils-dev libssl-dev libelf-dev libstdc++-12-dev libdw-dev libiberty-dev
           git clone https://github.com/SimonKagstrom/kcov.git
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
             os: ubuntu-24.04
             CC: gcc-12
             CXX: g++-12
-          - name: Linux (GCC 14.3.0)
+          - name: Linux (GCC 14.2.0)
             os: ubuntu-24.04
             CC: gcc-14
             CXX: g++-14
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,7 +156,7 @@ jobs:
         run: |
           sudo dpkg --add-architecture i386
           sudo apt update
-          sudo apt install cmake gcc-7-multilib g++-7-multilib zlib1g-dev:i386 libssl-dev:i386 libcurl4-openssl-dev:i386
+          sudo apt install cmake gcc-multilib g++-multilib zlib1g-dev:i386 libssl-dev:i386 libcurl4-openssl-dev:i386
 
         # https://github.com/actions/runner-images/issues/9491
       - name: Decrease vm.mmap_rnd_bit to prevent ASLR ASAN issues
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,17 +39,17 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: Linux (GCC 9, 32-bit)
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -39,38 +39,38 @@ jobs:
...
       matrix:
         include:
-          - name: Linux (GCC 7, 32-bit)
-            os: ubuntu-20.04
-            CC: gcc-7
-            CXX: g++-7
+          - name: Linux (GCC 9, 32-bit)
+            os: ubuntu-22.04
+            CC: gcc-9
+            CXX: g++-9
```
JoshuaMoelans, line 45: Had to change to `gcc-9` because `multilib` was not available otherwise.

.github/workflows/ci.yml
```diff
@@ -140,13 +140,22 @@ jobs:
           cache: "pip"
 
       - name: Installing Linux Dependencies
-        if: ${{ runner.os == 'Linux' && matrix.os != 'ubuntu-20.04' && !env['TEST_X86'] }}
+        if: ${{ runner.os == 'Linux' && matrix.os != 'ubuntu-22.04' && !env['TEST_X86'] }}
         run: |
           sudo apt update
-          sudo apt install cmake clang-14 clang-tools llvm kcov g++-12 valgrind zlib1g-dev libcurl4-openssl-dev
+          sudo apt install cmake clang-15 clang-tools llvm g++-12 valgrind zlib1g-dev libcurl4-openssl-dev
```
JoshuaMoelans, line 146: Changed to clang-15 (not sure why this was clang-14 before?)

.github/workflows/ci.yml
```diff
@@ -140,13 +140,22 @@ jobs:
...
+          # Install kcov from source
+          sudo apt-get install binutils-dev libssl-dev libelf-dev libstdc++-12-dev libdw-dev libiberty-dev
+          git clone https://github.com/SimonKagstrom/kcov.git
+          cd kcov
+          mkdir build
+          cd build
+          cmake ..
+          make
+          sudo make install
```
JoshuaMoelans, line 155: kcov is not available in 24.04 (yet)
see https://answers.launchpad.net/ubuntu/+source/kcov/+question/818841

JoshuaMoelans: I'm not sure if this manual installation is the way to go here, seems a bit out-of-place but might just be a necessary evil. Alternatively we could wait to update the `22.04` runners to `24.04` and just keep them as-is.

supervacuus: There is no need to have the `kcov` build run on `24.04`. This is primarily a coverage run, which should produce the same results on both distro versions. On the other hand, it looks like `24.04` will never get that package, and if building `kcov` is relatively trivial and works, then why not go with that build?

.github/workflows/ci.yml
supervacuus: Yeah, as discussed on discord, `22.04` has nothing below `gcc-10` on `jammy` and `gcc-9` on `jammy-updates`.

.github/workflows/ci.yml
supervacuus: Since we switched to `24.04`, you can now use `clang-18` (the default `clang` on that distro version).

.github/workflows/ci.yml
supervacuus: btw, if you think that the build script for `kcov` is unnecessarily detailed for the workflow spec, you can move it to the scripts folder and only invoke the script from the workflow.

JoshuaMoelans: alright, then I'll keep it as it is now (part of the `24.04` run, and since the `kcov` install is pretty minimal it can stay in the workflow file)

tests/test_integration_crashpad.py
```diff
@@ -114,7 +114,7 @@ def test_crashpad_proxy_set_empty(cmake, httpserver):
             cmake, httpserver, "http-proxy"
         )
 
-        with httpserver.wait(timeout=15) as waiting:
+        with httpserver.wait(timeout=30) as waiting:
```
supervacuus, line 117: I don't think whatever is bugging you here is an httpserver timeout issue. I think it makes sense to revert these.

supervacuus (approved):
besides removal of the two test-builds, lgtm

PR #1157, opened 2025-02-25
expand on what is sentry_capture_minidump for

bruno-garcia:
This has been causing confusion. 
More API doc hopefully to help clarify things.

#skip-changelog

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -1405,19 +1405,19 @@ SENTRY_API sentry_uuid_t sentry_capture_event(sentry_value_t event);
 
 /**
  * Allows capturing independently created minidumps.
- * 
- * You don't need this function if you're relying on Sentry to create the minidump.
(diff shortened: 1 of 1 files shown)


include/sentry.h
```diff
@@ -1405,6 +1405,10 @@ SENTRY_API sentry_uuid_t sentry_capture_event(sentry_value_t event);
 
 /**
  * Allows capturing independently created minidumps.
+ * 
+ * You don't need this function if you're relying on Sentry to create the minidump.
+ * This is useful when you have a minidump captured through a different mechanism,
+ * and you want Sentry to ingest it.
```
supervacuus, line 1411: This is great, but I would add it as a `Note: ` paragraph at the end of the inline comment. We do this with most meta descriptions in the header.

supervacuus (approved):

PR #1160, opened 2025-02-25
docs: improve `sentry_handle_exception()` description

### LINKED ISSUE #1158: Document usage of `sentry_handle_exception`
What would be an exception in this case?

Documenting usage on https://docs.sentry.io/platforms/native/usage/ would be very helpful
Or at least the inline doc, clarifying when to use it, when not to, etc

I see it was added here:
* https://github.com/getsentry/sentry-native/pull/201
With desc:
> Allows manually triggering of the backend given a completed ucontext

supervacuus:
fixes #1158.

#skip-changelog


bruno-garcia:
Thank you!

PR #1164, opened 2025-03-01
test: add path-prefix for tests that write to FS

vaind:
This helps with running unit tests on platforms that need to specify a custom directory to use for writing.

#skip-changelog - purely an internal change

### DIFF
--- a/tests/unit/sentry_testsupport.h
+++ b/tests/unit/sentry_testsupport.h
@@ -58,3 +58,14 @@
 #else
 #    define TEST_VISIBLE
 #endif
+
+#ifdef SENTRY_TEST_PATH_PREFIX
+#    define SENTRY_TEST_OPTIONS_NEW(Varname)                                   \
+        sentry_options_t *Varname = sentry_options_new();                      \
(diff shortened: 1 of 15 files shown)


tests/unit/sentry_testsupport.h
```diff
@@ -58,3 +58,14 @@
...
 #    define TEST_VISIBLE
 #endif
+
+#ifdef SENTRY_TEST_PATH_PREFIX
+#    define SENTRY_TEST_OPTIONS_NEW(Varname)                                   \
+        sentry_options_t *Varname = sentry_options_new();                      \
+        sentry_options_set_database_path(                                      \
+            Varname, SENTRY_TEST_PATH_PREFIX ".sentry-native");
+#else
+#    define SENTRY_TEST_PATH_PREFIX ""
```
vaind, line 68: Note: i've originally had
```c
#ifdef __ANDROID__
#    define PREFIX "/data/local/tmp/"
#else
#    define PREFIX ""
#endif
```
here, but it doesn't seem necessary, because other tests write to the current directory, even on Android

PR #1165, opened 2025-03-02
feat: switch compilation support

vaind:
Enables as to build the SDK on Switch. Doesn't actually depend on any Switch SDKs - just updates the whitelist the platform & updates a few places that use generic C APIs.

I've split out some self-contained small PRs that should be handled first:

* #1162 
* #1163 
* #1164

### DIFF
--- a/src/sentry_random.c
+++ b/src/sentry_random.c
@@ -1,10 +1,11 @@
 #include "sentry_boot.h"
 
-#include "sentry_random.h"
-#include "sentry_utils.h"
+#ifndef SENTRY_PLATFORM_NX
+#    include "sentry_random.h"
+#    include "sentry_utils.h"
 
-#ifdef SENTRY_PLATFORM_DARWIN
-#    include <stdlib.h>
+#    ifdef SENTRY_PLATFORM_DARWIN
+#        include <stdlib.h>
 
 static int
 getrandom_arc4random(void *dst, size_t bytes)
@@ -12,13 +13,13 @@ getrandom_arc4random(void *dst, size_t bytes)
     arc4random_buf(dst, bytes);
     return 0;
 }
-#    define HAVE_ARC4RANDOM
+#        define HAVE_ARC4RANDOM
 
-#endif
-#ifdef SENTRY_PLATFORM_UNIX
-#    include <errno.h>
-#    include <fcntl.h>
-#    include <unistd.h>
+#    endif
+#    ifdef SENTRY_PLATFORM_UNIX
+#        include <errno.h>
+#        include <fcntl.h>
+#        include <unistd.h>
 
 static int
 getrandom_devurandom(void *dst, size_t bytes)
@@ -45,9 +46,9 @@ getrandom_devurandom(void *dst, size_t bytes)
     return to_read > 0;
 }
 
-#    define HAVE_URANDOM
-#endif
-#ifdef SENTRY_PLATFORM_WINDOWS
+#        define HAVE_URANDOM
+#    endif
+#    ifdef SENTRY_PLATFORM_WINDOWS
 typedef BOOLEAN(WINAPI *sRtlGenRandom)(PVOID Buffer, ULONG BufferLength);
 
 static sRtlGenRandom pRtlGenRandom = NULL;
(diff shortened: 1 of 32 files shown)


tests/unit/test_basic.c
```diff
@@ -172,9 +172,6 @@ SENTRY_TEST(crashed_last_run)
         SENTRY_TEST_OPTIONS_NEW(options);
         TEST_CHECK(sentry__clear_crash_marker(options));
         sentry_options_free(options);
-
-        // -1 before sentry_init()
-        TEST_CHECK_INT_EQUAL(sentry_get_crashed_last_run(), -1);
```
vaind, line 177: Note: this was actually broken and only ever passed because unit-tests in this repo normally don't run as a single invocation with all tests but individual executions launched from a python script - one execution for each test - so that when the static variable was accessed here, it was -1.

src/sentry_random.c
```diff
@@ -1,24 +1,25 @@
 #include "sentry_boot.h"
 
-#include "sentry_random.h"
-#include "sentry_utils.h"
+#ifndef SENTRY_PLATFORM_NX
```
supervacuus, line 3: As described [on discord](https://discord.com/channels/621778831602221064/1343728776017739829/1345021291748393012), if you exclude the entire translation unit, it may be better to do it in CMake.

src/sentry_unix_pageallocator.c
```diff
@@ -5,12 +5,13 @@
 #include <sys/mman.h>
 #include <unistd.h>
 
+#ifndef SENTRY_PLATFORM_NX
```
supervacuus, line 8: Same here.

tests/unit/test_concurrency.c
```diff
@@ -75,7 +75,11 @@ SENTRY_TEST(concurrent_init)
 {
     long called = 0;
 
-#define THREADS_NUM 100
+#ifdef SENTRY_PLATFORM_NX
+#    define THREADS_NUM 90
```
supervacuus, line 79: This is fine, but I am curious to understand the difference.

supervacuus:
@vaind, I see nothing problematic in the proposed changes. Even the `#ifdef`s are only minor concern.

Have you tested dynamic mutex initialization downstream? Does it work? It might be better if I clean up that PR and we merge it so you have these PR changes isolated from the mutex init stuff.

vaind:
> Have you tested dynamic mutex initialization downstream? Does it work?

I have not tested this manually in any way but I have run all unit tests (unless they're explicitly skipped, as seen in this PR) and they all pass. are there specific tests that would fail if that mutex implementation didn't work?

tests/unit/test_concurrency.c
vaind: There's a limit on the number of threads and our original tests triggered it. We can either lower the number for all tests or do it as above with ifdef

src/sentry_random.c
vaind: OK, excluded in cmake now

tests/unit/test_concurrency.c
supervacuus: Okay, that makes sense. Let's go ahead and leave it like this.

src/CMakeLists.txt
```diff
@@ -56,18 +54,27 @@ sentry_target_sources_cwd(sentry
...
+		sentry_random.h
 		sentry_windows_dbghelp.c
 		sentry_windows_dbghelp.h
 		path/sentry_path_windows.c
 		symbolizer/sentry_symbolizer_windows.c
 	)
+elseif(NX)
+	sentry_target_sources_cwd(sentry
+		sentry_unix_spinlock.h
+		path/sentry_path_unix.c
```
supervacuus, line 67: Okay, so just to be clear, you will also add your random implementation in a build script downstream, not here. You could refer here to a file that does not exist upstream.

supervacuus (approved):
Would you still like to keep this off the changelog?

src/CMakeLists.txt
vaind: Yes, we're including sentry-native in the downstream cmake & adding sources to the original target using `target_sources(sentry ...), currently there's backend, random, path, modulefinder and os. 
I don't really want to enumerate them downstream as there will be changes in their relative locations, names, langauge (c/cpp), etc. so I wouldn't want to have to touch upstream just to do that.

PR #1166, opened 2025-03-03
Add `before_breadcrumb` hook

tustanivsky:
`before_breadcrumb` function is called with a breadcrumb object before it is added to be sent in case of an event. It can return a modified version of the breadcrumb or nothing to skip adding one entirely. This function can also be useful for triggering custom client logic when a breadcrumb meets certain criteria (i.e. capturing a separate event when an incoming breadcrumb has the error type).

Similar callback is already implemented in [Android](https://docs.sentry.io/platforms/android/configuration/options/#before-breadcrumb) and [Cocoa](https://docs.sentry.io/platforms/apple/configuration/options/#before-breadcrumb) SDKs.

Related to  https://github.com/getsentry/sentry-unreal/issues/807

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,6 +4,7 @@ about: Create a report to help us improve
 title: ""
 labels: "Platform: Native"
 assignees: ""
+type: Bug
 ---
 
 **Description**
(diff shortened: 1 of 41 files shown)


tustanivsky:
Superseded by https://github.com/getsentry/sentry-native/pull/1534

Issue #1167, opened 2025-03-03
Unable to use the Native SDK to set up Sentry Tracing custom instrumentation for Android


sillycelli:
**Description**
Hello!

TLDR: We are unable to use `sentry-native` to set up custom Sentry Tracing for Android. Would really appreciate some information and context as to whether or not this is expected.

I recently did some work to set up custom instrumentation for the Sentry Tracing feature. Our project supports multiple platforms, one of which is Android.

I initially tried to use `sentry-native` to set up Android custom instrumentation. I verified that the sample rate was set to 100% in my Android project. I also confirmed via breakpoints that all the necessary methods for Sentry Tracing are being called when expected. However, no Sentry transactions are sent from Android when using `sentry-native`.

Upon further investigation, we discovered a log line that seems related:
`sentry-native           com.mypackage                  I  throwing away transaction due to sample rate or user-provided sampling value in transaction context`

It appears that there may be a bug with `sentry-native` that does not respect the sample rate set for Android. Using the native SDK for our desktop platforms works great, so this seems to be an Android-specific issue. We are also able to use features such as Crash Reporting with Android using `sentry-native`, so it's perplexing that Tracing does not also work.

It would be great to be able to use `sentry-native` for this purpose to avoid unnecessary bridging. Any context or info you might be able to provide on this issue would be greatly appreciated.

**When does the problem happen**

- [ ] During build
- [x] During run-time
- [ ] When capturing a hard crash

**Environment**

- OS: Android
- IDE: Android Studio

**Steps To Reproduce**

The setup is more involved than just a code snippet, but we followed the instructions pretty much exactly for setting up custom instrumentation for the [native SDK](https://docs.sentry.io/platforms/native/tracing/instrumentation/custom-instrumentation/). We are also setting the sample rate in our Android project in the AndroidManifest.xml:

`<meta-data
			android:name="io.sentry.traces.sample-rate"
			android:value="1.0"
			/>`

We also tried setting it through our C++ code, but this did not work either.

**Log output**
I can provide other logs if needed, but this appears to be the most relevant log line when trying to run Android Sentry Tracing with `sentry-native`:

`sentry-native           com.mypackage                  I  throwing away transaction due to sample rate or user-provided sampling value in transaction context`

supervacuus:
Thanks for the report @sillycelli!

[...]

PR #1176, opened 2025-03-12
feat(ndk): add `traces_sample_rate` synchronization

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/1167

- [x] add option
- [ ] write test(s) for the NDK
- [ ] add to docs
- [ ] add to changelog
#skip-changelog for now

### DIFF
--- a/ndk/sample/build.gradle.kts
+++ b/ndk/sample/build.gradle.kts
@@ -11,7 +11,7 @@ android {
 
     defaultConfig {
         applicationId = "io.sentry.ndk.sample"
-        minSdk = 19
+        minSdk = 21
         targetSdk = 34
         versionCode = 2
         versionName = project.version.toString()
--- a/ndk/sample/src/main/cpp/ndk-sample.cpp
+++ b/ndk/sample/src/main/cpp/ndk-sample.cpp
@@ -22,4 +22,29 @@ JNIEXPORT void JNICALL Java_io_sentry_ndk_sample_NdkSample_message(JNIEnv *env, 
     sentry_capture_event(event);
 }
 
+JNIEXPORT void JNICALL Java_io_sentry_ndk_sample_NdkSample_transaction(JNIEnv *env, jclass cls) {
+    __android_log_print(ANDROID_LOG_WARN, TAG, "Sending transaction.");
+
+    sentry_transaction_context_t *tx_ctx
+            = sentry_transaction_context_new("little.teapot",
+                                             "Short and stout here is my handle and here is my spout");
+    sentry_value_t custom_sampling_ctx = sentry_value_new_object();
+    sentry_value_set_by_key(
+            custom_sampling_ctx, "b", sentry_value_new_int32(42));
+    sentry_transaction_t *tx
+            = sentry_transaction_start(tx_ctx, custom_sampling_ctx);
+    sentry_transaction_set_data(
+            tx, "url", sentry_value_new_string("https://example.com"));
+    sentry_span_t *child
+            = sentry_transaction_start_child(tx, "littler.teapot", NULL);
+    sentry_span_t *grandchild
+            = sentry_span_start_child(child, "littlest.teapot", NULL);
+
+    sentry_span_set_data(
+            child, "span_data_says", sentry_value_new_string("hi!"));
+    sentry_span_finish(grandchild);
+    sentry_span_finish(child);
+    sentry_transaction_finish(tx);
+
+}
 }
--- a/ndk/build.gradle.kts
+++ b/ndk/build.gradle.kts
@@ -21,7 +21,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath("com.android.tools.build:gradle:7.4.2")
+        classpath("com.android.tools.build:gradle:8.7.1")
         classpath(kotlin("gradle-plugin", version = "1.8.0"))
         classpath("com.vanniktech:gradle-maven-publish-plugin:0.18.0")
         // dokka is required by gradle-maven-publish-plugin.
--- a/ndk/gradle/wrapper/gradle-wrapper.properties
+++ b/ndk/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.9-bin.zip
 networkTimeout=10000
 validateDistributionUrl=true
 zipStoreBase=GRADLE_USER_HOME
--- a/ndk/lib/build.gradle.kts
+++ b/ndk/lib/build.gradle.kts
@@ -11,7 +11,7 @@ android {
     namespace = "io.sentry.ndk"
 
     defaultConfig {
-        minSdk = 19
+        minSdk = 21
 
         externalNativeBuild {
             cmake {
--- a/ndk/lib/src/main/java/io/sentry/ndk/NdkOptions.java
+++ b/ndk/lib/src/main/java/io/sentry/ndk/NdkOptions.java
@@ -14,6 +14,7 @@ public final class NdkOptions {
   private final @Nullable String sdkName;
   private NdkHandlerStrategy ndkHandlerStrategy =
       NdkHandlerStrategy.SENTRY_HANDLER_STRATEGY_DEFAULT;
+  private final float tracesSampleRate;
 
   public NdkOptions(
       @NotNull String dsn,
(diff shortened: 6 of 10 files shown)


ndk/sample/src/main/cpp/ndk-sample.cpp
```diff
@@ -22,4 +22,29 @@ JNIEXPORT void JNICALL Java_io_sentry_ndk_sample_NdkSample_message(JNIEnv *env,
     sentry_capture_event(event);
 }
 
+JNIEXPORT void JNICALL Java_io_sentry_ndk_sample_NdkSample_transaction(JNIEnv *env, jclass cls) {
```
JoshuaMoelans, line 25: Since the NDK has no transport, we can't really test this just by looking at Sentry dashboards. A way to check if the envelope is created properly, is by looking at `data/data/io.sentry.ndk.sample/files/outbox`. 

This is probably how we want to this test on the native repo (and then have a integration test similar to [this one from native](https://github.com/getsentry/sentry-native/blob/master/tests/test_integration_http.py#L525) on the java-SDK side)

ndk/sample/build.gradle.kts
```diff
@@ -11,7 +11,7 @@ android {
 
     defaultConfig {
         applicationId = "io.sentry.ndk.sample"
-        minSdk = 19
+        minSdk = 21
```
JoshuaMoelans, line 14: Bumped bc Android Studio was complaining about the old AGP version; seems fine to do this, since [sentry-Java also bumped 4 months ago](https://github.com/getsentry/sentry-java/pull/3852).

JoshuaMoelans:
back to draft, forgot to change the new variable to getter-setter only (it should not be part of constructor)

JoshuaMoelans:
I'm not sure if there is a proper place to add this feature to the native docs, or if it makes more sense to only add it to the Android docs once we implement the feature to pass through the `tracesSampleRate` over there.

ndk/lib/src/androidTest/jni/ndk-test.cpp
```diff
@@ -0,0 +1,50 @@
...
+extern "C" {
+
+JNIEXPORT void JNICALL Java_io_sentry_ndk_sample_NdkSample_crash(JNIEnv *env, jclass cls) {
+    __android_log_print(ANDROID_LOG_WARN, TAG, "About to crash.");
+    char *ptr = 0;
+    *ptr += 1;
+}
+
+JNIEXPORT void JNICALL Java_io_sentry_ndk_sample_NdkSample_message(JNIEnv *env, jclass cls) {
+    __android_log_print(ANDROID_LOG_WARN, TAG, "Sending message.");
```
supervacuus, line 16: i know that this is only test code but why `ANDROID_LOG_WARN`? Same for the other occurrences. I think we should default to `INFO` and only log differently when the context requires it (`WARN` for `"About to crash."` is valid,`ERROR` would be the right for `"Transaction failed to send."`).

supervacuus (changes requested):

supervacuus:
> I'm not sure if there is a proper place to add this feature to the native docs, or if it makes more sense to only add it to the Android docs once we implement the feature to pass through the `tracesSampleRate` over there.

Currently, this doesn't require much documentation. The ask is primarily for sentry-native to honor the setting from the sentry-android layer. Most people would expect it just to work, and it only pops up because it doesn't. Of course, the core question remains: Do we think it _should_ work?

So, the getter/setter doesn't require attention from the docs. What should be documented, and probably on the Android side (but not necessarily by their team alone), is how transactions should work if enabled on both layers. This PR doesn't answer this since it only opens the pass-through down and back up (which is fine, but as such, it can't provide any clarifying insights either).

ndk/lib/src/androidTest/jni/ndk-test.cpp
JoshuaMoelans: We just copied over the existing functions from `ndk-sample.cpp` without giving it too much thought, but given that there already are other log levels, it makes sense to change them to the appropriate ones for each log statement.

ndk/lib/src/androidTest/java/io/sentry/ndk/sample/NdkSample.java
```diff
@@ -0,0 +1,13 @@
+package io.sentry.ndk.sample;
+
+public class NdkSample {
```
markushi, line 3: How about renaming this to `NdkTestHelper`? To make it's purpose more clear.

ndk/lib/src/androidTest/java/io/sentry/ndk/SentryNdkTest.java
```diff
@@ -0,0 +1,141 @@
...
+            100,
+            "io.sentry.ndk");
+
+    // set tracesSampleRate to 0
+    options.setTracesSampleRate(0);
+
+    // when initialized
+    SentryNdk.init(options);
+
+    // then it does not crash
```
markushi, line 126:
```suggestion
    // and a transaction is captured
```

ndk/lib/src/androidTest/java/io/sentry/ndk/SentryNdkTest.java
```diff
@@ -0,0 +1,141 @@
...
+    NdkSample.transaction();
+
+    File[] files = outboxPath.listFiles();
```
markushi, line 129:
```suggestion
    // the corresponding transaction should stored on disk
    File[] files = outboxPath.listFiles();
```

ndk/lib/src/androidTest/java/io/sentry/ndk/sample/NdkSample.java
```diff
@@ -0,0 +1,13 @@
...
+  static {
+    System.loadLibrary("sentry-android-test");
+  }
+
+  public static native void crash();
```
markushi, line 8: Right now we don't have tests to test `crash()` or `message()`. I'd either remove them or add tests, similar to the blackbox text for `transaction()`

ndk/lib/src/test/java/io/sentry/ndk/NdkOptionsTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class NdkOptionsTest {
+
+  @Test
+  public void tracesSampleRate() {
```
markushi, line 13: maybe add one more test which verifies that the default value for `options.getTracesSampleRate()` is `0.0`?

ndk/sample/build.gradle.kts
markushi: Should be fine, as it's just the sample anyway.

ndk/sample/build.gradle.kts
```diff
@@ -60,6 +60,16 @@ android {
...
+        getByName("debug") {
+            isMinifyEnabled = false
+            signingConfig = signingConfigs.getByName("debug")
+            addManifestPlaceholders(
+                mapOf(
+                    "sentryDebug" to true,
+                    "sentryEnvironment" to "debug",
+                ),
+            )
+        }
```
markushi, line 72: Any specific reason why we need this change? I don't see `sentryDebug` / `sentryEnvironment` being used, so I'd simply remove this block.

markushi (changes requested):
Looking good, nice! I left a few minor/medium comments, just ping me as soon as I should take a second look 🚀

ndk/sample/build.gradle.kts
JoshuaMoelans: Good catch, this was left over from the CI testing we did (but is unnecessary to remain within this PR).

ndk/sample/src/main/cpp/ndk-sample.cpp
supervacuus: Yeah, this is fine. The `outbox` is the primary interface back to `sentry-android` and, as such, the ideal test boundary. Adding additional machinery beyond parsing the output won't provide better feedback.

supervacuus (approved):

ndk/lib/src/test/java/io/sentry/ndk/NdkOptionsTest.java
```diff
@@ -0,0 +1,47 @@
...
+            "https://key@sentry.io/proj",
+            true,
+            "out",
+            "1.0.0",
+            "production",
+            "dist",
+            100,
+            "io.sentry.ndk");
+    // set tracesSampleRate to 0
+    options.setTracesSampleRate(0);
```
markushi, line 43: I'd remove this
```suggestion
```

ndk/lib/src/test/java/io/sentry/ndk/NdkOptionsTest.java
```diff
@@ -0,0 +1,47 @@
...
+            100,
+            "io.sentry.ndk");
+    // set tracesSampleRate to 1
+    options.setTracesSampleRate(1);
+
+    assertEquals(1.0f, options.getTracesSampleRate(), 0.0f);
+  }
+
+  @Test
+  public void tracesSampleRateZero() {
```
markushi, line 31:
```suggestion
  public void tracesSampleRateDefaultsToZero() {
```

markushi (approved):
LGTM!

Issue #880, opened 2023-09-06
ARM64 support for windows


cstavitsky:
(Opening on behalf of a customer)

After chatting with Arpad and Karl internally, they mentioned that Sentry on devices using ARM64 architecture (Windows) is currently untested. They mentioned testing would require buying additional specific hardware, and that there is some candidate hardware they have in mind.

Overall, this Github issue represents:
- the request to test and confirm whether Sentry works as one would expect with ARM64 architecture. 
- (if Sentry is tested and does not work) a feature request to get it working.

supervacuus:
@kahest: let's talk about this in the next sync.

supervacuus:
@kahest, short update: with the [0.7.0](https://github.com/getsentry/sentry-native/releases/tag/0.7.0) release we're cross-compiling crashpad for ARM64 on Windows in our CI. We fixed many build issues, and thebrowsercompany is using that version for the upcoming Windows version of Arc.

We still have no test environment, but at least we know when upstream changes (or ours) to the crashpad fork will break ARM64 builds on Windows. So, while official support is unwarranted, users who want to experiment with that option should be enabled to do so. I don't know if that is relevant to the unnamed customer.

DDxPlague:
Given that Github runners are now available with Windows and Arm64, is this something that could be pushed back up the priority list? Can you test with the Github Windows on Arm runners?

kahest:
@DDxPlague thanks for reaching out, we'll discuss if we can bump prio on this

JoshuaMoelans:
We're keeping an eye on the availability of the Windows CI runners, as they are [not here yet](https://github.com/orgs/community/discussions/19197#discussioncomment-11946571).

JoshuaMoelans:
While GitHub-hosted ARM64 runners for Windows are available since September 2024, they rely on so called "Partner" images. There is currently no Windows ARM64 "Partner" image available for deployment. 

We are waiting for the [Windows ARM64 partner-runner-image to become available](https://github.com/actions/partner-runner-images?tab=readme-ov-file#windows-11-on-arm), so we'll be keeping an eye on when the [new Windows-ARM64 image](https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md) is available by following this [PR](https://github.com/actions/partner-runner-images/pull/71/files).

decriptor:
@JoshuaMoelans It looks like that https://github.com/actions/partner-runner-images/pull/71 was merged yesterday.

[...]

PR #1178, opened 2025-03-14
chore: add Linux arm64 runner for tests

JoshuaMoelans:
As they are now [publicly available](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/) we can try and see how far we can get with Native on Linux arm.

Ideally we also have this for Windows, but those runners are [not available yet](https://github.com/orgs/community/discussions/19197#discussioncomment-11946571). 

Related to https://github.com/getsentry/sentry-native/issues/880

#skip-changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,11 @@ The SDK bundle contains the following folders:
 The SDK currently supports and is tested on the following OS/Compiler variations:
 
 - 64bit Linux with GCC 14
+- Arm64 Linux with GCC 14
 - 64bit Linux with GCC 12
 - 64bit Linux with GCC 9
 - 64bit Linux with clang 19
+- Arm64 Linux with clang 19
 - 32bit Linux with GCC 9 (cross compiled from 64bit host)
 - 32bit Windows with MSVC 2019
 - 64bit Windows with MSVC 2022
--- a/README.md
+++ b/README.md
@@ -61,15 +61,13 @@ The SDK bundle contains the following folders:
 
 The SDK currently supports and is tested on the following OS/Compiler variations:
 
-- 64bit Linux with GCC 14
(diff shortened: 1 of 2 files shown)


JoshuaMoelans:
We can keep this PR open until the Windows Arm runners are also available, or just merge this for now. I'm fine with keeping this around, since the main issue(s) raised were always about Windows.

supervacuus:
> We can keep this PR open until the Windows Arm runners are also available, or just merge this for now. I'm fine with keeping this around, since the main issue(s) raised were always about Windows.

Let's merge this. #880 shouldn't be closed, but I see no reason not to have this running as soon as possible.

supervacuus:
It might also be fair to include a clang-based build with ASAN and LLVM-COV (since there is code that only runs on specific architectures).

README.md
```diff
@@ -62,9 +62,11 @@ The SDK bundle contains the following folders:
 The SDK currently supports and is tested on the following OS/Compiler variations:
 
 - 64bit Linux with GCC 14
+- Arm64 Linux with GCC 14
```
supervacuus, line 65: Okay, sorry, but if we do that, let's write `amd64` or `x64` where we currently say 64bit and place `x64`/`arm64` on the same line.

JoshuaMoelans: Is this better?
```
- x64/arm64 Linux with GCC 14
- x64 Linux with GCC 12
- x64 Linux with GCC 9
- x64/arm64 Linux with clang 19
- x86 Linux with GCC 9 (cross compiled from x64 host)
- x86 Windows with MSVC 2019
- x64 Windows with MSVC 2022
```

supervacuus (approved):

Issue #962, opened 2024-03-06
[#inc-517] Add basic crashing test with asan (HW Asan, GWP-ASan)


kahest:
```[tasklist]
### 🛑 Blocked by 
- [ ] https://github.com/getsentry/sentry-java/pull/3189
- [ ] https://github.com/getsentry/sentry-native/pull/944
```

JoshuaMoelans:
# Testing Address Sanitizers
 
This comment acts as a "what-we-already-tried" benchmark to work from, which can guide our next steps in resolving this issue. During the investigation of https://github.com/getsentry/sentry-native/issues/974 I had a look into the following 3 ASAN types:

## ASan
([Android Docs](https://source.android.com/docs/security/test/asan))
This first Address Sanitizer gave me actual results when trying to build an Android app with native C memory bug detection. For future reference, this might make the most sense to add as a first "basic crashing test".

To run ASan on an emulated Pixel 9 Pro (running Android 14.0 Default Android System Image, so we [have Root access](https://stackoverflow.com/questions/43923996/adb-root-is-not-working-on-emulator-cannot-run-as-root-in-production-builds)) I added the following to the sample app ([found here](https://github.com/getsentry/sentry-native/tree/master/ndk)):
In `/sample/cmakelists.txt`
```make
target_compile_options(ndk-sample PRIVATE -fsanitize=address -fno-omit-frame-pointer)
target_link_options(ndk-sample PRIVATE -fsanitize=address)

set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=address -fno-omit-frame-pointer")
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address -fno-omit-frame-pointer")
set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -fsanitize=address")
```

And these to `/sample/build.gradle.kts` and `sentry-native/ndk/lib/build.gradle.kts`
```kts
externalNativeBuild {
    cmake {
        arguments.add(0, "-DANDROID_STL=c++_shared")
        arguments.add(0, "-DSENTRY_NATIVE_SRC=$sentryNativeSrc")
        arguments.add("-DCMAKE_VERBOSE_MAKEFILE=ON")
        abiFilters.addAll(listOf("x86_64", "arm64-v8a"))
        abiFilters.forEach { abi ->
            println("ABI=$abi")
[...]

PR #1180, opened 2025-03-18
add NDK testing setup

JoshuaMoelans:
#skip-changelog
sidequest to be merged into https://github.com/getsentry/sentry-native/pull/1176

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -255,14 +255,6 @@ jobs:
         working-directory: ndk
         run: make check
 
-      - name: Run NDK tests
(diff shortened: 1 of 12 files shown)


JoshuaMoelans:
To avoid overlapping efforts with https://github.com/getsentry/sentry-native/issues/962 , we'll limit this PR to only set up NDK tests without running them in CI.

supervacuus:
> To avoid overlapping efforts with #962 , we'll limit this PR to only set up NDK tests without running them in CI.

If that is the path forward and everything works as you expect it, feel free to change the merge to `master` and rebase #1176 on top of that merge.

PR #1189, opened 2025-03-27
chore: Added .NET specific directories to `.gitignore`

bitsandfoxes:
Added these because the repo will always be "dirty" when used as a submodule as part of the .NET or Unity SDK.

#skip-changelog

Sidenote: 
Should these not be `/build*` and `/install*` instead?
https://github.com/getsentry/sentry-native/blob/ce9d742d5374d1902f52aa303d776a5d4dee2785/.gitignore#L8-L9
If I read that correct, the pattern matching is already excluding `/bin` inadvertently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ local.properties
 .cxx
 build
 
+# .NET
+bin/
+obj/


supervacuus:
> Sidenote:
> Should these not be `/build*` and `/install*` instead?

No, since these are explicitly meant for the `build` and `install` directories for CMake (in the project root), which often have both a prefix and a suffix.

supervacuus:
> If I read that correct, the pattern matching is already excluding `/bin` inadvertently.

I'm not sure what you mean. Are you asking whether the pattern `bin/` also includes a top-level `bin` directory? If yes, then the answer is also "yes".

Issue #1188, opened 2025-03-26
Allow specifying `attachment_type` for attachments


limbonaut:
I am currently implementing the View Hierarchy feature for the Godot SDK ([GH-143](https://github.com/getsentry/sentry-godot/pull/143)). As per [RFC](https://github.com/getsentry/rfcs/blob/main/text/0033-view-hierarchy.md), we need to set `attachment_type` to `event.view_hierarchy` and `content_type` to `application/json`. However, it appears that there is no direct way to do this in sentry-native.

Failing to do so, the View Hierarchy section is not shown in the Issues. I.e., just attaching the JSON file doesn't actually enable the corresponding UI view in Sentry.

PR #1191, opened 2025-03-27
feat: add option to attach view hierarchy

JoshuaMoelans:
Closes https://github.com/getsentry/sentry-native/issues/1188

- [x] add new option
- [x] make it work for Crashpad <- through Relay inference
- [x] infer attachment_type in Relay ([implement](https://github.com/getsentry/relay/blob/90f67c6fbfb8728168f871a9296d9253ef9b0574/relay-server/src/endpoints/minidump.rs#L122-L130) [test](https://github.com/getsentry/relay/blame/90f67c6fbfb8728168f871a9296d9253ef9b0574/relay-server/src/endpoints/minidump.rs#L355)) https://github.com/getsentry/relay/pull/4624
- [x] add to docs -> warning about needing the filename `view-hierarchy.json` or it won't work on Crashpad, linking to [RFC#33](https://github.com/getsentry/rfcs/blob/main/text/0033-view-hierarchy.md) https://github.com/getsentry/sentry-docs/pull/13191
- [x] update changelog
- [x] update contributing.md
- [x] add tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## Unreleased
 
 **Features**:
-- Provide the option to attach a view hierarchy file. ([#1191](https://github.com/getsentry/sentry-native/pull/1191))
+
+- Provide an option for downstream SDKs to attach a view hierarchy file. ([#1191](https://github.com/getsentry/sentry-native/pull/1191))
 
 ## 0.8.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add option to attach screenshots on Windows to fatal error events. ([#1170](https://github.com/getsentry/sentry-native/pull/1170))
 - Add an option for `Crashpad` on Linux to delay application shutdown until the upload of the crash report in the `crashpad_handler` is complete. This is useful for deployment in `Docker` or `systemd`, where the life cycle of additional processes is bound by the application life cycle. ([#1153](https://github.com/getsentry/sentry-native/pull/1153), [crashpad#121](https://github.com/getsentry/crashpad/pull/121))
+- Allow specifying `attachment_type` and `content_type` for attachments. ([#1191](https://github.com/getsentry/sentry-native/pull/1191))
 
 ## 0.8.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Add option to attach screenshots on Windows to fatal error events. ([#1170](https://github.com/getsentry/sentry-native/pull/1170))
 - Add an option for `Crashpad` on Linux to delay application shutdown until the upload of the crash report in the `crashpad_handler` is complete. This is useful for deployment in `Docker` or `systemd`, where the life cycle of additional processes is bound by the application life cycle. ([#1153](https://github.com/getsentry/sentry-native/pull/1153), [crashpad#121](https://github.com/getsentry/crashpad/pull/121))
-- Allow specifying `attachment_type` and `content_type` for attachments. ([#1191](https://github.com/getsentry/sentry-native/pull/1191))
+- Provide the option to attach a view hierarchy file. ([#1191](https://github.com/getsentry/sentry-native/pull/1191))
 
 ## 0.8.2
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -156,7 +156,8 @@ The example currently supports the following commands:
 - `socks5-proxy`: Uses a localhost `SOCKS5` proxy on port 1080.
 - `capture-transaction`: Captures a transaction.
 - `traces-sampler`: Installs a traces sampler callback function when used alongside `capture-transaction`.
-
+- `attach-view-hierarchy`: Adds an attachment as if it was a `view-hierarchy.json` file, giving it the proper `attachment_type` and `content_type`. It is currently defined as the
+  `CMakeCache.txt` file, which is part of the CMake build folder.
  
 Only on Linux using crashpad:
 - `crashpad-wait-for-upload`: Couples application shutdown to complete the upload in the `crashpad_handler`. 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -156,8 +156,8 @@ The example currently supports the following commands:
 - `socks5-proxy`: Uses a localhost `SOCKS5` proxy on port 1080.
 - `capture-transaction`: Captures a transaction.
 - `traces-sampler`: Installs a traces sampler callback function when used alongside `capture-transaction`.
-- `attach-view-hierarchy`: Adds an attachment as if it was a `view-hierarchy.json` file, giving it the proper `attachment_type` and `content_type`. It is currently defined as the
-  `CMakeCache.txt` file, which is part of the CMake build folder.
+- `attach-view-hierarchy`: Adds a `view-hierarchy.json` attachment file, giving it the proper `attachment_type` and `content_type`. 
+ This file can be found in `./tests/fixtures/view-hierachy.json`.
  
 Only on Linux using crashpad:
 - `crashpad-wait-for-upload`: Couples application shutdown to complete the upload in the `crashpad_handler`. 
--- a/examples/example.c
+++ b/examples/example.c
@@ -291,10 +291,10 @@ main(int argc, char **argv)
     }
 
     if (has_arg(argc, argv, "attach-view-hierarchy")) {
-        // TODO we assume the file lives one directory up;
-        //  should we add it to GH?
-        sentry_options_add_typed_attachment(options, "../view-hierarchy.json",
-            VIEW_HIERARCHY, "application/json");
+        // assuming the example / test is run directly from the cmake build
+        // directory
+        sentry_options_add_typed_attachment(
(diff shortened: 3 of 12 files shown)


examples/example.c
```diff
@@ -290,6 +290,13 @@ main(int argc, char **argv)
         sentry_options_set_crashpad_wait_for_upload(options, true);
     }
 
+    if (has_arg(argc, argv, "attach-view-hierarchy")) {
+        // TODO we assume the file lives one directory up;
+        //  should we add it to GH?
+        sentry_options_add_typed_attachment(options, "../view-hierarchy.json",
+            VIEW_HIERARCHY, "application/json");
```
supervacuus, line 297: I don't know whether we should specifically add a test for view hierarchy. I would upload whatever we have handy.

src/sentry_core.c
```diff
@@ -545,6 +561,15 @@ sentry__prepare_event(const sentry_options_t *options, sentry_value_t event,
         if (!item) {
             continue;
         }
+        if (attachment->type) {
```
supervacuus, line 564: While this is technically correct, since the backend seemingly auto-assigns `event.attachment` to envelopes without an `attachment_type`, I am not sure that relying on the numeric representation is the right approach.

We actually have to check `content_type` because it could be `NULL`, which has a meaning (we want auto-detection in the backend).

For `attachment_type`, we rely on the value being there so we can add the header unconditionally. If you want to express that we don't need to explicitly send the `attachment_type` if it is `event.attachment` because it is the default, then let's compare explicitly:

```suggestion
        if (attachment->type != ATTACHMENT) {
```

examples/example.c
```diff
@@ -290,6 +290,13 @@ main(int argc, char **argv)
         sentry_options_set_crashpad_wait_for_upload(options, true);
     }
 
+    if (has_arg(argc, argv, "attach-view-hierarchy")) {
+        // assuming the example / test is run directly from the cmake build
+        // directory
+        sentry_options_add_typed_attachment(
+            options, "./CMakeCache.txt", VIEW_HIERARCHY, "application/json");
+    }
+
```
JoshuaMoelans, line 299: We want to test this (similar to how we already test attachments); I think it makes sense to add a test for each backend that checks whether the attachment exists & has the `attachment_type:event.view_hierarchy` and `content_type:application/json`.

include/sentry.h
```diff
@@ -1186,6 +1186,20 @@ SENTRY_API void sentry_options_add_attachment(
...
+    MINIDUMP,
+    VIEW_HIERARCHY,
+} sentry_attachment_type_t;
+
+SENTRY_API void sentry_options_add_typed_attachment(sentry_options_t *opts,
+    const char *path, sentry_attachment_type_t attachment_type,
+    const char *content_type);
+SENTRY_API void sentry_options_add_typed_attachment_n(sentry_options_t *opts,
+    const char *path, size_t path_len, sentry_attachment_type_t attachment_type,
+    const char *content_type);
```
JoshuaMoelans, line 1201: We have [wchar versions ](https://github.com/getsentry/sentry-native/blob/ce9d742d5374d1902f52aa303d776a5d4dee2785/include/sentry.h#L1270-L1276) of `add_attachment`, do we also need this for these new functions?

supervacuus: + `content_type` needs a `content_type_len` parameter for `_n`

limbonaut:
Just letting you know that I tried it out with [GH-143](https://github.com/getsentry/sentry-godot/pull/143) and it worked! Thanks!

examples/example.c
JoshuaMoelans: I adapted the existing tests to also send the view-hierarchy attachment & check for its existence (see commit https://github.com/getsentry/sentry-native/pull/1191/commits/27f938f86ff4185f60b1d0aaccd45fc22b0587c9). These already work for backends Inproc/Breakpad/None, so only Crashpad testing is left.

examples/example.c
```diff
@@ -290,6 +290,12 @@ main(int argc, char **argv)
         sentry_options_set_crashpad_wait_for_upload(options, true);
     }
 
+    if (has_arg(argc, argv, "attach-view-hierarchy")) {
+        // assuming the example / test is run directly from the cmake build
+        // directory
+        sentry_options_add_view_hierarchy(options, "./CMakeCache.txt");
+    }
```
supervacuus, line 297: Since this is now an option specifically for that use case, I think it is fair to provide an actual `view-hierarchy.json` as a test fixture. See how you can bind the update of that file in the build directory to the `sentry_example` target: 

https://github.com/getsentry/sentry-native/blob/dc4d53dd9a5c225063bd5edfcb8387b2bb2e4912/CMakeLists.txt#L659-L660

You can use a JSON as minimal as in the relay test. There is no need to parse in our integration test (although I am certainly not against it either).

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+**Features**:
+- Provide the option to attach a view hierarchy file. ([#1191](https://github.com/getsentry/sentry-native/pull/1191))
```
JoshuaMoelans, line 6: Should the Changelog also point to the Relay update (https://github.com/getsentry/relay/pull/4624)?

include/sentry.h
```diff
@@ -1186,6 +1186,26 @@ SENTRY_API void sentry_options_add_attachment(
...
     sentry_options_t *opts, const char *path, size_t path_len);
 
+/**
+ * The attachment_type.
+ */
+typedef enum {
+    ATTACHMENT,
+    MINIDUMP,
+    VIEW_HIERARCHY,
+} sentry_attachment_type_t;
```
JoshuaMoelans, line 1196: Because we no longer provide an API that needs to expose the attachment types, maybe we should move this to where it is actually used?

CHANGELOG.md
supervacuus: It is okay if it's only visible from the PR. I only add our forked/vendored dependencies to the changelog. But you can add it if you'd like to.

tests/test_integration_http.py
```diff
@@ -284,7 +285,7 @@ def test_inproc_crash_http(cmake, httpserver, build_args):
     child = run(
         tmp_path,
         "sentry_example",
-        ["log", "start-session", "attachment", "crash"],
+        ["log", "start-session", "attachment", "attach-view-hierarchy", "crash"],
```
supervacuus, line 288: While I do understand that nothing changed in this PR wrt `crashpad` backend, I still think it makes sense to add a test for crashpad as well:

1. to automatically verify it works :-)
2. to ensure that future changes to `crashpad` attachment handling won't break it
3. to have an explicitly stated contract in form of test, that says: crashpad doesn't send `application/json` and we verified that it works in relay.

include/sentry.h
```diff
@@ -1186,6 +1186,18 @@ SENTRY_API void sentry_options_add_attachment(
 SENTRY_API void sentry_options_add_attachment_n(
     sentry_options_t *opts, const char *path, size_t path_len);
 
+/**
+ * Adds a new view hierarchy attachment to be sent along.
+ *
+ * `path` is assumed to be in platform-specific filesystem path encoding.
+ * API Users on windows are encouraged to use
+ * `sentry_options_add_view_hierarchyw` instead.
```
supervacuus, line 1194: From https://github.com/getsentry/sentry-docs/pull/13191#pullrequestreview-2735095217

Let's extend this to say that this interface's primary use case is its use in downstream SDKs. I couldn't think of an immediate reason any direct SDK user would use this option.

tests/test_integration_http.py
JoshuaMoelans: I've added it to the Crashpad tests in https://github.com/getsentry/sentry-native/pull/1191/commits/aef02036d2ca6145f672f7d2fe003908da732df5 .

include/sentry.h
JoshuaMoelans: I extended it, and am also open to not adding the docs page entry for this feature.

CONTRIBUTING.md
```diff
@@ -156,7 +156,8 @@ The example currently supports the following commands:
 - `socks5-proxy`: Uses a localhost `SOCKS5` proxy on port 1080.
 - `capture-transaction`: Captures a transaction.
 - `traces-sampler`: Installs a traces sampler callback function when used alongside `capture-transaction`.
-
+- `attach-view-hierarchy`: Adds an attachment as if it was a `view-hierarchy.json` file, giving it the proper `attachment_type` and `content_type`. It is currently defined as the
+  `CMakeCache.txt` file, which is part of the CMake build folder.
```
supervacuus, line 160: The latter part is no longer valid; please update it.

CHANGELOG.md
supervacuus, line 6: Sorry for being pedantic, but let's explicitly say for downstream SDKs.
```suggestion
- Provide an option for downstream SDKs to attach a view hierarchy file. ([#1191](https://github.com/getsentry/sentry-native/pull/1191))
```

supervacuus (approved):
~Two~ Three minor updates are needed before it's ready to merge.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+**Features**:
```
supervacuus, line 5:
```suggestion
**Features**:

```

Issue #1193, opened 2025-04-02
crash inside nvidia drivers since 0.7.6


pcorbineau:
**Description**
We are facing an issue since we update sentry-native.
A StackOverflow is catched inside nvwgf2umx.dll since 0.7.6. Last version 0.8.3 not working also.
The last working version was 0.6.5 ( we did not test version between 0.6.5 and 0.7.6)

**When does the problem happen**

- [ ] During build
- [x] During run-time
- [ ] When capturing a hard crash

**Environment**

- OS: Windows 11
- Compiler: [e.g. MSVC 19] : MSVC 19
- CMake version and config: [e.g. 3.17.2, SENTRY_BACKEND=inproc]: 3.31
building with options : `-DSENTRY_BUILD_RUNTIMESTATIC=OFF -DSENTRY_BUILD_TESTS=OFF -DSENTRY_BUILD_EXAMPLES=OFF`

**Steps To Reproduce**

Hard to produce a minimum setup. But if we change sentry.dll from 0.7.6 to 0.6.5 our app runs.

Unlike rjamieson [from this forum thread](https://forums.developer.nvidia.com/t/stack-overflow-exception-in-nvwgf2umx-dll/327461/2) , if have issue with both `-DSENTRY_BUILD_RUNTIMESTATIC=OFF` and `-DSENTRY_BUILD_RUNTIMESTATIC=ON`

**Log output**

```
Exception thrown at 0x00007FFCD691A3DD (nvwgf2umx.dll) : 0xC00000FD: Stack overflow (parameters: 0x0000000000000001, 0x000000B1F1701FB8).

nvwgf2umx.dll!00007ffcd691a3dd()
nvwgf2umx.dll!00007ffcd729842b()
nvwgf2umx.dll!00007ffcd6943a95()
nvwgf2umx.dll!00007ffcd687fb44()
nvwgf2umx.dll!00007ffcd688132d()
nvwgf2umx.dll!00007ffcd6d22b07()
nvwgf2umx.dll!00007ffcd6d11780()
nvwgf2umx.dll!00007ffcd6d1cdef()
nvwgf2umx.dll!00007ffcd6d1cd4f()
nvwgf2umx.dll!00007ffcd6d1162d()
nvwgf2umx.dll!00007ffcd71b0d2a()
nvwgf2umx.dll!00007ffcd76f42fa()
kernel32.dll!BaseThreadInitThunk()
ntdll.dll!RtlUserThreadStart()
```

[...]

PR #1196, opened 2025-04-08
fix: more defensive auto thread stack guarantee

supervacuus:
This builds on top of #1049, provides a fix for #1193, and adds further parameterization for users who want to tune this to their use cases.

### The PR does the following

* Expose a compile-time parameter `SENTRY_HANDLER_STACK_SIZE` that sets the required stack size for the handler (for Linux/macOS using `inproc` and Windows using any backend, it defaults to 64KiB, which is also the current default).
* Expose a compile-time parameter `SENTRY_THREAD_STACK_GUARANTEE_FACTOR` which is the factor that the stack reserve of a given thread must at least be larger than the specified `SENTRY_HANDLER_STACK_SIZE` to ask for a guarantee (only Windows, defaults to 10x).
* Expose a compile-time option `SENTRY_THREAD_STACK_GUARANTEE_AUTO_INIT` that toggles between turning on and off auto-initialization of the thread stack guarantee for each thread after the sentry library has been loaded (if building as a dynamic library). This option only initializes the thread that calls `sentry_init()` when doing a static build (only Windows, defaults to `On`).
* Expose a compile-time option `SENTRY_THREAD_STACK_GUARANTEE_VERBOSE_LOG` that toggles between turning on and off `INFO` level logging of successful setting of thread stack guarantees (only Windows, defaults to `Off`).
* Provide a public interface (`sentry_set_thread_stack_guarantee()`) to allow users to set the stack guarantee for each thread manually if the auto-initialization doesn't behave as the use case requires it
* Adapt the auto-initialization to be much more defensive about which threads it attempts to request a stack guarantee for so we don't eat into the stack budget of threads with tiny reserves. The reserve and above meta-parameters play into the decision, which essentially boils down to this:
**if a thread doesn't have at least `SENTRY_HANDLER_STACK_SIZE x SENTRY_THREAD_STACK_GUARANTEE_FACTOR` of stack reserve, we won't request a guarantee from that thread.**

### Reasoning

* Most threads in users' processes will have enough stack reserve (1MiB) for our default guarantee (64KiB), so there is no reason to disable auto-initialization because most users will never be confronted with any issue
* However, no auto initialization should itself lead to a stack overflow, and thus, it makes sense to introduce a boundary at which guarantees will no longer be attempted
* The trade-off here is that users who have threads in their process that do not fit that boundary will have some threads without a handler guarantee (i.e. if these overflow, they will crash the application, but the crash won't be reported)
* This might sound like a bad trade-off; however, most of these threads will be outside the control of our users, and as such, any reports will be non-actionable.
* Further, we now expose parameters for the tiny group of users that control threads with limited stacks or with enough insight into third-party-controlled threads so that reports can be actionable if they are overflowing:
  * they can disable auto-initialization entirely and define guarantees for each thread as they see fit
  * they can change the handler stack size according to their usage of hooks (with or without auto-init)
  * they can change the thread stack guarantee factor so that threads with limited stacks will still be automatically initialized with a guarantee they have tested
* We keep the current default guarantee of 64KiB to avoid breaking the current handler assumptions in the field (particularly concerning hooks being called from the handler).


README.md
```diff
@@ -246,32 +243,58 @@ using `cmake -D BUILD_SHARED_LIBS=OFF ..`.
...
+- `SENTRY_SDK_NAME` (Default: `sentry.native` or `sentry.native.android`):
   Sets the SDK name that should be included in the reported events. If you're overriding this, also define
   the same value using `target_compile_definitions()` on your own targets that include `sentry.h`.
 
+- `SENTRY_HANDLER_STACK_SIZE` (Default: `64`):
+  This specifies the size in KiB of the stack reserved for the crash handler (including hooks like `on_crash` and
+  `before_send`) on Windows, Linux and the `inproc` backend in macOS. Reserving the stack is necessary in case of a
+  stack-overflow, where the handler could otherwise no longer execute. This parameter allows users to specify their
+  target stack size in KiB, because some applications might require a different value from our default. This
+  value can be as small as 16KiB (on `crashpad` and `breakpad`) for handlers to work, but we recommend 32KiB on 64-bit
```
JoshuaMoelans, line 273: Why do we recommend 32KiB but set the default to 64? Is 32KiB the minimum recommendation?

include/sentry.h
```diff
@@ -1298,6 +1298,30 @@ SENTRY_API void sentry_options_set_database_pathw(
...
     sentry_options_t *opts, const wchar_t *path, size_t path_len);
+
+/**
+ * Allows users to define a thread stack guarantee manually on each thread. This
+ * is necessary for crash handlers to run after a thread crashed due to a
+ * stack-overflow on Windows.
+ *
+ * By default the Native SDK automatically sets this value for you, but in some
+ * cases it might be preferable to set these values yourself for each thread you
+ * manage. Ensure to disable the `SENTRY_THREAD_STACK_GUARANTEE_AUTO_INIT` CMake
```
JoshuaMoelans, line 1309:
```suggestion
 * manage. Make sure to disable the `SENTRY_THREAD_STACK_GUARANTEE_AUTO_INIT` CMake
```

src/sentry_core.c
```diff
@@ -192,6 +192,13 @@ sentry_init(sentry_options_t *options)
     sentry_integration_setup_qt();
 #endif
 
+#if defined(SENTRY_PLATFORM_WINDOWS)                                           \
+    && (!defined(SENTRY_BUILD_SHARED) || defined(_GAMING_XBOX_SCARLETT))
+    // This function must be positioned so that any dependents on its cached
+    // functions are invoked after it.0
```
JoshuaMoelans, line 198:
```suggestion
    // functions are invoked after it.
```

src/sentry_os.c
```diff
@@ -168,16 +170,138 @@ sentry__get_os_context(void)
...
+        "Couldn't load `SetThreadStackGuarantee`: "
+        "`sentry_set_thread_stack_guarantee()` won't work. (error-code: "
+        "`%lu`)");
+
+    // Retrieve `GetCurrentThreadStackLimits()` for Windows 8+ targets.
+    LOAD_FUNCTION(kernel32, "GetCurrentThreadStackLimits",
+        void(WINAPI *)(PULONG_PTR, PULONG_PTR),
+        g_kernel32_GetCurrentThreadStackLimits,
+        "Couldn't load `GetSystemTimePreciseAsFileTime`. Falling back on "
+        "`GetSystemTimeAsFileTime`. (error-code: `%lu`)");
```
JoshuaMoelans, line 230: This doesn't seem like the correct message here

JoshuaMoelans (commented):
Some first nitpicks/typos/questions

README.md
supervacuus: Yes, it is meant as a minimum recommendation. I'll make sure to explain that.

include/sentry.h
supervacuus: Can I ask why? I consider "ensure" to be more formal and shorter, both great attributes for documentation.

JoshuaMoelans: For some reason "ensure to" sounds a bit weird, but I cannot find grammatical backings for this suggestion so I'm fine with keeping it as-is.

supervacuus:
@stima, I just wanted to ping you with the soon-to-be-released approach regarding the previous conversations. In particular, this last comment also led to the current reasoning: https://github.com/getsentry/sentry-native/pull/1049#issuecomment-2426962688. In your scenario, you could leave the auto-init enabled to cover the trivial threads and only manually set the threads you know might be excluded anyway (or not assign a guarantee for those).

@pcorbineau, @robert-jamieson, and @MaxLeb: I will release this next week as a fix. If you have time to try out the PR, feedback would be welcome. The primary concern is that no stack overflow happens in your current setup without additional configuration. 

Of course, I would also be happy to know if the exposed parameters make sense to you. In your scenario, at least based on what I gathered from your previous feedback, you could reduce the handler stack size to 32KiB and only require a factor of two. Of course, this may not be enough for these threads in all cases, so if overflows from these threads (however unlikely) aren't actionable due to missing symbols, you can leave it at the defaults, which won't set any guarantees for nVidia's driver threads.

supervacuus:
> We did the tests

Thanks a lot for the feedback!

> 1. We do not have any stack overflow happening with default settings

Great!

> 2. Changing options to reduce to 32KiB with a factor of 2 does work but we indeed have errors when trying to reserve the stack (which might be ok for us)

The default settings should work well for you, too. The parameterization is primarily there if you see an issue with the particular threads not being "covered" or if you own such a thread and have detailed stack usage bounds.

JoshuaMoelans (approved):
lgtm!

tests/unit/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ add_executable(sentry_test_unit
 	test_modulefinder.c
 	test_mpack.c
 	test_options.c
-	test_os_release.c
+    test_os.c
```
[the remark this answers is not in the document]
supervacuus: Thanks. That was on my to-do list, but somehow, I scratched it before doing it. There are multiple tabs vs. spaces errors from recent changes that I have now fixed. The primary cause seems to be that CLion ignores the `editorconfig` for CMake.

PR #1198, opened 2025-04-11
feat: NX support for curl transport

vaind:
This enables the downstream NX SDK to configure curl with custom options.

#skip-changelog - this is an internal change for the downstream NX SDK

### DIFF
--- a/tests/unit/test_failures.c
+++ b/tests/unit/test_failures.c
@@ -1,11 +1,7 @@
 #include "sentry_core.h"
+#include "sentry_options.h"
 #include "sentry_testsupport.h"
 
-#ifdef SENTRY_PLATFORM_NX
-SENTRY_TEST(init_failure)
-{
-    sentry_options_t *options = sentry_options_new();
-#else
 static int
 transport_startup_fail(
     const sentry_options_t *UNUSED(options), void *UNUSED(state))
@@ -26,9 +22,16 @@ SENTRY_TEST(init_failure)
 
     SENTRY_TEST_OPTIONS_NEW(options);
     sentry_options_set_transport(options, transport);
-#endif
     sentry_options_set_dsn(options, "https://foo@sentry.invalid/42");
     int rv = sentry_init(options);
 
+#ifdef SENTRY_PLATFORM_NX
+    // On NX a failing transport must not fail initialization.
+    TEST_CHECK(rv == 0);
+    SENTRY_WITH_OPTIONS (runtime_options) {
+        TEST_CHECK(runtime_options->transport == NULL);
+    }
+#else
     TEST_CHECK(rv != 0);
+#endif
 }


tests/unit/test_failures.c
```diff
@@ -1,6 +1,11 @@
 #include "sentry_core.h"
 #include "sentry_testsupport.h"
 
+#ifdef SENTRY_PLATFORM_NX
+SENTRY_TEST(init_failure)
+{
+    sentry_options_t *options = sentry_options_new();
+#else
```
supervacuus, line 8: I don't understand what this test is doing. I prefer a clean separation of preprocessor conditionals here instead of splitting two tests at a sensible but hard-to-read boundary. Something like this:

```C
#include "sentry_core.h"
#include "sentry_testsupport.h"

#ifdef SENTRY_PLATFORM_NX

SENTRY_TEST(init_failure)
{
    sentry_options_set_dsn(options, "https://foo@sentry.invalid/42");
    int rv = sentry_init(options);

    TEST_CHECK(rv != 0);
}

#else

static int
transport_startup_fail(
    const sentry_options_t *UNUSED(options), void *UNUSED(state))
{
    return 1;
}

static void
noop_send(const sentry_envelope_t *UNUSED(envelope), void *UNUSED(data))
{
}

SENTRY_TEST(init_failure)
{
    sentry_transport_t *transport
        = sentry_new_function_transport(noop_send, NULL);
    sentry_transport_set_startup_func(transport, transport_startup_fail);

    sentry_options_set_transport(options, transport);
    sentry_options_set_dsn(options, "https://foo@sentry.invalid/42");
    int rv = sentry_init(options);

    TEST_CHECK(rv != 0);
}

#endif
```

The duplication is fine, making the two tests much more readable, especially if things change later.

However, I still don't get what the NX version of that test does. You don't inject failure, so initializing with only a parseable but otherwise invalid (unreachable?) DSN  leads to `sentry_init()` failing, which is not visible from the rest of your changes (you even continue if the transport startup failed). 

If that is what you are testing, I would like a small comment like `// We fail init on NX if the DSN is unreachable` (or whatever doesn't reveal too much information) since under normal conditions that DSN or options setup wouldn't fail `sentry_init()`.

supervacuus (changes requested):
It looks good; just a minor clean-up and question regarding the provided test.

tests/unit/test_failures.c
vaind: dsn isn't actually the important bit, it's the part where we use sentry_options_new() without setting an explicit database directory (which will fail on NX because the current directory is not writable). My bad for not writing this down in the test as comments. 

However, your prompt pushed me to rething this and actually, we should be testing that the failing transport still works but is ignored. I've changed the test code to match that.

supervacuus (approved):

PR #1199, opened 2025-04-15
feat: add windows arm64 CI runner

JoshuaMoelans:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,6 +111,8 @@ jobs:
             TEST_X86: 1
           - name: Windows (latest)
             os: windows-latest
+          - name: Windows arm64
+            os: windows-11-arm
           - name: Windows ClangCL (latest)
             os: windows-latest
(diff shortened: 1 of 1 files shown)


supervacuus:
> currently failing due to [Cargo being missing](https://github.com/getsentry/sentry-native/actions/runs/14469109376/job/40578334469?pr=1199#step:19:109) . Looking at the runner inventory it is indeed [not listed](https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md) (although on the Windows2022 image [it is present](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#rust-tools)). Since it is not listed as "omitted" on the arm runner image, it might just not be available (yet).

we can install this ourselves, if only for the ARM64 Windows runner. however, why do we even need cargo/rust?

supervacuus:
> however, why do we even need cargo/rust?

ugh, `mitmproxy` > `pyca/cryptography`

PR #1200, opened 2025-04-15
fix: trace sync improvements

JoshuaMoelans:
Related to the investigation of https://github.com/getsentry/sentry-native/issues/1167

The aim of this PR is to rework the way we handle trace syncing (e.g. with `set_trace`) and how this affects data inside events & spans.

There's `ToDos` left in the code to indicate some parts of the implementation left to discuss.

- [x] use `propagation_context` for `set_trace`
- [x] apply `propagation_context` for new/scoped spans
- [x] adapt tests & document expectations 
- [x] add tests (span children, order of operations wrt scoping/creation/`set_trace`)
- [x] write docs (e.g. develop docs, similar to [JS Tracing docs](https://develop.sentry.dev/sdk/platform-specifics/javascript-sdks/browser-tracing/#tracing-model)) https://github.com/getsentry/sentry-docs/pull/13599

#skip-changelog for now


tests/unit/test_tracing.c
```diff
@@ -1377,37 +1378,165 @@ SENTRY_TEST(set_trace_id_with_txn)
...
+        = sentry_span_start_child(span_child, "op_g", "desc_g");
 
     // set the direct trace first
     const char *direct_trace_id = "aaaabbbbccccddddeeeeffff00001111";
     const char *direct_parent_span_id = "f0f0f0f0f0f0f0f0";
     sentry_set_trace(direct_trace_id, direct_parent_span_id);
 
     // events should get that trace applied
+    // TODO reason why they should get that trace applied
+    //  (tx started before set_trace, but trace was set so events get that data)
```
JoshuaMoelans, line 1413: (idea) Trace was set after Spans started, so these should not have the `trace_id` (unless they now get scoped).

tests/unit/test_tracing.c
```diff
@@ -1377,37 +1378,165 @@ SENTRY_TEST(set_trace_id_with_txn)
...
 
-    // events should get the transaction's trace applied
+    // check if trace_id was passed down properly
+    // TODO maybe refactor these span checks into function
+    TEST_CHECK_STRING_EQUAL(
+        sentry_value_as_string(sentry_value_get_by_key(tx->inner, "trace_id")),
+        direct_trace_id);
+    // TODO we only set the transaction on the scope; how about the child spans?
+    //  only apply the trace IF they also get scoped? or should they now
+    //  inherit?
```
JoshuaMoelans, line 1428: (question) When a transaction gets put on the scope, should the existing child spans take over the `trace_id` the same way the now-scoped transaction does? Or do child spans only take the `trace_id` whence they themselves get scoped?

tests/unit/test_tracing.c
```diff
@@ -1377,37 +1378,165 @@ SENTRY_TEST(set_trace_id_with_txn)
...
+    sentry_set_span(span_child);
+    TEST_CHECK_STRING_EQUAL(sentry_value_as_string(sentry_value_get_by_key(
+                                span_child->inner, "trace_id")),
+        direct_trace_id);
+    // TODO grandchild does not get scoped, so what should its trace_id be?
+    TEST_CHECK_STRING_NOT_EQUAL(sentry_value_as_string(sentry_value_get_by_key(
+                                    span_grandchild->inner, "trace_id")),
+        direct_trace_id);
```
JoshuaMoelans, line 1436: This assumes the 2nd option described above; child spans created before `set_trace` only inherit the `set_trace` values if they get scoped (and do not get it from scoping the root span)

tests/unit/test_tracing.c
```diff
@@ -1377,37 +1378,165 @@ SENTRY_TEST(set_trace_id_with_txn)
         = "2674eb52d5874b13b560236d6c79ce8a-a0f9fdf04f1a63df-1";
     const char *txn_trace_id = "2674eb52d5874b13b560236d6c79ce8a";
     const char *txn_parent_span_id = "a0f9fdf04f1a63df";
+    sentry_transaction_context_t *tx_ctx_scoped
+        = sentry_transaction_context_new("wow!", NULL);
+    sentry_transaction_context_update_from_header(
+        tx_ctx_scoped, "sentry-trace", trace_header);
```
JoshuaMoelans, line 1384: We should also test doing this after `set_trace` was called; IMO it should get priority over the value gotten from `set_trace`.

supervacuus: Yes, iirc that was the idea (i.e., explicit traces from external contexts overriding the inherited trace) from our last sync on the topic. Let's add any test that raises similar questions until we hit an implementation blocker.

tests/unit/test_tracing.c
supervacuus: This is where it gets complicated and why I believe applying the propagation context when scoping can be over-eager: the application can make sense if it's a transaction, but why would a span be connected to another trace just because the user scoped it?

I know that I proposed this "trace-application at scoping time", but the cleanest approach still is only to do it during creation. It is also a problem that, while we can lock the scope, we can't lock the span (or txn) that a user retains a reference to (so it is also racy).

tests/unit/test_tracing.c
supervacuus: ditto.

tests/unit/test_tracing.c
supervacuus: I mean, just to be clear: I am also not quite sure what it means if a user introduces an external trace that overwrites the traces introduced by SDKs that run "above" the Native SDK. And I mean this concretely for the Native SDK and generally. The latter seems to be answered by "Sentry does it that way", but for the Native SDK specifically, it also highlights the fact that we only have a single scope. The above behavior makes perfect sense if a thread continues an external trace, while all others stay inside the propagated trace.

src/sentry_scope.c
```diff
@@ -76,6 +76,7 @@ get_scope(void)
     g_scope.tags = sentry_value_new_object();
     g_scope.extra = sentry_value_new_object();
     g_scope.contexts = sentry_value_new_object();
+    g_scope.propagation_context = sentry_value_new_object();
```
JoshuaMoelans, line 79: From the [TwP docs
](https://develop.sentry.dev/sdk/telemetry/traces/tracing-without-performance/#trace-duration-and-storage)

> On a related note, the propagationContext should be populated with a random traceId and spanId if no incoming trace is present. This—in combination with the [sentry-trace header specification](https://develop.sentry.dev/sdk/telemetry/traces/#header-sentry-trace) requiring a spanId—has an important implication on the Sentry product: A non-existing spanId will be propagated along with the trace and attached to events. While not ideal, we accept this limitation as the Sentry product can and should handle non-existing (parent) spans anyway.

So upon creation, a random `trace_id` and `span_id` should be stored in our `propagation_context`. However, it seems like the main use of this would be to propagate them in the `sentry-trace` header to downstream services, which I don't think we are doing (related to [`tracePropagationTargets`](https://develop.sentry.dev/sdk/telemetry/traces/#tracepropagationtargets)). 

I think there is probably no harm in doing this, but also no real benefit, so I'm not convinced implementing this is necessary.

supervacuus: We do have `sentry_transaction_iter_headers()`, which is used to propagate a trace from a transaction to downstream services. Although we currently do not do this for TwP, I see no reason not to do the right thing here.

src/sentry_core.c
```diff
@@ -182,6 +182,18 @@ sentry_init(sentry_options_t *options)
...
+                scope->propagation_context, "trace", sentry_value_new_object());
+            sentry_uuid_t trace_id = sentry_uuid_new_v4();
+            sentry_uuid_t span_id = sentry_uuid_new_v4();
+            sentry_value_set_by_key(
+                sentry_value_get_by_key(scope->propagation_context, "trace"),
+                "trace_id", sentry__value_new_internal_uuid(&trace_id));
+            sentry_value_set_by_key(
+                sentry_value_get_by_key(scope->propagation_context, "trace"),
+                "span_id", sentry__value_new_span_uuid(&span_id));
+        }
```
JoshuaMoelans, line 196: (related to [this conversation](https://github.com/getsentry/sentry-native/pull/1200#discussion_r2058015839))

The [TwP docs](https://develop.sentry.dev/sdk/telemetry/traces/tracing-without-performance/#trace-duration-and-storage) tell us to set the propagation context such that events in a span-less world still have a trace. Hence we now check if we are running native without spans, and only then set the propagation context.

This avoids the issue where setting these values for all Native SDK inits leads to Transactions picking up the same data from the propagation context which is also used by default in any Event (that has no scoped transactions) leading to non-scoped transactions having the same `trace_id` as these Events (which should not be the case).

supervacuus: Given this auto-init, should these values be applied to any event as long as no trace was set from above or transaction created otherwise? I.e., generalizing the workflow in a direction where everything always inherits from here (even the transactions constructed outside of an incoming trace boundary, whether TwP or from transaction continuation). 

I am not sure if that is the intention, but it seemed like with TwP, everything should be connected to _some_ trace. That could be the reason for the auto-init, too. Or was that what you meant, and I was too stupid to comprehend?

JoshuaMoelans: The way I understood it from the JS SDK folks I talked to, they always use the propagation context as a single source-of-truth for the currently active `trace_id`. In TwP (which this auto-init assumes) we would then have a generated `trace_id` and `span_id` (and these will only change when `set_trace` gets called) such that events have *some* trace indeed. This also allows users to define trace boundaries inside Native even without Performance (which I think wasn't really possible without `set_trace` before in a TwP setting).

If we now did the auto-init for the non-TwP case, we would run into the issue where even a transaction that has not been scoped will get the `propagation_context`'s value for the `trace_id` set, but that value would also get used when we capture an event (which then connects the `tx` and the event even though it was not scoped). In JS it seems like they always use the propagation context, but that doesn't seem to work for us, hence I moved the init inside the Twp-if-clause.

So if we do have Performance enabled, we only set the `trace_id` on events IF there was a scoped Span (inheriting the id from there) OR if there was no scoped span, we look into the `propagation_context` (which would then only have a value if `set_trace` was called).

(I don't think understanding this is about stupidity, it's just a rather confusing topic 😅 )
Also, it seems like none of this is standardized across SDKs, so it's all just an (in this case, my) interpretation of the somewhat vague develop docs.

supervacuus: > If we now did the auto-init for the non-TwP case, we would run into the issue where even a transaction that has not been scoped will get the propagation_context's value for the trace_id set, but that value would also get used when we capture an event (which then connects the tx and the event even though it was not scoped).

To be honest, from my current understanding, this is acceptable behavior. My current understanding is that TwP is not a separate feature, but the need to always have traces even when performance is off.

There is no reason that an event doesn't belong to the same trace as a later-scoped transaction. It shouldn't be within the same span, but it can certainly be within the same trace. That would also be the case when `set_trace` was called from above before an event was captured, where the event would get the trace context from that trace, and a transaction being created within that trace, but later scope would also share the trace with the event that wasn't captured within its bounds:

set_trace ->
capture_event "A" -> event "A" would get trace-context that connects to that trace)
transaction_context_new/transaction_start/set_transaction_object -> transaction inherits the trace
capture_event "B" -> event "B" would get the trace-context from the transaction which inherits the trace, but has a different span_id

In this scenario, events "A" and "B" and the transaction belong to the same trace, but "A" is associated with the TwP span, whereas "B" is associated with the transaction span.

Using auto-init is the same, except that no one from above set those explicitly.

supervacuus: > I don't think understanding this is about stupidity, it's just a rather confusing topic 😅

I don't even think it's overly confusing. It is mainly that the implementations deviate quite a bit from a relatively underspecified spec, and we now have to clarify a few behaviors. However, writing the cases down for clarity is not an effortless task, as understanding what was written is also challenging, given the heavy vocabulary used in the topic.

We can certainly review the example from above again with your graphical scenarios.

supervacuus: A parallel visualization to your use-case graphs would be to show exactly where the data comes from, which I attempted to do with the text flow above. It is one thing to understand the scenarios and another to demonstrate the data flow that enables us to model the desired behavior (especially: what is the source of the trace-context at each point in each scenario)?

src/sentry_core.c
```diff
@@ -195,6 +195,17 @@ sentry_init(sentry_options_t *options)
...
+        sentry_value_set_by_key(
+            scope->propagation_context, "trace", sentry_value_new_object());
+        sentry_uuid_t trace_id = sentry_uuid_new_v4();
+        sentry_uuid_t span_id = sentry_uuid_new_v4();
+        sentry_value_set_by_key(
+            sentry_value_get_by_key(scope->propagation_context, "trace"),
+            "trace_id", sentry__value_new_internal_uuid(&trace_id));
+        sentry_value_set_by_key(
+            sentry_value_get_by_key(scope->propagation_context, "trace"),
+            "span_id", sentry__value_new_span_uuid(&span_id));
```
supervacuus, line 208: This block looks like a great static helper function.

supervacuus (approved):
This looks good! Thanks, @JoshuaMoelans. Do you still want to add integration tests for some of the modelled use cases, or do you think the unit tests cover things well enough?

Have you tried to run this inside sentry-android (using a local version of the ndk package)?

JoshuaMoelans:
I think integration tests make sense; currently the unit tests only ensure that spans get created properly, not that they leave the native SDK with the expected values. I will add those before merging 👍 

Local testing with `sentry-android` seems very promising, as now transactions/spans are trace-connected to the Android layer running above.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Breaking changes**:
+
+- Use `propagation_context` as the single source of `trace_id` for spans and events. Transactions no longer create a new trace, but inherit the trace from the `propagation_context`. If needed, this context can be updated manually through `sentry_set_trace()` ([#1200](https://github.com/getsentry/sentry-native/pull/1200))
```
supervacuus, line 7:
```suggestion
- Use `propagation_context` as the single source of `trace_id` for spans and events. Transactions no longer create a new trace, but inherit the trace from the `propagation_context` created during SDK initialization. This context can be later modified through `sentry_set_trace()` (primarily used by other SDKs). ([#1200](https://github.com/getsentry/sentry-native/pull/1200))
```

supervacuus (approved):

PR #1211, opened 2025-04-23
test: fix tests if release is set, e.g. in an env var

vaind:
While preparing another PR (#1204), I've faced an issue that if the downstream SDK sets release in sentry_init(), some tests that match captured envelopes using a custom transport, start failing. The same could be checked by running sentry_test_unit after seetting `SENTRY_RELEASE` env var.

I've tested the changes manually - should I also change tests or CI to test this?

#skip-changelog


supervacuus (approved):
It is fine to turn off session tracking here, as none of the affected tests require it, and with this change, it is now encoded and documented. Whether `SENTRY_RELEASE` is applied is covered in integration tests. I would leave this for now.

PR #1215, opened 2025-04-29
chore: tweak benchmark charts

jpnurmi:
By default, every "customSmallerIsBetter" chart is `#ff3838`. Add distinct colors per OS:

- Windows: `#357ec7` (Windows blue)
- macOS: `#717378` (Space gray)
- Linux: `#e95420` (Ubuntu orange)

This makes it easier to visually distinguish between OSes in the benchmark charts.

| Before | After |
|---|---|
| image | image |

https://github.com/benchmark-action/github-action-benchmark?tab=readme-ov-file#customizing-the-benchmarks-result-page

#skip-changelog

### DIFF
--- a/index.html
+++ b/index.html
@@ -126,6 +126,12 @@
           _: '#333333'
         };
 
+        const benchColors = {
+          Windows: '#357ec7',
+          macOS: '#717378',
+          Linux: '#e95420',
+        };
+
         function init() {
           function collectBenchesPerTestCase(entries) {
             const map = new Map();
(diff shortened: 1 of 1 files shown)


supervacuus (changes requested):
LGTM! Can you add a section to the `README.md` that links to the benchmark graphs and explains what we are benchmarking and how we gather them? And then, a paragraph on how to run the benchmarks locally. Thx!

JoshuaMoelans:
not sure if this is the place, but since we're restyling the graphs here as well, I had a two related nits: could we truncate the commit message such that the data at the bottom of the hover-field doesn't get cut off? 
image

Second nit; maybe we should label the y-axis with 'mean SDK startup time (ms)' so it's clear which measure we are displaying.

supervacuus (approved):

JoshuaMoelans (approved):

PR #1221, opened 2025-04-30
Add FreeBSD platform detection to sentry.h

mq1n:
Extend platform detection logic to recognize FreeBSD as a supported platform.

Define SENTRY_PLATFORM_FREEBSD and SENTRY_PLATFORM_UNIX macros when compiling on FreeBSD. It's sufficient for the overall build to work without any additional modifications to the cpp files as FreeBSD is compatible with the existing Unix platform code.

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -52,6 +52,9 @@ extern "C" {
 #elif defined(__linux) || defined(__linux__)
 #    define SENTRY_PLATFORM_LINUX
 #    define SENTRY_PLATFORM_UNIX
+#elif defined(__FreeBSD__)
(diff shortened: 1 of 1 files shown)


supervacuus:
Hi @mq1n. Can you explain what you want to achieve with these changes? I detailed the changes for FreeBSD to become a community-supported platform (support from Sentry would at least require the tests to run in CI) for the Native SDK: https://github.com/getsentry/sentry-native/pull/1113#issuecomment-2685474047

By just adding a marker for FreeBSD and making it a UNIX, a considerable part of the Native SDK features will not work.

PR #1224, opened 2025-05-04
chore: PS downstream SDK support

vaind:
Enables a private sentry-playstation SDK to build for PlayStation. I'll follow up with test updates in another PR once I get them to run on PS (or update this one if it's still not reviewed)

#skip-changelog because it's only used internally.

### DIFF
--- a/src/sentry_utils.c
+++ b/src/sentry_utils.c
@@ -389,7 +389,7 @@ sentry__usec_time_to_iso8601(uint64_t time)
     size_t buf_len = sizeof(buf);
     time_t secs = time / 1000000;
     struct tm *tm;
-#if defined(SENTRY_PLATFORM_WINDOWS) || defined(SENTRY_PLATFORM_PROSPERO)
+#if defined(SENTRY_PLATFORM_WINDOWS) || defined(SENTRY_PLATFORM_PS)
     tm = gmtime(&secs);
 #else
     struct tm tm_buf;
@@ -462,7 +462,7 @@ sentry__iso8601_to_usec(const char *iso)
     tm.tm_sec = s;
 #ifdef SENTRY_PLATFORM_WINDOWS
     time_t time = _mkgmtime(&tm);
-#elif defined(SENTRY_PLATFORM_PROSPERO)
+#elif defined(SENTRY_PLATFORM_PS)
     time_t time = mktime(&tm); // TODO is this correct?
 #elif defined(SENTRY_PLATFORM_AIX)
     /*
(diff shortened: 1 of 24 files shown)


src/sentry_utils.c
```diff
@@ -462,6 +462,8 @@ sentry__iso8601_to_usec(const char *iso)
     tm.tm_sec = s;
 #ifdef SENTRY_PLATFORM_WINDOWS
     time_t time = _mkgmtime(&tm);
+#elif defined(SENTRY_PLATFORM_PS)
+    time_t time = mktime(&tm); // TODO is this correct?
```
supervacuus, line 466: Is the question directed at me? `mktime()` assumes `tm` contains a local-time offset (whatever that is). It is equivalent to `timelocal()`. So, since that part of the code parses an ISO8601 string that always ends with `Z`, you will potentially introduce a round-trip error between it and `sentry__usec_time_to_iso8601()`. The function is currently only used in tests, by the way.

If you want to be sure, you must find some way to access the local time offset (but I am not sure what POSIX interfaces PS exposes, given that `timegm` isn't available).

Of course, if PS guarantees to only have local-time offsets of 0, at that level, then you're good too.

vaind: thx for the feedback. I've had to move the code to the downstream repo to make it work properly.

vendor/acutest.h
```diff
@@ -914,9 +914,18 @@ test_do_run__(const struct test__* test, int index)
...
 
 #ifdef __cplusplus
-    try {
+#ifdef __has_feature
+#if __has_feature(cxx_exceptions)
+    #define ACUTEST_HAS_EXCEPTIONS
+#endif
+#else
+    #define ACUTEST_HAS_EXCEPTIONS
+#endif
```
vaind, line 923: Adds support for `-fno-exceptions`

vendor/acutest.h
```diff
@@ -914,9 +914,18 @@ test_do_run__(const struct test__* test, int index)
     test_begin_test_line__(test);
 
 #ifdef __cplusplus
-    try {
+#ifdef __has_feature
+#if __has_feature(cxx_exceptions)
```
vaind, line 918:
```suggestion
// Support for `-fno-exceptions` 
#if __has_feature(cxx_exceptions)
```

vaind:
@supervacuus do you want to have another look? There've been some changes (that came from testing on devkit) since your last review. Feel free to merge if it's all good from your POV

supervacuus (approved):

Issue #1225, opened 2025-05-05
User Context is Case Sensitive


sdzhong:
You can set the user as per https://docs.sentry.io/platforms/native/enriching-events/identify-user/, which shows

`sentry_value_set_by_key(user, "email", sentry_value_new_string("jane.doe@example.com"));`

This is case sensitive--needs to be lowercase `user` and `email` in the above example.

However, the Sentry UI shows user contexts with capitalization, like User, Email, IP Address, and Username
![Image](https://github.com/user-attachments/assets/98993008-601d-49a7-a87d-95f6874c762d)

This is confusing to users.

Suggestions:
1. case should not be load bearing but
2. if it is, they should match, and also
3. if it is, this should be handled internally / just ToLower() all user, context and tag elements

kahest:
Hey @sdzhong not sure I fully understand - did you try setting a user property starting with an uppercase letter, e.g. `sentry_value_set_by_key(user, "Email", sentry_value_new_string("jane.doe@example.com"));` and it didn't work?

If so, it's a bug, but we might want to fix it in backend/relay, during normalization, depending on whether SDKs are expected to lowercase known keys.

If it works, then everything works as intended, because we have some logic on the Frontend that makes sure that certain known context properties are rendered "nice" (e.g. email->Email, ip_address->IP Address,...).

zsd4yr:
Hey Kahest! that's correct, we tried uppercase Username and Email first and the User.Display came out as the ip address instead

kahest:
@zsd4yr can you share a sentry.io link to such an event (either here or to karl.struggl@sentry.io)?

PR #1228, opened 2025-05-06
feat: add `sentry_value_new_user`

JoshuaMoelans:
Came up in discussion of https://github.com/getsentry/sentry-native/issues/1225 ; when creating a user object, it's not immediately obvious which fields are expected and how they should be cased. Hence we add a `sentry_value_new_user(id, username, email, ip_address)` function that constructs the user object with the correct keys (as defined in the [develop docs](https://develop.sentry.dev/sdk/data-model/event-payloads/user/)).

The function should be used as follows:
```c
sentry_value_t user = sentry_value_new_user(NULL, "John Doe", "john.doe@example.com", NULL);
sentry_value_set_by_key(user, "custom_key", sentry_value_new_string("custom_value"));
sentry_set_user(user);
```

### ToDo:
- [x] implement function
- [x] add tests
- [x] add to docs https://github.com/getsentry/sentry-docs/pull/13624


src/sentry_value.c
```diff
@@ -431,6 +431,34 @@ sentry__value_new_object_with_size(size_t size)
     }
 }
 
+sentry_value_t
+sentry_value_new_user(const int32_t *id, const char *username,
+    const char *email, const char *ip_address)
```
JoshuaMoelans, line 436: I'm not 💯 on passing in a pointer for the `id` here, but if we change the signature to be 
```c
sentry_value_new_user(int32_t id, const char *username, const char *email, const char *ip_address)
```
instead, the `id` can never be NULL, which would mean we enforce it to not be optional (while the [docs](https://docs.sentry.io/platforms/native/enriching-events/identify-user/) say each of them should be).

supervacuus: It is your lucky day: `id` is also a string.

src/sentry_value.c
```diff
@@ -431,6 +431,34 @@ sentry__value_new_object_with_size(size_t size)
     }
 }
 
+sentry_value_t
+sentry_value_new_user(const int32_t *id, const char *username,
```
supervacuus, line 435: Please also provide a `new_user_n` implementation and implement non-`_n` in terms of `_n'.

src/sentry_core.c
```diff
@@ -702,7 +702,7 @@ sentry_set_user(sentry_value_t user)
         }
         sentry__options_unlock();
     }
-
+    // TODO do we always want to overwrite the user even if it is null?
```
supervacuus, line 705: Yes, this is unsetting the user.

supervacuus:
The only thing left is updating the docs to eliminate manual key-setting.

PR #1233, opened 2025-05-09
fix: support musl on Linux

### LINKED ISSUE #1232: support linux musl
Relates to:
* https://github.com/getsentry/sentry-dotnet/issues/4127

jpnurmi:
This PR includes:
- `libunwind` -based unwinder by @supervacuus ([fix/use_libunwind_as_macos_unwinder](https://github.com/getsentry/sentry-native/tree/fix/use_libunwind_as_macos_unwinder))
- build system changes to select `SENTRY_WITH_LIBUNWIND` for musl
- crashpad update: https://github.com/getsentry/crashpad/pull/125
- ci changes to test with musl and `libunwind`

Closes #1232


.github/workflows/ci.yml
```diff
@@ -149,16 +162,25 @@ jobs:
...
+          apk update
+          apk add bash build-base cargo cmake curl curl-dev git icu linux-headers mitmproxy perl python3-dev sudo ${{ matrix.APK_PACKAGES }}
+          curl -sSL --retry 5 https://dot.net/v1/dotnet-install.sh | bash -eo pipefail /dev/stdin --channel 8.0 --install-dir /usr/share/dotnet
+          ln -s /usr/share/dotnet/dotnet /usr/local/bin/dotnet
+
       - uses: actions/checkout@v4
         with:
           submodules: recursive
       - uses: actions/setup-python@v5
+        if: ${{ !matrix.container }}
```
jpnurmi, line 177: `setup-python` doesn't work in arbitrary containers but basically only supports the runner images: https://github.com/actions/python-versions

.github/workflows/ci.yml
```diff
@@ -252,6 +274,16 @@ jobs:
           cat /etc/hosts
         shell: bash
 
+      - name: Prepare env for Alpine Linux
+        if: ${{ contains(matrix.container, 'alpine') }}
+        run: |
+          apk add moreutils
+          # comment out "::1 localhost ..." to avoid conflicts with proxy tests
+          sed '/^::1/ s/^/#/' /etc/hosts | sponge /etc/hosts
+          python3 -m venv .venv --system-site-packages
```
jpnurmi, line 283: is passing `--system-site-packages` ok? installing `mitmproxy` with apk takes a few seconds, whereas building it from scratch for venv took ~4 minutes.

.github/workflows/ci.yml
```diff
@@ -149,16 +162,25 @@ jobs:
       VS_GENERATOR_TOOLSET: ${{ matrix.VS_GENERATOR_TOOLSET }}
 
     steps:
+      - name: Installing Alpine Linux Dependencies
+        if: ${{ contains(matrix.container, 'alpine') }}
+        run: |
+          apk update
+          apk add bash build-base cargo cmake curl curl-dev git icu linux-headers mitmproxy perl python3-dev sudo ${{ matrix.APK_PACKAGES }}
+          curl -sSL --retry 5 https://dot.net/v1/dotnet-install.sh | bash -eo pipefail /dev/stdin --channel 8.0 --install-dir /usr/share/dotnet
+          ln -s /usr/share/dotnet/dotnet /usr/local/bin/dotnet
```
jpnurmi, line 171: `dotnet` for `test_dotnet_signals.py`

.github/workflows/ci.yml
```diff
@@ -149,16 +162,25 @@ jobs:
       VS_GENERATOR_TOOLSET: ${{ matrix.VS_GENERATOR_TOOLSET }}
 
     steps:
+      - name: Installing Alpine Linux Dependencies
+        if: ${{ contains(matrix.container, 'alpine') }}
+        run: |
+          apk update
+          apk add bash build-base cargo cmake curl curl-dev git icu linux-headers mitmproxy perl python3-dev sudo ${{ matrix.APK_PACKAGES }}
```
jpnurmi, line 169: Note: git needs to be installed before actions/checkout

CMakeLists.txt
```diff
@@ -504,6 +515,15 @@ if(SENTRY_WITH_LIBUNWINDSTACK)
 	endif()
 endif()
 
+if(SENTRY_WITH_LIBUNWIND)
+	if(LINUX)
+		find_path(LIBUNWIND_INCLUDE_DIR libunwind.h PATH_SUFFIXES libunwind REQUIRED)
+		find_library(LIBUNWIND_LIBRARY unwind REQUIRED)
```
jpnurmi, line 521: Note: manual find instead of pkg-config because `llvm-libunwind` doesn't ship a .pc file

.github/workflows/ci.yml
supervacuus: It is fine.

.github/workflows/ci.yml
supervacuus: Ugh, the container integration of GHA workflows is a little weird, to be honest. 

Two things: 

`setup-python` is only necessary because we haven't switched to using `uv` for the test setup yet. So this is not a long-term special case. 

Suppose the container setup (essentially, now parallel conditional build-steps in the workflow yaml) gets out of hand at some point. In that case, we can also evaluate switching to Dockerfiles written on top of the Alpine image, manually creating images that already contain what we want, and executing the pytest there.

.github/workflows/ci.yml
supervacuus: ditto: https://github.com/getsentry/sentry-native/pull/1233#discussion_r2086210671

.github/workflows/ci.yml
```diff
@@ -86,6 +86,18 @@ jobs:
...
+          - name: Linux (GCC + musl + libunwind)
+            os: ubuntu-latest
+            container: alpine:3.21
+            APK_PACKAGES: libunwind-dev libunwind-static
+            CC: gcc
+            CXX: g++
+          - name: Linux (clang + musl + llvm-libunwind)
+            os: ubuntu-latest
+            container: alpine:3.21
+            APK_PACKAGES: clang llvm-libunwind-dev llvm-libunwind-static
```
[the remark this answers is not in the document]
jpnurmi: I was still using `libunwind-dev` during the initial experiments when I took the screenshots and only later realized there's `llvm-libunwind-dev` available too. I see the same behavior on x86_64 where the stacktrace is empty during crash handling. Stacktraces unwound in unit tests look fine, though. :thinking:

bruno-garcia:
PR title has:

> * crashpad update: https://github.com/getsentry/sentry-native/pull/1233

Did you mean to link to: 
* https://github.com/getsentry/crashpad/pull/124

jpnurmi:
> PR title has:
> 
> > * crashpad update: [fix: support musl on Linux #1233](https://github.com/getsentry/sentry-native/pull/1233)
> 
> Did you mean to link to:
> 
> * [feat: add CrashpadClient::Add/RemoveAttachment() for Windows and Linux crashpad#124](https://github.com/getsentry/crashpad/pull/124)

Oops, thanks! It was meant to link to:
- https://github.com/getsentry/crashpad/pull/125

.github/workflows/ci.yml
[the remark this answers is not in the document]
jpnurmi: Passing the signal handler context works in the old `libunwind` because `unw_context_t` is an alias to `ucontext_t`. Apparently, this is not the case with `llvm-libunwind`. Shall we drop support for it then? `libunwind-static` should be fine for sentry-dotnet.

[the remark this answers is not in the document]
supervacuus: Yes, that is expected. `llvm-libunwind` is a stripped-down version of `libunwind`, primarily used as an exception unwinder. We don't need to support it.

I would even go one step further and explicitly use `unw_init_local2()` with the `UNW_INIT_SIGNAL_FRAME` flag for the case where you init with the `ucontext`. This interface is also not available in `llvm-libunwind`.

src/unwinder/sentry_unwinder_libunwind.c
```diff
@@ -0,0 +1,46 @@
...
+sentry__unwind_stack_libunwind(
+    void *addr, const sentry_ucontext_t *uctx, void **ptrs, size_t max_frames)
+{
+    if (addr) {
+        return 0;
+    }
+
+    unw_cursor_t cursor;
+    if (uctx) {
+        int ret = unw_init_local(&cursor, (unw_context_t *)uctx->user_context);
```
supervacuus, line 16: If we drop `llvm-libunwind`, we can use `unw_init_local2` here to be even more explicit (and support a broader range of `libunwind` versions and target platforms):
```suggestion
        int ret = unw_init_local2(&cursor, (unw_context_t *)uctx->user_context, UNW_INIT_SIGNAL_FRAME);
```

CMakeLists.txt
```diff
@@ -504,6 +515,15 @@ if(SENTRY_WITH_LIBUNWINDSTACK)
...
+		target_include_directories(sentry PRIVATE ${LIBUNWIND_INCLUDE_DIR})
+		target_link_libraries(sentry PRIVATE ${LIBUNWIND_LIBRARY})
```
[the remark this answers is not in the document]
supervacuus: Thanks, that's a very sensible default approach.

.github/workflows/ci.yml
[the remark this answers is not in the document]
supervacuus: Btw, even if we drop `llvm-libunwind`, I would keep the clang configuration (and just use `libunwind` in that case). Having checks for both compilers on a specific target-image is never a bad idea.

supervacuus:
If only commit suggestions could have a clang-format pass 😅

jpnurmi:
No worries :)

supervacuus (approved):

PR #1234, opened 2025-05-12
fix: introduce malloc/MAKE rv checks if missing

supervacuus:
https://github.com/getsentry/sentry-native/pull/1224 raises the issue that there are still some unchecked returns from `SENTRY_MAKE` and `sentry_malloc`. This PR (now) fixes this first layer, but we should also fix the internal uses of all functions affected by early returns. For instance, here:

https://github.com/getsentry/sentry-native/blob/6a6e42fcd88724c329e1154dda4d767770eafabe/src/path/sentry_path.c#L18-L23

`path_iter` could be `NULL`. Evaluating our coverage and keeping it would require a bit more work.


src/symbolizer/sentry_symbolizer_unix.c
```diff
@@ -155,6 +157,9 @@ dladdr(void *s, Dl_info *i)
             i->dli_fbase = tb; /* Includes XCOFF header */
             /* library filename + ( + member + ) + NUL */
             char *libname = (char *)sentry_malloc(AIX_PRINTED_LIB_LEN);
+            if (!libname) {
+                return 0;
+            }
```
JoshuaMoelans, line 162: (from the function's docstring above) why is a null-`libname` a success?
``` c
 /* Returns 1 on failure and 0 on success. "s" is the address of the symbol,
 * and "i" points to a Dl_info structure to fill. Note that i.dli_fname is
 * not const, and should be freed.
 */
```

tests/unit/test_sync.c
```diff
@@ -33,7 +33,7 @@ SENTRY_TEST(background_worker)
 {
     for (size_t i = 0; i < 100; i++) {
         sentry_bgworker_t *bgw = sentry__bgworker_new(NULL, NULL);
-        TEST_CHECK(!!bgw);
+        TEST_ASSERT(!!bgw);
```
JoshuaMoelans, line 36: q: Why do we only switch to `TEST_ASSERT` here? (Considering the [docstring for it](https://github.com/getsentry/sentry-native/blob/ba3d4167ff408f11245c986802ccf5ba6c073f31/vendor/acutest.h#L86-L110), aborting when no backgroundworker was created makes sense, but I just wonder if there might be more places where this change could be useful).

tests/unit/test_tracing.c
```diff
@@ -1201,14 +1213,17 @@ SENTRY_TEST(txn_name_n)
...
     sentry_transaction_set_name_n(txn, txn_name, sizeof(txn_name));
 
     sentry_value_t txn_name_value
         = sentry_value_get_by_key(txn->inner, "transaction");
     TEST_CHECK(
         sentry_value_get_type(txn_name_value) == SENTRY_VALUE_TYPE_STRING);
-    TEST_CHECK_STRING_EQUAL(sentry_value_as_string(txn_name_value), "the_txn");
+    const char *txn_name_str = sentry_value_as_string(txn_name_value);
+    TEST_ASSERT(!!txn_name_str);
+    TEST_CHECK_STRING_EQUAL(txn_name_str, "the_txn");
```
JoshuaMoelans, line 1226: There are a lot of new `TEST_ASSERT(!!var);` entries (thanks for that!), but there are still some situations similar to the snippet below which we might want to split up into a similar `get_by_key+value_as_string->test_assert->test_check_string_equal` flow.

```c
TEST_CHECK_STRING_EQUAL(sentry_value_as_string(
  sentry_value_get_by_key(tx->inner, "trace_id")), txn_trace_id);
```

JoshuaMoelans (changes requested):
This looks good, I just have one remark on a success/fail return value & two questions about applying the test changes more broadly.

src/symbolizer/sentry_symbolizer_unix.c
supervacuus: Tbh, I didn't look at the inline docs. I examined the surrounding code, where every early exit returns `0`, and the only success case I could find returns `1`. Maybe I misread?

tests/unit/test_sync.c
supervacuus: Switching from `CHECK` to `ASSERT` generally means that continuing and aggregating errors could lead to a crash in the test itself, which is almost always much more complicated to pinpoint than a failed assertion. 

* `TEST_ASSERT` and similar macros should be used whenever further execution could lead to trouble. 
* `TEST_CHECK` should be used whenever a result is wrong, but the remaining test can still execute cleanly. 

The latter is preferred because this way we get to fix all errors in one iteration. However, the former is a necessity for running tests safely.

tests/unit/test_tracing.c
supervacuus: There are other places where a separation like this one would make sense. However, "fixing" the tests in this way was not my primary concern. 

The added `TEST_ASSERT` locations were identified by running the entire test suite against a `sentry_malloc()` that always returns `NULL`. 

The case you mentioned (and similar ones) is entirely safe:

* `sentry_value_get_by_key()` could look up a key that doesn't exist, in which case it returns a null value
* `sentry_value_as_string()` handles null values safely and wouldn't crash
* `sentry_value_as_string()` returns an empty string `""` when converting a null value and not a `NULL` pointer!
* `TEST_CHECK_STRING_EQUAL` would pass an empty string to `strcmp()` thus leading to safe behavior

src/symbolizer/sentry_symbolizer_unix.c
JoshuaMoelans: Seems like the docstring is just wrong, and it returns `1` on success and `0` on early exit/failure, so your addition is functionally correct 👌

PR #1235, opened 2025-05-12
chore: ignore CI/benchmark runs for all markdown file changes

JoshuaMoelans:
#skip-changelog


JoshuaMoelans:
Seems like `paths-ignore` is currently borked on `pull_request` events  (https://github.com/actions/runner/issues/2324) ; as soon as any non-matching file is part of the PR, the workflow will just run even if the next commit has only changes to files in the paths-ignore list.

Issue #1214, opened 2025-04-25
Prevent sending running Transactions/Spans if app has been put to sleep


Angelodaniel:
Currently there is no way on the Native side to stop/prevent/drop transactions reactively if the app has been put on sleep mode. 

Why I need this is that putting the app to sleep while capturing transactions greatly impacts the Span durations and thus also my durations of the app. 

It would be great if I can exclude those/ not capture them in the first place as these spans are useless. Something similar to [Android](https://docs.sentry.io/platforms/android/configuration/filtering/#using-before-send-transaction)/[iOS](https://docs.sentry.io/platforms/apple/guides/ios/configuration/filtering/#filtering-spans) `beforeSendTransaction` would be a way to implement this.

PR #1236, opened 2025-05-13
feat: Add `before_send_transaction`

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/1214 (and related to https://github.com/getsentry/sentry-native/issues/864#issuecomment-1649461251)

ToDo
- [x] add option
- [x] call inside transaction event preparation
- [x] add to example
- [x] answer todo questions in-code
- [x] add tests
- [x] changelog
- [x] docs https://github.com/getsentry/sentry-docs/pull/13715

## questions (resolved)
- Why do we have invoke_before_send passed into prepare_event ([see code](https://github.com/getsentry/sentry-native/blob/e511a8ed0be28e4fd9826854032df370ac0acd77/src/sentry_core.c#L554-L555)), and why is it true for sentry_core occurrences ([1](https://github.com/getsentry/sentry-native/blob/e511a8ed0be28e4fd9826854032df370ac0acd77/src/sentry_core.c#L466), [2](https://github.com/getsentry/sentry-native/blob/e511a8ed0be28e4fd9826854032df370ac0acd77/src/sentry_core.cL1407-L1408)), but passed as !options->on_crash_func for [breakpad](https://github.com/getsentry/sentry-native/blob/e511a8ed0be28e4fd9826854032df370ac0acd77/src/backends/sentry_backend_breakpad.cpp#L145-L146)?

  > For crashes, we only `before_send` if there is no `on_crash` hook.
- Where do we callback? Before or after applying scope?
  
  > As late as possible, to give the user maximal context.
- (void* hint) parameter (which is unused for the other callbacks, but exists afaict), [see here](https://github.com/getsentry/sentry-native/pull/1236/files#diff-9c77242571e813d699b88663fdc7d7ad7bb5b481a3c1af8d34e1a9e242cbf076R1765-R1766)
  
  > Not relevant for `before_send_transaction`.
- Do we want to add `before_transaction_data` to modify transactions before sending? [see here](https://github.com/getsentry/sentry-native/pull/1236/files#diff-9c77242571e813d699b88663fdc7d7ad7bb5b481a3c1af8d34e1a9e242cbf076R1767)
  
  > Yes.

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -862,7 +862,7 @@ SENTRY_API void sentry_options_set_send_default_pii(
  * sent.
  */
 typedef sentry_value_t (*sentry_event_function_t)(
-    sentry_value_t event, void *hint, void *closure);
+    sentry_value_t event, void *hint, void *user_data);
 
 /**
  * Sets the `before_send` callback.
@@ -870,7 +870,7 @@ typedef sentry_value_t (*sentry_event_function_t)(
  * See the `sentry_event_function_t` typedef above for more information.
  */
 SENTRY_API void sentry_options_set_before_send(
-    sentry_options_t *opts, sentry_event_function_t func, void *closure);
+    sentry_options_t *opts, sentry_event_function_t func, void *user_data);
 
 /**
  * Type of the `on_crash` callback.
@@ -923,7 +923,7 @@ SENTRY_API void sentry_options_set_before_send(
  * sent.
  */
 typedef sentry_value_t (*sentry_crash_function_t)(
-    const sentry_ucontext_t *uctx, sentry_value_t event, void *closure);
+    const sentry_ucontext_t *uctx, sentry_value_t event, void *user_data);
 
 /**
  * Sets the `on_crash` callback.
@@ -1763,7 +1763,7 @@ typedef struct sentry_transaction_s sentry_transaction_t;
  * return a `sentry_value_new_null()` instead.
  */
 typedef sentry_value_t (*sentry_transaction_function_t)(
-    sentry_value_t transaction, void *data);
+    sentry_value_t transaction, void *user_data);
 
 /**
  * Sets the `before_transaction` callback.
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -870,7 +870,7 @@ typedef sentry_value_t (*sentry_event_function_t)(
  * See the `sentry_event_function_t` typedef above for more information.
  */
 SENTRY_API void sentry_options_set_before_send(
-    sentry_options_t *opts, sentry_event_function_t func, void *data);
+    sentry_options_t *opts, sentry_event_function_t func, void *closure);
 
 /**
  * Type of the `on_crash` callback.
@@ -1761,19 +1761,15 @@ typedef struct sentry_transaction_s sentry_transaction_t;
  * that same transaction. In case the transaction should be discarded, the
  * callback needs to call `sentry_value_decref` on the provided transaction, and
  * return a `sentry_value_new_null()` instead.
- * TODO ensure this docstring is correct & complete
- * TODO do we want a (void* hint) parameter too (see sentry_event_function_t)?
- *  this is always passed as NULL to the on_crash/on_send callbacks :thinking:
- *  TODO before_send_data (like for on_crash/on_send)?
  */
 typedef sentry_value_t (*sentry_transaction_function_t)(
-    sentry_value_t transaction);
+    sentry_value_t transaction, void *data);
 
 /**
  * Sets the `before_transaction` callback.
  */
 SENTRY_EXPERIMENTAL_API void sentry_options_set_before_transaction(
-    sentry_options_t *opts, sentry_transaction_function_t func);
+    sentry_options_t *opts, sentry_transaction_function_t func, void *data);
 
 /**
  * A sentry Span.
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -1755,6 +1755,27 @@ struct sentry_transaction_s;
 typedef struct sentry_transaction_s sentry_transaction_t;
 
 /**
+ * Type of the `before_transaction` callback.
+ *
+ * The callback takes ownership of the `transaction`, and should usually return
+ * that same transaction. In case the transaction should be discarded, the
+ * callback needs to call `sentry_value_decref` on the provided transaction, and
+ * return a `sentry_value_new_null()` instead.
+ * TODO ensure this docstring is correct & complete
+ * TODO do we want a (void* hint) parameter too (see sentry_event_function_t)?
+ *  this is always passed as NULL to the on_crash/on_send callbacks :thinking:
+ *  TODO before_send_data (like for on_crash/on_send)?
+ */
+typedef sentry_value_t (*sentry_transaction_function_t)(
+    sentry_value_t transaction);
+
+/**
+ * Sets the `before_transaction` callback.
+ */
+SENTRY_EXPERIMENTAL_API void sentry_options_set_before_transaction(
+    sentry_options_t *opts, sentry_transaction_function_t func);
+
+/**
  * A sentry Span.
  *
  * See https://develop.sentry.dev/sdk/event-payloads/span/
--- a/src/sentry_core.c
+++ b/src/sentry_core.c
@@ -634,10 +634,9 @@ sentry__prepare_transaction(const sentry_options_t *options,
     }
 
     if (options->before_transaction_func) {
-        // TODO why no invoke_before_send in func arguments (vs prepare_event)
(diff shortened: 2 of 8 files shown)


src/sentry_core.c
```diff
@@ -633,6 +633,18 @@ sentry__prepare_transaction(const sentry_options_t *options,
         sentry__scope_apply_to_event(scope, options, transaction, mode);
     }
 
+    if (options->before_transaction_func) {
+        // TODO why no invoke_before_send in func arguments (vs prepare_event)
+        //      (also why is it always `true` for prepare_event)
+        SENTRY_DEBUG("invoking `before_transaction` hook");
+        transaction = options->before_transaction_func(transaction);
```
JoshuaMoelans, line 640: I'm not 💯 on calling the `before_transaction_func` here, because at this point the transaction is already an event (and has way more data than a `sentry_transaction_t` has). 

Maybe it makes more sense to apply the callback in the `sentry_transaction_finish` (e.g. [here](https://github.com/getsentry/sentry-native/blob/e511a8ed0be28e4fd9826854032df370ac0acd77/src/sentry_core.c#L1040-L1118)) where users only have access to the following keys: "trace_id", "span_id", "parent_span_id", "op", "description", "status", "data"

(versus a full json like below:)
<details>

```json
{
  "event_id" : "3236d6f4-75bd-44d7-4b77-4f2bbc700137",
  "platform" : "native",
  "transaction" : "little.teapot",
  "start_timestamp" : "2025-05-13T13:58:26.352429Z",
  "spans" : [ {
    "op" : "littlest.teapot",
    "span_id" : "98ed455fca274524",
    "status" : "ok",
    "trace_id" : "ce3e2f93331243724a8cbc02555d2cae",
    "parent_span_id" : "4ebf9dbce31d408d",
    "description" : null,
    "start_timestamp" : "2025-05-13T13:58:26.352437Z",
    "timestamp" : "2025-05-13T13:58:26.352439Z"
  }, {
    "op" : "littler.teapot",
    "span_id" : "4ebf9dbce31d408d",
    "status" : "ok",
    "trace_id" : "ce3e2f93331243724a8cbc02555d2cae",
    "parent_span_id" : "1dafa81af1e34160",
    "description" : null,
    "start_timestamp" : "2025-05-13T13:58:26.352434Z",
    "data" : {
      "span_data_says" : "hi!"
    },
    "timestamp" : "2025-05-13T13:58:26.352442Z"
  } ],
  "type" : "transaction",
  "timestamp" : "2025-05-13T13:58:26.352444Z",
  "level" : "info",
  "contexts" : {
    "trace" : {
      "trace_id" : "ce3e2f93331243724a8cbc02555d2cae",
      "span_id" : "1dafa81af1e34160",
      "op" : "Short and stout here is my handle and here is my spout",
      "status" : "ok",
      "data" : {
        "url" : "https://example.com"
      }
    },
    "os" : {
      "name" : "macOS",
      "version" : "15.4.1",
      "build" : "24E263",
      "kernel_version" : "24.4.0"
    },
    "runtime" : {
      "type" : "runtime",
      "name" : "testing-runtime"
    }
  },
  "release" : "test-example-release",
  "environment" : "development",
  "user" : {
    "id" : "42",
    "username" : "some_name"
  },
  "sdk" : {
    "name" : "sentry.native",
    "version" : "0.8.4",
    "packages" : [ {
      "name" : "github:getsentry/sentry-native",
      "version" : "0.8.4"
    } ]
  },
  "tags" : {
    "expected-tag" : "some value"
  },
  "extra" : {
    "extra stuff" : "some value",
    "…unicode key…" : "őá…–🤮🚀¿ 한글 테스트"
  },
  "breadcrumbs" : [ {
    "timestamp" : "2025-05-13T13:58:26.350755Z",
    "message" : "default level is info"
  }, {
    "timestamp" : "2025-05-13T13:58:26.352420Z",
    "type" : "http",
    "message" : "debug crumb",
    "category" : "example!",
    "level" : "debug",
    "data" : {
      "url" : "https://example.com/api/1.0/users",
      "method" : "GET",
      "status_code" : 200,
      "reason" : "OK"
    }
  }, {
    "timestamp" : "2025-05-13T13:58:26.352423Z",
    "message" : "lf\ncrlf\r\nlf\n...",
    "category" : "something else"
  } ]
}
```

</details>

supervacuus: > because at this point the transaction is already +- an event (and has way more data than a sentry_transaction_t has, since the scope was applied to the event).

Why is this a bad thing? The more context a user gets, the closer they are to sending, the better. No?

JoshuaMoelans: It's not bad, just wanted to make sure this is the intended use & that we have no reason to restrict access to this kind of event data. Since the `before_send_func` callback only gets called for non-transaction events, I agree with you that it does make sense to give the user as much context as possible.

supervacuus:
> Why do we have invoke_before_send passed into prepare_event ([see code](https://github.com/getsentry/sentry-native/blob/e511a8ed0be28e4fd9826854032df370ac0acd77/src/sentry_core.c#L554-L555)), and why is it true for sentry_core occurrences ([1](https://github.com/getsentry/sentry-native/blob/e511a8ed0be28e4fd9826854032df370ac0acd77/src/sentry_core.c#L466), [2](https://github.com/getsentry/sentry-native/blob/e511a8ed0be28e4fd9826854032df370ac0acd77/src/sentry_core.cL1407-L1408)), but passed as !options->on_crash_func for [breakpad](https://github.com/getsentry/sentry-native/blob/e511a8ed0be28e4fd9826854032df370ac0acd77/src/backends/sentry_backend_breakpad.cpp#L145-L146)?

Because this function is used for regular event capture, and during the capture of crashes from the handler. We want to invoke the `before_send` hook with regular events, because that is what we use it for. For crashes, we only run the `before_send` hook if no `on_crash` hook has been defined. This is documented in the respective hook docs.

> Where do we callback? Before or after applying scope?

As late as possible: after applying the scope. What would your argument be against that?

> (void* hint) parameter (which is unused for the other callbacks, but exists afaict), [see here](https://github.com/getsentry/sentry-native/pull/1236/files#diff-9c77242571e813d699b88663fdc7d7ad7bb5b481a3c1af8d34e1a9e242cbf076R1765-R1766)

The `hint` parameter isn't really applicable to the Native SDK, but we could pass in the ucontext or exception_ptrs if it is used in the crash context as we do for the `on_crash` hook (but in an untyped fashion). I don't think it is relevant to the `before_send_transaction`.

> Do we want to add before_transaction_data to modify transactions before sending? [see here](https://github.com/getsentry/sentry-native/pull/1236/files#diff-9c77242571e813d699b88663fdc7d7ad7bb5b481a3c1af8d34e1a9e242cbf076R1767)

I think so, but similar to the concrete place of calling the callback I would compare with other SDKs.

src/sentry_core.c
supervacuus: I wouldn't restrict filter options without a "designed" reason. This is the only place where users can make changes and should have as much access as possible. However, other SDKs should be cross-checked.

include/sentry.h
```diff
@@ -870,7 +870,7 @@ typedef sentry_value_t (*sentry_event_function_t)(
  * See the `sentry_event_function_t` typedef above for more information.
  */
 SENTRY_API void sentry_options_set_before_send(
-    sentry_options_t *opts, sentry_event_function_t func, void *data);
+    sentry_options_t *opts, sentry_event_function_t func, void *closure);
```
supervacuus, line 873: I don't follow: why the change to `closure` from `data`, but below you use `data`? Similarly, in the example callback definition, you name the parameter `closure`.

I'm happy with any name, but we should use the same name for the same thing. So, if we prefer to use `data` for a parameter users can pass in at callback registration to later access it from within the callback, then let's use that name everywhere. When I say "everywhere", I mean across interfaces that provide a parameter for a similar purpose, as well as across internal names and callback definitions inside example or test code.

I don't think `closure` is a particularly good name, even though closures are used in other languages for that purpose. We don't _close_ over any lexical scope. I would probably call it `user_data`, but think `data` is fine in that context.

By the way, changing parameters is thankfully not a breaking change in C (neither for ABI nor API).

supervacuus (changes requested):
Looks great. Just wonder about the parameter naming and whether we should use the PR to make its use consistent.

include/sentry.h
JoshuaMoelans: Thanks for bringing this up. I was also confused where the term 'closure' came from, but since it had been in use forever, I just changed my `data` parameter to `closure` (but forgot an occurrence, which luckily started this conversation)

I'll rename them all to `data`, since in the options we store them as `before_send_data`/`on_crash_data`/`before_transaction_data` anyway.

supervacuus (approved):

PR #1242, opened 2025-05-14
chore: bump codecov-action to 5.4.2

supervacuus:
This change is primarily motivated by the failing codecov invocation in the Linux ARM64 llvm-cov CI config (it works in all other configs, besides `clang-cl`, but this is another issue).

This seemingly was fixed in major 5 of the codecov-action: https://github.com/codecov/codecov-action/issues/1554

codecov works for the first time since you introduced the build config: https://github.com/getsentry/sentry-native/actions/runs/15024781906/job/42222733350?pr=1242#step:23:62

#skip-changelog


JoshuaMoelans (approved):
lgtm!

PR #1244, opened 2025-05-14
test: define individual unit tests for CTEST

vaind:
During development, I'm used to running individual unit tests via Test Explorer in VS Code.
However, this has not been possible for sentry-native because of its use of add_test().

This PR changes the behavior - instead of adding a single `sentry_test_unit` to CTEST, we add individual unit tests by reading the existing `tests.inc` file.

I've also noticed that the `sentry_fuzz_json` and `sentry_example` have been added to CTEST but they're actually not test executables, so removing those.

Note: this does not influence anyone launching executable directly (what we also do in CI in this repo), only `ctest` command users , such as IDE integrations. Maybe worth adding a changelog item anyway if some SDK consumer launched ctest with the target filter `sentry_test_unit`. 

Before 
before
and after:
 after

### DIFF
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -107,7 +107,7 @@ if(SENTRY_CTEST_INDIVIDUAL)
 		endif()
 	endforeach()
 else()
-	add_test(NAME sentry/all COMMAND sentry_test_unit)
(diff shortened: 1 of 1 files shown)


supervacuus:
I have no clue why the test setup was so weirdly defined in CMake. I do not interface with CTest often, but this is undoubtedly an improvement for anyone using integrated test discovery.

Issue #1226, opened 2025-05-06
Local scopes


jpnurmi:
We would like to add support for local scopes for the needs of downstream SDKs (sentry-unreal).

API proposal for capturing events with local scopes, inspired by other Sentry SDKs:

```c
void sentry_capture_event_with_scope(sentry_value_t event, sentry_scope_callback_t callback, void *userdata)
```

- Clones the current scope and pushes it on a (thread-local?) stack
- Calls the scope configuration callback
  `typedef void (*sentry_scope_callback_t)(sentry_scope_t *scope, void *userdata)`
- Flushes the local scope
- Captures the event
- Discards the local scope

Ideally, [GenericPlatformSentryScope](https://github.com/getsentry/sentry-unreal/blob/main/plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryScope.cpp) in sentry-unreal would be a thin wrapper around an opaque `sentry_scope_t*`. To achieve this, sentry-native could expose getters and setters for all the data stored in the scope.

```c
typedef struct sentry_scope_s {
    char *transaction;
    sentry_value_t fingerprint;
    sentry_value_t user;
    sentry_value_t tags;
    sentry_value_t extra;
    sentry_value_t contexts;
    sentry_value_t breadcrumbs;
    sentry_level_t level;
    sentry_value_t client_sdk;
    sentry_transaction_t *transaction_object;
    sentry_span_t *span;
} sentry_scope_t;
```

- transaction
  - `const char *sentry_scope_get_transaction(sentry_scope_t *scope)`
  - `void sentry_scope_set_transaction(sentry_scope_t *scope, const char *transaction)` + `_n`
- fingerprint
  - `sentry_value_t sentry_scope_get_fingerprint(sentry_scope_t *scope)`
  - `void sentry_scope_set_fingerprint(sentry_scope_t *scope, const char *fingerprint, ...)` + `_n`
  - `void sentry_scope_remove_fingerprint(sentry_scope_t *scope)`
- user
  - `sentry_value_t sentry_scope_get_user(sentry_scope_t *scope)`
  - `void sentry_scope_set_user(sentry_scope_t *scope, sentry_value_t user)`
  - `void sentry_scope_remove_user(sentry_scope_t *scope)`
- tags
  - `sentry_value_t sentry_scope_get_tags(sentry_scope_t *scope)`
  - `const char *sentry_scope_get_tag(sentry_scope_t *scope, const char *key)`
[...]

PR #1248, opened 2025-05-20
WIP: Local scopes

jpnurmi:
WIP: #1226


jpnurmi:
@supervacuus This is ready for a first pass. Things are looking promising for sentry-unreal:
- https://github.com/getsentry/sentry-unreal/pull/928

src/sentry_scope.c
```diff
@@ -139,6 +152,29 @@ sentry__scope_flush_unlock(void)
     }
 }
 
+sentry_scope_t *
+sentry_local_scope_new(void)
+{
+    sentry_scope_t *scope = sentry_malloc(sizeof(sentry_scope_t));
```
supervacuus: Please use `SENTRY_MAKE` for type-sized constructions.

include/sentry.h
```diff
@@ -1502,6 +1502,26 @@ SENTRY_API void sentry_user_consent_reset(void);
...
+ *
+ * See https://develop.sentry.dev/sdk/telemetry/scopes/
+ */
+struct sentry_scope_s;
+typedef struct sentry_scope_s sentry_scope_t;
+
+/**
+ * Creates a local scope.
+ *
+ * The return value must be freed with `sentry_scope_free`.
```
supervacuus: So, this is surprising. I would have guessed, given the discussion on the issue, that the choice now was made in a similar vein to what existed in `sentry-unreal`, i.e., limit the "local" scope to a particular capture. In that case, I would have expected `sentry_capture_event_with_scope()` to take ownership, essentially binding the scope's lifetime to the event being captured.

If we leave the local scope ownership in the hands of the users, that would imply reuse, which can totally make sense, but given the drastic move towards a simpler solution, also reflected in the choice of exposed setters, I wonder if scope-instance reuse is a more significant driver than the convenience that ownership transfer to capture facilitates.

I can imagine both making sense; however, keeping a local scope around without access to its entirety may be less valuable. Transferring ownership also replicates the current capture behavior regarding events, although it is unclear whether those two should share lifetime bounds.

In any case, that is a decision that we need to make nowish, but there could be a sensible upgrade path. For instance, if we start with default ownership transfer now, we can later, if the need arises on the user side, make ownership a member of the "local" scope and create two constructors for local scopes. Capture would then decide based on that whether it frees the scope or requires the user to do so.

src/sentry_scope.c
```diff
@@ -265,6 +301,89 @@ sentry__scope_get_span_or_transaction(void)
...
+        idx_b++;
+    }
+
+    // merge the remaining breadcrumbs in timestamp order
+    while (idx_a < len_a && idx_b < len_b) {
+        sentry_value_t item_a = sentry_value_get_by_index(list_a, idx_a);
+        sentry_value_t item_b = sentry_value_get_by_index(list_b, idx_b);
+
+        if (cmp_breadcrumb(item_a, item_b) <= 0) {
+            sentry_value_append(
```
supervacuus: Given that we create the list above, there is only one failure case for the append functions: we have a memory allocation issue. In that case, we should exit by returning a null value.

src/sentry_scope.c
```diff
@@ -265,6 +301,89 @@ sentry__scope_get_span_or_transaction(void)
...
+        || sentry_value_is_null(timestamp_b)) {
+        return 0;
+    }
+
+    return strcmp(sentry_value_as_string(timestamp_a),
+        sentry_value_as_string(timestamp_b));
+}
+
+static sentry_value_t
+merge_breadcrumbs(sentry_value_t list_a, sentry_value_t list_b, size_t max)
```
supervacuus, line 337: This function's control flow is not easy on the eyes, but it makes sense not to split it up (neither splitting across skip + merge, nor extracting the repeating iteration structure).

I suppose the "cheap" solution, where you concatenate and sort first and skip later, was not performant enough and was not much easier to read either (including implementation questions related to a list sorter)?

src/sentry_scope.c
```diff
@@ -265,6 +301,89 @@ sentry__scope_get_span_or_transaction(void)
...
 #endif
 
+static int
+cmp_breadcrumb(sentry_value_t a, sentry_value_t b)
+{
+    sentry_value_t timestamp_a = sentry_value_get_by_key(a, "timestamp");
+    sentry_value_t timestamp_b = sentry_value_get_by_key(b, "timestamp");
+    if (sentry_value_is_null(timestamp_a)
+        || sentry_value_is_null(timestamp_b)) {
+        return 0;
```
supervacuus: Is that the right choice here? This would mean that if either of the timestamps doesn't exist (which would be an inconsistent state), we continue to iterate on list a.

If we don't want to fail in this case (inconsistent state), I prefer to skip the item without a timestamp. That would mean to return `0` if `timestamp_a` were a null value and `1` if `timestamp_b` were a null value. If both are, then this would be rectified during the comparison of the next iteration.

tests/unit/test_scope.c
```diff
@@ -0,0 +1,604 @@
+#include "sentry.h"
+#include "sentry_scope.h"
+#include "sentry_testsupport.h"
+#include "sentry_utils.h"
+
+SENTRY_TEST(scope_contexts)
```
supervacuus, line 6: Thanks for all of these :heart:

src/sentry_scope.c
jpnurmi: I started with sort, went down the rabbit hole of `qsort_s` on Windows, `qsort_r` with arguments in different order in glibc vs. bsd, and eventually found out there's only non-thread-safe `qsort` available on Android. :frowning_face: At that point, I thought maybe we could do better and merge them on the fly up until max_breadcrumbs to avoid allocating in worst case 2 x max_breadcrumbs for sorting and then throw half of it away...

include/sentry.h
jpnurmi: I misunderstood https://github.com/getsentry/sentry-native/issues/1226#issuecomment-2880408968 that we should make it possible to scope I/O and tasks, but I realize now that comment came before the eventual empty write-only scope proposal. Taking ownership is fine by me and also what @tustanivsky proposed at https://github.com/getsentry/sentry-unreal/pull/928#discussion_r2113878692.

src/sentry_scope.c
jpnurmi: Thanks, I've updated it to return `0` and `1` for null a and b, respectively.

I have mixed feelings about whether it's better to deal with missing timestamps like this or simply fail. While breadcrumbs created by the SDK are guaranteed to have timestamps, nothing prevents one from creating breadcrumb values by hand.

src/sentry_scope.c
supervacuus: To be clear, I love the implementation, and it is much preferable to a "cheap" solution with overallocation and `qsort` to then discard cloned and sorted items. It is just a control-flow monster with a considerable amount of repetition that isn't easily abstracted, making it more difficult to digest. But this is still more than acceptable.

src/sentry_scope.c
supervacuus: > I have mixed feelings about whether it's better to deal with missing timestamps like this or simply fail. While breadcrumbs created by the SDK are guaranteed to have timestamps, nothing prevents one from creating breadcrumb values by hand.

I can help you with that 😅 

_It is much better to discard invalid breadcrumbs than to fail outright_. We must make the early return, because otherwise we would pass NULL pointers to `strcmp`, leading to an even more confusing crash. It doesn't really matter how those non-timestamped entries were entered. 

If our breadcrumbs appear without timestamps, something has likely gone terribly wrong (most likely a memory corruption). If users modified them, intentionally or not, then they should still get the breadcrumbs that weren't altered. It will make it much easier to contextualize, rather than having no breadcrumbs at all.

This is different from errors we receive from the allocator, where we could choose to return an incomplete result, but signaling that something was wrong is probably more prudent, since the merged breadcrumbs will never materialize in an envelope.

jpnurmi: For comparison, I think sentry-python throws an exception :open_mouth: 

https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/scope.py#L1382-L1394

supervacuus: Does it? I only see that it swallows any exceptions and converts them to debug logs. Logging could also make sense in our case.

I understand that this prevents any sorting from being applied, but the program continues and would send the breadcrumbs as they existed before in the event dictionary.

Our approach isn't too different except for logging, which is a valuable addition for this case. 

However, in the Python example, many more failure modes (attribute, key, type, and value errors), within that short try block, are swallowed into a single "problems with sorting" debug log.

jpnurmi: Ehh, never mind. It was a brainfart... `pass` vs. `raise` :grin:

jpnurmi:
Thanks! :tada:

PR #1253, opened 2025-05-22
fix: respect event data when applying/merging scope data

jpnurmi:
This PR fixes the issue that when scope data is merged into events, the scope data currently takes precedence for tags, contexts, and extra data.

For example, if a scope has the following tags/contexts/extra:
```json
{ "both": "scope", "scope-only": "scope" }
```

And an event has the following data in the respective field:
```json
{ "both": "event", "event-only": "event" }
```

Expected result (event data takes precedence): 
```json
{ "both": "event", "event-only": "event", "scope-only": "scope" }
```

Actual result (scope data takes precedence):
```json
{ "both": "scope", "event-only": "event", "scope-only": "scope" }
```


supervacuus (commented):
Thx, @jpnurmi. Since this is a breaking change (even if it fixes towards expected behavior), can you please:

* provide a changelog section that mentions this as a breaking change (in addition to the fix, but referencing the same PR)
* link that "breaking" to the docs that support this as the expected outcome

jpnurmi:
I wanted to offload this from the local scope PR, but it's ok to hold this off until local scopes are ready. Even if it's the correct behavior, as far as I know, nobody has asked for this, and there's always a risk someone somewhere is relying on the wrong behavior. The breaking change will be easier to sell together with local scopes that justify the behavior change. :)

supervacuus (approved):

Issue #1141, opened 2025-02-04
Adding Dynamic Sampling Context


JoshuaMoelans:
_(this issue is still a work in progress, used to collect all relevant material about DSC implementation in Native)_

[Link to develop docs](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/#sdks)

[When to propagate this](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/#dsc-propagation):
- `trace` envelope header
- downstream services via the baggage header

# Related issues/PRs
These should be turned into native-specific sub-issues (or PRs)
- [ ] https://github.com/getsentry/projects/issues/316
    (example implementation in sentry-cocoa https://github.com/getsentry/sentry-cocoa/issues/4687)
- [ ] https://github.com/getsentry/team-sdks/issues/117
- [x] https://github.com/getsentry/sentry-native/pull/1137

kahest:
related issue for Rust: https://github.com/getsentry/sentry-rust/issues/735

PR #1254, opened 2025-05-23
feat: add dynamic sampling context

JoshuaMoelans:
fixes https://github.com/getsentry/sentry-native/issues/1141
implemented following the [develop docs](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/#sdks)

- [x] add `sample_rand` functionality ( ‼️ breaking ‼️ )
- [x] collect all attributes (from [spec](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/#dsc-specification))
- [x] add DSC to transaction events
- [x] add DSC to error events
- [x] ~~make sure we don't need to send it to downstream services with `baggage` header ([docs](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/#baggage-header))~~
  - [x] create issue to backlog the `baggage` aspect of the DSC https://github.com/getsentry/sentry-native/issues/1267 
- [x] add tests 
  - [x] new `org_id` extraction
  - [x] proper DSC handling
  - [x] new `sample_rand` feature
- [ ] add to docs, also update example in [Custom Instrumentation](https://docs.sentry.io/platforms/native/tracing/trace-propagation/custom-instrumentation/) to explain how to send the headers after `sentry_transaction_iter_headers`


src/sentry_utils.c
```diff
@@ -227,6 +230,7 @@ sentry__dsn_new_n(const char *raw_dsn, size_t raw_dsn_len)
     memset(&url, 0, sizeof(sentry_url_t));
     size_t path_len;
     char *project_id;
+    char org_id[64] = ""; // TODO figure out the max. size?
```
JoshuaMoelans: From [relay](https://github.com/getsentry/relay/blob/a75423ed9503463deb5a58fd5a8b1b94d737bc7d/relay-base-schema/src/organization.rs#L8) seems like this is a u64, so 20 characters should suffice.. (+1 for null-termination)

tests/unit/test_tracing.c
```diff
@@ -300,11 +302,45 @@ SENTRY_TEST(transport_sampling_transactions)
...
+            = sentry_transaction_start(tx_ctx, sentry_value_new_null());
+        sentry_uuid_t event_id = sentry_transaction_finish(tx);
+        if (!sentry_uuid_is_nil(&event_id)) {
+            sent_transactions += 1;
+        }
+    }
+
+    sentry_close();
+    // within one trace, either all or no transactions get sampled
+    TEST_CHECK(called_transport == 0 || called_transport == 100);
```
JoshuaMoelans, line 343: Not sure I like this test structure, since we coin flip on every test run whether `sample_rand` < `sample_rate = 0.5`. 

I put the test here just to showcase the single-trace `sample_rand` behavior, but maybe we can leave it out for the existing `transport_sampling_transactions_set_trace`.

src/sentry_core.c
```diff
@@ -211,6 +246,8 @@ sentry_init(sentry_options_t *options)
         }
         sentry_value_freeze(scope->client_sdk);
         initialize_propagation_context(&scope->propagation_context);
+        // TODO pass in scope, or make it call SENTRY_WITH_SCOPE?
+        set_dynamic_sampling_context(scope);
```
JoshuaMoelans: I'm not 100% sure what function signature makes more sense here; for now I felt like passing in the scope was most logical.

tests/unit/test_tracing.c
supervacuus: What is the downside of this structure? Why would you leave it out?

src/sentry_core.c
supervacuus: Good question! While all our mutexes are recursive, we shouldn't overdo them. If you have a static helper that is invoked from a function that has already retrieved shared memory, please always pass references to it down. We should actively seek and eliminate unnecessary shared memory retrievals by argument passing.

The primary reason we allow/need recursive mutexes is that we internally reuse API entry points that require access to global SDK state.

tests/unit/test_tracing.c
JoshuaMoelans: Leaving it out might not make sense indeed, but I was thinking that it might be reasonable to have a test which runs this multiple times, such that probabilistically both cases get hit at least once in one test-suite-execution (instead of only the one coin flip whether this run it'll be 0 or 100).

supervacuus:
Let's put this on hold until we have clarity regarding #1270.

src/sentry_envelope.c
```diff
@@ -254,6 +255,44 @@ sentry__envelope_add_event(sentry_envelope_t *envelope, sentry_value_t event)
...
+    SENTRY_WITH_OPTIONS (options) {
+        traces_sample_rate = options->traces_sample_rate;
+    }
+    SENTRY_WITH_SCOPE (scope) {
+        sentry_value_t dsc = scope->dynamic_sampling_context;
+        sentry_value_t trace_id = sentry_value_get_by_key(
+            sentry_value_get_by_key(
+                sentry_value_get_by_key(event, "contexts"), "trace"),
+            "trace_id");
+        if (!sentry_value_is_null(trace_id)) {
```
supervacuus, line 268: According to the [`DSC` spec](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/#dsc-specification), `trace_id` is strictly required, so there should be some log that informs the user of a missing `trace_id`, especially if we don't fail at this point but continue to fill the `DSC`. Not having a `trace` context or a `trace_id` inside is likely a programmer error and should be caught during development.

I am not sure if it is necessary to log for each missing k/v, but a counter path to this conditional with a `WARN`/`ERROR` that says:

> "couldn't retrieve `trace_id` from scope to apply to dynamic sampling context"

should be in there.

src/sentry_envelope.c
```diff
@@ -254,6 +255,44 @@ sentry__envelope_add_event(sentry_envelope_t *envelope, sentry_value_t event)
...
+            sentry_value_incref(trace_id);
+            sentry_value_set_by_key(dsc, "trace_id", trace_id);
+        }
+        sentry_value_t sample_rand = sentry_value_get_by_key(
+            sentry_value_get_by_key(scope->propagation_context, "trace"),
+            "sample_rand");
+        if (!sentry_value_is_null(sample_rand)) {
```
supervacuus, line 278: Not sure about this one wrt the [`DSC` spec](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/#dsc-specification), but `sampled` is also "required", but only set if `sample_rand` exists inside the propagation context trace. I guess you don't explicitly send it because that is the default anyway?

tests/unit/test_tracing.c
supervacuus: I still don't understand what you want to achieve. This test says:

> out of 100 transactions that belong to the same trace, either all or none will be sampled, depending on the sampling decision for that trace

Another test could be written that allows us to change the trace as often as necessary, so that all 1..n transactions within each trace are sampled at least once, and not sampled at all. Is that the test you had in mind?

tests/unit/test_utils.c
```diff
@@ -114,6 +130,7 @@ SENTRY_TEST(dsn_parsing_project_id_without_path)
     TEST_CHECK_STRING_EQUAL(dsn->host, "example.com");
     TEST_CHECK_INT_EQUAL(dsn->port, 443);
     TEST_CHECK_STRING_EQUAL(dsn->public_key, "username");
+    TEST_CHECK_STRING_EQUAL(dsn->org_id, "");
```
supervacuus, line 179: I think `dsn_parsing_project_id_with_path_prefix` should also get one of these checks.

src/sentry_utils.c
```diff
@@ -250,6 +255,13 @@ sentry__dsn_new_n(const char *raw_dsn, size_t raw_dsn_len)
     }
 
     dsn->host = url.host;
+    const char *org_id_dot = strchr(url.host, '.');
+    if (org_id_dot && url.host[0] == 'o') {
+        size_t length = (size_t)(org_id_dot - url.host - 1); // leave the o
+        strncpy(org_id, url.host + 1, MIN(length, 20));
+        org_id[length] = '\0'; // Null-terminate the string
+    }
+    dsn->org_id = sentry__string_clone(org_id);
```
supervacuus: So, we do expect `"o"` to be followed by a numeric string, but we never check that this is the case. Right now, we could have a host that starts with `"orange."` that would give us an `org_id` of `"range"`. 

This may be intentional, as we might be aware that some users have non-numeric IDs in their locally hosted systems. However, if that weren't the case, then I prefer to check that we can convert the string to a `u64` (`strtoull()`). Later, when this gets merged with the `u64` value type, we can even switch `org_id` to a `uint64_t` instead of a string.

supervacuus (changes requested):
This looks largely good, with a few nits on edge cases and error handling.

src/sentry_utils.c
JoshuaMoelans: I also think checking for `u64` makes more sense than expecting non-numeric IDs in self-hosted Sentry instances.

src/sentry_envelope.c
JoshuaMoelans: Yes I think my rationale here was that `sample_rand` gets initialized upon SDK start, so therefor we always expect a value to exist. But since we're checking if it is `null` just to be sure anyway, I see no harm in adding an additional `SENTRY_WARN()` too.

tests/unit/test_tracing.c
JoshuaMoelans: Looking at this now, I think we can keep this test in as-is, since we also have the test structure pasted below as an additional test which tests the 'multiple times' idea I had.

```c
SENTRY_TEST(transport_sampling_transactions_set_trace)
{
    uint64_t called_transport = 0;

    sentry_options_set_dsn(options, "https://foo@sentry.invalid/42");
    // Disable sessions or this test would fail if env:SENTRY_RELEASE is set.
    sentry_options_set_auto_session_tracking(options, 0);

    sentry_transport_t *transport
        = sentry_transport_new(send_transaction_envelope_test_basic);
    sentry_transport_set_state(transport, &called_transport);
    sentry_options_set_transport(options, transport);

    sentry_options_set_traces_sample_rate(options, 0.75);
    sentry_init(options);

    uint64_t sent_transactions = 0;
    for (int i = 0; i < 100; i++) {
        // empty set_trace just to re-roll `sample_rand`
        sentry_set_trace("", "");
        sentry_transaction_context_t *tx_ctx
            = sentry_transaction_context_new("honk", "beep");
        sentry_transaction_t *tx
            = sentry_transaction_start(tx_ctx, sentry_value_new_null());
        sentry_uuid_t event_id = sentry_transaction_finish(tx);
        if (!sentry_uuid_is_nil(&event_id)) {
            sent_transactions += 1;
        }
    }

    sentry_close();

    // exact value is nondeterministic because of rng
    TEST_CHECK(called_transport > 50 && called_transport < 100);
    TEST_CHECK(called_transport == sent_transactions);
}
```

supervacuus: If this were merged with `master`, could we use `sentry_regenerate_trace()` here too, or does this only work with `sentry_set_trace()`?

src/sentry_envelope.c
supervacuus: I am less concerned about logging a warning here and more about whether `sampled` should also be set in the negative case.

tests/unit/test_tracing.c
JoshuaMoelans: Good idea, `sentry_set_trace` should work too (and be a cleaner test). Adapting it now!

supervacuus (approved):

PR #1256, opened 2025-05-23
chore: update gradle scripts to current AGP/SDK usage

supervacuus:
related to #962 and #974

#skip-changelog

PR #1258, opened 2025-05-27
fix(ndk): correct interpolation of new `project.layout.buildDirectory` property

supervacuus:
This is a follow-up to https://github.com/getsentry/sentry-native/pull/1256, where I replaced the deprecated `project.buildDir` with its lazy variant  `project.layout.buildDirectory`. It forgot to change the interpolation when constructing the distribution-zip path, since the new property is an entirely different type (duh).

This currently breaks the release archive job on `master` which isn't executed during the PR CI runs.


markushi (approved):
LGTM!

PR #1260, opened 2025-06-02
fix: close file and return 0 on success when writing raw envelopes

jpnurmi:
Fixes two issues when writing raw envelopes with `sentry_envelope_write_to_file()`:
- the file was not closed
- the return value was 1 on success

These issues came up while experimenting with a [custom disk transport](https://github.com/getsentry/sentry-dotnet/pull/4233) for sentry-dotnet that tries to move the file after it has been written, which fails on Windows:

> System.IO.IOException: The process cannot access the file because it is being used by another process.,    at System.IO.FileSystem.MoveFile(String, String, Boolean)


supervacuus (approved):
Nice catch. Thanks!

Issue #1240, opened 2025-05-14
Alpine Linux Docker image


jpnurmi:
The Alpine Linux Docker image is bare bones (< 5MB!) and requires quite a bit of boilerplate to make it usable for sentry-native's CI workflows. Consider preparing an image that has Bash, Git, CMake, Python etc. pre-installed to reduce the noise in ci.yml.

> Suppose the container setup (essentially, now parallel conditional build-steps in the workflow yaml) gets out of hand at some point. In that case, we can also evaluate switching to Dockerfiles written on top of the Alpine image, manually creating images that already contain what we want, and executing the pytest there.

_Originally posted by @supervacuus in https://github.com/getsentry/sentry-native/pull/1233#discussion_r2086210671_

PR #1261, opened 2025-06-02
ci: use Alpine Linux Docker image

jpnurmi:
Reverts most of the Alpine-specific steps introduced in #1233 by using a prebuilt Alpine Linux Docker image with relevant tools and dependencies pre-installed:
- https://github.com/getsentry/sentry-native/pkgs/container/sentry-native-alpine

Close: #1240

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,11 +34,7 @@ jobs:
           cmake -B sentry-native-xcode -GXcode -DCMAKE_SYSTEM_NAME=iOS
           xcodebuild build -project sentry-native-xcode/Sentry-Native.xcodeproj -sdk iphonesimulator
 
-  docker:
-    uses: getsentry/sentry-native/.github/workflows/docker.yml@ci/alpine
-
   test:
-    needs: [docker]
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -7,7 +7,6 @@ on:
     paths:
       - '.github/docker/**'
   workflow_dispatch:
-  workflow_call:
 
 jobs:
   alpine:


.github/workflows/ci.yml
```diff
@@ -273,19 +265,11 @@ jobs:
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
         run: |
           echo "127.0.0.1   sentry.native.test" | sudo tee -a /etc/hosts
+          # remove "::1 localhost ..." to avoid conflicts with proxy tests (musl)
+          sed '/^::1/d' /etc/hosts | sudo tee /etc/hosts
```
jpnurmi, line 275: `/etc/hosts` is a special bind-mounted file in GitHub Actions. I tried doing this in a Docker ENTRYPOINT script, but it was too early because the changes were lost at this point...

jpnurmi:
> Can't you just manually dispatch the docker workflow from the PR branch and then rerun the CI workflow?

For security reasons, the workflow must exist in the default branch to be able to trigger `workflow_dispatch` but it made me wonder if `workflow_call` from the PR branch is allowed...

.github/workflows/docker.yml
```diff
@@ -0,0 +1,41 @@
+name: Docker
+
+on:
+  push:
+    paths:
+      - '.github/docker/**'
```
supervacuus, line 8: Couldn't that lead to accidental overwrites? I understand how it would be convenient to always have the latest changes to the `Dockerfile` reflected in the image, but in this case, it could be from a pull request (PR), right? The image is shared with all other CI workflow invocations.

I don't want to overcomplicate this right away. Still, this automatic trigger "on push" together with an image name/tag that only uses the Alpine version as a signifier for change, seems like a problematic recipe.

Perhaps filtering here on master and releases is the right approach, as any PR issuer could still manually trigger this if they want to test the changes in the PR.

Appending additional version information might be sensible (although not as a first step); let's keep this in mind. For instance, if someone wants to test this within the PR, a manual workflow trigger could expose a parameter that appends to the version, preventing the corresponding image with tag `3.21` from being overwritten immediately in the registry. 

Versioning can also help prevent breakage if the `Dockerfile` needs to be changed due to a Native SDK version-dependent update (for instance, a new or replaced dependency). This doesn't have to be an Alpine x Native SDK version cross-product that updates with every Native SDK bump; a counter would suffice (and could be combined with the manual workflow parameter).

jpnurmi: Good point! Let's filter on the branch, at the very least.

I was considering this more like a one-time action. Once the image is there, ideally, one doesn't need to touch it, just like the ready-made runners. However, I didn't consider that with this, a random PR could indeed accidentally overwrite the image... 🤦

jpnurmi: Actually, I never thought individual PRs would start fiddling with the Docker image. PRs can always install packages straight in the workflow with `apk add` like it did before, right?

I wonder if I went too far by moving everything to the Dockerfile. What if we leave some minimal `apk add libunwind-dev libunwind-static` step in the CI workflow, just to make it easy to discover where to install Alpine packages if necessary?

supervacuus: > Actually, I never thought individual PRs would start fiddling with the Docker image. PRs can always install packages straight in the workflow with apk add like it did before, right?

Yes they can, but then that installation will only affect the current run.

> I wonder if I went too far by moving everything to the Dockerfile. What if we leave some minimal apk add libunwind-dev libunwind-static step in the CI workflow, just to make it easy to discover where to install Alpine packages if necessary?

I am not sure I understand. Do you want to enable contributors to add or remove packages? (If discoverability is a concern, consider commenting on the container entries in the matrix.) The boundary you've drawn in this PR is clean. I would only remove the automatic deployment of images. My concern is less about abuse (although that is also a concern), but more about honest mistakes. It's also like this: if someone overwrites the registry image, it's easy to recover a more sensible state. Last but not least, manual workflow dispatch is guarded so that it can only be run by authorized users.

jpnurmi: I pushed an update to make the experience closer to the ready-made Ubuntu runners. The most important explicitly linked sentry-native dependencies are now installed from the CI workflow. Should one need to experiment with new dependencies in PRs, the process is the same as for the Ubuntu runners.

PR #1262, opened 2025-06-02
feat: allow PS SDK to implement custom transport

vaind:
Similar to a custom backend that's already supported, this adds support for downstream SDKs (e.g. PlayStation) to implement custom transport with proprietary APIs of the target platform.

#skip-changelog - this is intended for internal use

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,8 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
 	set(LINUX TRUE)
 elseif(CMAKE_SYSTEM_NAME STREQUAL "AIX" OR CMAKE_SYSTEM_NAME STREQUAL "OS400")
 	set(AIX TRUE)
+elseif(CMAKE_SYSTEM_NAME STREQUAL "Prospero")
+	set(PROSPERO TRUE)
 endif()
 
 #setup sentry library type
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -123,17 +123,23 @@ else()
...
+		"The HTTP transport that sentry uses to submit events to the sentry server, can be either 'none', 'curl' or 'winhttp' on Windows.")
+endif()
 
 if(SENTRY_TRANSPORT STREQUAL "winhttp")
 	set(SENTRY_TRANSPORT_WINHTTP TRUE)
 elseif(SENTRY_TRANSPORT STREQUAL "curl")
 	set(SENTRY_TRANSPORT_CURL TRUE)
 elseif(SENTRY_TRANSPORT STREQUAL "none")
 	set(SENTRY_TRANSPORT_NONE TRUE)
+elseif(SENTRY_TRANSPORT STREQUAL "custom")
```
supervacuus: Do you think this is necessary? We already expose the transport configuration at runtime to our users. If you want to implement a different transport, you only have to select `SENTRY_TRANSPORT=none` and configure yours via the exposed API. This differs from the backend, where currently no runtime interface exists to register an alternative backend.

vaind: That's true and I've tried doing so in a custom sentry_init() override in the downstream SDK:
```c
int
sentry_init(sentry_options_t *options)
{
    if (options && !options->transport) {
        options->transport = sentry__transport_new_ps();
    }

    // ...
    
    int code = sentry__native_init(options);
    if (code != 0) {
        return code;
    }

    // Note: optios object is not available anymore after sentry__native_init()
    options = nullptr;

    // ...

    return 0;
}
```

However, this way users have no way to actually disable the custom transport and leave only crash handling (which is handled by the PS SDK itself). 

In order to do that, I'd need to add another ifdef to sentry-native - to be able to override `sentry_options_new(void)` and set the PS transport there. Do you prefer that change?

supervacuus: > However, this way users have no way to actually disable the custom transport and leave only crash handling (which is handled by the PS SDK itself).

Ok, I wasn't aware of that need.

> In order to do that, I'd need to add another ifdef to sentry-native - to be able to override sentry_options_new(void) and set the PS transport there. Do you prefer that change?

I am not sure, tbh. How would adding the `#ifdef` to `sentry_options_new()` allow users to disable your transport downstream? I mainly raised this issue because extending the SENTRY_TRANSPORT build parameter with a "custom" value seems like adding an option for something that already works without it. So I see a potential for confusion.

Couldn't you add a PS-specific compile-time flag downstream that introduces an `#ifdef` around 

```C
    if (options && !options->transport) {
        options->transport = sentry__transport_new_ps();
    }
```

in the downstream `sentry_init()`? That way, your users have a choice of using your transport implementation, none, or their own?

Alternatively, simply rename "custom" to "ps" or "downstream", etc., to avoid a concept name clash with how users currently define custom transports.

vaind: > How would adding the #ifdef to sentry_options_new() allow users to disable your transport downstream

Because the default transport is currently set over in `sentry_options_new()` - we could override that with the PS default transport.

> Couldn't you add a PS-specific compile-time flag downstream that introduces an #ifdef around

I could, but introducing another cmake variable in addition to `SENTRY_TRANSPORT` would just cause more confusion IMO. Instead, I came up with an alternative proposal - see the latest commit - it enumerates supported platforms for PS as here in the sentry-native repo and let the downstream SDK define the source.

supervacuus: > Instead, I came up with an alternative proposal - see the latest commit - it enumerates supported platforms for PS as here in the sentry-native repo and let the downstream SDK define the source.

👍 Makes sense!

Issue #1263, opened 2025-06-02
CI: spammy Git errors in "Linux (clang 19 + kcov)" during tests


jpnurmi:
Not sure if this has any impact on the job being so slow, but the `Test` step of the `CI / Linux (clang 19 + kcov)` job prints 1000+ Git errors while running tests:

```
[100%] Linking C executable ../../sentry_test_unit
[100%] Built target sentry_test_unit
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
Test assert_sdk_name...                         [ OK ]
PASSED
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
tests/test_unit.py::test_unit[assert_sdk_user_agent] Test assert_sdk_user_agent...                   [ OK ]
PASSED
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
tests/test_unit.py::test_unit[assert_sdk_version] Test assert_sdk_version...                      [ OK ]
PASSED
...
```

https://github.com/getsentry/sentry-native/actions/runs/15388061935/job/43315287311

PR #1265, opened 2025-06-03
ci: pin kcov to v43

supervacuus:
The behavior described in #1263 leads back to a [recent version of `kcov`](https://github.com/SimonKagstrom/kcov/commit/0ecc2dc754765f77c72e2637a244c579d3c63cf2) integrating a `git` metadata collection for the `coveralls` writer.

Our test execution happens in temporary directories rather than a `git` repository, so that all invocations will fail. Additionally, we don't use `coveralls` as a coverage service, but `kcov` doesn't allow selecting its writers (they are all registered on startup). 

Since it is generally a good idea to pin CI dependencies, I think this is a sane fix to flooding the `kcov` jobs with irrelevant `git` errors.

An upstream solution to the problem would be to silence `stderr` on the `git` invocations.

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,6 +190,7 @@ jobs:
           sudo apt-get install binutils-dev libssl-dev libelf-dev libstdc++-12-dev libdw-dev libiberty-dev
           git clone https://github.com/SimonKagstrom/kcov.git
           cd kcov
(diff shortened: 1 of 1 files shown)


jpnurmi:
how about using the latest release v43 unless a more recent is needed? ubuntu 22.04 had v38 and in 25.04, kcov is back at v43.

https://packages.ubuntu.com/plucky/kcov

Issue #586, opened 2021-08-27
Changing attachments after `sentry_init`


thanhldnv:
**Description**
I'm reading the source of Sentry SDK, and I see a comment relate using `sentry_options_t *options`.
```
/**
 * Initializes the Sentry SDK with the specified options.
 *
 * This takes ownership of the options.  After the options have been set
 * they cannot be modified any more.
 * Depending on the configured transport and backend, this function might not be
 * fully thread-safe.
 * Returns 0 on success.
 */
SENTRY_API int sentry_init(sentry_options_t *options);
```

My application need to attach the log file when a crash occurs. My logger works like `unix logrotate`, it will generate a new file every day or when the file size exceeds a limit. Then I need to update the new log file path via `sentry_options_add_attachment` so I understand this function will change the `options object`. 
So can I use `sentry_options_add_attachment` after `sentry_init` and is it safe ?
 As I tested this way Sentry still work well.

brianmichel:
Just wanted to add my plus one to this, since I'm discovering the same thing, and also wanting to have the ability to change attachments at runtime. As we're building https://github.com/thebrowsercompany/swift-sentry we want to mirror the Cocoa APIs to have a somewhat drop in replacement. I was a bit confused when I saw that adding an attachment required a `sentry_options_t` which we discard after initialization.

After reading the comments available on https://github.com/getsentry/sentry-native/pull/433 it's unclear  what API concerns still exist, or what is blocking it (other than the questions that exist which have no reply).

It would be exceptional to have this API!

supervacuus:
> After reading the comments available on #433 it's unclear what API concerns still exist, or what is blocking it (other than the questions that exist which have no reply).

Hi @brianmichel, the API concerns are mostly related to downstream SDKs that include the Native SDK to handle crashes on the "native" layer. Those SDKs must provide attachments via "byte-arrays" (in addition to lazily loaded paths). Since the PR was created in response to the needs of downstream SDKs, this still needs to be solved.

The more significant issue for users of our crashpad backend is that the `crashpad_handler` has no way of updating attachments after the process starts. This happens in `sentry_init()`, which makes attachments for crash events when using this backend essentially equally flexible to our current situation. 

Of course, changing this in crashpad is an option, but it would require a cross-platform IPC mechanism that we'd have to introduce between the client and the handler, which increases the maintenance effort of our fork considerably.

[...]

Issue #978, opened 2024-04-11
Extend envelope API to allow event-specific attachments


hongjiaheng:
Doesn't anyone have this need? Attach specific files when sending events. Currently, the SDK only provides a global interface for attaching files:
![image](https://github.com/getsentry/sentry-native/assets/72791827/fb75e81d-1278-493b-b866-b404ae650f6a)

Is there a way to attach different files for different events? I think the SDK support this, but it hasn't exposed the corresponding interface, because now it only provide one header file called sentry.h, but there are some userfull function define in sentry_envelope.h i cannot use it.
![image](https://github.com/getsentry/sentry-native/assets/72791827/407fd9c6-52d7-4dcb-a42a-f7b95d7cf0ca)

supervacuus:
Hi @hongjiaheng. While we expose some `envelope` interfaces in our header, these mainly deal with loading events/transactions from persisted envelopes.

Exposing envelopes in the send path would require opening up either with another 
envelope-centered hook or splitting that path and exposing the parts to the public API. 

This would allow users to extend the envelope further after we applied scope-related changes (including global attachments) and invoked the `before_send()` hook. Then, it would make sense to expose further envelope-related interfaces in the public API to allow, for instance, event-specific attachments.

Let's leave this here so users can vote.

hongjiaheng:
Thanks for your reply. I think this is a important requirement. I asked my Android colleagues, and their Sentry SDK supports this feature.

JoshuaMoelans:
Hi @simonvanbernem , is there a reason for you not to add the logfile/diagnostic data as an attachment in general (using `sentry_options_add_attachment(options, "/var/server.log");` during SDK init)? These should get picked up for game crashes without needing to use the local scoped attachments.

jpnurmi:
> I need to upload 3 files in total on crash

Would calling `sentry_attach_file` from `on_crash` do the trick?

JoshuaMoelans:
Ah thanks for clarifying! We only add attachments to events captured manually with `sentry_capture_event` (or to crash events, which you probably always want the attachments for). Logs are lightweight, so they only contain [minimal context](https://docs.sentry.io/platforms/native/logs/#default-attributes) & can be extended with custom attributes.

PR #1266, opened 2025-06-05
feat: Support modifying attachments after init (continued)

jpnurmi:
TODO:
- [x] Rebase starting point by @Swatinem
  - https://github.com/getsentry/sentry-native/pull/433
- [x] Crashpad IPC (Windows & Linux)
  - https://github.com/getsentry/crashpad/pull/124
- [ ] Add attachment API for local scopes (#1248)
- [ ] Tests, docs, etc.

See: #586


src/backends/sentry_backend_crashpad.cpp
```diff
@@ -660,6 +661,30 @@ crashpad_backend_prune_database(sentry_backend_t *backend)
     crashpad::PruneCrashReportDatabase(data->db, &condition);
 }
 
+#if defined(SENTRY_PLATFORM_WINDOWS) || defined(SENTRY_PLATFORM_LINUX)
+static void
+crashpad_backend_add_attachment(
+    sentry_backend_t *backend, const sentry_path_t *attachment)
+{
+    auto *data = static_cast<crashpad_state_t *>(backend->data);
+    if (!data->client) {
```
supervacuus: We should also check `data` here. I know there is a 100% chance that if the add attachment interface has been registered, `data` will not be `NULL`, just from following our initialization, but with all global data, I think it is fair to add the check.
````suggestion
    if (!data || !data->client) {
```

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -660,6 +661,30 @@ crashpad_backend_prune_database(sentry_backend_t *backend)
...
+    }
+    data->client->AddAttachment(base::FilePath(attachment->path));
+}
+
+static void
+crashpad_backend_remove_attachment(
+    sentry_backend_t *backend, const sentry_path_t *attachment)
+{
+    auto *data = static_cast<crashpad_state_t *>(backend->data);
+    if (!data->client) {
```
supervacuus: We should also check `data` here:
```suggestion
    if (!data || !data->client) {
``````

File: src/sentry_attachment.c [hunk 3]
```diff
@@ -0,0 +1,155 @@
...
+
+out:
+    sentry__path_free(path);
+}
+
+/**
+ * Reads the attachments from disk and adds them to the `envelope`.
+ */
+void
+sentry__apply_attachments_to_envelope(
```
supervacuus: So if this is meant as `sentry__` (i.e., module-public) interface, the usage docs should be in the module header rather than here (which they are, so that we only need to remove this here). Inline documentation here is also acceptable, but that should be implementation-specific details that answer questions about "why was this implemented this way?".

src/sentry_attachment.c
```diff
@@ -0,0 +1,155 @@
...
+    case VIEW_HIERARCHY:
+        return "event.view_hierarchy";
+    default:
+        UNREACHABLE("Unknown attachment type");
+        return "event.attachment";
+    }
+}
+
+static void
+sentry__attachment_free(sentry_attachment_t *attachment)
```
supervacuus: `static` (i.e., module-private) functions should not have a `sentry__` prefix.

src/sentry_core.c
```diff
@@ -1449,3 +1427,91 @@ sentry_capture_minidump_n(const char *path, size_t path_len)
 
     return sentry_uuid_nil();
 }
+
```
supervacuus, line 1430: I think it is fair to move all attachment-related public functions to the `sentry_attachment` module (except if you did that to break a harmful cyclic dependency).

I know that there are a couple of public functions in the core that could also be moved to respective modules, but I don't think that most of them are a great example that should be copied. 

The "core" should reflect the stuff that isn't in the modules, not necessarily a collection of public interfaces. If "core" helps to break cyclics between modules, then I am all for adding functions to core, but not in general (and attachment is probably a far-out leaf in the module dependency tree). Modules can and should contain implementations of public functions.

supervacuus:
> Hmm, is there a potential need for the user to be able to specify attachment or content types?

We recently discussed this in the context of view hierarchies. There is no sensible reason to expose the attachment type, since those should be handled by the SDK internally. 

Content type could be something to expose if users use non-standard extensions that the Python MIME type checker doesn't recognize.

src/sentry_core.c
jpnurmi: The structural changes were inherited from #433 for calling `sentry__attachment_add/remove` from both `sentry_options` and `sentry_core`, but I noticed some duplicate code in `sentry__apply_attachments_to_envelope` vs. `sentry__envelope_add_attachment` which is fixed by the last commit.

supervacuus:
> I was worried about introducing API that would steal the best names without being easy to extend, but it would be unfortunate to lose the convenient path-based API. If attachment-specific API is ever needed, changing `sentry_add_attachment` to return the attachment could be a nice way to retain that convenience.

I understand that worry. I think if you want to future-proof this, then we could decide on the following naming:

* `sentry_add_attachment_path` for the API that you expose in this PR
* `sentry_add_attachment_buffer` for the API that reads an in-memory buffer
* `sentry_add_attachment` for the generic thing that constructs an attachment object around an opaque handle (which the other two use internally)

With the naming set in stone, you could now release `sentry_add_attachment_path` first without killing the option of introducing the other interface names. There is no breakage concerning the limited path interface (which reflects the current option API) if we decide to add the others later. Of course, if you want to provide the generic one immediately, that is fine as well, but it's certainly not a blocker for me.

jpnurmi:
I checked sentry-docs and other SDKs for inspiration. It appears that in-memory buffers are generally referred to as "data" and/or "bytes".

If we reserve `sentry_add_attachment(obj)` for the object-based API, how do you like:

1. `sentry_add_attachment_from_file(path)` vs. `sentry_add_attachment_from_data(bytes, len)`, or
2. `sentry_attach_file(path)` vs. `sentry_attach_data(bytes, len)`

for the path and data-based convenience APIs?

## Current API

For reference. Could be deprecated and removed in 1.0 in favor of the new global APIs.

### Options

```c
void sentry_options_add_attachment(sentry_options_t *opts, const char *path);
void sentry_options_add_attachment_n(sentry_options_t *opts, const char *path, size_t path_len);
void sentry_options_add_attachmentw(sentry_options_t *opts, const wchar_t *path);
void sentry_options_add_attachmentw_n(sentry_options_t *opts, const wchar_t *path, size_t path_len);
```

## Proposed API

Here's a draft of the attachment API with some potential future additions.

### Global

```c
// later
// void sentry_add_attachment(sentry_attachment_t *attachment);

// alternative names: sentry_attach_file, sentry_add_attachment_path
void sentry_add_attachment_from_file(const char *path);
void sentry_add_attachment_from_file_n(const char *path, size_t path_len);
void sentry_add_attachment_from_filew(const wchar_t *path);
void sentry_add_attachment_from_filew_n(const wchar_t *path, size_t path_len);

// later (alternative names: sentry_attach_data, sentry_add_attachment_data)
// void sentry_add_attachment_from_data(const uint8_t *bytes, size_t bytes_len, const char *filename); // + _n + w

void sentry_remove_attachment(const char *path); // or filename for in-memory buffers
void sentry_remove_attachment_n(const char *path, size_t path_len);
void sentry_remove_attachmentw(const wchar_t *path);
void sentry_remove_attachmentw_n(const wchar_t *path, size_t path_len);
```

### Scope

```c
// later
// void sentry_scope_add_attachment(sentry_scope_t *scope, sentry_attachment_t *attachment);

// alternative names: sentry_scope_attach_file, sentry_scope_add_attachment_path
void sentry_scope_add_attachment_from_file(sentry_scope_t *scope, const char *path);
void sentry_scope_add_attachment_from_file_n(sentry_scope_t *scope, const char *path, size_t path_len);
void sentry_scope_add_attachment_from_filew(sentry_scope_t *scope, const wchar_t *path);
void sentry_scope_add_attachment_from_filew_n(sentry_scope_t *scope, const wchar_t *path, size_t path_len);

// later (alternative names: sentry_scope_attach_data, sentry_scope_add_attachment_from_data)
// void sentry_scope_add_attachment_data(sentry_scope_t *scope, const uint8_t *bytes, size_t bytes_len, const char *filename); + _n + w
```

### Attachment (later)

```c
// alternative names: sentry_attachment_new_path
sentry_attachment_t *sentry_attachment_from_file(const char *path); // + _n + w

// alternative names: sentry_attachment_new_data
sentry_attachment_t *sentry_attachment_from_data(const uint8_t *bytes, size_t bytes_len, const char *filename); // + _n + w

// void sentry_attachment_set_content_type(sentry_attachment_t *attachment, const char *content_type);
// void sentry_attachment_set_filename(sentry_attachment_t *attachment, const char *filename);  // + _n + w
// ...
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Features**:
+
+- The `sentry_attach_file`, `sentry_scope_attach_file` (and their wide-string variants), and `sentry_remove_attachment` have been added to modify the list of attachments that are sent along with sentry events after a call to `sentry_init`. ([#1266](https://github.com/getsentry/sentry-native/pull/1266))
```
supervacuus, line 7: Please clarify that this does not work on macOS. I know there is a remark in the readme, but it should also be stated in the changelog.

supervacuus (approved):
Beauty, thx! That got quite a bit of polish during the last rounds. I agree that using the actual opaque pointer as a handle to refer to attachments is better than going via the filename.

From my side this is ready.

supervacuus (approved):
Beauty, thx! That got quite a bit of polish during the last rounds. I agree that using the actual opaque pointer as a handle to refer to attachments is better than going via the filename.

From my side this is ready.

supervacuus:
Nothing is worth being said if you can't do it at least three times. 😅 I guess yesterday's outage is still ongoing.

jpnurmi:
Thanks for the feedback and guidance! I'm also happy with the result :)

> This also provides a solution for #978 if one uses a local scope together with the dynamic attachment. Adding this use case to the documentation might be a good idea.

I opened a doc PR here:

- https://github.com/getsentry/sentry-docs/pull/14019

PR #1269, opened 2025-06-10
chore: add crash-specific info to ucontext on PS

vaind:
#skip-changelog - internal change

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -517,7 +517,22 @@ typedef struct sentry_ucontext_s {
 #ifdef _WIN32
     EXCEPTION_POINTERS exception_ptrs;
 #elif defined(SENTRY_PLATFORM_PS)
-    int data;
+    /**
+     * ID of the thread that caused the crash,
+     * or -1 if an async exception occured,
+     * or -2 if the user triggered the crash dump manually.
+     */
+    pthread_t thread;
+    /**
+     * The reason code of the crash, refer to Kernel Reference.
+     */
+    int reason_code;
+    /**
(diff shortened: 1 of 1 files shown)


include/sentry.h
```diff
@@ -517,7 +517,22 @@ typedef struct sentry_ucontext_s {
...
+    /**
+     * The reason code of the crash, refer to Kernel Reference.
+     */
+    int reason_code;
+    /**
+     * Non-zero if the cause of the crash was a GPU protection fault.
+     * Because GPU may still continue to run after the crash, this
+     * timestamp is used to pinpoint the exact time of the crash.
+     */
+    uint64_t gpu_timestamp;
```
supervacuus: How stable is this set? This is not a blocker, and since it only affects downstream components, I'll leave this up to you, of course. However, if ABI stability is a concern for PS users, and the fields here could be extended down the road, it might be sensible to only expose an opaque pointer with getters (not sure if setters are needed for the crash context in the public header).

The other fields here are extremely stable because they are core to the respective operating systems. And even for those we thought about hiding them somehow (although here, primarily to prevent the requirement on `windows.h`, which requires a different solution).

supervacuus (approved):

include/sentry.h
vaind: I don't think ABI stability is a requirement for the SDK but sure, it doesn't hurt anything to make this opaque

vaind: although, these are actually just set by calling PS SDK APIs so it's safe to remove them completely and let consumers call said APIs themselves if they need to.

PR #1270, opened 2025-06-10
feat: manage trace life-cycle

supervacuus:
With the recently introduced support for traces without performance (`TwP`) and how they interact with performance-related traces, we broke with an undocumented assumption: _there should be no endless traces_.

Endless traces would have consequences for the visualization in the web UI as well as sampling decisions, as we'd like to introduce in #1254 (where events/transactions along a single trace should follow the same dice roll, to obtain complete traces).

Since all events inherit their traces from the propagation context, which the SDK initializes with a random trace and dummy span, this trace would be active forever. When `sentry_set_trace()` is used (either by the user directly or in a downstream SDK), this also means that the caller assumes the burden of managing the trace life-cycle. However, if this doesn't happen, it is the Native SDK's duty.

After a short discussion during last week's sync, we established transactions as a boundary that breaks traces (leaving the question open of what happens without performance, which I tried to address in the comments).

Related docs PR: https://github.com/getsentry/sentry-docs/pull/14586


supervacuus:
This change is no longer relevant, since traces can now have arbitrary length.

Summary of the quick discussion we had about this: if this ever comes up again, then the right choice parameterization is to define the number of captures until a trace is invalidated.

JoshuaMoelans:
After a chat with @cleptric , we actually do want Transactions to be the default trace boundary in the native SDK. 

The idea of having multiple transactions in one trace only holds for platforms where there actually are specific trace-boundaries (e.g. mobile screen load, web page load, unity scene load, ...). Hence, we reopen this PR to be merged before release `0.10.0`.

+ I think we can still keep the [manual trace boundary API](https://github.com/getsentry/sentry-native/pull/1293) around for Tracing Without Performance (e.g. for a custom boundary between events/logs/...)

src/sentry_core.c
```diff
@@ -1131,6 +1132,34 @@ sentry_transaction_finish_ts(
...
+            //   propagation_context? We can't use the transaction's, because we
+            //   want to be outside the transaction's span hierarchy. Could we
+            //   just keep the initially created dummy or should generate a new
+            //   one?
+            // TODO: if there are no transactions, all the events will be still
+            //   part of the same trace. If this needs to be fixed too, then we
+            //   could expose a capture counter that resets the trace every N
+            //   events being captured. This might also be more
+#endif
+        }
```
JoshuaMoelans: leftover cleanup of the 'alternative option'

```suggestion
        // if the SDK manages the trace (rather than the user or a downstream
        // SDK) we break propagation context traces at transaction boundaries.
        if (scope->trace_managed) {
          // change the trace in the propagation context to the one
          // the transaction created
          sentry_value_t txn_trace_id
              = sentry_value_get_by_key(tx, "trace_id");
          sentry_value_incref(txn_trace_id);

          sentry_value_set_by_key(
              sentry_value_get_by_key(scope->propagation_context, "trace"),
              "trace_id", txn_trace_id);
        }
```

src/sentry_core.c
```diff
@@ -1131,6 +1132,34 @@ sentry_transaction_finish_ts(
...
+                "trace_id", txn_trace_id);
+            // TODO: what should we do with the span_id in the
+            //   propagation_context? We can't use the transaction's, because we
+            //   want to be outside the transaction's span hierarchy. Could we
+            //   just keep the initially created dummy or should generate a new
+            //   one?
+            // TODO: if there are no transactions, all the events will be still
+            //   part of the same trace. If this needs to be fixed too, then we
+            //   could expose a capture counter that resets the trace every N
+            //   events being captured. This might also be more
```
JoshuaMoelans: Discussed this w/ Michi, and this is fine: we still allow `sentry_regenerate_trace()` in case people find their traces to be too convoluted/long-running.

src/sentry_core.c
```diff
@@ -1131,6 +1132,34 @@ sentry_transaction_finish_ts(
...
+            sentry_value_incref(txn_trace_id);
+
+            sentry_value_set_by_key(
+                sentry_value_get_by_key(scope->propagation_context, "trace"),
+                "trace_id", txn_trace_id);
+            // TODO: what should we do with the span_id in the
+            //   propagation_context? We can't use the transaction's, because we
+            //   want to be outside the transaction's span hierarchy. Could we
+            //   just keep the initially created dummy or should generate a new
+            //   one?
```
JoshuaMoelans: I think keeping the dummy one is fine.

JoshuaMoelans (approved):
LGTM; only needs fixing the one 'alternative' `if 0` path.

src/sentry_core.c
supervacuus: When I created this PR, we had no `sentry_regenerate_trace()` yet. So I wrote the comment partially from that POV.

However, one aspect still holds: the argument why we split traces at transaction boundaries still holds for use cases without performance, but in those cases, users must do it manually, and they have to know about it.

This is a weird asymmetry, especially given that the driver for the trace management is `TwP`. So the docs should be very clear about this and might even justify an entry in the "develop" docs regarding the difference to other SDKs that have trace-boundaries matching platform-specific points of interest.

tests/unit/test_tracing.c
```diff
@@ -1762,9 +1779,33 @@ SENTRY_TEST(propagation_context_init)
...
+    TEST_ASSERT(!!tx_ctx2);
+    sentry_transaction_t *tx2
+        = sentry_transaction_start(tx_ctx2, sentry_value_new_null());
+    TEST_ASSERT(!!tx2);
+    char *tx_trace_id2 = sentry__string_clone(sentry_value_as_string(
+        sentry_value_get_by_key(tx2->inner, "trace_id")));
+
+    // regenerating the trace turned off automatic trace management, so we that
+    // the transaction has the same trace_id as the one being applied after
+    // regenerating the trace previously.
```
JoshuaMoelans:
```suggestion
    // regenerating the trace turned off automatic trace management, so
    // the transaction has the same trace_id as the one being applied after
    // regenerating the trace previously.
```

JoshuaMoelans (approved):
LGTM, just very minor nit on a typo

src/sentry_core.c
```diff
@@ -21,6 +21,11 @@
 #include "sentry_transport.h"
 #include "sentry_value.h"
 
+#ifdef SENTRY_PLATFORM_WINDOWS
+#    include "sentry_os.h"
+#    include "sentry_screenshot.h"
+#endif
```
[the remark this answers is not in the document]
supervacuus: Yes, this seemingly was a merge accident from the GH auto-merge.

src/sentry_core.c
```diff
@@ -1131,6 +1136,17 @@ sentry_transaction_finish_ts(
...
+        // SDK) we break propagation context traces at transaction boundaries.
+        if (scope->trace_managed) {
+            sentry_value_t txn_trace_id
+                = sentry_value_get_by_key(tx, "trace_id");
+            sentry_value_incref(txn_trace_id);
+
+            sentry_value_set_by_key(
+                sentry_value_get_by_key(scope->propagation_context, "trace"),
+                "trace_id", txn_trace_id);
+        }
```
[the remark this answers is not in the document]
supervacuus: That is intentional until further specification counters the decision. From our current POV, only the `trace_id` should be changed, and the dummy hierarchy created during initialization can survive without compromising the effect of a trace boundary. I raised this topic during development and this resulted in the current decision. We can change this anytime.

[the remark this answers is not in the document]
supervacuus: https://github.com/getsentry/sentry-native/blob/0788b371c27995f33f92471f53264d2ac9f1a104/src/sentry_core.c#L1083-L1087

Issue #1210, opened 2025-04-22
Sentry Structured Logging for Native (C/C++)


AbhiPrasad:
https://develop.sentry.dev/sdk/telemetry/logs/

Sentry is adding support for [structured logging](https://github.com/getsentry/sentry/discussions/86804). Let's add it to the Native SDKs!

1. Define the [logs protocol](https://develop.sentry.dev/sdk/telemetry/logs/#logs-protocol) and log envelope item in the SDK
1. Add the Public API (SDK options and methods) as per the [docs](https://develop.sentry.dev/sdk/telemetry/logs/#public-api)
1. Make sure the the SDK follows the [documented behavior](https://develop.sentry.dev/sdk/telemetry/logs/#sdk-behavior)
1. Attach [default attributes](https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes) to the SDK as per docs
1. Instrument popular logging libraries to send logs to Sentry. SDK maintainers can best decide what they should support, but we should definitely support the same integrations that we do in the SDK today.

AbhiPrasad:
This was released with `0.11.1` of the Native SDK. Docs for this can be found here: https://docs.sentry.io/platforms/native/logs

Please give it a try and let us know what you think!

PR #1271, opened 2025-06-12
feat: structured logging 

JoshuaMoelans:
Closes https://github.com/getsentry/sentry-native/issues/1210

Will be structured into sub-PRs that merge into this branch
- [x] add option `enable_logs`
- [x] add `logs` module with functions like `sentry_log_debug("Log number %i : %s", 3, "I'm logged up")` https://github.com/getsentry/sentry-native/pull/1272
  - [x] send log envelopes according to [spec ](https://develop.sentry.dev/sdk/data-model/envelope-items/#log)
- [x] provide named parameter API 
  - (ON HOLD at https://github.com/getsentry/sentry-native/pull/1297 ) As mentioned in [this comment](https://github.com/getsentry/sentry-native/pull/1272#pullrequestreview-2954506464), we could have an alternative interface that takes in a list of `sentry_value_t` objects containing `{name, value}` tuples, allowing us to make use of `sentry.message.parameter.X` to the fullest extent (and not just relying on parameter index). Example API would look like `sentry_log_info("some %d data %p that we substitute %s", my_list);` where we then extract objects from `my_list` (a `sentry_value_t` list) to construct the string & store the named parameters.
- [x] add size/time-based buffering for logs https://github.com/getsentry/sentry-native/pull/1338
  - size-based can trigger whenever log is appended to `logs` list (and it goes over the max)
  - time-based flush every 5s (timer only starts after first log is added)
- [x] filter out `message.template` if there are no parameters (see https://github.com/getsentry/sentry-docs/pull/14759/files) part of https://github.com/getsentry/sentry-native/pull/1355
- [x] add `before_send_log` callback https://github.com/getsentry/sentry-native/pull/1355
- [x] add tests
  - [x] interaction with tracing (connected by `trace_id` and `span_id`
  - [x] tests for batching
- [x] add to changelog
- [ ] add to docs & update [Supported SDKs](https://docs.sentry.io/product/explore/logs/getting-started/) https://github.com/getsentry/sentry-docs/pull/15021
- [ ] reach out to [interested parties](https://github.com/getsentry/sentry/discussions/86804#discussioncomment-12681716) on the [Logs Thread](https://github.com/getsentry/sentry/discussions/86804)
- [x] update TODOs of https://github.com/getsentry/sentry-native/pull/1320 w.r.t. `int64` (see [commit](https://github.com/getsentry/sentry-native/pull/1320/commits/fb54090e604e44edd2732f6d81ce8a77261af5b8)) -> tracked on https://github.com/getsentry/sentry-native/pull/1328

Also, next to the initial implementation some future work remains. These should become their own issues after this PR is finished (and the merit to these improvements has been discussed as being worthwhile to implement)
- [ ] use more advanced [Batch Processor](https://develop.sentry.dev/sdk/telemetry/spans/batch-processor/) to buffer log sending
- [ ] ensure logs get sent/cached to disk on crash
- [ ] add integrations
  - [ ] `sentry__logger_func()` (e.g. `SENTRY_WARNF()`) ❗ (could be part of initial feature release, but consider cost; should people 'pay' for logs created by sdk internals? (more useful to us for debugging))
  - [ ] console logging (cout/printf)
  - [ ] others?
- [ ] think about prioritized sending (already brought up for Android/Cocoa)
  - In these SDKs, it's about prioritizing sending/cleaning up offline envelopes (e.g., there is a max storage after which they should ring-buffer-remove the oldest items, but then prioritizing keeping Errors around becomes a consideration)

#skip-changelog for now

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 **Features**:
 
-- Add support for structured logs. It is currently experimental, and one can enable it by setting `sentry_options_set_enable_logs`. When enabled, you can capture a log using `sentry_log_info()` (or another log level). Logs can be filtered by setting the `before_send_log` hook. More details and examples can be found in the docs. ([#1271](https://github.com/getsentry/sentry-native/pull/1271/))
+- Add support for structured logs. It is currently experimental, and one can enable it by setting `sentry_options_set_enable_logs`. When enabled, you can capture a log using `sentry_log_info()` (or another log level). Logs can be filtered by setting the `before_send_log` hook. ([#1271](https://github.com/getsentry/sentry-native/pull/1271/))
 
 ## 0.11.0
 


src/sentry_logs.c
```diff
@@ -0,0 +1,759 @@
...
+        SENTRY_WITH_OPTIONS (options) {
+            if (options->before_send_log_func) {
+                log = options->before_send_log_func(
+                    log, options->before_send_log_data);
+                if (sentry_value_is_null(log)) {
+                    SENTRY_DEBUG(
+                        "log was discarded by the `before_send_log` hook");
+                    discarded = true;
+                }
+            }
```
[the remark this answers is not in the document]
JoshuaMoelans: We provide the following warning in the `sentry_before_send_log_function_t` docstring:
```
/**
 * Type of the `before_send_log` callback.
 *
 * The callback takes ownership of the `log`, and should usually return
 * that same log. In case the log should be discarded, the
 * callback needs to call `sentry_value_decref` on the provided log, and
 * return a `sentry_value_new_null()` instead.
 */
```
So just like the other `before_X` callbacks, the user is responsible for cleaning up the passed-in log in case they want to return a NULL value.

examples/example.c
```diff
@@ -275,6 +305,32 @@ create_debug_crumb(const char *message)
...
 
+#ifdef SENTRY_PLATFORM_WINDOWS
+DWORD WINAPI
+log_thread_func(LPVOID lpParam)
+{
+    (void)lpParam;
+    int LOG_COUNT = 100;
+    for (int i = 0; i < LOG_COUNT; i++) {
+        sentry_log_debug(
+            "thread log %d on thread %lu", i, get_current_thread_id());
```
supervacuus, line 339: I would still introduce minimal waits here, since this should also be an example. You can expose the sleep duration in milliseconds as a `#define` and allow people to play with it, but I think it is a good idea to highlight the potential for losses.

examples/example.c
```diff
@@ -275,6 +305,32 @@ create_debug_crumb(const char *message)
...
+}
+#else
+void *
+log_thread_func(void *arg)
+{
+    (void)arg;
+    int LOG_COUNT = 100;
+    for (int i = 0; i < LOG_COUNT; i++) {
+        sentry_log_debug(
+            "thread log %d on thread %llu", i, get_current_thread_id());
```
supervacuus, line 351: ditto

src/sentry_core.c
```diff
@@ -312,6 +318,16 @@ sentry_flush(uint64_t timeout)
 int
 sentry_close(void)
 {
+    // Shutdown logs system before locking options to ensure logs are flushed
+    // TODO is this the only way? we got a potential deadlock on the options
+    // otherwise
+    //  (for envelope creation, see sentry__envelope_new)
```
supervacuus: Not sure what this means. `sentry_close()` will be called from `sentry_init()`.

src/sentry_envelope.c
```diff
@@ -401,6 +401,35 @@ sentry__envelope_add_transaction(
...
+        return NULL;
+    }
+
+    sentry_jsonwriter_t *jw = sentry__jsonwriter_new_sb(NULL);
+    if (!jw) {
+        return NULL;
+    }
+
+    sentry__jsonwriter_write_value(jw, logs);
+    item->payload = sentry__jsonwriter_into_string(jw, &item->payload_len);
```
supervacuus, line 424: The return value here can also be `NULL`. Thus, the entire header assignment should be conditional on it not being `NULL`.

include/sentry.h
```diff
@@ -1880,6 +1881,51 @@ SENTRY_EXPERIMENTAL_API void sentry_options_set_traces_sampler(
...
+    sentry_options_t *opts, int enable_logs);
+SENTRY_EXPERIMENTAL_API int sentry_options_get_enable_logs(
+    const sentry_options_t *opts);
+
+typedef enum {
+    SENTRY_LOG_RETURN_SUCCESS = 0,
+    SENTRY_LOG_RETURN_DISCARD = 1,
+    SENTRY_LOG_RETURN_FAILED = 2,
+    SENTRY_LOG_RETURN_DISABLED = 3
+} log_return_value_t;
```
supervacuus, line 1914: I miss an explanation of these return values. In particular, `FAILED` might not be clear. You can use this opportunity to explain that the current implementation minimally blocks the client but is lossy in return.

src/sentry_logs.c
```diff
@@ -0,0 +1,759 @@
...
+    case 'X':
+    case 'o': {
+        unsigned long long int val = va_arg(*args_copy, unsigned long long int);
+        sentry_value_set_by_key(
+            param_obj, "value", sentry_value_new_uint64(val));
+        // TODO update once unsigned 64-bit can be sent
+        sentry_value_set_by_key(
+            param_obj, "type", sentry_value_new_string("string"));
+        break;
+    }
```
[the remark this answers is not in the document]
supervacuus: The `va_arg` part of this message is correct at least where the ABI doesn't save us (on 64-bit ABIs, the var arg slots are already 8-byte wide), and this needs fixing. We can't ignore the length format specifier for `va_arg()` if we want to be portable (on most 32-bit ABIs, it would read two ints at once).

This is independent of the fact that we do not need the format specifier for whatever we store (because we decide to store every parameter in the biggest value type possible).

We need all combinations of length and number targets we can support, give each the correct `va_arg()` type spec, and then store it in our container for that number representation.

However, one could argue that this is not a blocker for an initial release (you could request users to use 64-bit types for parameters on 32-bit systems).

include/sentry.h
```diff
@@ -1880,6 +1881,51 @@ SENTRY_EXPERIMENTAL_API void sentry_options_set_traces_sampler(
...
+SENTRY_EXPERIMENTAL_API log_return_value_t sentry_log_debug(
+    const char *message, ...);
+SENTRY_EXPERIMENTAL_API log_return_value_t sentry_log_info(
+    const char *message, ...);
+SENTRY_EXPERIMENTAL_API log_return_value_t sentry_log_warn(
+    const char *message, ...);
+SENTRY_EXPERIMENTAL_API log_return_value_t sentry_log_error(
+    const char *message, ...);
+SENTRY_EXPERIMENTAL_API log_return_value_t sentry_log_fatal(
+    const char *message, ...);
```
supervacuus, line 1951: I miss an explanation of the format string restrictions that we impose.

src/sentry_logs.c
```diff
@@ -0,0 +1,759 @@
...
+        if (log_idx < QUEUE_LENGTH) {
+            // got a slot, write log to the buffer and unblock flusher
+            active->logs[log_idx] = log;
+            sentry__atomic_fetch_and_add(&active->adding, -1);
+
+            // Check if active buffer is now full and trigger flush. We could
+            // introduce additional watermarks here to trigger the flush earlier
+            // under high contention.
+            if (log_idx == QUEUE_LENGTH - 1) {
+                sentry__cond_wake(&g_logs_state.request_flush);
```
supervacuus, line 186: Please write a comment that we replace this mechanism with a "level-triggered flag" in an upcoming release.

src/sentry_logs.c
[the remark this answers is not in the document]
supervacuus: There should be unit tests that trigger this current scenario at least on the 32-bit builds.

include/sentry.h
```diff
@@ -1890,12 +1890,44 @@ SENTRY_EXPERIMENTAL_API void sentry_options_set_enable_logs(
...
+ * - %f, %F, %e, %E, %g, %G - floating point numbers (treated as double)
+ * - %c - single character
+ * - %s - null-terminated string (null pointers are handled as "(null)")
+ * - %p - pointer value (formatted as hexadecimal string)
+ *
+ * Unsupported format specifiers will consume their corresponding argument
+ * but will be recorded as "(unknown)" in the structured log data.
+ * Length modifiers (h, l, L, z, j, t) are parsed but ignored.
+ *
+ * Because of this, please only use 64-bit types for parameters.
```
supervacuus:
```suggestion
 * Because of this, please only use 64-bit types/casts for your arguments.
```

include/sentry.h
```diff
@@ -1890,12 +1890,44 @@ SENTRY_EXPERIMENTAL_API void sentry_options_set_enable_logs(
 SENTRY_EXPERIMENTAL_API int sentry_options_get_enable_logs(
     const sentry_options_t *opts);
 
+/**
+ * The potential returns of calling any of the sentry_logs_X functions
```
supervacuus:
```suggestion
 * The potential returns of calling any of the sentry_log_X functions
```

include/sentry.h
```diff
@@ -1890,12 +1890,44 @@ SENTRY_EXPERIMENTAL_API void sentry_options_set_enable_logs(
...
 typedef enum {
     SENTRY_LOG_RETURN_SUCCESS = 0,
     SENTRY_LOG_RETURN_DISCARD = 1,
     SENTRY_LOG_RETURN_FAILED = 2,
     SENTRY_LOG_RETURN_DISABLED = 3
 } log_return_value_t;
+
+/**
+ * Structured logging interface. Minimally blocks the client trying to log,
+ * but is therefor lossy when enqueueing a log fails
```
supervacuus: I think this is a classic example of "therefore/therefor" as a conjunctive adverb meaning: "we do lossy, because we don't want to block you" and in that case it is the one with an "e" at the end:
```suggestion
 * but is therefore lossy when enqueueing a log fails
```

supervacuus (approved):
Minor nits, otherwise LGTM. We can also move the `TODO` into a follow-up issue (as described in the issue description for other open topics). But those two items (level-triggered wake + length specifier usage) are not yet in the list of open issues, so I leave this up to you.

src/sentry_envelope.c
```diff
@@ -401,6 +404,35 @@ sentry__envelope_add_transaction(
...
+
```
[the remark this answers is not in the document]
supervacuus: Jup, same issue as raised here: https://github.com/getsentry/sentry-native/pull/1271#discussion_r2344197913

PR #1272, opened 2025-06-12
feat(logs): add sentry log API + send first logs

JoshuaMoelans:
First step towards #1210 (Merging into https://github.com/getsentry/sentry-native/pull/1271)

Adds the initial logs API `sentry_log_X(...)` and constructs envelopes with the appropriate date & attributes. Currently sends these one at a time (for each API call), so next PR will be adding batching for logs.

Example usage:
```c
sentry_log_info("API call to %s completed in %d ms with %f success rate", "/api/products", 2500, 0.95);
```

The 'first logs' visible on [Sentry](https://sentry-sdks.sentry.io/explore/logs/?logsFields=timestamp&logsFields=message&logsSortBys=-timestamp&project=4506178389999616&statsPeriod=14d)

#skip-changelog

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -15,6 +15,8 @@ sentry_target_sources_cwd(sentry
 	sentry_json.h
 	sentry_logger.c
 	sentry_logger.h
+	sentry_logs.c
+	sentry_logs.h
 	sentry_options.c
 	sentry_options.h
 	sentry_os.c
@@ -51,8 +53,6 @@ sentry_target_sources_cwd(sentry
 	transports/sentry_disk_transport.h
 	transports/sentry_function_transport.c
 	unwinder/sentry_unwinder.c
-	sentry_logs.c
-	sentry_logs.h
 )
 
 # generic platform / path / symbolizer
--- a/src/sentry_logs.c
+++ b/src/sentry_logs.c
@@ -145,6 +145,17 @@ skip_precision(const char *fmt_ptr)
     return fmt_ptr;
 }
 
+static const char *
+skip_length(const char *fmt_ptr)
+{
+    while (*fmt_ptr
+        && (*fmt_ptr == 'h' || *fmt_ptr == 'l' || *fmt_ptr == 'L'
+            || *fmt_ptr == 'z' || *fmt_ptr == 'j' || *fmt_ptr == 't')) {
+        fmt_ptr++;
+    }
+    return fmt_ptr;
+}
+
 static void
 populate_message_parameters(
     sentry_value_t attributes, const char *message, va_list args)
@@ -172,13 +183,7 @@ populate_message_parameters(
             fmt_ptr = skip_flags(fmt_ptr);
             fmt_ptr = skip_width(fmt_ptr);
             fmt_ptr = skip_precision(fmt_ptr);
-
-            // Skip length modifiers
-            while (*fmt_ptr
-                && (*fmt_ptr == 'h' || *fmt_ptr == 'l' || *fmt_ptr == 'L'
-                    || *fmt_ptr == 'z' || *fmt_ptr == 'j' || *fmt_ptr == 't')) {
-                fmt_ptr++;
-            }
+            fmt_ptr = skip_length(fmt_ptr);
 
             // Get the conversion specifier
             char conversion = *fmt_ptr;
--- a/src/sentry_value.c
+++ b/src/sentry_value.c
@@ -317,7 +317,7 @@ sentry_value_new_int32(int32_t value)
 sentry_value_t
 sentry_value_new_double(double value)
 {
-    thing_t *thing = sentry_malloc(sizeof(thing_t));
+    thing_t *thing = SENTRY_MAKE(thing_t);
     if (!thing) {
         return sentry_value_new_null();
     }
@@ -333,7 +333,7 @@ sentry_value_new_double(double value)
 sentry_value_t
 sentry_value_new_int64(int64_t value)
 {
-    thing_t *thing = sentry_malloc(sizeof(thing_t));
+    thing_t *thing = SENTRY_MAKE(thing_t);
     if (!thing) {
         return sentry_value_new_null();
     }
@@ -349,7 +349,7 @@ sentry_value_new_int64(int64_t value)
 sentry_value_t
 sentry_value_new_uint64(uint64_t value)
 {
-    thing_t *thing = sentry_malloc(sizeof(thing_t));
+    thing_t *thing = SENTRY_MAKE(thing_t);
     if (!thing) {
         return sentry_value_new_null();
     }
@@ -666,6 +666,9 @@ sentry__value_as_uuid(sentry_value_t value)
 char *
 sentry__value_stringify(sentry_value_t value)
 {
+    // returns empty string if snprintf fails
+    // (returning -1, so casting this to size_t it becomes > the buffer size)
+    // or if the value is too large for the buffer
 #define STRINGIFY_NUMERIC(fmt, value_fn)                                       \
     do {                                                                       \
         char buf[24];                                                          \
--- a/tests/unit/CMakeLists.txt
+++ b/tests/unit/CMakeLists.txt
@@ -28,7 +28,7 @@ add_executable(sentry_test_unit
 	test_fuzzfailures.c
 	test_info.c
 	test_logger.c
-    test_logs.c
+	test_logs.c
 	test_modulefinder.c
 	test_mpack.c
 	test_options.c


examples/example.c
```diff
@@ -359,6 +361,30 @@ main(int argc, char **argv)
 
     sentry_init(options);
 
+    // TODO incorporate into test
+    if (options->enable_logs) {
+        sentry_logger_trace(
```
JoshuaMoelans: rename to `sentry_log_X(...)`

include/sentry.h
```diff
@@ -1750,6 +1750,14 @@ SENTRY_EXPERIMENTAL_API void sentry_options_set_enable_logs(
 SENTRY_EXPERIMENTAL_API int sentry_options_get_enable_logs(
     const sentry_options_t *opts);
 
+// TODO think about API; functions or MACROs?
```
JoshuaMoelans: usually, functions are preferred over Macros (only use them if you have reason for it; for example syntactic convenience (e.g. varargs ending with NULL to specify the end))

src/sentry_logs.c
```diff
@@ -0,0 +1,381 @@
...
+#include <string.h>
+
+#include <stdarg.h>
+#include <stdio.h>
+
+char *
+log_level_as_string(sentry_log_level_t level)
+{
+    switch (level) {
+    case SENTRY_LOG_LEVEL_TRACE:
```
JoshuaMoelans: add TRACE to normal sentry_level_t and use that instead

src/sentry_logs.c
```diff
@@ -0,0 +1,381 @@
...
+    int size = vsnprintf(NULL, 0, message, args_copy) + 1;
+    va_end(args_copy);
+    char *fmt_message = sentry_malloc(size);
+    if (!fmt_message) {
+        return sentry_value_new_null();
+    }
+
+    vsnprintf(fmt_message, size, message, args);
+
+    sentry_value_set_by_key(log, "body", sentry_value_new_string(fmt_message));
```
JoshuaMoelans, line 241: q: Why do we need to evaluate the string format + args into body on the SDK side? Could this not be done during ingestion/at the frontend?

src/sentry_logs.c
```diff
@@ -0,0 +1,381 @@
...
+                // Escaped '%', not a format specifier
+                fmt_ptr++;
+                continue;
+            }
+
+            // Get the conversion specifier
+            char conversion = *fmt_ptr;
+            if (conversion) {
+                char key[64];
+                snprintf(key, sizeof(key), "sentry.message.parameter.%d",
```
JoshuaMoelans, line 192: Look into why other SDKs are using this too (instead of parameter.name); could be useful to track the variable over time across log statements. Probably hard to parse this in Native though.

~~-> insofar no usage of `sentry.message.parameter.X` found where `X` is not an integer. Will have to ask around internally.~~ -> Python does this, see comment below.

src/sentry_logs.c
```diff
@@ -0,0 +1,381 @@
...
+        if (!options->enable_logs)
+            return;
+    }
+    SENTRY_INFOF("Logging level: %i\n", level);
+    vprintf(message, args);
+    // create log from message
+    sentry_value_t log = construct_log(level, message, args);
+
+    // TODO split up the code below for batched log sending
+    //    e.g. could we store logs on the scope?
```
JoshuaMoelans: or just global static; ask around where the other SDKs put this (currently breadcrumbs are scoped, but would they have to be? maybe thread local makes more sense?)

src/sentry_logs.c
```diff
@@ -0,0 +1,381 @@
...
+    sentry_value_t logs = sentry_value_new_object();
+    sentry_value_t logs_list = sentry_value_new_list();
+    sentry_value_append(logs_list, log);
+    sentry_value_set_by_key(logs, "items", logs_list);
+    // sending of the envelope
+    // TODO ensure envelope starts out correctly; we get {dsn:...} as a header
```
JoshuaMoelans: check if we actually need this as a header for events/tx too

src/sentry_logs.c
JoshuaMoelans: like mentioned in the [code comment](https://github.com/getsentry/sentry-native/pull/1272/files#diff-3c65bca63aabeb5b604d2494d7b51a68981c406c55ef0a7a144563f5d7974214R14-R17) above, the problem with replacing it with `sentry_level_t` is that that one has no TRACE level & the `warn` level is specified as `warning` instead (as per the [docs](https://develop.sentry.dev/sdk/data-model/event-payloads/#optional-attributes)). 
Either we keep the separation, or introduce `TRACE` to `sentry_level_t` and conditionally return `warn` instead of `warning` when calling it from the logs side.

src/sentry_logs.c
JoshuaMoelans: seems like it is not required, but recommended https://develop.sentry.dev/sdk/data-model/envelopes/#envelope-headers
we already add it to events (`dsn`)/transactions(`dsn` + `sent_at`), so can't hurt to add the `dsn` here too.

src/sentry_logs.c
JoshuaMoelans: No real reason to put these on the scope, so we can just use a global static variable for storing our `logs` object.

src/sentry_logs.c
supervacuus: I think we answered this during our sync. The SDK knows and supports the format strings, whereas the backend would have to support various formatting, interpolation, and substitution approaches.

src/sentry_logs.c
supervacuus: i think the conditional thing is fine, but would add that one of these uses (probably `warn`) will be deprecated (target whenever 1.x).

examples/example.c
```diff
@@ -359,6 +359,26 @@ main(int argc, char **argv)
 
     sentry_init(options);
 
+    // TODO incorporate into test
```
JoshuaMoelans, line 362: Leaving this in as a TODO for one of the future Logs PRs (after batching is implemented). This PR already provides some Unit tests (which will also need updating after), but making the integration test can wait for the 'full implementation' IMHO.

include/sentry.h
```diff
@@ -370,6 +370,7 @@ SENTRY_API char *sentry_value_to_json(sentry_value_t value);
  * Sentry levels for events and breadcrumbs.
  */
 typedef enum sentry_level_e {
+    SENTRY_LEVEL_TRACE = -2,
```
JoshuaMoelans, line 395: Even though the level is `-2`, when the logger evaluates whether to return early from `sentry__logger_log` in [this snippet](https://github.com/getsentry/sentry-native/blob/master/src/sentry_logger.c#L93-L96), when the log level is set to debug (e.g. with `sentry_options_set_debug()`) we will still output `SENTRY_TRACE(...)` calls.

```c
    if (g_logger.logger_level != SENTRY_LEVEL_DEBUG
        && level < g_logger.logger_level) {
        return;
    }
```

We could update it so the logger defaults to `sentry_logger_t logger = { NULL, NULL, SENTRY_LEVEL_TRACE };` as the 'lowest level' instead, or alternatively we only output trace logger calls when `sentry_options_set_logger_level(options, SENTRY_LEVEL_TRACE);` was explicitly set (even if running with debug?)

src/sentry_logs.c
```diff
@@ -0,0 +1,388 @@
...
+        return "warn";
+    case SENTRY_LEVEL_ERROR:
+        return "error";
+    case SENTRY_LEVEL_FATAL:
+        return "fatal";
+    case SENTRY_LEVEL_INFO:
+    default:
+        return "unknown";
+    }
+}
```
JoshuaMoelans: (wrt https://github.com/getsentry/sentry-native/pull/1272#discussion_r2154485002) we already have a similar `sentry_level_t -> string` [conversion function](https://github.com/getsentry/sentry-native/blob/master/src/sentry_logger.c#L71-L88) specific to the `sentry_logger`, so I think keeping one for the `logs` too seems fine (and then we don't need conditional shenanigans to output `warn` vs `warning`).

include/sentry.h
supervacuus: > Even though the level is `-2`, when the logger evaluates whether to return early from `sentry__logger_log` in [this snippet](https://github.com/getsentry/sentry-native/blob/master/src/sentry_logger.c#L93-L96), when the log level is set to debug (e.g. with `sentry_options_set_debug()`) we will still output `SENTRY_TRACE(...)` calls.

Why does this happen? If we trace-level log, then the incoming `level` would be `-2`, and that should be smaller than `g_logger.logger_level`, which should be set to `-1` if the debug option is enabled, meaning early exit, no?

> We could update it so the logger defaults to `sentry_logger_t logger = { NULL, NULL, SENTRY_LEVEL_TRACE };` as the 'lowest level' instead, or alternatively we only output trace logger calls when `sentry_options_set_logger_level(options, SENTRY_LEVEL_TRACE);` was explicitly set (even if running with debug?)

I would prefer not to have trace as the default debug log level. Trace should be something that someone wants to enable explicitly. These will be dramatically chatty (for instance, I could imagine adding trace-level logs for synchronization primitives or similar low-level things). So, yes, the second option.

JoshuaMoelans:
@sentry review

examples/example.c
```diff
@@ -359,6 +359,26 @@ main(int argc, char **argv)
...
+    if (sentry_options_get_enable_logs(options)) {
+        sentry_log_trace("We log it up  %i percent, %s style\n", 100, "trace");
```
supervacuus: what is the significance of the newline in these logs?

examples/example.c
```diff
@@ -359,6 +359,26 @@ main(int argc, char **argv)
...
+        sentry_log_info(
+            "API call to %s completed in %d ms with %f success rate",
+            "/api/products", 2500, 0.95);
+
+        sentry_log_warn("Processing %d items, found %u errors, pointer: %p",
+            100, 5u, (void *)0x12345678);
+
+        sentry_log_error("Character '%c' is invalid", 'X');
+    }
+
```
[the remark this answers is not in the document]
JoshuaMoelans: This will be part of the full implementation (after batching), hence the ToDo can remain.

src/sentry_logs.c
```diff
@@ -0,0 +1,389 @@
...
+            }
+
+            if (*fmt_ptr == '%') {
+                // Escaped '%', not a format specifier
+                fmt_ptr++;
+                continue;
+            }
+
+            // Get the conversion specifier
+            char conversion = *fmt_ptr;
```
supervacuus, line 189: I think not referencing `fmt_ptr` for another 80 lines is a clear sign that the code in between should be abstracted away. I think you could have a function `construct_param_from_conversion()` that takes the `key`, the `conversion`, and the `args_copy`, and would be called like this:

```C
char conversion = *fmt_ptr;
if (conversion) {
    char key[64];
    snprintf(key, sizeof(key), "sentry.message.parameter.%d", param_index);
    sentry_value_t param_obj = construct_param_from_conversion(key, conversion, args_copy);
    sentry_value_set_by_key(attributes, key, param_obj);
    param_index++;
    fmt_ptr++;
}
```

Consequently, the function would not concern itself with parsing the format string, but only with converting incoming values according to the already identified conversion specifier.

You could then also extract the corresponding `construct_param_[int|pointer|float|...]` function, although I am unsure how much that would benefit you.

src/sentry_logs.c
```diff
@@ -0,0 +1,389 @@
...
+                case 'u':
+                case 'x':
+                case 'X':
+                case 'o': {
+                    unsigned int val = va_arg(args_copy, unsigned int);
+                    sentry_value_set_by_key(param_obj, "value",
+                        sentry_value_new_int32((int32_t)val));
+                    sentry_value_set_by_key(
+                        param_obj, "type", sentry_value_new_string("integer"));
+                    break;
```
supervacuus: We must be cautious here: if you skip the length specifiers, a 64-bit value could be passed in. Even if the value is a 32-bit unsigned, there is no way the backend could know that this is an `int32` containing a `uint32` without an explicit protocol.

Essentially, if we accept unsigned integers of 32-bit, we'd have to store them in a double.

If we accept 64-bit unsigned integers, it must be clear that JSON will overflow numbers above 2^53-1

supervacuus: By the way, introducing `sentry_value_t` (a list of kv-objects) as a parameter to format strings would also introduce some normalization (or put the burden of providing parameters that map to the sentry payloads on the user, instead of the code having to guess) in terms of the data that users can pass in here.

src/sentry_logs.c
```diff
@@ -0,0 +1,389 @@
...
+
+                switch (conversion) {
+                case 'd':
+                case 'i': {
+                    int val = va_arg(args_copy, int);
+                    sentry_value_set_by_key(
+                        param_obj, "value", sentry_value_new_int32(val));
+                    sentry_value_set_by_key(
+                        param_obj, "type", sentry_value_new_string("integer"));
+                    break;
```
supervacuus: Same as below... since you are skipping length, this could be a 64-bit signed integer.

src/sentry_logs.c
```diff
@@ -0,0 +1,389 @@
...
+                && (*fmt_ptr == 'h' || *fmt_ptr == 'l' || *fmt_ptr == 'L'
+                    || *fmt_ptr == 'z' || *fmt_ptr == 'j' || *fmt_ptr == 't')) {
+                fmt_ptr++;
+            }
+
+            if (*fmt_ptr == '%') {
+                // Escaped '%', not a format specifier
+                fmt_ptr++;
+                continue;
+            }
```
supervacuus, line 181: I would skip this case first since the escape needs to be `%%` anyway (i.e., nothing can be in-between) and because it is the only skip that continues rather than progresses.

src/sentry_logs.c
```diff
@@ -0,0 +1,389 @@
...
+                    fmt_ptr++;
+                }
+            }
+
+            // Skip length modifiers
+            while (*fmt_ptr
+                && (*fmt_ptr == 'h' || *fmt_ptr == 'l' || *fmt_ptr == 'L'
+                    || *fmt_ptr == 'z' || *fmt_ptr == 'j' || *fmt_ptr == 't')) {
+                fmt_ptr++;
+            }
```
supervacuus: I think it is fair to extract each of these into either a macro or a function and call them:

* `skip_flags()`
* `skip_width()`
* `skip_precision()`
* `skip_length()` (or as mentioned in other comments, maybe even `extract_length()` 😅)

called as:

```C
fmt_ptr = skip_flags(fmt_ptr);
```

Which would make the entire function easier to read and test particular parts.

What I currently miss in this code is how we handle malformed format strings, but we don't need to address this issue in this PR.

supervacuus (changes requested):
That is a great start! I might give this another go tomorrow, but there are already a couple of points to address.

I did not explicitly mention the alternative interface for the format parameters (except for a concrete point where it overlaps with this PR, at least in terms of design effect)

include/sentry.h
JoshuaMoelans: I think I added the first clause in the if-statement as a `if debug, log everything` (even though before adding the `trace` level, this would result in the same behavior as not having that check). Now I think just removing the first clause, and only checking whether the level is lower than the set logger_level makes sense.

src/sentry_logs.c
JoshuaMoelans: The Python SDK seems to be be doing this for f-strings, but in some integrations it's not always showing up as `parameter.X` (see https://github.com/getsentry/sentry-python/issues/4525).

As mentioned in  https://github.com/getsentry/sentry-native/pull/1272#pullrequestreview-2954506464, providing a separate interface in Native that allows for these kinds of named parameters could be a potential improvement that will be part of the next Logs PR (after full initial implementation is finished, including batching).

examples/example.c
JoshuaMoelans: unfortunate remainder of the beginning, when all the API did was print it back to stdout 🙈 will clean this up, thanks :D

src/CMakeLists.txt
```diff
@@ -51,6 +51,8 @@ sentry_target_sources_cwd(sentry
 	transports/sentry_disk_transport.h
 	transports/sentry_function_transport.c
 	unwinder/sentry_unwinder.c
+	sentry_logs.c
+	sentry_logs.h
```
supervacuus: please keep these two in the sorted list above

src/sentry_value.c
```diff
@@ -324,6 +330,38 @@ sentry_value_new_double(double value)
     return rv;
 }
 
+sentry_value_t
+sentry_value_new_int64(int64_t value)
+{
+    thing_t *thing = sentry_malloc(sizeof(thing_t));
```
supervacuus:
```suggestion
    thing_t *thing = SENTRY_MAKE(thing_t);
```

src/sentry_value.c
```diff
@@ -324,6 +330,38 @@ sentry_value_new_double(double value)
...
+
+    sentry_value_t rv;
+    rv._bits = (uint64_t)(size_t)thing;
+    return rv;
+}
+
+sentry_value_t
+sentry_value_new_uint64(uint64_t value)
+{
+    thing_t *thing = sentry_malloc(sizeof(thing_t));
```
supervacuus:
```suggestion
    thing_t *thing = SENTRY_MAKE(thing_t);
```

src/sentry_value.c
```diff
@@ -624,6 +666,18 @@ sentry__value_as_uuid(sentry_value_t value)
...
 {
+#define STRINGIFY_NUMERIC(fmt, value_fn)                                       \
+    do {                                                                       \
+        char buf[24];                                                          \
+        size_t written                                                         \
+            = (size_t)sentry__snprintf_c(buf, sizeof(buf), fmt, value_fn);     \
+        if (written >= sizeof(buf)) {                                          \
+            return sentry__string_clone("");                                   \
+        }                                                                      \
+        buf[written] = '\0';                                                   \
```
supervacuus, line 680: `STRINGIFY_NUMERIC` could use a comment (above the `#define`) that signals that we willingly conflate the two error cases of `snprintf` (return value is an `int` because written can be < 0, which signifies an encoding error) into an overflowing `size_t`, because we handle them the same way (return an empty string).

tests/unit/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ add_executable(sentry_test_unit
 	test_fuzzfailures.c
 	test_info.c
 	test_logger.c
+    test_logs.c
```
supervacuus: Please indent with a tab.

src/sentry_logs.c
```diff
@@ -0,0 +1,416 @@
...
+            fmt_ptr = skip_flags(fmt_ptr);
+            fmt_ptr = skip_width(fmt_ptr);
+            fmt_ptr = skip_precision(fmt_ptr);
+
+            // Skip length modifiers
+            while (*fmt_ptr
+                && (*fmt_ptr == 'h' || *fmt_ptr == 'l' || *fmt_ptr == 'L'
+                    || *fmt_ptr == 'z' || *fmt_ptr == 'j' || *fmt_ptr == 't')) {
+                fmt_ptr++;
+            }
```
supervacuus: Since we now always store with the largest value types available, and there is no need to extract length modifiers, this could also be extracted to a `skip_length` function, right?

supervacuus (approved):
LGTM, a couple of minor things.

PR #1273, opened 2025-06-12
docs: sync return values of AIX `dladdr` with implementation

supervacuus:
#skip-changelog

@JoshuaMoelans followup to #1234


JoshuaMoelans (approved):
nice 🚀 😁

PR #1274, opened 2025-06-12
ci: drop windows-2019 runner images

supervacuus:
We currently use the `windows-2019` runner image for the "old" Windows configuration (which also builds the 32-bit binary). This image is deprecated: https://github.com/actions/runner-images/issues/12045.

All other Windows configurations use `windows-latest`, which currently is `windows-2022` but will be upgraded to `windows-2025`, so let's pin this one on 2022.


JoshuaMoelans (approved):
lgtm!

PR #1275, opened 2025-06-13
feat: support attaching bytes

jpnurmi:
~~Q: Should it be called `sentry_attach_buffer` or `sentry_attach_data`?
A: `sentry_attach_bytes`~~

TODO:
- [x] ~~Crashpad: IPC~~
  - ~~https://github.com/getsentry/crashpad/pull/127~~ (write files instead due to IPC limits)
- [x] Crashpad: filenames must be unique
  - https://github.com/getsentry/crashpad/pull/129
- [x] Integration tests
- [x] Docs
  - https://github.com/getsentry/sentry-docs/pull/14077
  - Consider copying [Size Limits](https://docs.sentry.io/platforms/native/guides/minidumps/#size-limits) and linking from the API docs?

### DIFF
--- a/src/sentry_attachment.c
+++ b/src/sentry_attachment.c
@@ -91,7 +91,7 @@ sentry__attachments_free(sentry_attachment_t *attachments)
 }
 
 /**
- * Compares attachments for equality to avoid adding duplicates and for removal.
+ * Compares attachments for equality to avoid adding duplicates.
  *
  * File attachments are considered equal if the paths are equal. Byte buffer
  * attachments are only considered equal to themselves - they are not required
@@ -156,7 +156,7 @@ sentry__attachments_remove(
     sentry_attachment_t **next_ptr = attachments_ptr;
 
     for (sentry_attachment_t *it = *attachments_ptr; it; it = it->next) {
-        if (attachment_eq(it, attachment)) {
+        if (it == attachment) {
             *next_ptr = it->next;
             sentry__attachment_free(it);
             return;


supervacuus:
If we want to stay consistent with other SDKs, `bytes` is fine since we also use it for the `UUID` API in a similar manner. `data` is somewhat unspecific. We use `data` in the context of `user_data` for callbacks (but only as parameter specs) and `data` as the field for transactions and spans (as part of the function names).

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -662,26 +663,93 @@ crashpad_backend_prune_database(sentry_backend_t *backend)
...
+                     "\" to \"%" SENTRY_PATH_PRI "\"",
+            sentry__path_filename(attachment->path),
+            path.BaseName().value().c_str());
+    }
+
+    sentry__path_free(attachment->path);
+    attachment->path = sentry__path_new(path.value().c_str());
+
+    return !crashpad::IsRegularFile(path);
+}
```
jpnurmi: @supervacuus Doing this in C is a bit painful due to `char` vs. `wchar`. My first idea was to help from Crashpad (https://github.com/getsentry/crashpad/pull/129), but then I realized Crashpad's file utils are also accessible here. Is it acceptable to use them like this? If not, `std::filesystem::path` could also be an option.

supervacuus: The `crashpad` backend implementation can utilize all that the C++17 standard library provides (which, yes, includes `std::filesystem`), as well as the entire `crashpad` interface.

jpnurmi: I switched to `std::filesystem` because the code looked almost nicer, and that way it won't be affected by potential future changes in `crashpad`.

include/sentry.h
```diff
@@ -1813,6 +1813,29 @@ SENTRY_API sentry_attachment_t *sentry_scope_attach_file(
...
+ * `sentry_scope_attach_bytesw` instead.
+ *
+ * The returned `sentry_attachment_t` is owned by the SDK and will remain valid
+ * until the attachment is removed with `sentry_remove_attachment` or
+ * `sentry_close` is called.
+ *
+ * See the NOTE on attachments above for restrictions of this API.
+ */
+SENTRY_API sentry_attachment_t *sentry_attach_bytes(
+    const char *buf, size_t buf_len, const char *filename);
```
jpnurmi, line 1845: `char *` may give a false impression that it's treated as a string. what if it was `(u)int8_t *` instead?

include/sentry.h
```diff
@@ -1813,6 +1813,29 @@ SENTRY_API sentry_attachment_t *sentry_scope_attach_file(
 SENTRY_API sentry_attachment_t *sentry_scope_attach_file_n(
     sentry_scope_t *scope, const char *path, size_t path_len);
 
+/**
+ * Attaches bytes to be sent along with events.
+ *
+ * `filename` is assumed to be in platform-specific filesystem path encoding.
+ * API Users on windows are encouraged to use `sentry_attach_bytesw` or
+ * `sentry_scope_attach_bytesw` instead.
```
jpnurmi, line 1825: there could be a note about that crashpad requires unique filenames, so the attachment may end up renamed to `basename-N.ext` if necessary.

src/CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ sentry_target_sources_cwd(sentry
 	sentry_tracing.c
 	sentry_tracing.h
 	path/sentry_path.c
+	path/sentry_path.cpp
```
jpnurmi: The VS and Unix makefile generators have no problems. Is there a risk of conflicts with other generators?

include/sentry.h
supervacuus: I think `char*` buffers are established enough in the C/C++ world as byte buffers. I see no need to introduce the `stdint.h` header for this.

include/sentry.h
supervacuus: Yes, this makes sense.

src/CMakeLists.txt
supervacuus: I prefer to keep C++ utilities within their respective C++ modules (which currently are strictly limited to the `crashpad` and `breakpad` backends) and avoid introducing new ones.

I understand that this function has potentially general utility, but users who do not use `crashpad` should not have to pay the price of linking with the C++ standard library (which this addition would force them to, and for no reason). 

If this makes sense as a utility for the entire code base, not just `crashpad`, we should rewrite it in terms of the C standard library.

src/backends/sentry_backend_breakpad.cpp
```diff
@@ -166,13 +167,12 @@ breakpad_backend_callback(const google_breakpad::MinidumpDescriptor &descriptor,
...
 
             if (options->attach_screenshot) {
-                sentry_path_t *screenshot_path
-                    = sentry__screenshot_get_path(options);
-                if (sentry__screenshot_capture(screenshot_path)) {
-                    sentry__envelope_add_attachment(
-                        envelope, screenshot_path, ATTACHMENT, nullptr);
+                sentry_attachment_t *screenshot = sentry__attachment_from_path(
+                    sentry__screenshot_get_path(options));
+                if (sentry__screenshot_capture(screenshot->path)) {
```
supervacuus: Please check `screenshot` for `NULL`.

src/path/sentry_path.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <filesystem>
+#include <string>
+
+namespace fs = std::filesystem;
+
+sentry_path_t *
+sentry__path_unique(sentry_path_t *path)
+{
+    fs::path unique(path->path);
```
supervacuus: We should check that `path->path` is not `NULL` here. Most implementations only debug-assert that the source is not `NULL`.

src/sentry_attachment.c
```diff
@@ -91,22 +149,49 @@ sentry__attachments_remove(
...
+        return NULL;
+    }
+
+    sentry_attachment_t *clone = SENTRY_MAKE(sentry_attachment_t);
+    if (!clone) {
+        return NULL;
+    }
+    memset(clone, 0, sizeof(sentry_attachment_t));
+
+    clone->path = sentry__path_clone(attachment->path);
```
supervacuus, line 182: There should be an error path for `clone->path == NULL` (like the one for `clone->buf` inside the if, maybe a `goto` error/fail).

src/sentry_attachment.c
```diff
@@ -43,33 +86,38 @@ sentry__attachments_free(sentry_attachment_t *attachments)
...
     }
 }
 
+static bool
+attachment_eq(const sentry_attachment_t *a, const sentry_attachment_t *b)
+{
+    if (a == b) {
+        return true;
+    }
+    if (!a || !b || a->buf || b->buf || a->type != b->type) {
```
supervacuus, line 106: Let's highlight here once more (as a comment) that buffer attachments are ignored because the linked list only stores path attachments. I know this is only a `static` helper, but its name suggests a more generic use.

Tbh, I still don't understand why the buffer attachments aren't stored in that list, since the scope functions attach buffer attachments.

src/CMakeLists.txt
supervacuus: Just to be clear (because i saw you reopening the crashpad PR) i meant restricting it to the crashpad backend module. Not to the git submodule. 

It is fine to have it here, just not in the global modules but behind the SENTRY_BACKEND=crashpad build guard (and this is easily achieved by making this a static helper in the backend module with exception for the unit test).

jpnurmi: New day - new proposal :sweat_smile: 

- organize buffer attachments in `sentry_backend_crashpad` so that they cannot conflict  with each other: `<run>/<uuid>/attachment.txt`
- let crashpad internally fix up the names if it needs to when it creates the flat directory structure for crash report attachments: `attachment.txt`, `attachment-1.txt`, ...

How do you like this approach?

src/sentry_attachment.c
jpnurmi: All attachments are stored in the same list. For byte attachments, we cannot use the filename alone for comparison because it's just the display name in Sentry. We don't want to memcmp the buffer either, so byte attachments are only considered equal if the pointer is equal.

supervacuus: Okay, then that should be documented in the private implementation (i don't think that this leads to any unclear effects in the public interface).

supervacuus: I think it would also have been okay to use the filename as a unique identifier for byte buffers and leave uniqueness concerns entirely up to the user. 

Since uniquness is user experience relevant, it doesn't really matter that we allow non-unique names in the envelope, users will have no idea which is which. Even with auto-numbering it might be hard to identify.

This is not a huge topic for me, but since you are iterating again, I just wanted to provide that step-back perspective.

jpnurmi: Yeah, I probably spent too much time on this... But hey, I quite like the latest iteration that keeps `sentry_backend_crashpad` pretty simple, and hides/delays the filename fixup until crashpad needs it while building the crash report.

The reason I ended up going this route was because duplicate filenames worked fine with the other backends, whereas with Crashpad it would result in somewhat obscure errors from the handler process that would be hard to notice, especially on Windows.

```
[54558:54558:20250620,080243.880352:ERROR file_io_posix.cc:153] open /home/jpnurmi/Projects/sentry/sentry-playground/build/.sentry-native/attachments/607f9f4c-32cc-48c2-9990-b097157276a8/test.tar.gz: File exists (17)
[54558:54558:20250620,080243.880363:ERROR crash_report_exception_handler.cc:286] attachment test.tar.gz couldn't be created, skipping
[54558:54558:20250620,080243.880374:ERROR file_io_posix.cc:153] open /home/jpnurmi/Projects/sentry/sentry-playground/build/.sentry-native/attachments/607f9f4c-32cc-48c2-9990-b097157276a8/test.tar.gz: File exists (17)
[54558:54558:20250620,080243.880376:ERROR crash_report_exception_handler.cc:286] attachment test.tar.gz couldn't be created, skipping
```

jpnurmi: I added a clarifying comment in https://github.com/getsentry/sentry-native/pull/1275/commits/070d933e7d72e7dae8d85e6b4834ac4d3f3921ed

src/sentry_attachment.c
```diff
@@ -43,33 +86,45 @@ sentry__attachments_free(sentry_attachment_t *attachments)
...
+        sentry__attachment_free(attachment);
     }
 }
 
+/**
+ * Compares attachments for equality to avoid adding duplicates and for removal.
+ *
+ * File attachments are considered equal if the paths are equal. Byte buffer
+ * attachments are only considered equal to themselves - they are not required
+ * to have unique filenames, and the bytes are not compared.
```
supervacuus: Upon re-reading this comment, the behavior still warrants a note in the public header documentation.

It should be noted that

* Neither the `filename` nor the `bytes` parameters' content is part of the identity for bytes-based attachments (whereas with file-based attachments, the `filename` is the identity).
* We nonetheless recommend using unique filenames (even if the public API does not enforce them), because they will be easier to differentiate in the Web UI, and (segue to) `crashpad` also requires unique filenames for the multipart upload (continue with crashpad note).

jpnurmi: Now that I'm re-reading it again myself, I'm not sure why I decided to use `attachment_eq` even for removal. The pointer comparison should be enough for removal and not cause any surprises.

src/sentry_attachment.c
```diff
@@ -156,7 +156,7 @@ sentry__attachments_remove(
     sentry_attachment_t **next_ptr = attachments_ptr;
 
     for (sentry_attachment_t *it = *attachments_ptr; it; it = it->next) {
-        if (attachment_eq(it, attachment)) {
+        if (it == attachment) {
```
supervacuus, line 159: Yeah, I just wanted to add that for removal, it is no longer helpful to fall back on path comparisons, since for file-based attachments, there can only be one handle with the same filename inside the list.

PR #1276, opened 2025-06-16
feat: add sentry_attachment_set_content_type()

jpnurmi:
A small follow-up to #1266 to support specifying the content type for attachments, as needed for [sentry-unreal](https://github.com/getsentry/sentry-unreal/blob/main/plugin-dev/Source/Sentry/Public/SentryAttachment.h).

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -1833,6 +1833,9 @@ SENTRY_API sentry_attachment_t *sentry_scope_attach_filew_n(
     sentry_scope_t *scope, const wchar_t *path, size_t path_len);
 #endif
 
+SENTRY_API void sentry_attachment_set_content_type(
+    sentry_attachment_t *attachment, const char *content_type);
+
 /* -- Session APIs -- */
 
 typedef enum {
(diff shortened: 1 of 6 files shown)


include/sentry.h
```diff
@@ -1833,6 +1833,9 @@ SENTRY_API sentry_attachment_t *sentry_scope_attach_filew_n(
     sentry_scope_t *scope, const wchar_t *path, size_t path_len);
 #endif
 
+SENTRY_API void sentry_attachment_set_content_type(
```
supervacuus, line 1836: I wonder if someone will ever need the `_n` function for the content type, but let's add it for consistency.

src/sentry_attachment.c
```diff
@@ -1,6 +1,22 @@
...
+    sentry_attachment_t *attachment, const char *content_type)
+{
+    if (!attachment) {
+        return;
+    }
+
+    if (attachment->content_type_owned) {
+        sentry_free((void *)attachment->content_type);
+    }
+    attachment->content_type = sentry__string_clone(content_type);
```
supervacuus: I'm not 100% sure about this, and it's not super crucial since it's not part of the contract, but it might be more sensible to just clone the content type in `sentry__attachments_add` as well and not have a special case for ownership.

PR #1279, opened 2025-06-17
fix: compilation error if SENTRY_TEST_PATH_PREFIX is defined

vaind:
This got broken accidently in #1234 but there is no code path that matches in this repo so I've only found this out when trying to integrate in a downstream SDK. Should we add a test python script or sth?

#skip-changelog


supervacuus:
Ouch, I wrongly assumed that this conditional path was somewhere stressed here. Yes, we should add a dummy test to include that path. While missing the macro-continuation backslashes was an accident, the removal of the last semicolon was intentional.

tests/test_unit.py
```diff
@@ -20,3 +20,13 @@ def test_unit_transport(cmake, unittest):
...
+
+
+def test_unit_with_test_path(cmake, unittest):
+    cwd = cmake(
+        ["sentry_test_unit"],
+        {"SENTRY_BACKEND": "none", "SENTRY_TRANSPORT": "none"},
+        cflags=["-DSENTRY_TEST_PATH_PREFIX=./"],
+    )
+    env = dict(os.environ)
+    run(cwd, "sentry_test_unit", ["--no-summary", unittest], check=True, env=env)
```
vaind, line 32: @supervacuus do you think this makes sense or it's better to add another job/step to CI?

supervacuus: No, this is fine 👍. This should only build to surface exactly what led you to create this PR.

tests/unit/sentry_testsupport.h
```diff
@@ -73,9 +73,9 @@
 #ifdef SENTRY_TEST_PATH_PREFIX
 #    define SENTRY_TEST_OPTIONS_NEW(Varname)                                   \
         sentry_options_t *Varname = sentry_options_new();                      \
-        TEST_ASSERT(!!Varname);
-sentry_options_set_database_path(
-    Varname, SENTRY_TEST_PATH_PREFIX ".sentry-native")
+        TEST_ASSERT(!!Varname);                                                \
+        sentry_options_set_database_path(                                      \
+            Varname, SENTRY_TEST_PATH_PREFIX ".sentry-native");
```
supervacuus: Please remove the trailing semicolon to ensure consistent call-sites.

vaind:
```suggestion
            Varname, SENTRY_TEST_PATH_PREFIX ".sentry-native")
```

PR #1287, opened 2025-06-25
meta: identify Xbox as a separate SDK name

### LINKED ISSUE #1259: Report Xbox in SDK name
Similar to Android: https://github.com/getsentry/sentry-native/blob/6ca83891fdea3f141df66eb6a2f148ef465463c2/include/sentry.h#L28

Is it possible when building for GDK to report instead `sentry.native.xbox` ? 

It's easier to debug stuff based on sdk name then having to parse the envelope to get the context.os etc

supervacuus:
fixes #1259

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -33,19 +33,6 @@
 extern "C" {
 #endif
 
-/* SDK Version */
-#ifndef SENTRY_SDK_NAME
-#    if defined(__ANDROID__)
-#        define SENTRY_SDK_NAME "sentry.native.android"
-#    elif defined(_GAMING_XBOX_SCARLETT)
-#        define SENTRY_SDK_NAME "sentry.native.xbox"
-#    else
-#        define SENTRY_SDK_NAME "sentry.native"
-#    endif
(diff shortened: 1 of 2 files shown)


include/sentry.h
```diff
@@ -35,8 +35,10 @@ extern "C" {
 
 /* SDK Version */
 #ifndef SENTRY_SDK_NAME
-#    ifdef __ANDROID__
+#    if defined(__ANDROID__)
 #        define SENTRY_SDK_NAME "sentry.native.android"
+#    elif defined(_GAMING_XBOX_SCARLETT)
+#        define SENTRY_SDK_NAME "sentry.native.xbox"
```
vaind: Unrelated to this PR but maybe we could move the SDK name+version+user-agent defines after platform defines and use these to decide what to set.

vaind (approved):

include/sentry.h
supervacuus: It is related. Done.

vaind (approved):

Issue #1289, opened 2025-06-27
Feature request: allow clearing attachments in the global scope


jpnurmi:
TLDR: Proposal to add `sentry_clear_attachments()` to help managing attachments in `sentry-unreal` and `sentry-godot`.

_Originally posted by @tustanivsky in https://github.com/getsentry/sentry-unreal/pull/982#discussion_r2172017274_

> Given the discrepancies between Native, Android, and Cocoa in terms of attachment API support it looks like we have the following options:
>
> 1. Add `clearAttachments` method to Native which is already available for Android/Cocoa and expose that for global scope instead of giving user the ability to remove individual attachments
> 2. Add `removeAttachment` to Android/Cocoa to enable removing individual attachments on mobile
> 3. Combine 1 and 2 to cover all possible use cases
> 4. Leave only `addAttachment` for global scope without the ability to remove it
> 5. Leave things as-is and document the current limitations

PR #1290, opened 2025-06-30
feat: add `sentry_clear_attachments()`

jpnurmi:
This PR complements the newly introduced attachments API with a function to allow clearing all previously added attachments in the global scope.

Close: #1289

### DIFF
--- a/src/sentry_core.c
+++ b/src/sentry_core.c
@@ -1475,8 +1475,8 @@ sentry_attach_bytes_n(
 void
 sentry_clear_attachments(void)
 {
-    SENTRY_WITH_SCOPE_MUT (scope) {
-        SENTRY_WITH_OPTIONS (options) {
+    SENTRY_WITH_OPTIONS (options) {
+        SENTRY_WITH_SCOPE_MUT (scope) {
             if (options->backend && options->backend->remove_attachment_func) {
                 for (sentry_attachment_t *it = scope->attachments; it;
                     it = it->next) {
@@ -1484,9 +1484,9 @@ sentry_clear_attachments(void)
                         options->backend, it);
                 }
             }
+            sentry__attachments_free(scope->attachments);
+            scope->attachments = NULL;
         }
-        sentry__attachments_free(scope->attachments);
-        scope->attachments = NULL;
     }
 }
 


src/sentry_core.c
```diff
@@ -1472,6 +1472,24 @@ sentry_attach_bytes_n(
         buf, buf_len, sentry__path_from_str_n(filename, filename_len)));
 }
 
+void
+sentry_clear_attachments(void)
+{
+    SENTRY_WITH_SCOPE_MUT (scope) {
+        SENTRY_WITH_OPTIONS (options) {
+            if (options->backend && options->backend->remove_attachment_func) {
```
jpnurmi, line 1480: consider a separate `sentry_backend_t::clear_attachments_func`
- to avoid nested locking of options and scope
- to avoid several IPC messages

jpnurmi: Even with a separate clear-function, the backend still needs access to scope->attachments to be able to clean up the temporary unique files for buffer attachments. :(

jpnurmi: An integration test revealed that clearing all attachments in Crashpad was a bit too effective, as it cleared `__sentry-event` and `__sentry-breadcrumbN` too. I switched it back to removing explicitly added Crashpad attachments one by one to avoid making Crashpad aware of "fixed" attachments.

src/sentry_core.c
```diff
@@ -1478,8 +1478,11 @@ sentry_clear_attachments(void)
     SENTRY_WITH_SCOPE_MUT (scope) {
         SENTRY_WITH_OPTIONS (options) {
```
supervacuus: _Referencing the nested locking assumption in the other comment_: a `SENTRY_WITH_OPTIONS` scope-macro only very shortly acquires a lock, and then only operates on the premise of an incremented ref-count (i.e., options won't be freed) during the scope, whereas `SENTRY_WITH_SCOPE[_MUT]` holds the lock for the entire lexical scope.

Nonetheless, we should change the order here and acquire the options first, and then the scope to have a consistent lock order.

supervacuus (approved):
LGTM, except for lock order.

supervacuus (approved):

PR #1292, opened 2025-06-30
ci: update kcov

jpnurmi:
Bump kcov to an un unreleased revision that contains both:
- https://github.com/SimonKagstrom/kcov/pull/482
- https://github.com/SimonKagstrom/kcov/pull/483

Linux (clang 19 + kcov):
| Before | After |
|---|---|
| [47m 18s](https://github.com/getsentry/sentry-native/actions/runs/15874686952/job/44759360055#logs) | [8m 12s](https://github.com/getsentry/sentry-native/actions/runs/15971824656/job/45044951744#logs) |

<details><summary>old details</summary>

When running the very first sentry-native unit test `assert_sdk_name`:
```sh
perf record -g -- kcov --include-path=/path/to/sentry-native/src /tmp/pytest-of-jpnurmi/pytest-20/cmake0/coverage ./sentry_test_unit --no-summary assert_sdk_name
```
It takes roughly 5.5 seconds, and 92% of the time is spent in [`get_real_path()`](https://github.com/SimonKagstrom/kcov/blob/ee728054aaf8942393b60973c166741761324126/src/utils.cc#L576-L596).

image

For that single test case alone, kcov does 1792620 repetitive failed and uncached lookups. Only 2214 of the lookups are unique. I'm not sure if it's safe to cache the failed lookups, but the purpose of this experiment is to see what kind of impact it would have on the sentry-native CI if the bottleneck was fixed in kcov in one way or another...
</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,9 +180,10 @@ jobs:
           sudo apt install cmake clang-19 llvm g++-12 valgrind zlib1g-dev libcurl4-openssl-dev
           # Install kcov from source
           sudo apt-get install binutils-dev libssl-dev libelf-dev libstdc++-12-dev libdw-dev libiberty-dev
-          git clone https://github.com/jpnurmi/kcov.git
+          git clone https://github.com/SimonKagstrom/kcov.git
           cd kcov
-          git checkout v43-opt # pin to a version without the current coveralls git integration
+          # pin to a known good version with the coveralls git integration and performance bottlenecks fixed
+          git checkout 8afe9f29c58ef575877664c7ba209328233b70cc
           mkdir build
           cd build
           cmake ..
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,9 +180,9 @@ jobs:
           sudo apt install cmake clang-19 llvm g++-12 valgrind zlib1g-dev libcurl4-openssl-dev
           # Install kcov from source
           sudo apt-get install binutils-dev libssl-dev libelf-dev libstdc++-12-dev libdw-dev libiberty-dev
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -180,9 +180,9 @@ jobs:
           sudo apt install cmake clang-19 llvm g++-12 valgrind zlib1g-dev libcurl4-openssl-dev
           # Install kcov from source
           sudo apt-get install binutils-dev libssl-dev libelf-dev libstdc++-12-dev libdw-dev libiberty-dev
-          git clone https://github.com/SimonKagstrom/kcov.git
+          git clone https://github.com/jpnurmi/kcov.git
           cd kcov
-          git checkout v43 # pin to a version without the current coveralls git integration
+          git checkout v43-opt # pin to a version without the current coveralls git integration
```
supervacuus: While I prefer to use pinning here, and no new version with a fix has been tagged or released yet, I already fixed the coveralls git integration upstream: https://github.com/SimonKagstrom/kcov/pull/482

jpnurmi:
Yep, this speed-up looks so nice that I'll open a PR to see what the maintainer thinks.

jpnurmi:
> To be clear: you're telling me kcov calls `realpath` two million times for paths that it failed to resolve solely during `assert_sdk_name`? Besides the potential for caching of failed resolution, it would be fascinating to understand what happens in this case. I could imagine this to be a GitHub runner filesystem pathology. I have a hard time understanding why so many calls to `realpath` are even necessary (independent of what is happening on the runner image, or that recurring lookups could be cached).

Looks like it repeatedly resolves the real path when processing line by line. Here's a better capture with debug symbols

image

PR #1294, opened 2025-07-01
feat: xbox one support

bruno-garcia:
Resolves: https://github.com/getsentry/sentry-xbox/issues/9

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -36,7 +36,7 @@ extern "C" {
 /* common platform detection */
 #ifdef _WIN32
 #    define SENTRY_PLATFORM_WINDOWS
-#    ifdef _GAMING_XBOX_SCARLETT
+#    ifdef _GAMING_XBOX_SCARLETT || _GAMING_XBOX_XBOXONE
 #        define SENTRY_PLATFORM_XBOX_SCARLETT
 #    endif
 #elif defined(__APPLE__)
--- a/vendor/acutest.h
+++ b/vendor/acutest.h
@@ -487,7 +487,7 @@ test_print_in_color__(int color, const char* fmt, ...)
         printf("\033[0m");
         return n;
     }
-#elif defined ACUTEST_WIN__ && !defined _GAMING_XBOX_SCARLETT
+#elif defined ACUTEST_WIN__ && (!defined _GAMING_XBOX_SCARLETT || !defined _GAMING_XBOX_XBOXONE)
     {
         HANDLE h;
         CONSOLE_SCREEN_BUFFER_INFO info;
--- a/toolchains/xbox/CMakeGDKXboxOne.cmake
+++ b/toolchains/xbox/CMakeGDKXboxOne.cmake
@@ -1,0 +1,231 @@
+# Adapted by Sentry from:
+# https://raw.githubusercontent.com/microsoft/Xbox-GDK-Samples/710f4bd9095d3796d07505249a7b383857e8a23f/Samples/Tools/CMakeExample/CMake/CMakeGDKXboxOne.cmake
+#
+# CMakeGDKXboxOne.cmake : CMake definitions for Microsoft GDK targeting Xbox One/Xbox One S/Xbox One X
+#
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+mark_as_advanced(CMAKE_TOOLCHAIN_FILE)
+
+if(_GDK_XBOX_ONE_TOOLCHAIN_)
+    return()
+endif()
+
+set(XBOX_CONSOLE_TARGET "xboxone" CACHE STRING "")
+
+include("${CMAKE_CURRENT_LIST_DIR}/DetectGDK.cmake")
+
+message("XdkEditionTarget = ${XdkEditionTarget}")
+
+set(CMAKE_TRY_COMPILE_PLATFORM_VARIABLES GDK_VERSION BUILD_USING_BWOI)
+
+#--- Windows SDK
+include("${CMAKE_CURRENT_LIST_DIR}/DetectWindowsSDK.cmake")
+
+set(CMAKE_SYSTEM_NAME WINDOWS)
+set(CMAKE_SYSTEM_VERSION ${WINDOWS_SDK_VER})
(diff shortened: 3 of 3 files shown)


toolchains/xbox/gxdk_xs_toolchain.cmake
```diff
@@ -21,8 +21,11 @@ set(CMAKE_TRY_COMPILE_PLATFORM_VARIABLES GDK_VERSION)
 set(CMAKE_SYSTEM_NAME WINDOWS)
 set(CMAKE_SYSTEM_VERSION ${GDK_VERSION})
 
-set(CMAKE_GENERATOR_PLATFORM "Gaming.Xbox.Scarlett.x64" CACHE STRING "" FORCE)
-set(CMAKE_VS_PLATFORM_NAME "Gaming.Xbox.Scarlett.x64" CACHE STRING "" FORCE)
+# set(CMAKE_GENERATOR_PLATFORM "Gaming.Xbox.Scarlett.x64" CACHE STRING "" FORCE)
```
bruno-garcia: @vaind @jpnurmi 

I hacked this to test.
What would you recommend I do in order to have this work for both Xbox One and Xbox Series X|S?
I imagine we don't want two toolchain files, right? 

Can we have it toggle based on "architecture"? (`Gaming.Xbox.XboxOne.x64` vs `Gaming.Xbox.Scarlett.x64`)?

Once this is ready for review I'll ask Mischan for his input of course but trying to spare him until this is in better shape

jpnurmi: how about a "base" toolchain file with the common xbox parts, included by the xbox scarlett and xbox one toolchain files?

```cmake
include(${CMAKE_CURRENT_LIST_DIR}/gxdk_base_toolchain.cmake)

set(CMAKE_GENERATOR_PLATFORM "Gaming.Xbox.XboxOne.x64" CACHE STRING "" FORCE)
set(CMAKE_VS_PLATFORM_NAME "Gaming.Xbox.XboxOne.x64" CACHE STRING "" FORCE)
set(CMAKE_CXX_FLAGS_INIT "$ENV{CFLAGS} -D_GAMING_XBOX_ONE ${CMAKE_CXX_FLAGS_INIT}" CACHE STRING "" FORCE)
```

jpnurmi: Passing in a variable from the command-line works too, but specialized toolchain files might be more convenient because you only need to select one toolchain file with help of auto-completion instead of passing an extra variable in addition to the toolchain file.

include/sentry.h
```diff
@@ -36,8 +36,8 @@ extern "C" {
 /* common platform detection */
 #ifdef _WIN32
 #    define SENTRY_PLATFORM_WINDOWS
-#    ifdef _GAMING_XBOX_SCARLETT
-#        define SENTRY_PLATFORM_XBOX_SCARLETT
+#    ifdef _GAMING_XBOX_SCARLETT || _GAMING_XBOX_XBOXONE
```
jpnurmi: > sentry.h(39,1): warning C4067: unexpected tokens following preprocessor directive - expected a newline

```suggestion
#    ifdef _GAMING_XBOX
```

vendor/acutest.h
```diff
@@ -487,7 +487,7 @@ test_print_in_color__(int color, const char* fmt, ...)
         printf("\033[0m");
         return n;
     }
-#elif defined ACUTEST_WIN__ && !defined _GAMING_XBOX_SCARLETT
+#elif defined ACUTEST_WIN__ && (!defined _GAMING_XBOX_SCARLETT || !defined _GAMING_XBOX_XBOXONE)
```
jpnurmi:
```suggestion
#elif defined ACUTEST_WIN__ && !defined _GAMING_XBOX_SCARLETT && !defined _GAMING_XBOX_XBOXONE
```

jpnurmi: please ignore the previous proposal, here's a simpler variant:

```suggestion
#elif defined ACUTEST_WIN__ && !defined _GAMING_XBOX
```

toolchains/xbox/gxdk_xs_toolchain.cmake
jpnurmi: These files rarely need to be touched. I propose taking a step back and following the original samples/structure by MS:

- https://github.com/microsoft/Xbox-GDK-Samples/blob/main/Samples/Tools/CMakeGDKExample/gxdk_toolchain.cmake
- https://github.com/microsoft/Xbox-GDK-Samples/blob/main/Samples/Tools/CMakeGDKExample/gxdk_xs_toolchain.cmake

bruno-garcia: Thanks J-P!!

bruno-garcia:
@supervacuus before we move this out of draft (just need some real-device testing IMO), could you take a peek and see if it's OK?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **Features**:
 
 - Add `sentry_clear_attachments()` to allow clearing all previously added attachments in the global scope. ([#1290](https://github.com/getsentry/sentry-native/pull/1290))
+- Compiles also on Xbox One ([#1294](https://github.com/getsentry/sentry-native/pull/1294))
```
supervacuus, line 8: The PR says: "feat: xbox one support" but the changelog only mentions compilation. Do you expect another change before Xbox One can be considered supported? I think it should be made clear from the changelog.

supervacuus (approved):
Glad to see that we can reuse the code paths entirely for both Xboxes. One thing that is apparent from the toolchain files for Scarlett and Xbox One is that they are highly similar, and this similarity can be leveraged to halve the required build-script code by extracting a handful of variables. Not necessarily in this PR, but as a sensible follow-up.

CHANGELOG.md
bruno-garcia: Good point, I made it clear that it's about compilation for now since I can't validate at this point the support. But we'll order a devkit and get this tested at some point and add to docs and communicate things better then.
Thank you!

PR #1298, opened 2025-07-03
chore: Xbox: disable select tests

vaind:
- [x] disable tests that require FS access
- [ ] ~fix module_finder code or tests - currently fails on debug_id check: ` test_modulefinder.c:30: Check sentry_value_get_type(debug_id) == SENTRY_VALUE_TYPE_STRING... failed`~ - moved to a followup issue #1300
- [x] disable module_finder test

#skip-changelog

### DIFF
--- a/tests/unit/test_basic.c
+++ b/tests/unit/test_basic.c
@@ -225,7 +225,7 @@ SENTRY_TEST(capture_minidump_basic)
 {
     // skipping on platforms that don't have access to fixtures on the local FS
 #if defined(SENTRY_PLATFORM_ANDROID) || defined(SENTRY_PLATFORM_NX)            \
-    || defined(SENTRY_PLATFORM_PS)
+    || defined(SENTRY_PLATFORM_PS) || defined(SENTRY_PLATFORM_XBOX)
     SKIP_TEST();
 #else
     SENTRY_TEST_OPTIONS_NEW(options);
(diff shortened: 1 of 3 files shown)


vaind:
@supervacuus you're marked as the one to contribute the module-finder code for xbox - any idea why debug images wouldn't have debug_id? I don't have xbox-devkit access right now to debug so I only know tests fail in CI

vaind:
> or maybe the test was accidentally built as release only?

Indeed I've changed this sometime in the process of trying to make this work and forget to go back. That was the culprit of module_finder test failure so I've reverted the change. However, to make it work with debug builds, a cmake toolchain update is necessary, so I've created a new PR #1302 to replace this one

supervacuus:
> > or maybe the test was accidentally built as release only?
> 
> Indeed I've changed this sometime in the process of trying to make this work and forget to go back. That was the culprit of module_finder test failure so I've reverted the change. However, to make it work with debug builds, a cmake toolchain update is necessary, so I've created a new PR #1302 to replace this one

Okay, thanks @vaind, that makes sense.

So, to be clear: the test can run as a `Release` build, but it needs debug information (i.e., `RelWithDebInfo`), because we extract the debug ID from the ClearView `PE` header, which only exist if `PDB` files have been created (which do not need to be transferred to the machine test executes on, but if they weren't built, there would simply be no header in the executable referencing them, and thus no debug ID).

PR #1301, opened 2025-07-04
feat(logs): add (u)int64 sentry_value_t type

JoshuaMoelans:
Cherry-picked from https://github.com/getsentry/sentry-native/pull/1297 (since we don't need the full alternative API for now). Merging into https://github.com/getsentry/sentry-native/pull/1272 so we can send integers as int64.

Will initially be used mainly for the new Logs feature, but we should introduce (mostly test) this for other data we send too.

---

Introduces `sentry_value_new_(u)int64()`

---

#skip-changelog


src/sentry_value.c
```diff
@@ -82,8 +82,8 @@ typedef struct {
     union {
         void *_ptr;
         double _double;
-        int64_t _int64;
-        uint64_t _uint64;
+        int64_t _int64_data; // to avoid shadowing __int64 (C2628 error)
```
JoshuaMoelans: For context:
- https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-2/compiler-error-c2628
- https://learn.microsoft.com/en-us/cpp/cpp/int8-int16-int32-int64

JoshuaMoelans:
@sentry review

JoshuaMoelans:
@sentry review

src/sentry_value.c
supervacuus: Can't we name them `_i64` and `_u64` respectively? From the type, it is clear what they represent and would prefer to have as short a name as possible for any union variant.

src/sentry_value.c
```diff
@@ -648,6 +688,26 @@ sentry__value_stringify(sentry_value_t value)
         buf[written] = '\0';
         return sentry__string_clone(buf);
     }
+    case SENTRY_VALUE_TYPE_INT64: {
+        char buf[24];
+        size_t written = (size_t)sentry__snprintf_c(
+            buf, sizeof(buf), "%lld", (long long)sentry_value_as_int64(value));
```
supervacuus: We should use `"%" PRIi64`. Is the cast necessary (or might that be because you used the old style conversion specifier)?

src/sentry_value.c
```diff
@@ -648,6 +688,26 @@ sentry__value_stringify(sentry_value_t value)
...
+        if (written >= sizeof(buf)) {
+            return sentry__string_clone("");
+        }
+        buf[written] = '\0';
+        return sentry__string_clone(buf);
+    }
+    case SENTRY_VALUE_TYPE_UINT64: {
+        char buf[24];
+        size_t written = (size_t)sentry__snprintf_c(buf, sizeof(buf), "%llu",
```
supervacuus: We should use `"%" PRIu64`.

src/sentry_value.c
```diff
@@ -648,6 +688,26 @@ sentry__value_stringify(sentry_value_t value)
         buf[written] = '\0';
         return sentry__string_clone(buf);
     }
+    case SENTRY_VALUE_TYPE_INT64: {
```
supervacuus: There is a significant amount of overlap in the cases for `DOUBLE`, `INT64`, and `UINT64`. That screams "macro", parameterizing the format string, the `sentry_value_as_X`, and possibly the cast (if really necessary).

src/sentry_value.c
```diff
@@ -681,6 +741,10 @@ sentry__value_clone(sentry_value_t value)
     case THING_TYPE_DOUBLE:
         sentry_value_incref(value);
         return value;
+    case THING_TYPE_INT64:
+    case THING_TYPE_UINT64:
+        sentry_value_incref(value);
+        return value;
```
supervacuus: Those could all share the same case.

src/sentry_value.c
```diff
@@ -893,6 +957,52 @@ sentry_value_as_double(sentry_value_t value)
...
+    const thing_t *thing = value_as_thing(value);
+    if (thing && thing_get_type(thing) == THING_TYPE_INT64) {
+        return thing->payload._int64_data;
+    }
+    if (thing && thing_get_type(thing) == THING_TYPE_UINT64) {
+        return (int64_t)thing->payload._uint64_data;
+    }
+    if (thing && thing_get_type(thing) == THING_TYPE_DOUBLE) {
+        return (int64_t)thing->payload._double;
+    }
```
supervacuus: I am pretty sure that this isn't something we should support, because a significant part of these conversions is undefined behavior:
* NaN casting is undefined
* while we can represent integers in the range +/-2^53 exactly with `double` precision, the overflow that happens beyond this doesn't mean we cannot represent numbers in `double` that are much larger than what we can represent with `int64_t` they only become very coarse... and those would overflow a signed integer, i.e. lead to UB
* Unsigned 64-bit integers will overflow 64-bit signed integers

Therefore, we should introduce checks here and only convert _finite_ doubles and unsigneds that fall within the `INT64_MIN` and `INT64_MAX` range, and document this decision; otherwise, we should remove these conversions. I would opt for the first.

src/sentry_value.c
```diff
@@ -893,6 +957,52 @@ sentry_value_as_double(sentry_value_t value)
...
+            : 0;
+    }
+    if (thing && thing_get_type(thing) == THING_TYPE_DOUBLE) {
+        // TODO no check for double out of uint64 range
+        // TODO double gets truncated (1.9 -> 1)
+        return thing->payload._double >= 0 ? (uint64_t)thing->payload._double
+                                           : 0;
+    }
+    return 0;
+}
```
supervacuus: This code shares the same issues as above (except for unsigned overflow being well-defined) and also introduces clamping towards zero, which discards information.

I think I would deliberately overflow here instead of clamping (or not convert and log a warning), because it is much clearer that something went wrong than a "number conversion" to 0. Keep in mind that the zero default case from the `value_as_double` function referred to conversions from non-numeric value types. However, by introducing an allowed conversion from other `numeric` types, making zero a default for a range issue is problematic.

As a significant number of problems have occurred in past programming interfaces due to overly eager type coercion, which led to the loss of information (was the 0 in the data, or was it due to type coercion when unboxing a value?). Let's not repeat this mistake. 😅

supervacuus: I recommend adding test cases for all undefined behavior and other problematic cases mentioned above before applying any changes to the code.

supervacuus (changes requested):
There are a couple of issues that require further deliberation. We can pair up on this if you'd like.

supervacuus (approved):

PR #1304, opened 2025-07-08
feat: implement the new User Feedback API

jpnurmi:
Sentry API: [User Feedback](https://develop.sentry.dev/sdk/data-model/envelope-items/#user-feedback) vs.  [User Report - Deprecated](https://develop.sentry.dev/sdk/data-model/envelope-items/#user-report---deprecated)

See also:
- https://github.com/getsentry/sentry-native/issues/885#issuecomment-2428023882

  > While the original function for capturing user feedback (that requires `event_id`) was called `CaptureUserFeedback`, we decided to name the new one `CaptureFeedback`.


include/sentry.h
```diff
@@ -2434,10 +2436,31 @@ SENTRY_API sentry_value_t sentry_value_new_user_feedback_n(
...
+ *
+ * See https://develop.sentry.dev/sdk/data-model/envelope-items/#user-feedback
+ */
+SENTRY_API sentry_value_t sentry_value_new_feedback(const char *message,
+    const char *contact_email, const char *name, const char *url,
+    const sentry_uuid_t *associated_event_id, const sentry_uuid_t *replay_id);
+SENTRY_API sentry_value_t sentry_value_new_feedback_n(const char *message,
+    size_t message_len, const char *contact_email, size_t contact_email_len,
+    const char *name, size_t name_len, const char *url, size_t url_len,
+    const sentry_uuid_t *associated_event_id, const sentry_uuid_t *replay_id);
```
jpnurmi: Four string arguments in a row can easily get mixed up.

include/sentry.h
```diff
@@ -2434,10 +2436,31 @@ SENTRY_API sentry_value_t sentry_value_new_user_feedback_n(
     const char *email, size_t email_len, const char *comments,
     size_t comments_len);
 
+/**
+ * Captures a deprecated User Report and sends it to Sentry.
+ * Deprecated: Please use `sentry_capture_feedback` instead.
+ */
+SENTRY_API void sentry_capture_user_feedback(sentry_value_t user_report);
```
jpnurmi, line 2471: `sentry_capture_feedback(sentry_value_t)` vs. `sentry_capture_user_feedback(sentry_value_t)`
   - Both operate on a `sentry_value_t`. Technically, it could be one function that checks for the existence of "message" vs. "comments", for example.
   - I renamed the old argument to `user_report` in some places to help highlight that it operates on the old User Report API.

include/sentry.h
jpnurmi: Not sure if we need `url` at all for sentry-native
> _String, optional._ The URL of the webpage the user was on when submitting feedback.
This may be used to search for or set alerts on feedback.

One possibility to make the API easier to read:
```c
sentry_value_t sentry_value_new_feedback(const char *message, const char *contact_email, const char *name);
void sentry_feedback_set_associated_event_id(sentry_value_t feedback, const sentry_uuid_t *associated_event_id);
void sentry_feedback_set_replay_id(sentry_value_t feedback, const sentry_uuid_t *replay_id);
```

JoshuaMoelans: Same goes for the `replay_id` ^

jpnurmi:
What if we don't change the sentry-native API at all but wire it up with the new Sentry Feedback API under the hood, and allow passing `NULL` as event ID?

jpnurmi:
### Option 1

Keep the old API as is. Add the new API alongside. (The way this PR started.)

```c
// Old API
sentry_value_t sentry_value_new_user_feedback(const sentry_uuid_t *uuid,
                                              const char *name,
                                              const char *email,
                                              const char *comments);
void sentry_capture_user_feedback(sentry_value_t user_feedback);

// New API
sentry_value_t sentry_value_new_feedback(const char *message,
                                         const char *contact_email,
                                         const char *name,
                                         const sentry_uuid_t *associated_event_id);

void sentry_capture_feedback(sentry_value_t feedback);
```

### Option 2

Wire up the existing API with the new Sentry Feedback API, and allow passing `NULL` for the event ID. Retain backward compatibility by checking feedback fields. (The way this PR is currently implemented.)

```c
sentry_value_t old_feedback = sentry_value_new_object();
sentry_value_set_by_key(old_feedback, "event_id", ...);
/* ... */
sentry_capture_feedback(old_feedback);

// vs.

sentry_value_t new_feedback = sentry_value_new_user_feedback(NULL, name, email, comments);
sentry_capture_feedback(new_feedback);
```

### Option 3

Something between options 1 and 2. For example, have one unified `sentry_value_new_user_feedback` but separate `sentry_capture_user_feedback` (old) and `capture_feedback` (new) to avoid unexpected behavioral changes for existing users.

```c
sentry_value_t feedback = sentry_value_new_user_feedback(NULL, name, email, comments);

sentry_capture_user_feedback(feedback);  // Old API
// vs.
sentry_capture_feedback(feedback); // New API
```

jpnurmi:
The [docs](https://develop.sentry.dev/sdk/data-model/envelope-items/#user-report---deprecated) claim that the old API should still work:

> This is an older, deprecated way of submitting user feedback. It continues to work so older SDKs won't break. New implementations should follow the feedback item type described above.

However, with the current sentry-native master based on the old feedback endpoint, I'm unable to get any feedback to show up at all in a self-hosted Sentry version 25.6.2. With the snippet copied from the example, an event shows up but there's no sign of feedback:

```c
    sentry_value_t event = sentry_value_new_message_event(
        SENTRY_LEVEL_INFO, "my-logger", "Hello user feedback!");
    sentry_uuid_t event_id = sentry_capture_event(event);

    sentry_value_t user_feedback = sentry_value_new_user_feedback(
        &event_id, "some-name", "some-email", "some-comment");

    sentry_capture_user_feedback(user_feedback);
```

Switching to the new endpoint helps.

examples/example.c
```diff
@@ -556,12 +556,8 @@ main(int argc, char **argv)
         sentry_capture_event(event);
     }
     if (has_arg(argc, argv, "capture-user-feedback")) {
-        sentry_value_t event = sentry_value_new_message_event(
-            SENTRY_LEVEL_INFO, "my-logger", "Hello user feedback!");
-        sentry_uuid_t event_id = sentry_capture_event(event);
-
         sentry_value_t user_feedback = sentry_value_new_user_feedback(
-            &event_id, "some-name", "some-email", "some-comment");
+            NULL, "some-name", "some-email", "some-message");
```
supervacuus: If we keep the old implementation and do not only replace it with the new one (which it seems like we could, without any effects to the user, since the new one takes an optional event), then we should keep the example and integration test for the old.

jpnurmi: Thanks, sounds like a good idea. I've restored the old integration test and made the example construct an old-style user report object by hand, similarly to sentry-unreal.

supervacuus:
> Apparently, old-style user feedback can get quietly dropped in multiple ways with sentry-native. Errors about invalid emails or too old events didn't surface when sending envelopes with the sentry-native API. Trying to send the same feedback data straight to the feedback endpoint with curl cli revealed the problems. Confusingly enough, the new API does accept invalid emails and has no limitation on how old an event it can be associated with.

I think all of what you describe hints even more at replacing the current implementation entirely with the new API. What is the reason to maintain this? What can't the new feedback API do that the old can (especially since we expose the same interfaces in the SDK, or at least the new is a strict superset of the old)?

jpnurmi:
Ok, what about converting old feedback objects to the new format? This would help to avoid breaking `sentry-unreal`, which doesn't use `sentry_value_new_user_feedback` but constructs feedback objects by hand:

https://github.com/getsentry/sentry-unreal/blob/c9b9d320cbef96dbb7dbd3f1d2e0190e43cb126f/plugin-dev/Source/Sentry/Private/GenericPlatform/GenericPlatformSentryUserFeedback.cpp

jpnurmi:
Some feedback from @bruno-garcia about repurposed `capture_user_feedback` vs. separate new `capture_feedback`:

> We’d like to standardize on the new name though.
> But more importantly, it’s a breaking change for self hosted sentry that don’t have support for the new envelope format
> so I’d add the new function. keep the old one for a while. sentry converts it to the new format on the backend. eventually we drop it with a breaking change notice
> meanwhile mark it as obsolete so folks know to migrate to the new one

With this in mind, I'm inclined to revert to option 1.

jpnurmi:
Thanks for an interesting suggestion. A deliberate break is indeed quite tempting. According to GitHub's public code search, you might be right that `sentry-unreal` is probably the only user of this API. 

https://github.com/search?q=sentry_capture_user_feedback+NOT+path%3A**%2Fsentry.h+NOT+path%3A**%2Fsentry_core.c+NOT+path%3A**%2Fexample.c&type=code&ref=advsearch

Nevertheless, since `sentry_capture_feedback` is wanted for consistency with other SDKs, adding `sentry_capture_user_report` on the side seems inconsistent. What if we just add the new API and mark the old API more visibly deprecated with https://github.com/getsentry/sentry-native/pull/1308?

supervacuus:
> Nevertheless, since sentry_capture_feedback is wanted for consistency with other SDKs, adding sentry_capture_user_report on the side seems inconsistent. What if we just add the new API and mark the old API more visibly deprecated with https://github.com/getsentry/sentry-native/pull/1308?

As I mentioned earlier, this wasn't intended as a new proposal. Given the relatively narrow requirements (_which this should have started with_), we should opt for option 1, even if it is far from ideal from the perspective of most users.

src/sentry_value.c
```diff
@@ -1350,9 +1350,11 @@ sentry_value_t
 sentry_value_new_user_feedback(const sentry_uuid_t *uuid, const char *name,
     const char *email, const char *comments)
 {
+    SENTRY_SUPPRESS_DEPRECATED
```
jpnurmi, line 1353: GCC and Clang allow calling a deprecated function in the same compilation unit, but MSVC throws a warning

```
D:\a\sentry-native\sentry-native\src\sentry_value.c(1353,12): error C2220: the following warning is treated as an error [C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\cmake0\sentry.vcxproj]
D:\a\sentry-native\sentry-native\src\sentry_value.c(1353,12): warning C4996: 'sentry_value_new_user_feedback_n': Use `sentry_value_new_feedback_n` instead [C:\Users\runneradmin\AppData\Local\Temp\pytest-of-runneradmin\pytest-0\cmake0\sentry.vcxproj]
```

supervacuus (approved):
Looks great! Thanks for going the rounds with me @jpnurmi!

mmalutinl:
@jpnurmi Should this PR have changed the docs as well or am I missing something?
I read this and see deprecated_by_this_pr example there?
https://docs.sentry.io/platforms/native/user-feedback/

JoshuaMoelans:
Hi @mmalutinl , good catch! We'll update the docs, thanks :)

edit: PR is open https://github.com/getsentry/sentry-docs/pull/18527

Issue #974, opened 2024-04-01
Android NDK Segfault crash related to `sentry_value_set_by_key_n`


aaronjosephsonos:
**Description**

Our Android project has seen crashes related to a Segfault in Sentry native code. There are **no** places in our code that are directly calling nativeAddBreadcrumb, so thinking something under the hood in Sentry library is making the call. Still, there seems to be a Segfault bug in sentry native code `sentry_value.c's` file. Thanks in advance for taking a look, and please let me know if there is additional information I can help provide.

**When does the problem happen**

- [ ] During build
- [X] During run-time
- [ ] When capturing a hard crash

**Environment**

- OS: Have seen crashes on Android 10,12,13,14
- Compiler: 
- CMake version and config:
- Sentry version - 7.4.0
- Sentry compose version - 7.4.0
- Sentry gradle plugin version - 4.3.0

Subset of Sentry options enabled
```
            options.enableAllAutoBreadcrumbs(true)
            options.isAnrEnabled = true
            options.isEnableNdk = true
            options.isAttachStacktrace = true
            options.isSendDefaultPii = false
            options.isEnableScopeSync = true
            options.maxBreadcrumbs = 500
            options.isEnableAutoSessionTracking = true
            options.isAttachThreads = true
            options.anrTimeoutIntervalMillis = 10000
            options.isAttachScreenshot = false
            options.tracesSampleRate = null
            options.profilesSampleRate = null
            when (configProvider.buildEnvironment) {
                Environment.Development -> {
                    options.maxBreadcrumbs = 700
                    options.isSendDefaultPii = true
                    options.isAttachScreenshot = true
                    options.tracesSampleRate = 1.0
                    options.profilesSampleRate = 1.0
                }
                Environment.Alpha, Environment.Beta -> {
                    options.tracesSampleRate = 0.4
                    options.profilesSampleRate = 0.4
                }
                Environment.Production -> {
                    // use Sonos (and Sentry) default values
                }
            }
```

**Steps To Reproduce**
I myself have not run into this issue and have not found a reliable way to reproduce. Our project has seen the crash approximately 43 times in the past 30 days across 32 users.
**Log output**
```
Exception Type: Unknown (SIGSEGV)

Application Specific Information:
Segfault

Thread 0 Crashed:
0   base.apk                        0x71ef138ce0        sentry_value_set_by_key_n (sentry_value.c:482)
[...]

PR #1309, opened 2025-07-10
ci: enable tsan

supervacuus:
related to: #962 #974


vaind:
Thanks for the info @supervacuus, I'll likely not get to do it today so feel free to make the changes and add me as a reviewer in the PR, I can verify downstream.

PR #1317, opened 2025-07-17
feat: add `sentry__path_get_mtime()`

jpnurmi:
For cleaning up old expired (feedback) envelopes. Extracted from #1303.

#skip-changelog


tests/unit/test_path.c
```diff
@@ -239,3 +247,30 @@ SENTRY_TEST(path_directory)
...
+    TEST_ASSERT(!!path);
+
+    TEST_CHECK(sentry__path_remove(path) == 0);
+    TEST_CHECK(!sentry__path_is_file(path));
+    TEST_CHECK(sentry__path_get_mtime(path) <= 0);
+
+    sentry__path_touch(path);
+    TEST_CHECK(sentry__path_is_file(path));
+
+    int before = sentry__path_get_mtime(path);
```
JoshuaMoelans:
```suggestion
    const time_t before = sentry__path_get_mtime(path);
```

tests/unit/test_path.c
```diff
@@ -239,3 +247,30 @@ SENTRY_TEST(path_directory)
...
+    TEST_CHECK(before > 0);
+
+    sleep_s(1);
+
+    sentry__path_write_buffer(path, "after", 5);
+    int after = sentry__path_get_mtime(path);
```
JoshuaMoelans:
```suggestion
    const time_t after = sentry__path_get_mtime(path);
```

src/path/sentry_path_unix.c
```diff
@@ -267,6 +267,17 @@ sentry__path_get_size(const sentry_path_t *path)
     }
 }
 
+time_t
+sentry__path_get_mtime(const sentry_path_t *path)
+{
+    struct stat buf;
+    if (stat(path->path, &buf) == 0) {
+        return (time_t)buf.st_mtime;
```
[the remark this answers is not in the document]
jpnurmi: `sentry__path_get_size()` would not work as expected on directories. it would return the size of the directory FS node, not the contents, which is why the extra check is there.

`sentry__path_get_mtime`, on the other hand, works fine on both files and directories. i think it makes no sense to add an arbitrary restriction

PR #1318, opened 2025-07-18
feat: add `sentry__process_spawn()`

jpnurmi:
For spawning a fully detached subprocess (feedback handler / crash reporter) that continues running independently from the (potentially crashed) parent process. Extracted from #1303.

### Linux

Classic double-`fork()` plus `setsid()` in between with stdin/out/err detached, and `execv()` or `execvp()` (same as what [Crashpad](https://github.com/getsentry/crashpad/blob/getsentry/util/posix/spawn_subprocess.cc) does too).

### macOS

Same as Linux except that `.app` bundles are launched with `open -a <bundle> --args <args..>` to make macOS activate the app correctly and bring it to the foreground.

### Windows

[`CreateProcessW`](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) + [`DETACHED_PROCESS`](https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags)

#skip-changelog


vaind (commented):
## Logic Error: False Success in Double-Fork Pattern

The Unix implementation has a fundamental design flaw in the double-fork pattern where `sentry__process_spawn()` returns `true` even when the target executable fails to launch.

**Logic Analysis:**
1. **Parent process** calls `fork()` → creates **first child** (pid1)
2. **First child** calls `setsid()` and `fork()` again → creates **second child** (pid2, the grandchild)
3. **Second child** (grandchild) calls `execv()` - this is where the target process actually runs
4. **First child** immediately calls `_exit(0)` after creating the grandchild
5. **Parent process** waits for the **first child** to exit with `waitpid(pid1, &status, 0)`

**The Problem:**
```c
// In spawn_process():
if (pid2 == 0) {
    // second child (grandchild): actual execv() happens here
    execv(argv[0], argv);
    SENTRY_ERRORF("execv failed: %s", strerror(errno));
    _exit(1);  // ← This failure is never seen by parent
} else {
    // first child exits immediately after fork
    _exit(0);  // ← Parent only sees this exit code
}
```

**Issue Flow:**
1. Parent waits for first child with `waitpid(pid1, &status, 0)`
2. First child exits with code 0 after successful second fork
3. If `execv()` fails in grandchild, it exits with code 1
4. But parent already received exit code 0 from first child → returns `true`

**Impact:** Function returns `true` for non-existent executables, making the boolean return value misleading.

**Note:** This is inherent to fully detached processes - you can't reliably detect launch success without IPC. Consider either:
- Documenting that `true` only means "detachment succeeded"
- Changing return type to `void` 
- Adding optional launch verification mechanism

This affects the Unix implementation only (lines 84-146).

—
*Comment by Claude Code*

vaind:
> `SENTRY_TEST(process_spawn)` seems somehow unstable or problematic when running with Valgrind:
> 
> * passed in the original PR: [getsentry/sentry-native/actions/runs/16290908192/job/46000449822](https://github.com/getsentry/sentry-native/actions/runs/16290908192/job/46000449822)
> * failed when the same code was extracted into this PR: [getsentry/sentry-native/actions/runs/16370255127/job/46256680933?pr=1318](https://github.com/getsentry/sentry-native/actions/runs/16370255127/job/46256680933?pr=1318)

Claude says it's because of the previous comment about return values. 

The valgrind failure I identified earlier is still valid because:

  1. The same logic flaw remains: The function still returns true when the first child exits successfully (line 144), even if the actual execv() in the
  grandchild fails.
  2. The test still fails: The test expects sentry__process_spawn() to return true when the process is successfully spawned, but due to the logic flaw, it may
  return true even when the cp command fails.
  3. Memory leak still occurs: Because the test fails at the assertion, it doesn't reach the cleanup code that would free the allocated paths.

  Current Status

  The CI is still failing with the same valgrind error and the same fundamental issue:

  - Test failure: TEST_CHECK(sentry__process_spawn(cp, exe->path, dst->path, NULL)) fails
  - Memory leak: Because the test fails, sentry__path_free(exe) is never reached
  - Root cause: The double-fork logic doesn't properly detect when the target process fails to launch

  The code changes I can see (NULL pointer checks) were good fixes, but the fundamental double-fork logic issue remains. The function will return true as long as
   the fork operations succeed, regardless of whether the target executable actually runs.

  To fix this properly, the implementation would need to either:

  1. Accept the limitation: Document that the return value only indicates fork success, not process launch success
  2. Use a different approach: Implement some form of IPC to detect actual launch failures
  3. Change the API: Make the function return void since reliable success detection isn't possible with full process detachment

  The current implementation creates a false expectation that launch failures can be detected, which they cannot be with the double-fork pattern.

vaind (approved):
LGTM

Issue #1316, opened 2025-07-16
Support Offline Caching of Envelopes


kpujjigit:
As of now it seems offline caching of envelopes is only supported in Java, JS and Flutter; not Native. Good for use cases when network is interrupted or down.

There exists this feature request: [Issue #727](https://github.com/getsentry/sentry-unreal/issues/727) ← in sentry-unreal which enables offline caching, and is built atop the Native SDK. Does that imply completing this case will enable it for sentry-native too?

bruno-garcia:
Possibly related as surfaced by Karl:
* https://github.com/getsentry/sentry-native/pull/1152
* https://github.com/getsentry/crashpad/pull/120

PR #1321, opened 2025-07-21
Fix WinHTTP initialization timing to prevent error 12007

techwmazur:
Resolve an issue in Microsoft GDK titles where WinHttpSendRequest intermittently failed with error code 12007 in Release Package builds. This fix delays the initialization of the HTTP session until the network stack is fully ready and the first HTTP request is made

https://learn.microsoft.com/pl-pl/gaming/gdk/docs/features/console/networking/initialization-connectivity-networking


src/transports/sentry_transport_xnetworking.cpp
```diff
@@ -0,0 +1,67 @@
+#ifdef _GAMING_XBOX_SCARLETT
```
jpnurmi: This could be left out because the file is only compiled for Xbox.

jpnurmi (commented):
Any idea if this works for Xbox One, too? 

Maybe we could use:

1. `if(XBOX)` in CMake

    https://github.com/getsentry/sentry-native/blob/70f38a25655323f2e1e336a8771725c335c41319/CMakeLists.txt#L14

2. `#ifdef SENTRY_PLATFORM_XBOX` in C

    https://github.com/getsentry/sentry-native/blob/70f38a25655323f2e1e336a8771725c335c41319/include/sentry.h#L40

?

src/transports/sentry_transport_xnetworking.cpp
techwmazur: Sure thing. I've missed [this](https://github.com/getsentry/sentry-native/pull/1294) change

jpnurmi: Thanks for catching up with the recent changes made for Xbox One.

Furthermore, we don't need the `#ifdef SENTRY_PLATFORM_XBOX` guard in `sentry_transport_xnetworking.cpp` because the file is not compiled at all for any other platform besides Xbox:

https://github.com/getsentry/sentry-native/blob/a3f3de6f3be43e4059c4535c0ae6ab42a2cc5c4f/src/CMakeLists.txt#L115-L119

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 
 - Update Xbox toolchain to include `UseDebugLibraries` fix for Debug builds. ([#1302](https://github.com/getsentry/sentry-native/pull/1302))
 - Fix GDK version selection for Xbox by propagating `XdkEditionTarget` to MSBuild. ([#1312](https://github.com/getsentry/sentry-native/pull/1312))
+- Fix WinHTTP initialization timing to prevent error 12007 on Xbox platform
```
jpnurmi: This should help with Danger:

```suggestion
- Fix WinHTTP initialization timing to prevent error 12007 on Xbox platform. ([#1321](https://github.com/getsentry/sentry-native/pull/1321))
```

src/transports/sentry_transport_winhttp.c
```diff
@@ -118,28 +147,14 @@ sentry__winhttp_transport_start(
...
-        // On windows 8.0 or lower, WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY does
-        // not work on error we fallback to WINHTTP_ACCESS_TYPE_DEFAULT_PROXY
-        if (!state->session) {
-            state->session = WinHttpOpen(state->user_agent,
-                WINHTTP_ACCESS_TYPE_DEFAULT_PROXY, WINHTTP_NO_PROXY_NAME,
-                WINHTTP_NO_PROXY_BYPASS, 0);
-        }
-    }
+#ifndef SENTRY_PLATFORM_XBOX
+    sentry__winhttp_sesssion_start(state);
```
bruno-garcia: typo on session

src/transports/sentry_transport_xnetworking.cpp
```diff
@@ -0,0 +1,64 @@
+#include "sentry_boot.h"
```
vaind, line 1: I think we should keep the contents here in the *winhttp.c file instead of creating a new one. So far, the mapping is file=transport. This would break that by only extending what's needed for winhttp transport.

src/transports/sentry_transport_xnetworking.cpp
```diff
@@ -0,0 +1,64 @@
...
+                    = CreateEvent(nullptr, TRUE, FALSE, nullptr);
+                if (network_initialized_event != nullptr) {
+                    XTaskQueueRegistrationToken token;
+                    hr = XNetworkingRegisterConnectivityHintChanged(queue,
+                        network_initialized_event,
+                        sentry__transport_network_connectivity_hint_changed_callback,
+                        &token);
+                    if (SUCCEEDED(hr)) {
+                        DWORD result = WaitForSingleObjectEx(
+                            network_initialized_event, INFINITE, FALSE);
```
vaind, line 43: This may block indefinitely. Is that the right thing to do? I guess it's alright if we're on the background thread, but might be something to keep in mind in relation to #1316 - at the very least, add a TODO mentioning that issue

src/transports/sentry_transport_winhttp.c
```diff
@@ -187,6 +202,20 @@ sentry__winhttp_send_task(void *_envelope, void *_state)
     sentry_envelope_t *envelope = (sentry_envelope_t *)_envelope;
     winhttp_bgworker_state_t *state = (winhttp_bgworker_state_t *)_state;
 
+#ifdef SENTRY_PLATFORM_XBOX
+    if (!state->session) {
+        SENTRY_DEBUG(
+            "ensuring xbox network is initialized for WinHttp transport");
+        sentry__transport_ensure_network_initialized();
+        sentry__winhttp_session_start(state);
```
vaind, line 210: Just to make sure - are both of these safe to call from a background thread?

src/transports/sentry_transport_winhttp.c
```diff
@@ -187,6 +202,20 @@ sentry__winhttp_send_task(void *_envelope, void *_state)
     sentry_envelope_t *envelope = (sentry_envelope_t *)_envelope;
     winhttp_bgworker_state_t *state = (winhttp_bgworker_state_t *)_state;
 
+#ifdef SENTRY_PLATFORM_XBOX
+    if (!state->session) {
```
vaind, line 206: We're only checking session once, however, the suspend/resume cycle would influence the connectivity. Do we need to ensure connection every time we make a request instead?

src/transports/sentry_transport_xnetworking.cpp
vaind: Similar to what we do for Switch in https://github.com/getsentry/sentry-native/blob/d0d732e3b4b13072ddfbe8decee3064e2a7e5d6e/src/transports/sentry_transport_curl.c#L192-L196

vaind (changes requested):
Thanks for the contribution. I've got some concerns listed below:

src/transports/sentry_transport_xnetworking.cpp
techwmazur: I've created dedicated file because xnetworking requires c++ compiler.  Having that in mind, do you still think that it should be merged to winhttp.c?

techwmazur: @vaind ping

vaind: Then it's probably good as is, I'll leave this up to the maintainers

vaind: Also, it may be worthwile to just move this PR to sentry-xbox repository. Then this being a cpp file would be a non-issue

PR #1326, opened 2025-07-29
feat: add `(u)int64 sentry_value_t` type

JoshuaMoelans:
Extracted from https://github.com/getsentry/sentry-native/pull/1301 as this is relevant not only to Logs, but also for the read/capture envelopes feature (see [comment](https://github.com/getsentry/sentry-native/pull/1320#discussion_r2230413583)). We can merge this to main before logs is finished.

The feature is mostly done already, but needs some more testing (e.g. seeing if values end up in the UI)

Also [this](https://github.com/getsentry/sentry-native/pull/1301#issuecomment-3117011947) needs to be tackled:
> Looks like sentry__value_from_json also needs some changes to avoid sentry__strtod_c and be able to handle > 2^53.


jpnurmi (commented):
I'd try to avoid manual integer parsing (`is_integer_string`). I suppose we need to check for `'-'` to determine whether unsigned values are allowed but besides that, I'd only operate with `strtoxxx` together with the appropriate errno and range checks. Can we keep it closer to the original code and basically just add extra checks to the "else" branch where it previously had the double fallback? Also, would one `strtoull` call be enough and then we'll do the cast manually based on the sign and range?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Add `sentry_value_new_feedback` and `sentry_capture_feedback` to allow capturing [User Feedback](https://develop.sentry.dev/sdk/data-model/envelope-items/#user-feedback). ([#1304](https://github.com/getsentry/sentry-native/pull/1304))
   - Deprecate `sentry_value_new_user_feedback` and `sentry_capture_user_feedback` in favor of the new API.
 - Add `sentry_envelope_read_from_file`, `sentry_envelope_get_header`, and `sentry_capture_envelope`. ([#1320](https://github.com/getsentry/sentry-native/pull/1320))
+- add `(u)int64` `sentry_value_t` type ([#1326](https://github.com/getsentry/sentry-native/pull/1326))
```
jpnurmi:
```suggestion
- Add `(u)int64` `sentry_value_t` type. ([#1326](https://github.com/getsentry/sentry-native/pull/1326))
```

include/sentry.h
```diff
@@ -388,6 +400,16 @@ SENTRY_API size_t sentry_value_get_length(sentry_value_t value);
  */
 SENTRY_API int32_t sentry_value_as_int32(sentry_value_t value);
 
+/**
+ * Converts a value into a 64 bit signed integer.
```
jpnurmi:
```suggestion
 * Converts a value into a 64-bit signed integer.
```

include/sentry.h
```diff
@@ -388,6 +400,16 @@ SENTRY_API size_t sentry_value_get_length(sentry_value_t value);
...
+ */
+SENTRY_API int64_t sentry_value_as_int64(sentry_value_t value);
+
+/**
+ * Converts a value into a 64 bit unsigned integer.
```
jpnurmi:
```suggestion
 * Converts a value into a 64-bit unsigned integer.
```

src/sentry_json.c
```diff
@@ -367,6 +368,28 @@ sentry__jsonwriter_write_int32(sentry_jsonwriter_t *jw, int32_t val)
...
+        snprintf(buf, sizeof(buf), "%" PRId64, val);
+        write_str(jw, buf);
+    }
+}
+
+void
+sentry__jsonwriter_write_uint64(sentry_jsonwriter_t *jw, uint64_t val)
+{
+    // TODO currently all uint64 values should be sent as strings; update when
+    // this changes
```
jpnurmi: what happened? what is this needed for? i was't part of the original reviews and might be missing some background context but i tried reverting "always send uint64 as string" and the unit tests pass. :)

JoshuaMoelans: Ah seems that I forgot to link to the develop docs; basically for Logs it is specified that we cannot ingest unsigned 64-bit integers yet (see [this section](https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes:~:text=For%2064%2Dbit%20unsigned%20integers%2C%20use%20the%20string%20type%20to%20avoid%20overflow%20issues.)), and I tried whether this is the case in general by trying to send some transactions with 64-bit unsigned integer data. Turns out we cannot handle this in the Sentry UI (yet), so for now I think it makes sense to send all uint64 values as strings instead.

In the Logs PR we just [store the uint64 arguments as string ](https://github.com/getsentry/sentry-native/pull/1272/files#diff-3c65bca63aabeb5b604d2494d7b51a68981c406c55ef0a7a144563f5d7974214R49-R56)`sentry_value_t` types, but with this more general change we could have logs store them as uint64, and then we only need to update the`sentry__jsonwriter_write_uint64` function once these values can get handled.

Reverting this should make the `value_uint64` [test](https://github.com/getsentry/sentry-native/pull/1326/files#diff-4aa0def1e40856a78920fdefd717906eb347c6dd7c01bd01ab2060950cc1c110R132-R163) fail, since it expects "" around the numbers in the json representation.

---

To test the UI handling of these values, I added this in the `example.c` before sending a transaction:

```c
        sentry_transaction_set_data(tx, "bigger", sentry_value_new_int32(INT64_MIN));
        sentry_transaction_set_data(tx, "biggest", sentry_value_new_uint64(UINT64_MAX));
```

And without the uint64->string conversion, the UI shows this:
Screenshot 2025-07-30 at 10 43 40

Adding the conversion makes it show up correctly.

tests/unit/test_value.c
```diff
@@ -347,6 +454,59 @@ SENTRY_TEST(value_freezing)
...
+    STRINGIFY_AND_CHECK(rv, "3.14");
+
+    rv = sentry_value_new_double(1000000000000000);
+    STRINGIFY_AND_CHECK(rv, "1e+15");
+
+    rv = sentry_value_new_double(INFINITY);
+    STRINGIFY_AND_CHECK(rv, "inf");
+
+    rv = sentry_value_new_double(NAN);
+    STRINGIFY_AND_CHECK(rv, "nan");
```
JoshuaMoelans: These tests fail on Linux because of INFINITY mapping to `Inf` and NAN mapping to `NaN`, and on Windows because NAN mapping to `-nan(ind)`. We could either introduce platform-specific strings just for testing this edge-case, or just remove these two tests.
```suggestion
    rv = sentry_value_new_double(INFINITY);
#ifdef SENTRY_PLATFORM_LINUX
    STRINGIFY_AND_CHECK(rv, "Inf");
#else
    STRINGIFY_AND_CHECK(rv, "inf");
#endif

    rv = sentry_value_new_double(NAN);
#ifdef SENTRY_PLATFORM_LINUX
    STRINGIFY_AND_CHECK(rv, "NaN");
#elifdef SENTRY_PLATFORM_WINDOWS
    STRINGIFY_AND_CHECK(rv, "-nan(ind)");
#else
    STRINGIFY_AND_CHECK(rv, "nan");
#endif
```

JoshuaMoelans: Since this is apparently implementation-defined (see C standard below), might make more sense to not test for this explicitly. We could maybe check that the output string just doesn't look number-like, but since currently `value_stringify` is only ever used internally for converting a string `sentry_value_t`, I'm also fine with skipping these edge-cases.

---

Screenshot 2025-07-30 at 11 36 08

(from https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf , page 290 in the pdf)

tests/unit/test_value.c
```diff
@@ -71,6 +72,112 @@ SENTRY_TEST(value_int32)
...
+    TEST_CHECK(sentry_value_is_frozen(val));
+    sentry_value_decref(val);
+
+    // We do convert int32 to int64
+    val = sentry_value_new_int32(42);
+    TEST_CHECK(sentry_value_as_int64(val) == 42LL);
+    sentry_value_decref(val);
+
+    // We don't convert uint64 to int64
+    val = sentry_value_new_uint64(-42LL);
```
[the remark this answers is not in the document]
JoshuaMoelans: Good catch, not super relevant for the thing we test here (the uint64 just underflows, and we only want to check that a uint64 cannot convert into an int64), but still an inconsistency that might be confusing later on.

src/sentry_value.c
```diff
@@ -838,9 +900,56 @@ sentry_value_as_double(sentry_value_t value)
...
+uint64_t
+sentry_value_as_uint64(sentry_value_t value)
+{
+    if ((value._bits & TAG_MASK) == TAG_INT32) {
+        SENTRY_WARN("Cannot convert int32 into uint64, returning 0");
+        return 0;
+    }
+
+    const thing_t *thing = value_as_thing(value);
+    if (thing && thing_get_type(thing) == THING_TYPE_UINT64) {
```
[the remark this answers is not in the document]
JoshuaMoelans: We decided against doing this just to keep it consistent, but fair consideration.

tests/unit/test_value.c
jpnurmi: Could it be an option to make the test relaxed enough? Case-insensitive and vague "starts with" or "contains" type of checks?

src/sentry_json.c
```diff
@@ -592,17 +615,39 @@ tokens_to_value(jsmntok_t *tokens, size_t token_count, const char *buf,
...
+                    } else {
+                        rv = sentry_value_new_uint64((uint64_t)ull_val);
+                    }
+                    success = true;
+                }
+            }
+
+            // Both failed, fallback to double
+            if (!success) {
+                double val = sentry__strtod_c(buf + root->start, NULL);
```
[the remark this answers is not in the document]
JoshuaMoelans: I think this is fine as-is, since we check after doing the `strtoull` function whether the return fits inside int32, int64 or uint64 (and only then create the `sentry_value_t`)

jpnurmi (approved):
LGTM :+1:

Only minor changes after already having been thoroughly reviewed in #1301

tests/unit/test_value.c
```diff
@@ -347,6 +454,69 @@ SENTRY_TEST(value_freezing)
...
+    do {                                                                       \
+        char *stringified = sentry__value_stringify(rv);                       \
+        for (char *p = stringified; *p; ++p)                                   \
+            *p = tolower(*p);                                                  \
+        TEST_CHECK(strstr(stringified, Expected) != NULL);                     \
+        sentry_free(stringified);                                              \
+        sentry_value_decref(rv);                                               \
+    } while (0)
+
+    sentry_value_t rv = sentry_value_new_list();
```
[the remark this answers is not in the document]
JoshuaMoelans: Good catch 👌

PR #1329, opened 2025-08-01
chore: move xbox toolchain to sentry-xbox repository

vaind:
Having the xbox support spread out between multiple repositories (sentry-native and sentry-xbox) induces a maintenance burden. Also, it's a bit of a pain  when including toolchains in the downstream sentry-xbox due to having first fetch this repo before being able to use them in cmake. 

This PR removes the toolchains from this repo. Additionally, I've removed changelog items from the unreleased versionnot relevant here anymore.


supervacuus (approved):

toolchains/xbox/xbox_build.md
bruno-garcia, line 1: I'd keep this file, but change the contents to:

"This SDK is kept compatible with Xbox but the actual toolchain, CI and integration tests are on a private repository because they rely on GDKX components that can't be made available on this repository and to avoid maintenance complexity everything was moved to the private repository. 
We'll send an invite to the repository once you follow the steps documented: See [Xbox documentation](https://docs.sentry.io/platforms/xbox/) to get access."


Something like this ^ 

tl;dr: "Sorry, should still work, if. u need to figure it out u can, if u want our stuff, ask for access to the private repo. Thanks"

vaind: done

include/sentry.h
```diff
@@ -72,6 +72,8 @@ extern "C" {
 #ifndef SENTRY_SDK_NAME
 #    if defined(SENTRY_PLATFORM_ANDROID)
 #        define SENTRY_SDK_NAME "sentry.native.android"
+#    elif defined(SENTRY_PLATFORM_XBOX)
+#        define SENTRY_SDK_NAME "sentry.native.xbox"
```
supervacuus: I didn't highlight this because it would be fine to do this in a closed repo. So, with the toolchain defs removed, no one could use the open _alone_ repo anymore anyway, so having the def here wouldn't provide a lot in return. However, I am fine either way. You guys delineate which parts should be where.

include/sentry.h
```diff
@@ -185,8 +187,6 @@ typedef enum {
     SENTRY_VALUE_TYPE_NULL,
     SENTRY_VALUE_TYPE_BOOL,
     SENTRY_VALUE_TYPE_INT32,
-    SENTRY_VALUE_TYPE_INT64,
-    SENTRY_VALUE_TYPE_UINT64,
```
supervacuus: Why?

include/sentry.h
```diff
@@ -250,16 +250,6 @@ SENTRY_API sentry_value_t sentry_value_new_null(void);
...
-/**
- * Creates a new 64-bit signed integer value.
- */
-SENTRY_API sentry_value_t sentry_value_new_int64(int64_t value);
-
-/**
- * Creates a new 64-bit unsigned integer value.
- */
-SENTRY_API sentry_value_t sentry_value_new_uint64(uint64_t value);
-
```
supervacuus: Why?

include/sentry.h
```diff
@@ -398,16 +388,6 @@ SENTRY_API size_t sentry_value_get_length(sentry_value_t value);
...
-/**
- * Converts a value into a 64-bit signed integer.
- */
-SENTRY_API int64_t sentry_value_as_int64(sentry_value_t value);
-
-/**
- * Converts a value into a 64-bit unsigned integer.
- */
-SENTRY_API uint64_t sentry_value_as_uint64(sentry_value_t value);
-
```
supervacuus: Why?

include/sentry.h
```diff
@@ -682,16 +662,6 @@ typedef struct sentry_envelope_s sentry_envelope_t;
...
-/**
- * Given an Envelope, returns the header if present.
- *
- * This returns a borrowed value to the headers in the Envelope.
- */
-SENTRY_API sentry_value_t sentry_envelope_get_header(
-    const sentry_envelope_t *envelope, const char *key);
-SENTRY_API sentry_value_t sentry_envelope_get_header_n(
-    const sentry_envelope_t *envelope, const char *key, size_t key_len);
-
```
supervacuus: Why?

include/sentry.h
```diff
@@ -728,44 +698,6 @@ SENTRY_API int sentry_envelope_write_to_file(
...
-    const wchar_t *path);
-SENTRY_API sentry_envelope_t *sentry_envelope_read_from_filew_n(
-    const wchar_t *path, size_t path_len);
-#endif
-
-/**
- * Submits an envelope, first checking for consent.
- */
-SENTRY_API void sentry_capture_envelope(sentry_envelope_t *envelope);
-
```
supervacuus: Why?

supervacuus (changes requested):
Just blocking this so it doesn't get accidentally merged.

include/sentry.h
vaind: that's an accident, I've done `git checkout master ./include/sentry.h` yesterday when restoring the sdk name change but looks like I didn't have latest master fetched.

include/sentry.h
vaind: I've been thinking along the same line but @bruno-garcia asked me to put this back because it's possible sometimes can still use sentry-native directly if they use their own toolchain file (e.g. if they pull this directly to their build system instead of precompiling), which is a good point. This is a small piece of code anyway so should be alright to keep here

CMakeLists.txt
```diff
@@ -60,6 +51,8 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "AIX" OR CMAKE_SYSTEM_NAME STREQUAL "OS400")
 	set(AIX TRUE)
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Prospero")
 	set(PROSPERO TRUE)
+elseif("${CMAKE_GENERATOR_PLATFORM}" MATCHES "Gaming\\.Xbox\\.(Scarlett|XboxOne)\\.x64")
+	set(XBOX TRUE)
```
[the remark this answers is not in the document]
vaind: that's not true, there's no CMAKE_SYSTEM_NAME == windows condition in this if-elseif chain

include/sentry.h
supervacuus: This certainly makes sense.

include/sentry.h
supervacuus: I expected as much, but I have a habit of excessively marking obvious accidents just in case, even if CI fails.

supervacuus (approved):

PR #1334, opened 2025-08-08
(outdated) logs - initial batching prototype

JoshuaMoelans:
superseded by https://github.com/getsentry/sentry-native/pull/1338 and https://github.com/getsentry/sentry-native/pull/1365

---
VERY WORK-IN-PROGRESS

to be merged into https://github.com/getsentry/sentry-native/pull/1271

## ToDo
- [ ] implement batching logic, which flushes a logs queue every 100 logs / every 5 seconds
- [ ] ensure that this implementation is thread-safe, and blocks as little as possible
- [ ] adapt existing tests & add integration test checking log envelope contents

## Progress

- initial implementation is just a single logs buffer/queue that fills up, flushes when it reaches 100, and spawns a thread with a 5s timer.
    - Makes use of atomic fetching of counter values to avoid collision.

- double buffer approach (like in .NET) seems more promising

#skip-changelog


JoshuaMoelans:
Closing this. We already launched logs in [v.0.11.1](https://github.com/getsentry/sentry-native/releases/tag/0.11.1), and this initial prototype got reworked into the [batching PR](https://github.com/getsentry/sentry-native/pull/1338) that was part of this release.

PR #1336, opened 2025-08-11
feat: Implement the GPU Info gathering within the Native SDK

mujacica:
* Implement GPU Info Context for Native SDK
* Support for Mac/Windows/Unix (Vulkan, Dynamically loaded)
* Support for several Vendors
* Added SENTRY_WITH_GPU_INFO option, ON by default on supported platforms
* Include vulkan-headers as submodule

### DIFF
--- a/src/gpu/sentry_gpu_vulkan.c
+++ b/src/gpu/sentry_gpu_vulkan.c
@@ -41,6 +41,8 @@ static PFN_vkGetPhysicalDeviceMemoryProperties
 
 static SENTRY_LIBRARY_HANDLE vulkan_library = NULL;
 
+static void unload_vulkan_library(void);
+
 static bool
 load_vulkan_library(void)
 {


README.md
```diff
@@ -295,6 +295,12 @@ using `cmake -D BUILD_SHARED_LIBS=OFF ..`.
   tuning the thread stack guarantee parameters. Warnings and errors in the process of setting thread stack guarantees
   will always be logged.
 
+- `SENTRY_WITH_GPU_INFO` (Default: `OFF`):
+  Enables GPU information collection and reporting. When enabled, the SDK will attempt to gather GPU details such as
+  GPU name, vendor, memory size, and driver version, which are included in event contexts. The implementation uses
+  platform-specific APIs: DXGI and Direct3D9 on Windows, IOKit on macOS, and PCI/DRM on Linux. Setting this to
+  `OFF` disables GPU information collection entirely, which can reduce dependencies and binary size.
```
bruno-garcia: Do we intend to turn this on by default at some point? Might be worth adding a note

src/gpu/sentry_gpu_common.c
```diff
@@ -0,0 +1,93 @@
...
+    case 0x121A:
+        return sentry__string_clone("3dfx Interactive");
+    case 0x18CA:
+        return sentry__string_clone("XGI Technology");
+    case 0x1039:
+        return sentry__string_clone("Silicon Integrated Systems [SiS]");
+    case 0x126F:
+        return sentry__string_clone("Silicon Motion");
+    default:
+        return sentry__string_clone("Unknown");
```
bruno-garcia: How about something like this so we can look at those by usage and expand?
```suggestion
        return sentry__string_clone("Unknown - " + vendor_id);
```

src/gpu/sentry_gpu_unix.c
```diff
@@ -0,0 +1,410 @@
...
+
+    sentry_gpu_info_t *gpu_info = NULL;
+    struct dirent *entry;
+
+    while ((entry = readdir(drm_dir)) != NULL) {
+        if (strncmp(entry->d_name, "card", 4) != 0) {
+            continue;
+        }
+
+        char name_path[512];
```
bruno-garcia: "can't ever big bigger" famous last words :)
But if it is, just truncates and then we can't find the path?

JoshuaMoelans: `d_name` is luckily limited to 256 characters https://man7.org/linux/man-pages/man3/readdir.3.html , so this shouldn't be an issue

src/gpu/sentry_gpu_common.c
mujacica: Good idea! Added it.

src/gpu/sentry_gpu_unix.c
mujacica: Using MAX_PATH now ;) Nice catch

README.md
mujacica: I rewrote it a bit, so now it will be enabled by default on all platforms that have required libraries that we can link against (DirectX, IOKit, etc. depending on the platform).

tests/assertions.py
```diff
@@ -49,7 +49,107 @@ def assert_user_feedback(envelope):
     assert user_feedback is not None
     assert user_feedback["name"] == "some-name"
     assert user_feedback["contact_email"] == "some-email"
-    assert user_feedback["message"] == "some-message"
+
```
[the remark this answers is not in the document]
mujacica: Will do, Cursor!

bruno-garcia:
@sentry review

src/gpu/sentry_gpu_vulkan.c
```diff
@@ -0,0 +1,260 @@
...
+// Dynamic function pointers
+static PFN_vkCreateInstance pfn_vkCreateInstance = NULL;
+static PFN_vkDestroyInstance pfn_vkDestroyInstance = NULL;
+static PFN_vkEnumeratePhysicalDevices pfn_vkEnumeratePhysicalDevices = NULL;
+static PFN_vkGetPhysicalDeviceProperties pfn_vkGetPhysicalDeviceProperties
+    = NULL;
+static PFN_vkGetPhysicalDeviceMemoryProperties
+    pfn_vkGetPhysicalDeviceMemoryProperties
+    = NULL;
+
```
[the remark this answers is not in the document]
supervacuus: Since this only happens during scope initialization, which is explicitly locked, it is fair to assume single-threaded access. However, a small comment that clarifies this is not wasted effort.

src/gpu/sentry_gpu_common.c
```diff
@@ -0,0 +1,153 @@
...
+        snprintf(unknown_vendor, sizeof(unknown_vendor), "Unknown (0x%04X)",
+            vendor_id);
+        return sentry__string_clone(unknown_vendor);
+    }
+    }
+}
+
+static sentry_value_t
+create_gpu_context_from_info(sentry_gpu_info_t *gpu_info)
+{
```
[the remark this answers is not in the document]
supervacuus: Since this gets called on a list that we populate right before, the chance of getting a `NULL` `gpu_info` is relatively small. A check at the start with an error log (not sure about the level, since you still consider it optional) and an early exit seems fair nonetheless.

src/gpu/sentry_gpu_vulkan.c
```diff
@@ -0,0 +1,260 @@
...
+    uint64_t total_memory = 0;
+    for (uint32_t i = 0; i < memory_properties.memoryHeapCount; i++) {
+        if (memory_properties.memoryHeaps[i].flags
+            & VK_MEMORY_HEAP_DEVICE_LOCAL_BIT) {
+            total_memory += memory_properties.memoryHeaps[i].size;
+        }
+    }
+    gpu_info->memory_size = total_memory;
+
```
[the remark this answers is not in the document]
supervacuus: That really depends on whether places that dereference these info fields have `NULL` checks. We generally don't null-check if a field is optional (i.e., where read access must constantly be guarded). Because we want to collect as much information as possible, and it is better to have some fields empty (even if due to memory allocation errors) than none at all. 

However, if the entire context doesn't make sense, if the field is missing, or we know that follow-up allocations are likely to fail as well, then we should exit early and free the `gpu_info` (and any other successful allocations before).

src/gpu/sentry_gpu_vulkan.c
```diff
@@ -0,0 +1,260 @@
...
+static sentry_gpu_info_t *
+create_gpu_info_from_device(VkPhysicalDevice device)
+{
+    VkPhysicalDeviceProperties properties;
+    VkPhysicalDeviceMemoryProperties memory_properties;
+
+    pfn_vkGetPhysicalDeviceProperties(device, &properties);
+    pfn_vkGetPhysicalDeviceMemoryProperties(device, &memory_properties);
+
+    sentry_gpu_info_t *gpu_info = sentry_malloc(sizeof(sentry_gpu_info_t));
```
supervacuus:
```suggestion
    sentry_gpu_info_t *gpu_info = SENTRY_MAKE(sentry_gpu_info_t);
```

src/gpu/sentry_gpu_vulkan.c
```diff
@@ -0,0 +1,260 @@
...
+    result = pfn_vkEnumeratePhysicalDevices(instance, &device_count, devices);
+    if (result != VK_SUCCESS) {
+        SENTRY_DEBUGF("Failed to get Vulkan physical devices: %d", result);
+        sentry_free(devices);
+        pfn_vkDestroyInstance(instance, NULL);
+        unload_vulkan_library();
+        return NULL;
+    }
+
+    sentry_gpu_list_t *gpu_list = sentry_malloc(sizeof(sentry_gpu_list_t));
```
supervacuus:
```suggestion
    sentry_gpu_list_t *gpu_list = SENTRY_MAKE(sentry_gpu_list_t);
```

README.md
```diff
@@ -299,6 +299,13 @@ using `cmake -D BUILD_SHARED_LIBS=OFF ..`.
   tuning the thread stack guarantee parameters. Warnings and errors in the process of setting thread stack guarantees
   will always be logged.
 
+- `SENTRY_WITH_GPU_INFO` (Default: `OFF`):
```
supervacuus, line 306: From the CMake script, I would qualify this, because by default it's `ON` on all supported platforms
```suggestion
- `SENTRY_WITH_GPU_INFO` (Default: `ON` on Windows, macOS, and Linux, otherwise `OFF`):
```

CMakeLists.txt
```diff
@@ -563,6 +585,12 @@ if(NOT XBOX)
 	endif()
 endif()
 
+# handle Vulkan headers for GPU info  
+if(SENTRY_WITH_GPU_INFO)
+	target_include_directories(sentry PRIVATE 
+		"$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/external/vulkan-headers/include>")
```
supervacuus, line 618: Okay, so we use a vendored snapshot of the _Vulkan_ headers and assume they are stable with what we dynamically load in the field. I'm sorry if this has already been mentioned elsewhere. It might be a good idea to document it in the module header of the context gathering. 

Typically, when we use dynamic loading, we make ourselves independent of any headers, not only for function prototypes but also for data definitions (so that we do not rely on the availability of a specific SDK, which might often not expose what we want, although we know it is there). I understand that this may be impractical for Vulkan.

However, in the case of the proposed concrete support, we rely on the libraries being installed in a known location, but we use only the headers we vendor. Does this sound right?

tests/unit/test_gpu.c
```diff
@@ -0,0 +1,395 @@
...
+            TEST_CHECK(strstr(vendor_name, "Silicon Motion") != NULL);
+            break;
+        case 0x0000:
+        case 0xFFFF:
+            TEST_CHECK(strstr(vendor_name, "Unknown") != NULL);
+            TEST_CHECK(strstr(vendor_name, "0x") != NULL);
+            break;
+        default:
+            TEST_CHECK(strstr(vendor_name, "Unknown") != NULL);
+            TEST_CHECK(strstr(vendor_name, "0x") != NULL);
```
supervacuus, line 150: Btw, if you have lots of those, feel free to add something like this:

```C
#define TEST_CHECK_IS_SUBSTRING(Haystack, Needle)                              \
    do {                                                                       \
        TEST_CHECK(strstr(Haystack, Needle) != NULL);                          \
        TEST_MSG("Haystack: %s", Haystack);                                    \
        TEST_MSG("Needle: %s", Needle);                                        \
    } while (0)
```

to `sentry_testsupport.h`

tests/unit/test_gpu.c
```diff
@@ -0,0 +1,395 @@
...
+            break;
+        }
+
+        sentry_free(vendor_name);
+    }
+
+    // Test with actual GPU info if available
+    if (gpu_list && gpu_list->count > 0) {
+        for (unsigned int i = 0; i < gpu_list->count; i++) {
+            sentry_gpu_info_t *gpu_info = gpu_list->gpus[i];
```
supervacuus, line 160: Here, you can use `TEST_ASSERT` to get early feedback if something is wrong, rather than a `SIGSEGV`.
```suggestion
            sentry_gpu_info_t *gpu_info = gpu_list->gpus[i];
            TEST_ASSERT(!!gpu_info);
```

tests/unit/test_gpu.c
```diff
@@ -0,0 +1,395 @@
...
+    // Test the common vendor ID to name mapping function with all supported
+    // vendors
+    unsigned int test_vendor_ids[]
+        = { 0x10DE, 0x1002, 0x8086, 0x106B, 0x1414, 0x5143, 0x1AE0, 0x1010,
+              0x1023, 0x102B, 0x121A, 0x18CA, 0x1039, 0x126F, 0x0000, 0xFFFF };
+
+    for (size_t i = 0; i < sizeof(test_vendor_ids) / sizeof(test_vendor_ids[0]);
+        i++) {
+        char *vendor_name = sentry__gpu_vendor_id_to_name(test_vendor_ids[i]);
+        TEST_CHECK(vendor_name != NULL);
```
supervacuus: Should probably be `TEST_ASSERT`.

tests/unit/test_gpu.c
```diff
@@ -0,0 +1,395 @@
...
+#ifdef SENTRY_WITH_GPU_INFO
+    // When GPU support is enabled, we should get some GPU information (at least
+    // on most systems)
+    if (gpu_list && gpu_list->count > 0) {
+        printf("Found %u GPU(s):\n", gpu_list->count);
+
+        // Check that at least one GPU has populated fields
+        bool has_info = false;
+        for (unsigned int i = 0; i < gpu_list->count; i++) {
+            sentry_gpu_info_t *gpu_info = gpu_list->gpus[i];
```
supervacuus, line 20: Here, you can use `TEST_ASSERT` to get early feedback if something is wrong, rather than a `SIGSEGV`.
            
```suggestion
            sentry_gpu_info_t *gpu_info = gpu_list->gpus[i];
            TEST_ASSERT(!!gpu_info);
```

src/gpu/sentry_gpu_common.c
```diff
@@ -0,0 +1,153 @@
...
+    sentry_value_t gpu_context = sentry_value_new_object();
+    if (sentry_value_is_null(gpu_context)) {
+        return gpu_context;
+    }
```
supervacuus, line 63: This is generally not necessary to do with a `sentry_value_t`. Any of the following functions will either fail early and free any incoming allocations or act like a no-op. So, purely from a safety perspective, going ahead would be okay. If you want to prevent doing unnecessary computation, returning early is still fine, of course.

src/gpu/sentry_gpu_vulkan.c
```diff
@@ -0,0 +1,260 @@
...
+    }
+#else
+    vulkan_library = SENTRY_LOAD_LIBRARY("libvulkan.so.1");
+    if (!vulkan_library) {
+        vulkan_library = SENTRY_LOAD_LIBRARY("libvulkan.so");
+    }
+#endif
+
+    if (!vulkan_library) {
+        SENTRY_DEBUG("Failed to load Vulkan library");
```
supervacuus: Failing to load Vulkan is worth more than a debug log. Not sure though what it should be in its current state.

src/gpu/sentry_gpu_vulkan.c
```diff
@@ -0,0 +1,260 @@
...
+        = (PFN_vkGetPhysicalDeviceProperties)SENTRY_GET_PROC_ADDRESS(
+            vulkan_library, "vkGetPhysicalDeviceProperties");
+    pfn_vkGetPhysicalDeviceMemoryProperties
+        = (PFN_vkGetPhysicalDeviceMemoryProperties)SENTRY_GET_PROC_ADDRESS(
+            vulkan_library, "vkGetPhysicalDeviceMemoryProperties");
+
+    if (!pfn_vkCreateInstance || !pfn_vkDestroyInstance
+        || !pfn_vkEnumeratePhysicalDevices || !pfn_vkGetPhysicalDeviceProperties
+        || !pfn_vkGetPhysicalDeviceMemoryProperties) {
+        SENTRY_DEBUG("Failed to load required Vulkan functions");
```
supervacuus: Ditto, we could mark the debug logs that are currently on failure paths to switch to a higher log level. I understand that you don't want to flood users without interest in the GPU context with warnings or errors, but perhaps after we detect a Vulkan library, not being able to resolve required functions is still worth a chatty warning.

README.md
```diff
@@ -98,7 +98,7 @@ per platform, and can also be configured for cross-compilation.
 System-wide installation of the resulting sentry library is also possible via
 CMake.
 
-The prerequisites for building differ depending on the platform and backend. You will always need `CMake` to build the code. Additionally, when using the `crashpad` backend, `zlib` is required. On Linux and macOS, `libcurl` is a prerequisite. For more details, check out  the [contribution guide](./CONTRIBUTING.md).
+The prerequisites for building differ depending on the platform and backend. You will always need `CMake` to build the code. Additionally, when using the `crashpad` backend, `zlib` is required. On Linux and macOS, `libcurl` is a prerequisite. When GPU information gathering is enabled (`SENTRY_WITH_GPU_INFO=ON`), the **Vulkan SDK** is required for cross-platform GPU detection. For more details, check out  the [contribution guide](./CONTRIBUTING.md).
```
supervacuus: This mentions more details in the contribution docs, but no update yet in this PR. Maybe that would be a good place for documenting this: https://github.com/getsentry/sentry-native/pull/1336/files#r2410896961.

supervacuus (changes requested):
Somehow my review message got lost... this looks great. Only minor topics that can be "fixed" or clarified.

src/gpu/sentry_gpu_vulkan.c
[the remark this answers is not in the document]
mujacica: Even if these fail, we still want to get the device metadata we can, that's the reason why no explicit checks were added here.

src/gpu/sentry_gpu_common.c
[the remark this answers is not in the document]
mujacica: Added SENTRY_WARN here to notify that the GPU info won't be provided.

src/gpu/sentry_gpu_vulkan.c
mujacica: Changed to Warn, thanks.

src/gpu/sentry_gpu_vulkan.c
mujacica: Agreed, changed to Warn.

CMakeLists.txt
mujacica: That sounds exactly righy!

src/gpu/sentry_gpu_vulkan.c
```diff
@@ -0,0 +1,260 @@
...
+
+    if (gpu_list->count == 0) {
+        sentry_free(gpu_list->gpus);
+        sentry_free(gpu_list);
+        unload_vulkan_library();
+        return NULL;
+    }
+
+    // Clean up the dynamically loaded Vulkan library since we're done with it
+    unload_vulkan_library();
```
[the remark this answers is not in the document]
mujacica: This function is only called once during scope initialization, and the data it gathers will be reused in subsequent scenarios.

src/gpu/sentry_gpu_common.c
```diff
@@ -0,0 +1,153 @@
...
+#include "sentry_string.h"
+
+char *
+sentry__gpu_vendor_id_to_name(unsigned int vendor_id)
+{
+    switch (vendor_id) {
+    case 0x10DE:
+        return sentry__string_clone("NVIDIA Corporation");
+    case 0x1002:
+        return sentry__string_clone("Advanced Micro Devices, Inc. [AMD/ATI]");
```
JoshuaMoelans, line 13: I also saw https://devicehunt.com/view/type/pci/vendor/1022 but not sure these are consumer-grade devices (Vendor name is `Advanced Micro Devices, Inc. [AMD]`).

src/gpu/sentry_gpu_vulkan.c
```diff
@@ -0,0 +1,260 @@
...
+            gpu_list->gpus[gpu_list->count] = gpu_info;
+            gpu_list->count++;
+        }
+    }
+
+    sentry_free(devices);
+    pfn_vkDestroyInstance(instance, NULL);
+
+    if (gpu_list->count == 0) {
+        sentry_free(gpu_list->gpus);
```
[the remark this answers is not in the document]
mujacica: Not correct. If the function fails, that single GPU will not be added, but others might. We are calling free if the count is 0, which means no leak is possible.

src/gpu/sentry_gpu_common.c
```diff
@@ -0,0 +1,153 @@
...
+    case 0x8086:
+        return sentry__string_clone("Intel Corporation");
+    case 0x106B:
+        return sentry__string_clone("Apple Inc.");
+    case 0x1414:
+        return sentry__string_clone("Microsoft Corporation");
+    case 0x5143:
+        return sentry__string_clone("Qualcomm");
```
JoshuaMoelans, line 22: 5143 doesn't seem to have devices listed: https://devicehunt.com/view/type/pci/vendor/5143
but 17CB does: https://devicehunt.com/view/type/pci/vendor/17CB (although they don't look like GPUs)

There are also a few other Qualcomm IDs:
https://pcilookup.com/?ven=qualcomm&dev=&action=submit

src/gpu/sentry_gpu_vulkan.c
```diff
@@ -0,0 +1,266 @@
...
+    uint64_t total_memory = 0;
+    for (uint32_t i = 0; i < memory_properties.memoryHeapCount; i++) {
+        if (memory_properties.memoryHeaps[i].flags
+            & VK_MEMORY_HEAP_DEVICE_LOCAL_BIT) {
+            total_memory += memory_properties.memoryHeaps[i].size;
+        }
+    }
+
+    // Sentry expects memory size in MB, and Vulkan reports in bytes
+    gpu_info->memory_size = total_memory / (1024 * 1024);
```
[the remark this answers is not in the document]
mujacica: memory_size field is expected to store MB's.

tests/test_integration_gpu.py
```diff
@@ -0,0 +1,259 @@
...
+                memory_size = gpu["memory_size"]
+                assert isinstance(
+                    memory_size, int
+                ), f"{context_key} memory_size should be an integer"
+                assert memory_size > 0, f"{context_key} memory_size should be positive"
+                # Should be at least 1MB (very conservative)
+                assert (
+                    memory_size >= 1024 * 1024
+                ), f"{context_key} memory size seems too small"
+
```
JoshuaMoelans, line 162: This seems to do pretty much the same as `assert_gpu_context` (this just has some additional checks for memory size, `vendor_id.isdigit()`, `name != "UNKNOWN"`); would make sense to consolidate the two imo

src/gpu/sentry_gpu_common.c
mujacica: 5143 and 17CB should cover all Adreno devices we need

src/gpu/sentry_gpu_common.c
```diff
@@ -0,0 +1,169 @@
...
+            = create_gpu_context_from_info(gpu_list->gpus[i]);
+        if (!sentry_value_is_null(gpu_context)) {
+            // Use context_index (not i) so keys are contiguous: gpu, gpu2, gpu3
+            char context_key[32];
+            if (context_index == 0) {
+                snprintf(context_key, sizeof(context_key), "gpu");
+            } else {
+                snprintf(context_key, sizeof(context_key), "gpu%u",
+                    context_index + 1);
+            }
```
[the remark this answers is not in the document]
mujacica: gpu is used per-default - there is no such Specification

PR #1337, opened 2025-08-11
fix: correctly apply dynamic mutex initialization in unit-tests

supervacuus:
With the tsan introduction, we also introduced mutexes into the unit tests to properly synchronize some of the test-runners (e.g., acutest has internals that are not thread-safe, for instance, updating records of check results until the test finishes). Sadly, I forgot to cover dynamic mutexes in the two test modules, which are currently not integrated in CI.

This PR only provides fixes for the failing downstream builds. However, we can restructure the mutex setup to eliminate the need to cover dynamic mutex initialization, thereby avoiding a separate test path for mutex initialization.

Raised here: https://github.com/getsentry/sentry-native/pull/1309#issuecomment-3173578709

#skip-changelog

### DIFF
--- a/tests/unit/test_concurrency.c
+++ b/tests/unit/test_concurrency.c
@@ -3,11 +3,17 @@
 
 #include <sentry_sync.h>
 
+#ifdef SENTRY__MUTEX_INIT_DYN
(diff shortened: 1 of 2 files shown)


JoshuaMoelans (approved):
LGTM

PR #1338, opened 2025-08-13
logs: single-queue batching

JoshuaMoelans:
Initial 'simple' queue implementation for logs batching.

- when first log is enqueued, starts a timer thread; this listens for a manual flush, or times out after 5s to trigger a flush.
- while flushing, all incoming logs are dropped. On high volume (e.g. 50 threads concurrently sending logs) we lose about 20%. 
- Flushing takes about ~0.5-1.5ms on average, and does **not** block the client thread

We could still extend this to a double buffer if needed, but might not be necessary to get us started with logs on Native.

### DIFF
--- a/src/sentry_core.c
+++ b/src/sentry_core.c
@@ -8,6 +8,7 @@
 #include "sentry_core.h"
 #include "sentry_database.h"
 #include "sentry_envelope.h"
+#include "sentry_logs.h"
 #include "sentry_options.h"
 #include "sentry_os.h"
 #include "sentry_path.h"
@@ -323,6 +324,9 @@ sentry_close(void)
             SENTRY_DEBUG("shutting down backend");
             options->backend->shutdown_func(options->backend);
         }
+
+        // Shutdown logs system before transport to ensure logs are flushed
+        sentry__logs_shutdown(options->shutdown_timeout);
 
         if (options->transport) {
             if (sentry__transport_shutdown(
--- a/src/sentry_logs.c
+++ b/src/sentry_logs.c
@@ -55,39 +55,24 @@ static void
 flush_logs(void)
 {
     // Use atomic compare-and-swap to ensure only one thread flushes at a time
-    long expected = 0;
-    // TODO platform-agnostic?
-    if (!__sync_bool_compare_and_swap(&lq.flushing, expected, 1)) {
+    // TODO is this platform-agnostic?
+    if (!__sync_bool_compare_and_swap(&lq.flushing, 0, 1)) {
         // Another thread is already flushing
         return;
     }
 
     // Determine which buffer to flush
-    long current_active = sentry__atomic_fetch(&lq.active_buffer);
-    long flush_buffer_idx
-        = current_active; // Start by trying to flush the active buffer
+    long current_active_buffer_idx = sentry__atomic_fetch(&lq.active_buffer);
+    // Active buffer has logs, switch to the other buffer for new writes
+    long new_active = 1 - current_active_buffer_idx;
+    sentry__atomic_store(&lq.active_buffer, new_active);
 
-    // Check if active buffer has any logs to flush
-    if (sentry__atomic_fetch(&lq.buffers[current_active].count) == 0) {
-        // Active buffer is empty, try the other buffer
-        // TODO do we want this?
-        flush_buffer_idx = 1 - current_active;
-        if (sentry__atomic_fetch(&lq.buffers[flush_buffer_idx].count) == 0) {
-            // Both buffers are empty, nothing to flush
-            sentry__atomic_store(&lq.flushing, 0);
-            return;
-        }
-    } else {
-        // Active buffer has logs, switch to the other buffer for new writes
-        long new_active = 1 - current_active;
-        sentry__atomic_store(&lq.active_buffer, new_active);
-    }
-
-    log_buffer_t *flush_buffer = &lq.buffers[flush_buffer_idx];
+    log_buffer_t *flush_buffer = &lq.buffers[current_active_buffer_idx];
     long count = sentry__atomic_fetch(&flush_buffer->count);
 
-    if (count == 0) { // TODO when can this happen?
-        // Buffer became empty while we were setting up, nothing to do
+    // Check if active buffer has any logs to flush
+    if (count == 0) {
+        // no need to swap
         sentry__atomic_store(&lq.flushing, 0);
         return;
     }
@@ -97,6 +82,9 @@ flush_logs(void)
     sentry_value_t logs_list = sentry_value_new_list();
 
     for (long i = 0; i < count; i++) {
+        // TODO reason about using a `sentry_value_t` list instead, as we could
+        // then just
+        //  set the logs["items"] to that value
         sentry_value_append(logs_list, flush_buffer->logs[i]);
     }
 
@@ -112,10 +100,11 @@ flush_logs(void)
     }
 
     // Free the logs object since envelope doesn't take ownership
+    // TODO if we replace by `sentry_value_t`, we could still decref +
+    // sentry_value_new_list?
     sentry_value_decref(logs);
 
-    // Reset the flushed buffer - this is now safe since we switched active
-    // buffer
+    // Reset the flushed buffer
     sentry__atomic_store(&flush_buffer->count, 0);
 
     // Clear flushing flag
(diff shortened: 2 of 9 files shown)


src/sentry_logs.c
```diff
@@ -4,9 +4,225 @@
...
+timer_task_func(void *task_data, void *worker_state)
+{
+    SENTRY_DEBUG("Starting timer_task_func");
+    (void)task_data; // unused parameter
+    (void)worker_state; // unused parameter
+    while (true) {
+        // check if timer_running is false (only on shutdown)
+        if (sentry__atomic_fetch(&g_logs_single_state.timer_running) == 0) {
+            break;
+        }
```
supervacuus: That could be the loop condition then, right?

src/sentry_logs.c
```diff
@@ -4,9 +4,225 @@
...
+        sentry_mutex_t task_lock;
+        sentry__mutex_init(&task_lock); // Initialize the mutex
+        sentry__mutex_lock(&task_lock); // Lock before cond_wait
```
supervacuus: You don't need to initialize the mutex every loop iteration. So, you can move this outside the loop.

However, in the current configuration, where you supposedly already rectified synchronized access to the flush buffer, you can keep the mutex locked `outside` the loop. Whenever the wait starts, it unlocks the lock and re-locks after that. So, as long as you don't need to have mutually exclusive access to _something_, there is no need to do anything with that lock.

src/sentry_logs.c
```diff
@@ -4,9 +4,225 @@
...
+
+        // Sleep for 5 seconds or until request_flush hits
+        sentry__cond_wait_timeout(
+            &g_logs_single_state.request_flush, &task_lock, 5000);
```
supervacuus: After waiting, you usually want to check why you woke and if the loop invariant still holds.

src/sentry_logs.c
```diff
@@ -4,9 +4,225 @@
...
+
+        sentry__mutex_unlock(&task_lock); // Unlock after cond_wait returns
```
supervacuus: again in the current config, can both happen after the loop.

src/sentry_logs.c
```diff
@@ -4,9 +4,225 @@
...
+#    define QUEUE_LENGTH 5
+#else
+#    define QUEUE_LENGTH 100
+#endif
+#define FLUSH_TIMER 5
+
+typedef struct {
+    sentry_value_t logs[QUEUE_LENGTH];
+    volatile long index;
+    volatile long adding;
```
supervacuus: These volatiles are not necessary and might prevent specific optimizations that would be safe to do. Doing atomic operations on integers will never be considered side-effect-free by a compiler. The real use case for `volatile` in C is memory-mapped I/O because the compiler cannot know the side effect.

src/sentry_logs.c
```diff
@@ -4,9 +4,225 @@
...
+
+// TODO look at mutex init for inspiration on how to dynamically init
+//  logs as sentry_value_t list using sentry_value_new_list_with_size(...)
+//  (at runtime since function call is not compile-time available)
+static struct {
+    log_single_buffer_t queue;
+    sentry_bgworker_t *timer_worker;
+    volatile long timer_task_submitted;
+    volatile long timer_running;
+    volatile long flushing;
```
supervacuus: ditto

src/sentry_logs.c
```diff
@@ -4,9 +4,237 @@
...
+    sentry_cond_t request_flush;
+    sentry_cond_t adding_done;
+    sentry_mutex_t adding_lock;
```
supervacuus: Why are those two necessary? The `adding` topic is entirely an atomic concern. Let's reduce the number of sync variables to a minimum. Adding the condvar + mutex on top of an atomic counter will lead to all kinds of complexities and is unnecessary, since the flusher already spins on the counter (i.e., the counter acts as a signal).

src/sentry_logs.c
```diff
@@ -4,9 +4,237 @@
...
+static void timer_task_func(void *task_data, void *worker_state);
+
+static void
+flush_logs_single_queue(void)
+{
+    long already_flushing
+        = sentry__atomic_fetch_and_add(&g_logs_single_state.flushing, 1);
+    if (already_flushing
+        || !sentry__atomic_fetch(&g_logs_single_state.queue.index)) {
+        sentry__atomic_fetch_and_add(&g_logs_single_state.flushing, -1);
```
supervacuus: If we are already flushing, shouldn't that variable be set by the thread that is flushing? These two conditions should better be separate if statements, because `flushing` just acts as a reentrancy guard, while a 0 `index` probably should reset `flushing` if it was set, because it would mean the timer triggered the flush.

src/sentry_logs.c
```diff
@@ -4,9 +4,237 @@
...
+    uint64_t before_flush = sentry__usec_time();
+    // TODO add logic to avoid empty-queue flush
+
+    // Wait for all adding operations to complete using condition variable
+    sentry__mutex_lock(&g_logs_single_state.adding_lock);
+    while (sentry__atomic_fetch(&g_logs_single_state.queue.adding) > 0) {
+        sentry__cond_wait(
+            &g_logs_single_state.adding_done, &g_logs_single_state.adding_lock);
+    }
+    sentry__mutex_unlock(&g_logs_single_state.adding_lock);
```
supervacuus: as mentioned above, the spin here should not be combined with a cond_var... you can add a TODO for a `sentry__cpu_relax()` (currently in spinlock, but only for unix).

src/sentry_logs.c
```diff
@@ -4,9 +4,237 @@
...
+        sentry_value_append(log_items, g_logs_single_state.queue.logs[i]);
+    }
+    sentry_value_set_by_key(logs, "items", log_items);
+    // sending of the envelope
+    sentry_envelope_t *envelope = sentry__envelope_new();
+    sentry__envelope_add_logs(envelope, logs);
+    SENTRY_WITH_OPTIONS (options) {
+        sentry__capture_envelope(options->transport, envelope);
+    }
+    sentry__atomic_store(&g_logs_single_state.queue.index, 0);
```
supervacuus: Maybe add a remark here that this could happen as a `cmp_xch` before we loop.

src/sentry_logs.c
```diff
@@ -4,9 +4,237 @@
...
+        if (sentry__bgworker_start(g_logs_single_state.timer_worker) != 0) {
+            SENTRY_WARN("Failed to start timer bgworker");
+            sentry__bgworker_decref(g_logs_single_state.timer_worker);
+            g_logs_single_state.timer_worker = NULL;
+            sentry__atomic_fetch_and_add(
+                &g_logs_single_state.timer_running, -1);
+        }
+    } else {
+        SENTRY_WARN("Failed to create timer bgworker");
+        sentry__atomic_fetch_and_add(&g_logs_single_state.timer_running, -1);
```
supervacuus: Is this a leftover from a previous attempt? Given the current implementation of the task as an endless loop, there is no need for you to use `bgworker`, right?

src/sentry_logs.c
```diff
@@ -4,9 +4,237 @@
...
+            sentry__atomic_fetch_and_add(
+                &g_logs_single_state.timer_running, -1);
+            sentry__atomic_store(&g_logs_single_state.timer_task_submitted, 0);
+        }
+    } else {
+        // Worker unavailable or task already submitted,
+        // reset the running flag
+        sentry__atomic_fetch_and_add(&g_logs_single_state.timer_running, -1);
+    }
+}
```
supervacuus: the same as above, the `timer_task` is not implemented as a "task" but as a thread-loop. This requires a `sentry__thread_spawn`, after the above did the `sentry__thread_new`.

src/sentry_core.c
```diff
@@ -324,6 +325,9 @@ sentry_close(void)
             options->backend->shutdown_func(options->backend);
         }
 
+        // Shutdown logs system before transport to ensure logs are flushed
+        sentry__logs_shutdown(options->shutdown_timeout);
+
```
supervacuus: I wonder if we shouldn't start the logs infra in init, too. Is there a spec that requires us to start the logs worker lazily?

src/sentry_logs.c
```diff
@@ -4,9 +4,237 @@
...
+    //  logs end up here anyway
+    if (log_idx < QUEUE_LENGTH) {
+        // Successfully got a slot, write the log
+        g_logs_single_state.queue.logs[log_idx] = log;
+        long adding_count = sentry__atomic_fetch_and_add(
+            &g_logs_single_state.queue.adding, -1);
+        // If this was the last adding operation, signal the condition
+        if (adding_count == 1) {
+            sentry__cond_wake(&g_logs_single_state.adding_done);
+        }
```
supervacuus: this can go

src/sentry_logs.c
```diff
@@ -4,9 +4,237 @@
...
+                    // start single task to run indefinitely
+                    setup_bgworker();
+                    generate_timer_task();
+                }
+            } else {
+                // Timer was already running, decrement back
+                sentry__atomic_fetch_and_add(
+                    &g_logs_single_state.timer_running, -1);
+            }
+        }
```
supervacuus: Even if we don't start the worker in `sentry_init()`, this code should be entirely outside of the enqueuing function

supervacuus (changes requested):
This is a first review and I guess it will still need a couple of rounds. Maybe we can work on these things a bit closer next.

src/sentry_core.c
JoshuaMoelans: Good point, I think we can be doing it less lazily (just checking whether `enable_logs` is true) during `sentry_init()`.

PR #1343, opened 2025-08-26
chore(internal): update mitmproxy

JoshuaMoelans:
Raised by dependabot: https://github.com/getsentry/sentry-native/security/dependabot/3

We only use mitmproxy for testing the Native SDK's proxy feature, so this is not shipped with the SDK itself. Still seems good to update this, as we didn't rely on the specific version.

#skip-changelog

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -37,7 +37,7 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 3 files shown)


JoshuaMoelans:
CI fails because `mitmproxy` version 12.1.2 needs Python version >3.12 https://pypi.org/project/mitmproxy/12.1.2/ (currently 3.11)

Issue #1345, opened 2025-08-28
`sentry_options_set_traces_sampler` is missing a user data argument


jpnurmi:
Unlike other Native SDK callbacks `sentry_options_set_traces_sampler` doesn’t provide a `user_data` parameter which normally acts as a closure allowing access to the subsystem implementation within the handler.

_Originally posted by @tustanivsky in https://github.com/getsentry/sentry-unreal/pull/1057#discussion_r2304518653_

PR #1346, opened 2025-08-28
fix: update `traces_sampler` to also take `user_data` argument

JoshuaMoelans:
fixes https://github.com/getsentry/sentry-native/issues/1345

### DIFF
--- a/tests/unit/test_sampling.c
+++ b/tests/unit/test_sampling.c
@@ -12,8 +12,10 @@ SENTRY_TEST(sampling_decision)
 
 static double
 traces_sampler_callback(const sentry_transaction_context_t *transaction_ctx,
-    sentry_value_t custom_sampling_ctx, const int *parent_sampled)
+    sentry_value_t custom_sampling_ctx, const int *parent_sampled,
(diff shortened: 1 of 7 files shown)


tests/unit/test_sampling.c
```diff
@@ -91,7 +93,8 @@ SENTRY_TEST(sampling_transaction)
     {
         // test the traces_sampler callback
         SENTRY_TEST_OPTIONS_NEW(options);
-        sentry_options_set_traces_sampler(options, traces_sampler_callback);
+        sentry_options_set_traces_sampler(
+            options, traces_sampler_callback, NULL);
```
jpnurmi: could this pass some random address and verify it in the callback?

jpnurmi (commented):
Thanks on behalf of downstream SDKs! :) This looks good, no need to inject internal values into the user's sampling context.

JoshuaMoelans:
(this would become part of the same release that includes the first Logs implementation for native, so I think bundling this in as a breaking change for native 0.11.0 is fine)

PR #1348, opened 2025-08-29
docs: replace legacy sentry_string_free with sentry_free

jpnurmi:
#skip-changelog

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -713,7 +713,7 @@ SENTRY_EXPERIMENTAL_API sentry_value_t sentry_envelope_get_transaction(
 /**
  * Serializes the envelope.
  *
- * The return value needs to be freed with sentry_free().
+ * The return value needs to be freed with `sentry_free`.
  */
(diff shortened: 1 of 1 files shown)


include/sentry.h
```diff
@@ -713,7 +713,7 @@ SENTRY_EXPERIMENTAL_API sentry_value_t sentry_envelope_get_transaction(
 /**
  * Serializes the envelope.
  *
- * The return value needs to be freed with sentry_string_free().
+ * The return value needs to be freed with sentry_free().
```
JoshuaMoelans: nit: for consistency
```suggestion
 * The return value needs to be freed with sentry_free.
```

jpnurmi: Could have backticks, too? Which do you prefer?
```suggestion
 * The return value needs to be freed with `sentry_free`.
```

JoshuaMoelans: oh yeah didn't even spot that those were missing 😅

PR #1366, opened 2025-09-09
feat: Add before crash hook to allow modifying other callbacks before crash

mujacica:
* Add before crash hook to allow modifying other callbacks before crash
* Hooks needed for game engines to prevent calling internal engine code while handling crashes
* differs from `on_crash_hook` by being the first thing to get called when handling a crash (which is necessary to avoid calling SENTRY_X with an internal engine logger for our in-crash-handling log statements).

### DIFF
--- a/tests/unit/test_before_crash.c
+++ b/tests/unit/test_before_crash.c
@@ -25,7 +25,7 @@ reset_before_crash_state(void)
 
 SENTRY_TEST(before_crash_func_call)
 {
-    void *user_data = (void *)0xDEADC0DE;
+    void *user_data = (void *)1;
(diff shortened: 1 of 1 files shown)


vaind:
I also think the logging part could be handled as a configuration, other than the reasons above, also because it's cleaner to know what's happening and that this must be checked first in the handler to configure logging. BTW, in sentry-playstation, I've done something similar WRT to logging - the handler replaces the handler: `sentry__logger_set_global({ nullptr, nullptr, SENTRY_LEVEL_ERROR });`

The logging changes could be done in a separate PR

PR #1370, opened 2025-09-15
test(logs): add 32-bit vargs test

JoshuaMoelans:


tests/unit/test_logs.c
```diff
@@ -111,3 +111,54 @@ SENTRY_TEST(formatted_log_messages)
...
+SENTRY_TEST(logs_param_conversion)
+{
+    // TODO this test shows the current limitation for parsing integers on
+    //  32-bit systems
+    int a = 1, b = 2, c = 3;
+#if defined(__i386__) || defined(_M_IX86) || defined(__arm__)
+    // Currently, on 32-bit platforms, we need to cast to long long since the
+    // parameter conversion expects long long for %d/%i format specifiers
+    test_param_conversion_helper(
+        "%lld %lld %lld", (long long)a, (long long)b, (long long)c);
```
supervacuus: Let's get rid of `%lld` usage as much as possible on our side of the logging code, because, while `%lld` correctly matches `long long`, `long long` isn't guaranteed to be only 64-bit (it can be broader). But in our logging pipeline, we can only support 64-bit numeric values (at best).

We would have the same problem as with 32-bit systems, on systems that support larger `long long`. That isn't the case very often, and it's not a problem on the platforms we currently support. Still, the Native SDK is continuously integrated on new platforms, and we should reduce the number of issues as much as possible by relying on a safe subset. This should be (applying will probably introduce wrong formatting):

```suggestion
    test_param_conversion_helper(
        "%" PRId64 " %" PRId64 " %" PRId64, (int64_t)a, (int64_t)b, (int64_t)c);
```

tests/unit/test_logs.c
```diff
@@ -111,3 +111,54 @@ SENTRY_TEST(formatted_log_messages)
...
+#else
+    test_param_conversion_helper("%d %d %d", a, b, c);
```
supervacuus, line 164: This should have a comment that raises:

* it is still undefined behavior (i.e., incorrect, given we read these params as 64-bit values)
* it only works because the 64-bit ABI provides 8-byte slots for variadic arguments

tests/unit/test_logs.c
```diff
@@ -111,3 +111,56 @@ SENTRY_TEST(formatted_log_messages)
...
+}
+
+SENTRY_TEST(logs_param_conversion)
+{
+    // TODO this test shows the current limitation for parsing integers on
+    //  32-bit systems
+    int a = 1, b = 2, c = 3;
+#if defined(__i386__) || defined(_M_IX86) || defined(__arm__)
+    // Currently, on 32-bit platforms, we need to cast to long long since the
+    // parameter conversion expects long long for %d/%i format specifiers
```
supervacuus: The comment is now out of sync with the code it is commenting on.

supervacuus (approved):

tests/unit/test_logs.c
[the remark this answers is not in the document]
supervacuus: This is clearly remarked in the test as undefined behavior that currently relies on the supported 64-bit ABIs. It is meant to highlight the reason why this works anyway.

tests/unit/test_logs.c
```diff
@@ -111,3 +111,54 @@ SENTRY_TEST(formatted_log_messages)
...
+    test_param_conversion_helper(
+        "%" PRId64 " %" PRId64 " %" PRId64, (int64_t)a, (int64_t)b, (int64_t)c);
+#else
+    test_param_conversion_helper("%d %d %d", a, b, c);
+#endif
```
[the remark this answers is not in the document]
supervacuus: Again, as mentioned above, this is precisely the reason why this test exists.

PR #1371, opened 2025-09-15
feat: Implement logging enable/disable feature, with crash handler options

mujacica:
* Added a feature to enable/disable logging
* Added a new option to enable/disable logging during crash handling
* Extended tests for the new functionality


src/backends/sentry_backend_breakpad.cpp
```diff
@@ -199,6 +207,13 @@ breakpad_backend_callback(const google_breakpad::MinidumpDescriptor &descriptor,
     }
     SENTRY_INFO("crash has been captured");
 
+    // Re-enable logging after crash handling if the option is set
+    SENTRY_WITH_OPTIONS (options) {
+        if (!options->handler_logging_enabled) {
+            sentry__logger_enable();
+        }
+    }
+
```
vaind: Probably doesn't make sense to reenable after the signal handler since we don't expect code to run afterwards. This applies to other handlers too.

supervacuus (approved):

include/sentry.h
```diff
@@ -1499,6 +1499,15 @@ SENTRY_API void sentry_options_set_system_crash_reporter_enabled(
...
 
+/**
+ * Enables or disables console logging during crash handling.
+ * When disabled, Sentry will not invoke logger callbacks during crash
+ * handler execution. This can be useful to avoid potential issues during
+ * crash handling. This is enabled by default.
+ */
+SENTRY_API void sentry_options_set_handler_logging_enabled(
+    sentry_options_t *opts, int enabled);
+
```
vaind: Seems to belong to the logger configuration options section. 
Maybe consider renaming to `void sentry_options_set_logger_is_signal_safe(sentry_options_t *opts, int value)`.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 **Internal:**
 
 - Support downstream Xbox SDK specifying networking initialization mechanism. ([#1359](https://github.com/getsentry/sentry-native/pull/1359))
+- Implement logging enable/disable feature, with crash handler options ([#1371](https://github.com/getsentry/sentry-native/pull/1371))
```
vaind: This adds an externally visible API/option so should be under Features, not Internal. Also, consider rewording, for example:
```suggestion
- Add a configuration to disable logging in crash handlers - `sentry_options_set_logger_is_signal_safe()` ([#1371](https://github.com/getsentry/sentry-native/pull/1371))
```

include/sentry.h
supervacuus: This is orthogonal to signal safety. It just excludes the logger from running during the handler, whether it is POSIX signals or anything else. Signal safety has a very concrete meaning in the context of POSIX signals, which we don't touch with this change.

I think it is fair to change the option to something connected to the `logger` "namespace", but I wouldn't go much further than something like `sentry_options_set_logger_enabled_in_handler` or `sentry_options_set_logger_enabled_in_crash_handler`.

src/backends/sentry_backend_breakpad.cpp
supervacuus: This is not entirely true. Not all paths immediately terminate any further execution. The chaining continues depending on the platform, backend, and hook returns. However, it isn't just handler chaining, but also backend handling code and concurrent threads that are still running after we return.

The question is much more: do we ever want to re-enable logging if _any_ code runs, given that the cause for which you wanted logging to be off in the first place is probably still in effect? So should this be: `sentry_options_set_logger_enabled_in_handler` or `sentry_options_set_logger_enabled_when_crashed`? 

Given the potential for concurrent threads to be executing code that could be logging, even after the crash handler returned, it is probably the latter.

mujacica: Changed to sentry_options_set_logger_enabled_when_crashed with latest commit

include/sentry.h
mujacica: Changed to sentry_options_set_logger_enabled_when_crashed with latest commit

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 - Add `user_data` parameter to `traces_sampler`. ([#1346](https://github.com/getsentry/sentry-native/pull/1346))
 
+**Features:**
+- Add a configuration to disable logging after crash has been detected - `sentry_options_set_logger_enabled_when_crashed()` ([#1371](https://github.com/getsentry/sentry-native/pull/1371))
```
supervacuus:
```suggestion
**Features:**

- Add a configuration to disable logging after a crash has been detected - `sentry_options_set_logger_enabled_when_crashed()`. ([#1371](https://github.com/getsentry/sentry-native/pull/1371))
```

examples/example.c
```diff
@@ -385,6 +401,21 @@ main(int argc, char **argv)
         sentry_options_add_view_hierarchy(options, "./view-hierarchy.json");
     }
 
+    if (has_arg(argc, argv, "test-logger")) {
```
supervacuus, line 404: Please add the newly added commands to the [contribution docs](https://github.com/getsentry/sentry-native/blob/master/CONTRIBUTING.md#integration-test-parameters).

tests/test_integration_logger.py
```diff
@@ -0,0 +1,162 @@
+import shutil
+import subprocess
+import sys
+
+import pytest
+
+import os
+
+from . import run
+from .conditions import has_breakpad, has_crashpad
```
supervacuus:
```suggestion
from .conditions import has_breakpad, has_crashpad, is_android
```

tests/test_integration_logger.py
```diff
@@ -0,0 +1,162 @@
...
+        ), "Should have SENTRY_LOG messages after crash when logging is enabled"
+
+
+@pytest.mark.parametrize(
+    "backend",
+    [
+        pytest.param(
+            "inproc",
+            marks=pytest.mark.skipif(
+                bool(os.environ.get("ANDROID_API")),
```
supervacuus: You can use `is_android` here:
```suggestion
                is_android,
```

tests/test_integration_logger.py
```diff
@@ -0,0 +1,162 @@
...
+    return parsed_data
+
+
+@pytest.mark.parametrize(
+    "backend",
+    [
+        pytest.param(
+            "inproc",
+            marks=pytest.mark.skipif(
+                bool(os.environ.get("ANDROID_API")),
```
supervacuus: You can use `is_android` here:
```suggestion
                is_android,
```

tests/test_integration_logger.py
```diff
@@ -0,0 +1,162 @@
...
+            "breakpad",
+            marks=pytest.mark.skipif(
+                not has_breakpad, reason="breakpad backend not available"
+            ),
+        ),
+        pytest.param(
+            "crashpad",
+            marks=pytest.mark.skipif(
+                not has_crashpad, reason="crashpad backend not available"
+            ),
```
supervacuus: You can also check for macOS here, since `crashpad` on macOS doesn't invoke any client-side handler:
```suggestion
            marks=[
                pytest.mark.skipif(
                    not has_crashpad, reason="crashpad backend not available"
                ),
                pytest.mark.skipif(
                    sys.platform == "darwin", reason="crashpad has no client handler on macOS"
                ),
            ],
```

tests/test_integration_logger.py
```diff
@@ -0,0 +1,162 @@
...
+    """Test that logging works during crash handling when enabled (default behavior)."""
+    data = _run_logger_crash_test(backend, cmake, "enable-logger-when-crashed")
+
+    # Verify that pre-crash logging worked
+    assert data["pre_crash_log_completed"], "Pre-crash log marker should be present"
+    assert len(data["sentry_logs"]) > 0, "Should have some SENTRY_LOG messages"
+
+    # When logging is enabled, we should see logs after the pre-crash marker
+    # Only check this on Linux, as other platforms don't reliably log during crash
+    if sys.platform == "linux" and backend != "crashpad":
```
supervacuus, line 126: So, only `inproc` and `breakpad` on Linux assert correctly here? Well, it is a start and exactly the reason to have the integration test in CI.

mujacica: Correct, crashpad "sometimes" does, and Windows/Mac are very random with the behavior. Something to take a look at, at some point.

examples/example.c
mujacica: Added with [c4d0fb4](https://github.com/getsentry/sentry-native/pull/1371/commits/c4d0fb47a1a844cd3def401e1db9ee84dfb8702e)

PR #1372, opened 2025-09-15
fix: resolve 'void function returning a value' compilation warnings

vaind:
Fixes compilation warnings about void functions returning values in test code.

#skip-changelog

🤖 Generated with [Claude Code](https://claude.ai/code)


tests/unit/tests.inc
```diff
@@ -205,6 +205,5 @@ XX(value_user)
 XX(value_wrong_type)
 XX(write_raw_envelope_to_file)
 XX(embedded_info_basic)
```
vaind:
```suggestion
XX(embedded_info_basic)
XX(embedded_info_disabled)
```

PR #1375, opened 2025-09-16
fix: prevent crashpad from leaking Objective-C ARC compile options

### LINKED ISSUE #1374: Broken MacOS build in 0.6.6 and later
**Description**

I've attempted to upgrade 0.6.2 currently used in our code base to the latest, 0.10.1. Windows seems to have gone smoothly, on Mac, the following error is being emitted while building Swift modules:
```
error: Driver threw unknown argument: '-fobjc-arc' without emitting errors. (in target 'redacted' from project 'Redacted')
```
This behavior starts in 0.6.6.

The problem has been introduced in [this commit](https://github.com/getsentry/crashpad/compare/432ff49ecccc1cdebf1a7646007bb0594ac3481f...f37da0f96ac60e919c0db6091d40000450ce7d6a#diff-a9fba704a6ab5ff33cb19bf292e697f6ba178c344a40851b96078a0cbde0e4a8) -- `-fobjc-arc`/`-fno-objc-arc-exceptions` flags are being set for APPLE unconditionally (and as PUBLIC options), which causes errors when Swift driver is used.

Please consider either keeping these options within local scope, or, if need to propagate those to the client code, limit to the relevant languages (ObjC/C/C++).

- [ x] During build
- [ ] During run-time

**Environment**

- OS: MacOS
- Compiler: XCode 16.2/16.4
- CMake version and config: 3.31.6

**Steps To Reproduce**

A project with mixed C/ObjC/Swift code base.

supervacuus:
Fixes #1374

Related `crashpad` PR: https://github.com/getsentry/crashpad/pull/134


supervacuus:
@jpnurmi, just fyi: this will pull in the external crash handler changes from the `crashpad` side. I guess they have no side effects unless things are "activated", but I wanted you to know anyway that those will become part of the next fix release.

PR #1377, opened 2025-09-17
fix: TOCTOU race between session life-cycle and event capture

supervacuus:
This issue more regularly appeared in the `concurrent_init` test in recent CI runs and was an actual race. 

We update the session if one is active inside a mutable options lock when capturing events. However, the check happened outside the lock. Concurrent session shutdown could mutate the session concurrently.

This PR fixes the race with a recheck inside the lock and a TSAN annotation around the optimistic read outside the lock.

### DIFF
--- a/src/sentry_core.c
+++ b/src/sentry_core.c
@@ -553,7 +553,7 @@ sentry__capture_event(sentry_value_t event, sentry_scope_t *local_scope)
                         envelope, mut_options->session);
                     // we're assuming that if a session is added to an envelope
                     // it will be sent onwards.  This means we now need to set
-                    // the init flag to false because we're no longer the
+                    // the init flag to false because we're no longer in the
                     // initial session update.
                     mut_options->session->init = false;
                 }


src/sentry_core.c
```diff
@@ -534,14 +535,28 @@ sentry__capture_event(sentry_value_t event, sentry_scope_t *local_scope)
...
-                // the init flag to false because we're no longer the
-                // initial session update.
-                mut_options->session->init = false;
+                // recheck inside the lock since our previous read is racy.
+                if (mut_options->session) {
+                    sentry__envelope_add_session(
+                        envelope, mut_options->session);
+                    // we're assuming that if a session is added to an envelope
+                    // it will be sent onwards.  This means we now need to set
+                    // the init flag to false because we're no longer the
```
mujacica:
```suggestion
                    // the init flag to false because we're no longer in the
```

mujacica: nit: Wording is not 100% clear, but if I understood the sentence correctly, 'in' is missing here.

PR #1378, opened 2025-09-17
tests: remove flaky logger test from transport suite

supervacuus:
Logger unit tests are always written from a single-threaded perspective. As such, they should be running inside the transport-based test suite since the transport could be concurrently logging and thus would lead to flaky test assertions.

#skip-changelog

### DIFF
--- a/tests/unit/test_logger.c
+++ b/tests/unit/test_logger.c
@@ -11,6 +11,8 @@ typedef struct {
 // Note: All logger unit-tests must only run from the transportless unit-test
 // suite, since the transport can concurrently log while we do our
 // single-threaded test assertions here, leading to flaky test runs.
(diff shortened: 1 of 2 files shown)


supervacuus:
Btw, the failing `concurrent_init` test on the `TSAN` build is fixed here: https://github.com/getsentry/sentry-native/pull/1377 if you want to take a look (and only if you have time).

tests/unit/test_logger.c
```diff
@@ -8,6 +8,12 @@ typedef struct {
     bool assert_now;
 } logger_test_t;
 
+// Note: All logger unit-tests must only run from the transportless unit-test
```
mujacica, line 11: Thanks for adding a comment, makes sense to exclude these.

mujacica (approved):

PR #1379, opened 2025-09-17
feat: Add comprehensive semver support for SENTRY_SDK_VERSION parsing

vaind:
## Summary

Adds comprehensive semantic versioning (semver) support to the SENTRY_SDK_VERSION parsing in CMake, enabling proper handling of prerelease identifiers and build metadata while maintaining backward compatibility.

### Version Extraction Enhancements
- **Full semver support**: Now correctly parses versions like `1.2.3-alpha.1+build.123`
- **Dual version variables**: 
  - `SENTRY_VERSION_FULL`: Complete version string (e.g., `1.2.3-alpha.1+build.123`)
  - `SENTRY_VERSION_BASE`: Major.minor.patch only (e.g., `1.2.3`)
  - Individual components: `SENTRY_VERSION_MAJOR`, `SENTRY_VERSION_MINOR`, `SENTRY_VERSION_PATCH`
- **Improved logging**: Shows both full and base versions during cmake configuration
- **Robust error handling**: Fixed condition that failed when version components were `0`

### Resource File Improvements  
- **Numeric fields**: Use base version components for Windows-compatible `FILEVERSION`/`PRODUCTVERSION`
- **String fields**: Use full version for `FileVersion`/`ProductVersion` display strings
- **Template updates**: Modified `sentry.rc.in` to use `@SENTRY_VERSION_FULL@` for string fields

### CMake Integration
- **PROJECT call**: Uses `SENTRY_VERSION_BASE` for maximum compatibility
- **write_basic_package_version_file**: Uses base version for proper package manager support
- **sentry_add_version_resource()**: Simplified to focus on resource generation only

## Manual Testing

Thoroughly tested with multiple version formats by modifying `SENTRY_SDK_VERSION` in `sentry.h` and running `cmake ..`:

### Test Results
- **Simple**: `0.10.1` → Base: `0.10.1`, Full: `0.10.1`
- **Prerelease**: `1.2.3-alpha.1` → Base: `1.2.3`, Full: `1.2.3-alpha.1`  
- **Build metadata**: `1.2.3+build.123` → Base: `1.2.3`, Full: `1.2.3+build.123`
- **Full semver**: `1.2.3-beta.2+build.456` → Base: `1.2.3`, Full: `1.2.3-beta.2+build.456`
- **Edge cases**: `0.0.0`, `0.0.1-alpha` → All handled correctly

### CMake Configuration Output Examples
```
-- Sentry SDK version (full): 1.5.2-beta.3+build.789
-- Sentry SDK version (base): 1.5.2
-- Sentry SDK version major='1' minor='5' patch='2'
```

<details>
<summary>Generated sentry.rc with prerelease semver (1.5.2-beta.3+build.789)</summary>

```rc
1 VERSIONINFO
FILEVERSION    1,5,2,0
PRODUCTVERSION 1,5,2,0
BEGIN
BLOCK "StringFileInfo"
BEGIN
    BLOCK "040904E4"
    BEGIN
        VALUE "FileDescription", "Client Library"
        VALUE "FileVersion", "1.5.2-beta.3+build.789"
        VALUE "InternalName", "sentry-native"
        VALUE "LegalCopyright", "https://sentry.io"
        VALUE "OriginalFilename", "sentry.dll"
        VALUE "ProductName", "Sentry Native SDK"
        VALUE "ProductVersion", "1.5.2-beta.3+build.789"
    END
END
BLOCK "VarFileInfo"
BEGIN
    VALUE "Translation", 0x409, 1252
END
END
```

**Key Success Points:**
- Numeric version fields (FILEVERSION/PRODUCTVERSION): Use base version `1,5,2,0` for Windows compatibility
- String version fields (FileVersion/ProductVersion): Use full semver `"1.5.2-beta.3+build.789"` preserving all metadata
</details>

## Compatibility

✅ **Backward compatible**: Existing simple versions like `0.10.1` work unchanged  
✅ **Forward compatible**: Supports full semver like `1.2.3-alpha.1+build.123`  
✅ **Edge cases handled**: Including `0.0.0` and other zero-component versions  
✅ **Windows resource files**: Proper numeric and string version separation

## Benefits

- **Console SDKs** can embed custom versioning schemes
- **Windows resource files** get proper numeric and string versions
- **Package managers** work with standard base versions  
- **CMake PROJECT** uses compatible version format
- **Full semver compliance** per https://semver.org/ specification

This enables console SDKs to use versions like `0.10.1+console.20250918` while maintaining full CMake and Windows resource compatibility.

🤖 Generated with [Claude Code](https://claude.ai/code)

#skip-changelog

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ endif()
 
 #read sentry-native version
 file(READ "include/sentry.h" _SENTRY_HEADER_CONTENT)
-string(REGEX MATCH "#define SENTRY_SDK_VERSION \"([0-9\.]+)\"" _SENTRY_VERSION_MATCH "${_SENTRY_HEADER_CONTENT}")
+string(REGEX MATCH "#define SENTRY_SDK_VERSION \"([0-9\.]+)([+\-][^\"]+)*\"" _SENTRY_VERSION_MATCH "${_SENTRY_HEADER_CONTENT}")
 set(SENTRY_VERSION "${CMAKE_MATCH_1}")
 unset(_SENTRY_HEADER_CONTENT)
 unset(_SENTRY_VERSION_MATCH)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,16 +9,30 @@ else()
 	cmake_policy(SET CMP0077 NEW)
 endif()
 
-#read sentry-native version
-file(READ "include/sentry.h" _SENTRY_HEADER_CONTENT)
-string(REGEX MATCH "#define SENTRY_SDK_VERSION \"([0-9\.]+)([+\-][^\"]+)*\"" _SENTRY_VERSION_MATCH "${_SENTRY_HEADER_CONTENT}")
-set(SENTRY_VERSION "${CMAKE_MATCH_1}")
-unset(_SENTRY_HEADER_CONTENT)
-unset(_SENTRY_VERSION_MATCH)
+# Extract version string from SENTRY_SDK_VERSION define in sentry.h
+file(READ "include/sentry.h" _VERSION_STR_TMP)
+# Supports full semver format including prerelease and build metadata
+string(REGEX MATCH "#define SENTRY_SDK_VERSION \"([^\"]+)\"" _VERSION_STR_TMP "${_VERSION_STR_TMP}")
+set(SENTRY_VERSION_FULL "${CMAKE_MATCH_1}")
+# Extract just the major.minor.patch part from the full version
+string(REGEX MATCH "^([0-9]+)\\.([0-9]+)\\.([0-9]+)" _VERSION_STR_TMP "${SENTRY_VERSION_FULL}")
+set(SENTRY_VERSION_MAJOR "${CMAKE_MATCH_1}")
+set(SENTRY_VERSION_MINOR "${CMAKE_MATCH_2}")
+set(SENTRY_VERSION_PATCH "${CMAKE_MATCH_3}")
+set(SENTRY_VERSION_BASE "${SENTRY_VERSION_MAJOR}.${SENTRY_VERSION_MINOR}.${SENTRY_VERSION_PATCH}")
+unset(_VERSION_STR_TMP)
+
+message(STATUS "Sentry SDK version (full): ${SENTRY_VERSION_FULL}")
+message(STATUS "Sentry SDK version (base): ${SENTRY_VERSION_BASE}")
+message(STATUS "Sentry SDK version major='${SENTRY_VERSION_MAJOR}' minor='${SENTRY_VERSION_MINOR}' patch='${SENTRY_VERSION_PATCH}'")
+
+if(NOT SENTRY_VERSION_FULL OR NOT SENTRY_VERSION_BASE OR SENTRY_VERSION_MAJOR STREQUAL "" OR SENTRY_VERSION_MINOR STREQUAL "" OR SENTRY_VERSION_PATCH STREQUAL "")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ endif()
 
 #read sentry-native version
 file(READ "include/sentry.h" _SENTRY_HEADER_CONTENT)
-string(REGEX MATCH "#define SENTRY_SDK_VERSION \"([0-9\.]+)\"" _SENTRY_VERSION_MATCH "${_SENTRY_HEADER_CONTENT}")
+string(REGEX MATCH "#define SENTRY_SDK_VERSION \"([0-9\.]+)([+\-][^\"]+)*\"" _SENTRY_VERSION_MATCH "${_SENTRY_HEADER_CONTENT}")
```
supervacuus: Thanks @vaind, I was always wondering why the regex was so strict. We can absolutely include

* semver [pre-release versioning](https://semver.org/#spec-item-9)
* semver [build metadata](https://semver.org/#spec-item-10)

in the regex, however, as you probably know and thus haven't changed, we can't use it for `SENTRY_VERSION` because CMake project and other output formats expect number-only versions (we could pass the full version to `write_basic_package_version_file()` so that the package is correctly versioned, it shouldn't break `find_package()`). I can see that you already considered this in your PR.

I would definitely love to have pre-release support for the Native SDK, which could have been used a couple of times for downstream purposes and direct users of the Native SDK. I hope there isn't a restriction in our current build or release (`craft`) setup that relies on the current strict versioning idea.

We could make the triple and appendage matcher more restrictive to prevent introducing whitespace or arbitrary ordering. The semver standard also includes a set of allowed characters for pre-release and build metadata. Are those too strong for your downstream uses? Something like:

```suggestion
string(REGEX MATCH "#define SENTRY_SDK_VERSION \"((([0-9]+)\\.([0-9]+)\\.([0-9]+))(-([0-9A-Za-z-]+(\\.[0-9A-Za-z-]+)*))?(\\+([0-9A-Za-z-]+(\\.[0-9A-Za-z-]+)*))?)\"" _SENTRY_VERSION_MATCH "${_SENTRY_HEADER_CONTENT}")
```

* where we actually match on a number _triple_ rather than any sequence of dotted numbers
* then have two optional terms (one for `-` and one for `+`) that can contain exactly the characters expressed in the semver spec, but which start without a leading dot

The only other thing that would have to change in this case is that we use `CMAKE_MATCH_2` for the `SENTRY_VERSION` rather than `CMAKE_MATCH_1`, since we still only want to pass down the triple in the script. 

You could also assign `CMAKE_MATCH_1` to `SENTRY_FULL_VERSION`, which we could pass to `write_basic_package_version_file()` and also use for the free text fields in `sentry.rc.in` (so that we have the triple in purely numeric version fields of the version header as we do now and use the full version in `FileVersion` and `ProductVersion`), but this isn't necessary in this PR.

vaind: I've updated the code to do all the version parsing in a single place. It doesn't strictly require the semver format (which would be `^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`) but keeps the simple enough to read while having some checks for the major.minor.patch parts of the version.

I see you've already approved but please have another look. This has been tested manually, you can see the results in the updated PR description.

PR #1384, opened 2025-09-18
fix: remove ASM language from the top-level CMake project

supervacuus:
We currently have no assembler requirement in the top-level project and thus were triggering CMake policy [CMP194](https://cmake.org/cmake/help/latest/policy/CMP0194.html). There should be no assembler configuration at all as long as we haven't added any of the subprojects.


jpnurmi:
Do we need to enable ASM in external/CMakeLists.txt for breakpad?

supervacuus:
> Do we need to enable ASM in external/CMakeLists.txt for breakpad?

Yes, but it is really only needed on Linux for the `getcontext` implementation. Since our `breakpad` fork isn't CMakeified, the entire CMake config happens in the top-level, and no `project()` triggers a language search in the subdirectory. However, it is really only there where we should do it, and not at the top level, where there are no restrictions concerning the backend, platform, or toolchain.

PR #1386, opened 2025-09-18
fix: add `crashpad_mpack` to the MSVC static runtime config

JanFellner:
+ properly aligned sorting order of wer lib to follow the leading code segment

#skip-changelog


jpnurmi (approved):
Thanks for the fix! 🙏

mujacica (approved):

PR #1387, opened 2025-09-18
test: add build test for static runtime with crashpad

supervacuus:
Triggered by #1386 (thanks again for the fix, @JanFellner). All other static runtime builds do not require any work in the top-level, but with `crashpad`, that is a regular topic if targets change. This static runtime build-only test should detect that the current master fails and verify that merging #1386 fixes the failed test.

#skip-changelog


supervacuus:
Test fails on all affected Windows builds (MSVC + ClangCL) correctly and expectedly.

PR #1392, opened 2025-09-29
fix: interop with managed .NET runtimes

jpnurmi:
Detect the situation when Mono/CLR converts a signal to a managed exception and transfers execution to the managed exception handler. In this case, Sentry Native should abort crash handling because the exception was already caught and handled in managed code.

See:
- #3954

### DIFF
--- a/src/backends/sentry_backend_inproc.c
+++ b/src/backends/sentry_backend_inproc.c
@@ -455,6 +455,53 @@ registers_from_uctx(const sentry_ucontext_t *uctx)
     return registers;
 }
 
+#ifdef SENTRY_PLATFORM_LINUX
+static uintptr_t
+get_stack_pointer(const sentry_ucontext_t *uctx)
+{
+#    if defined(__i386__)
+    return uctx->user_context->uc_mcontext.gregs[REG_ESP];
+#    elif defined(__x86_64)
+    return uctx->user_context->uc_mcontext.gregs[REG_RSP];
+#    elif defined(__ARM_EABI__)
+    return uctx->user_context->uc_mcontext.arm_sp;
+#    elif defined(__aarch64__)
+    return uctx->user_context->uc_mcontext.sp;
+#    elif defined(__mips__)
+    return uctx->user_context->uc_mcontext.gregs[MD_CONTEXT_MIPS_REG_SP];
+#    elif defined(__riscv)
+    return uctx->user_context->uc_mcontext.__gregs[MD_CONTEXT_RISCV_REG_SP];
+#    else
+    SENTRY_WARN("get_stack_pointer is not implemented for this architecture. "
+                "Signal chaining may not work as expected.");
+    return NULL;
+#    endif
+}
+
+static uintptr_t
+get_instruction_pointer(const sentry_ucontext_t *uctx)
+{
+#    if defined(__i386__)
+    return uctx->user_context->uc_mcontext.gregs[REG_EIP];
+#    elif defined(__x86_64)
+    return uctx->user_context->uc_mcontext.gregs[REG_RIP];
+#    elif defined(__ARM_EABI__)
+    return uctx->user_context->uc_mcontext.arm_pc;
+#    elif defined(__aarch64__)
+    return uctx->user_context->uc_mcontext.pc;
+#    elif defined(__mips__)
+    return uctx->user_context->uc_mcontext.pc;
+#    elif defined(__riscv)
+    return uctx->user_context->uc_mcontext.__gregs[MD_CONTEXT_RISCV_REG_PC];
+#    else
+    SENTRY_WARN(
+        "get_instruction_pointer is not implemented for this architecture. "
+        "Signal chaining may not work as expected.");
+    return NULL;
+#    endif
+}
+#endif
+
 static sentry_value_t
 make_signal_event(
     const struct signal_slot *sig_slot, const sentry_ucontext_t *uctx)
@@ -569,10 +616,23 @@ handle_ucontext(const sentry_ucontext_t *uctx)
             // the next signal coming in if we didn't "leave" here.
             sentry__leave_signal_handler();
 
+            uintptr_t ip = get_instruction_pointer(uctx);
+            uintptr_t sp = get_stack_pointer(uctx);
+
             // invoke the previous handler (typically the CLR/Mono
             // signal-to-managed-exception handler)
             invoke_signal_handler(
                 uctx->signum, uctx->siginfo, (void *)uctx->user_context);
+
+            // If the instruction or stack pointer changed, CLR/Mono converted
+            // the signal into a managed exception:
+            // https://github.com/dotnet/runtime/blob/6d96e28597e7da0d790d495ba834cc4908e442cd/src/mono/mono/mini/exceptions-arm64.c#L538
+            if (ip != get_instruction_pointer(uctx)
+                || sp != get_stack_pointer(uctx)) {
+                SENTRY_DEBUG("runtime converted the signal to a managed "
+                             "exception, we do not handle the signal");
+                return;
+            }
 
             // let's re-enter because it means this was an actual native crash
             sentry__enter_signal_handler();
--- a/tests/fixtures/dotnet_signal/Program.cs
+++ b/tests/fixtures/dotnet_signal/Program.cs
@@ -51,13 +51,16 @@ class Program
             {
                 Console.WriteLine("dereference a NULL object from managed code");
                 var s = default(string);
-                var c = s.Length;
+                var c = s!.Length;
             }
-            catch (NullReferenceException exception)
+            catch (NullReferenceException)
             {
-                Console.WriteLine("dereference another NULL object from managed code");
-                var s = default(string);
-                var c = s.Length;
+                if (args is ["managed-exception"])
+                {
+                    Console.WriteLine("dereference another NULL object from managed code");
+                    var s = default(string);
+                    var c = s!.Length;
+                }
             }
         }
     }
--- a/tests/fixtures/dotnet_signal/test_dotnet.csproj
+++ b/tests/fixtures/dotnet_signal/test_dotnet.csproj
@@ -4,5 +4,8 @@
     <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <PublishAot>true</PublishAot>
+    <RuntimeIdentifier>linux-x64</RuntimeIdentifier>
+    <Configuration>Release</Configuration>
   </PropertyGroup>
 </Project>
--- a/tests/test_dotnet_signals.py
+++ b/tests/test_dotnet_signals.py
@@ -35,12 +35,12 @@ def assert_run_dir_with_envelope(database_path):
     ), f"There is more than one crash envelope ({len(crash_envelopes)}"
 
 
-def run_dotnet(tmp_path, args):
+def run_dotnet(tmp_path, args=[]):
     env = os.environ.copy()
     env["LD_LIBRARY_PATH"] = str(tmp_path) + ":" + env.get("LD_LIBRARY_PATH", "")
     return subprocess.Popen(
-        args,
-        cwd=str(project_fixture_path),
+        [str(tmp_path / "bin/test_dotnet")] + args,
+        cwd=tmp_path,
         env=env,
         text=True,
         stdout=subprocess.PIPE,
@@ -49,11 +49,11 @@ def run_dotnet(tmp_path, args):
 
 
 def run_dotnet_managed_exception(tmp_path):
-    return run_dotnet(tmp_path, ["dotnet", "run"])
+    return run_dotnet(tmp_path, ["managed-exception"])
 
 
 def run_dotnet_native_crash(tmp_path):
-    return run_dotnet(tmp_path, ["dotnet", "run", "native-crash"])
+    return run_dotnet(tmp_path, ["native-crash"])
 
 
 @pytest.mark.skipif(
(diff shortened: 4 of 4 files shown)


src/backends/sentry_backend_inproc.c
```diff
@@ -455,6 +455,53 @@ registers_from_uctx(const sentry_ucontext_t *uctx)
...
+#    if defined(__i386__)
+    return uctx->user_context->uc_mcontext.gregs[REG_ESP];
+#    elif defined(__x86_64)
+    return uctx->user_context->uc_mcontext.gregs[REG_RSP];
+#    elif defined(__ARM_EABI__)
+    return uctx->user_context->uc_mcontext.arm_sp;
+#    elif defined(__aarch64__)
+    return uctx->user_context->uc_mcontext.sp;
+#    elif defined(__mips__)
+    return uctx->user_context->uc_mcontext.gregs[MD_CONTEXT_MIPS_REG_SP];
```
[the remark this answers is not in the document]
jpnurmi: https://github.com/getsentry/sentry-native/pull/1392/commits/10ba6bd8cfd60bd5d979ebd14e6015fcbea77624

src/backends/sentry_backend_inproc.c
```diff
@@ -569,11 +616,25 @@ handle_ucontext(const sentry_ucontext_t *uctx)
...
+            // If the instruction or stack pointer changed, CLR/Mono converted
+            // the signal into a managed exception and transferred execution to
+            // a managed exception handler.
+            // https://github.com/dotnet/runtime/blob/6d96e28597e7da0d790d495ba834cc4908e442cd/src/mono/mono/mini/exceptions-arm64.c#L538
+            if (ip != get_instruction_pointer(uctx)
+                || sp != get_stack_pointer(uctx)) {
+                SENTRY_DEBUG("runtime converted the signal to a managed "
+                             "exception, we do not handle the signal");
+                return;
+            }
```
supervacuus: The current implementation on `master` was written against the CLR JIT because it exhibited the same behavior (managed exceptions triggered by signals), required the same solution (invoking the signal chain at the beginning), and was the only code I had time to investigate in detail (so that we know what the runtime handler actually does). It was written with the assumption that downstream tests will compare it against other implementations to determine if they behave similarly.

For the CLR JIT (at least in versions 8 and 9), this added section _will still not be executed_, and I think (the comments in) this section and the PR in general should reflect that the changes are about the _current_ AOT behavior and not JIT. Even if the downstream usage is primarily about AOT deployments, the chance that the behavior will change or harmonize across implementations in the future is nonzero.

tests/fixtures/dotnet_signal/Program.cs
```diff
@@ -51,13 +51,16 @@ static void Main(string[] args)
...
-                var c = s.Length;
+                var c = s!.Length;
             }
-            catch (NullReferenceException exception)
+            catch (NullReferenceException)
             {
-                Console.WriteLine("dereference another NULL object from managed code");
-                var s = default(string);
-                var c = s.Length;
+                if (args is ["managed-exception"])
```
supervacuus: Why is this `if` necessary? I am not trying to be a stickler, but we should minimize the test code. If I see an `if` for the `"managed-exception"` inside the `catch`, I would assume that this code would be reached if we pass any other argument at the command line, which shouldn't be the case, right?

tests/test_dotnet_signals.py
```diff
@@ -83,20 +83,47 @@ def test_dotnet_signals_inproc(cmake):
...
+        dotnet_run = run_dotnet_managed_exception(tmp_path)
+        dotnet_run_stdout, dotnet_run_stderr = dotnet_run.communicate()
+
+        # the program does not handle the `NullReferenceException`, so the Native SDK will register a crash.
+        assert dotnet_run.returncode != 0
+        assert (
+            "NullReferenceException" in dotnet_run_stderr
+        ), f"Managed exception run failed.\nstdout:\n{dotnet_run_stdout}\nstderr:\n{dotnet_run_stderr}"
+        assert (database_path / "last_crash").exists(), "A crash was registered"
+
```
supervacuus: Again, this is not what is happening with CLR JIT. In CLR JIT, a signal handler will not return to its caller, and as such, the Native SDK will not handle a managed code exception as a crash, even with the changes proposed in this PR and even when the managed exception isn't handled inside the managed code.

I appreciate that CLR JIT might be an irrelevant topic for support, but it should be highlighted throughout the PR. Ideally, we would provide tests that track the behavior across the combination of (CLR+Mono, JIT+AOT). However, if we drop CLR JIT, we should highlight that all the code is written with AOT implementations in mind. Modifying the `ucontext` inside the signal chain is a rather curious choice and might be a moving target.

I am very grateful for these particular assertions, because they highlight the essential difference between the implementations. Is this actually the behavior you want downstream, since the Native SDK now actually handles a managed exception?

tests/test_dotnet_signals.py
```diff
@@ -35,12 +35,12 @@ def assert_run_dir_with_envelope(database_path):
     ), f"There is more than one crash envelope ({len(crash_envelopes)}"
 
 
-def run_dotnet(tmp_path, args):
+def run_dotnet(tmp_path, args=[]):
```
supervacuus: Please let's not use mutable default arguments, especially with a function that can be extended to various combinations later on:
```suggestion
def run_dotnet(tmp_path, args=None):
    if args is None:
        args = []
```

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+**Fixes**:
+- Fix interop with managed .NET runtimes. ([#1392](https://github.com/getsentry/sentry-native/pull/1392))
```
supervacuus: This should clearly highlight that the fix is related to AOT interop (and, by extension, Mono, although that isn't entirely obvious from the PR) and that the CLR JIT will still not report an unhandled managed exception (at least if the CLR JIT is still a topic and if it isn't maybe raise that the interop was changed to support primarily the AOT case as that is the relevant deployment downstream).

tests/fixtures/dotnet_signal/test_dotnet.csproj
```diff
@@ -4,5 +4,7 @@
     <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <PublishAot>true</PublishAot>
+    <Configuration>Release</Configuration>
```
supervacuus: Should we make those two test parametrizations (not necessarily `pytest` parameters)? If we want to track the behavior over time, it will definitely pay off. CLR JIT shows the same behavior in `Debug` and `Release`; I could imagine this also being the case in future versions of the other implementations.

supervacuus (changes requested):
Thanks for finding the diff to make the other implementations work @jpnurmi. Since this deviates significantly, we should document the change as clearly as possible (for our internal use) to ensure understanding that we have shifted the focus to the _current_ AOT signal/exception interface or adapt the tests to cover the relevant area.

tests/fixtures/dotnet_signal/Program.cs
jpnurmi: Currently, the test re-triggers a `NullReferenceException` from within the managed exception handler and leaks it into native code:

https://github.com/getsentry/sentry-native/blob/695f4a4dd21b27818ab5b0bbfcbbd9affea52f3b/tests/fixtures/dotnet_signal/Program.cs#L56-L61

I put it behind the `"managed-exception"` argument to be able to test the scenario where a managed exception is handled _without_ leaking it into native code, to let execution continue normally.

src/backends/sentry_backend_inproc.c
jpnurmi: Comment clarified in https://github.com/getsentry/sentry-native/pull/1392/commits/1bf0f524fe02c1023b9f68a95c834d1d5b8ecc48

tests/test_dotnet_signals.py
jpnurmi: Thanks. My intention is not to break the existing JIT support - just trying to fix the long-standing issue with duplicate crash events in AOT mode. :) I've restored the original JIT tests and added separate tests for AOT in https://github.com/getsentry/sentry-native/pull/1392/commits/060b18be6e3dbe1809fcdf108a957ee01824648a.

tests/fixtures/dotnet_signal/test_dotnet.csproj
jpnurmi: Good point. I've split the tests, and they are now passing these as arguments for AOT only:  https://github.com/getsentry/sentry-native/pull/1392/commits/060b18be6e3dbe1809fcdf108a957ee01824648a.

src/backends/sentry_backend_inproc.c
```diff
@@ -455,6 +455,53 @@ registers_from_uctx(const sentry_ucontext_t *uctx)
...
+#    elif defined(__aarch64__)
+    return uctx->user_context->uc_mcontext.sp;
+#    elif defined(__mips__)
+    return uctx->user_context->uc_mcontext.gregs[29]; // REG_SP
+#    elif defined(__riscv)
+    return uctx->user_context->uc_mcontext.__gregs[2]; // REG_SP
+#    else
+    SENTRY_WARN("get_stack_pointer is not implemented for this architecture. "
+                "Signal chaining may not work as expected.");
+    return NULL;
```
jpnurmi, line 474: is `#error` or a runtime warning a better choice here? i ended up with a warning because i thought it would be annoying to potentially break builds for irrelevant platforms that Sentry .NET doesn't even support

supervacuus: It is fine either way, but I would also err on giving a runtime response rather than failing at build time, as the entire execution path is optional.

tests/fixtures/dotnet_signal/Program.cs
supervacuus: However, this highlights a similar divergence to the one I previously raised with the test assertion. Because the initial reason for doing this was to have two managed exceptions, where one is caught in managed code and the other isn't, both would end up in our signal handler. 

Neither should create a native event, at least that was my assumption, since they both have nothing to do with an actual native crash, because you would get a stack trace of the runtime (or typically much less than that because whichever stackwalker is in effect will not have sufficient information to walk the runtime's stack).

In the AOT case, though, and apparently Mono too, since CLR AOT seems to be based on Mono AOT, you not only observe two `SIGSEGV`s (as in the CLR JIT), but also a final `SIGABRT` coming from the unhandled exception. This `SIGABRT` from my pov shouldn't trigger a native event either (for the same reason as above).

src/backends/sentry_backend_inproc.c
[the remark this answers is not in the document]
supervacuus: This is absolutely correct, but the only side-effect currently visible is for the logging toggle. Similarly, to how we "leave" the signal handler before chaining, we must also re-enable logging immediately after "leaving" and disable it again before re-entering, because if it were a managed code exception, we want logging to remain enabled. 

We can also move the entire `sig_slot` assignment down below the `CHAIN_AT_START` code, to make the path dependencies more obvious.

However, I think both have a lower priority than figuring out the signaling sequence of both runtimes and how we can align them.

tests/test_dotnet_signals.py
supervacuus: I am fine with dropping JIT, but we should still highlight it. If we keep it, that would be even better because it helps with tracking developments between the two implementations. In any case, thanks for the clarification.

supervacuus (changes requested):
I wonder if we can isolate the `SIGABRT` in case of an unhandled exception on AOT/Mono as well.

tests/fixtures/dotnet_signal/Program.cs
supervacuus: To clarify: I think it is acceptable for Mono to raise a `SIGABRT` at the end, but I wouldn't bypass this with an `if` statement inside the `catch`.

It is okay, because Sentry does install a top-level handler for .NET anyway, right? So the chance of triggering the `SIGABRT` handler of the Native SDK is relatively low. And if it triggers anyway, then that might highlight another issue.

But instead of preventing it from happening, I would let the C# code run as initially intended and then show inside the test that the serialized envelope _exists_ in the AOT/Mono case where two managed exceptions are raised, but that it is a SIGABRT and not the SIGSEGV that triggered the managed code exception. This way, you don't hide that behavior and explicitly show the difference between the two implementations in the test assertions.

Does that make sense?

I think developing a heuristic for ignoring that particular `SIGABRT` is rather an unnecessary investment at this point (except if you already know that it will be a problem downstream).

jpnurmi:
> > Were you able to observe this? Because this only happens when `crash_chaining` is disabled. I cannot imagine that crash or signal chaining is off by default (especially not on Android or Linux).
> 
> I tried creating a test case using `mcs` + `mono --aot` on Linux. Mono's native crash reporter kicks in when we call Mono's signal handler for a native crash, and execution ends there...

No wait, it's the newly added IP/SP check that prevents native crash handling, too. :facepalm: How the heck do we distinguish between these.......

supervacuus:
> to make it possible to swap the order in either managed or native code, respectively. However, that's just Linux, which is not relevant for Sentry .NET on Android or iOS. The problem is that there's no such type as `Mono.Runtime` on either Android or iOS...

We should do this in the Native SDK, similar to how we can change the invocation sequence in the handler; we can construct the signal chain up to a point during the setup of the signal handlers (rather than at signal-handling time). I can follow up on this topic next week.

tests/fixtures/dotnet_signal/Program.cs
jpnurmi: The intention is to test that the Native SDK does not capture any `SIGABRT` or `SIGSEGV` at all for such code:

```cs
try
{
    var s = default(string);
    var c = s.Length;
}
catch (NullReferenceException exception)
{
    // the exception is caught and handled in managed code. in AOT mode, execution
    // should continue normally without sentry-native's crash handling kicking in 
}
```

src/backends/sentry_backend_inproc.c
```diff
@@ -647,8 +709,10 @@ handle_ucontext(const sentry_ucontext_t *uctx)
     // forward as we're not restoring the page allocator.
     reset_signal_handlers();
     sentry__leave_signal_handler();
-    invoke_signal_handler(
-        uctx->signum, uctx->siginfo, (void *)uctx->user_context);
+    if (strategy != SENTRY_HANDLER_STRATEGY_CHAIN_AT_START) {
```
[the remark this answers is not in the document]
jpnurmi: only supported on Linux

src/backends/sentry_backend_inproc.c
```diff
@@ -456,6 +456,45 @@ registers_from_uctx(const sentry_ucontext_t *uctx)
...
+    return uctx->user_context->uc_mcontext.arm_pc;
+#    elif defined(__aarch64__)
+    return uctx->user_context->uc_mcontext.pc;
+#    else
+    SENTRY_WARN(
+        "get_instruction_pointer is not implemented for this architecture. "
+        "Signal chaining may not work as expected.");
+    return NULL;
+#    endif
+}
```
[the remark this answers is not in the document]
jpnurmi: mips, risc, and s390x were dropped because they are irrelevant for sentry-dotnet or .NET on Android, and there are no tests checking if they even work.

tests/fixtures/dotnet_signal/Program.cs
supervacuus: No, I understand that, but `SIGABRT` is not raised between those lines; it occurs when a managed code exception rises to the top without handling in C# code because the runtime then `abort()`s.

So by removing the second `NULL` dereference, you hide the fact that a `SIGABRT` is raised in the end. If that is your intention, then that is fine, but it should be clarified. In production code, the AOT executable will still `abort()` if a managed code exception raises to the top without handling, triggering the Native SDK signal handler. 

My point was that this might be unlikely to happen since `sentry-dotnet` probably acts as a top-level managed code exception handler anyway, and in case it didn't, you might still want an event.

However, the `SIGABRT` that we observe here is not, like the `SIGSEGV`, a signal we shouldn't see. It isn't something that gets converted to a managed code exception but propagates to the default handler.

supervacuus: We can also expose this as a third test option, if that makes more sense to you:

* a `NullReferenceException` that is handled inside managed code and thus absolutely must not be handled by the Native SDK
* a "native" crash that absolutely must be handled by the Native SDK
* a `NullReferenceException` that isn't handled inside the managed code (propagates to the top), which shouldn't be handled as a `SIGSEGV`, but where the application will return an error code in all runtimes, and in some might even raise a `SIGABRT` at the end.

The last one would be removed as a test-path if you don't execute the code in the catch clause for all managed exceptions. And given the broad spectrum of already known responses, it would be a limitation of our tests.

But I am totally fine to expose this as another parameter (`managed-exception`, `native-crash`, `unhandled-managed-expception`) you pass in, rather than automatically raising a second exception from the catch-clause. This way, you can handle them entirely separately with a separate set of assertions for each. Conflating them into one test case wasn't something I thought would last this long, given that the downstream integrations tests only exist now. However, the behavior of the unhandled managed exception should still be recorded and verified for tracking purposes.

Does that make more sense?

jpnurmi: Do you want to keep the nested exception i.e. should we continue throwing `NullReferenceException` from within a `NullReferenceException` handler or is it ok to separate those two cases like this: https://github.com/getsentry/sentry-native/pull/1392/commits/4a739075e215473c7bdf65ead3aabfbe58aa4be0 ?

supervacuus: As explained above, if we expose an additional parameter (aka another test dimension), I see no reason to keep the second throw in the catch of the first throw. There is no particular difference that I wanted to express by throwing from a catch; I could have thrown from outside as well. I only wanted to track the behavior of unhandled exceptions, too.

tests/fixtures/dotnet_signal/Program.cs
```diff
@@ -55,10 +55,13 @@ static void Main(string[] args)
...
             catch (NullReferenceException exception)
             {
-                Console.WriteLine("dereference another NULL object from managed code");
-                var s = default(string);
-                var c = s.Length;
             }
         }
+        else if (args.Contains("unhandled-managed-exception"))
+        {
+            Console.WriteLine("dereference another NULL object from managed code");
```
supervacuus:
```suggestion
            Console.WriteLine("dereference a NULL object from managed code (unhandled)");
```

src/backends/sentry_backend_inproc.c
```diff
@@ -494,10 +533,11 @@ make_signal_event(
...
         "captured backtrace from ucontext with %lu frames", frame_count);
-    // if unwinding from a ucontext didn't yield any results, try again with a
-    // direct unwind. this is most likely the case when using `libbacktrace`,
-    // since that does not allow to unwind from a ucontext at all.
-    if (!frame_count) {
+    // if unwinding from a ucontext didn't yield any results and we haven't
+    // chained signal handlers, try again with a direct unwind. this is most
+    // likely the case when using `libbacktrace`, since that does not allow to
+    // unwind from a ucontext at all.
+    if (!frame_count && strategy != SENTRY_HANDLER_STRATEGY_CHAIN_AT_START) {
```
supervacuus, line 541: I would clarify here why you exclude the "chain at start" strategy: because it crashes the handler.

Issue #1407, opened 2025-10-09
Handling of uin64_t values as strings causes processing errors during ingestion


mujacica:
Current handling of uint64_t values in Native SDK, or to be more precise, sending them as Strings, causes processing errors on Relay, for the fields that are expected to come as u64:

Native code: https://github.com/getsentry/sentry-native/blob/master/src/sentry_json.c#L381-L393
Relay: https://github.com/getsentry/relay/blob/c37efdb29e2c00b864e1b7a06212796b1d93e031/relay-event-schema/src/protocol/contexts/gpu.rs#L48

<img width="438" height="173" alt="Image" src="https://github.com/user-attachments/assets/d57d06ac-0f44-4f00-9354-3f6fd89a905e" />

PR #1408, opened 2025-10-09
fix: serialize `uint64_t` values as numerical instead of string

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/1407

We (mistakenly) assumed that the paragraph in the [logs documentation](https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes:~:text=Integers%20should%20be,them%20to%20strings.) stating `Make sure to always convert 64-bit unsigned integers to strings` applied to _all_ unsigned integers, but this is only an issue for log attributes. For all other values, we can safely ingest these.

We change the `jsonwriter` to write the actual numerical value again, and update `sentry_logs` to store unsigned integers as string `sentry_value_t` objects.

ToDo:
- ✅  update logs code to store unsigned ints as `sentry_value_t` strings instead

### DIFF
--- a/tests/unit/test_logs.c
+++ b/tests/unit/test_logs.c
@@ -247,10 +247,12 @@ test_param_conversion_types(const char *format, ...)
 
     // Validate %p (pointer) - should be string type with pointer representation
     TEST_CHECK_STRING_EQUAL(sentry_value_as_string(type5), "string");
-    // Pointer value should start with "0x" (platform dependent format)
+    // Pointer value format is platform-dependent, but should contain the hex
+    // digits (can be upper- or lower-cased)
     const char *ptr_str = sentry_value_as_string(value5);
     TEST_CHECK(ptr_str != NULL);
-    TEST_CHECK(strncmp(ptr_str, "0x", 2) == 0);
+    TEST_CHECK(strstr(ptr_str, "12345abc") != NULL
+        || strstr(ptr_str, "12345ABC") != NULL);
 
     // Validate %x (hex uint64) - should be string type with hex representation
     TEST_CHECK_STRING_EQUAL(sentry_value_as_string(type6), "string");
@@ -266,7 +268,7 @@ SENTRY_TEST(logs_param_types)
     double c = 3.14159;
     char d = 'A';
     const char *e = "test";
-    void *f = (void *)0x12345678;
+    void *f = (void *)0x12345abc;
     uint64_t g = 0xDEADBEEFDEADBEEF;
     test_param_conversion_types("%u %d %f %c %s %p %x", a, b, c, d, e, f, g);
 }


src/sentry_logs.c
```diff
@@ -335,9 +335,11 @@ static
     case 'X':
     case 'o': {
         unsigned long long int val = va_arg(*args_copy, unsigned long long int);
+        // TODO update once unsigned 64-bit can be sent as non-string
+        char buf[26];
+        snprintf(buf, sizeof(buf), "%" PRIu64, val);
         sentry_value_set_by_key(
-            param_obj, "value", sentry_value_new_uint64(val));
-        // TODO update once unsigned 64-bit can be sent
+            param_obj, "value", sentry_value_new_string(buf));
```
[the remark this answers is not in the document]
JoshuaMoelans: We specifically need to convert these for logs, since we cannot handle `uint64` values in log attributes. We can send non-log-attribute values as uint64 without issues.

src/sentry_logs.c
```diff
@@ -335,9 +335,11 @@ static
     case 'X':
     case 'o': {
         unsigned long long int val = va_arg(*args_copy, unsigned long long int);
+        // TODO update once unsigned 64-bit can be sent as non-string
+        char buf[26];
+        snprintf(buf, sizeof(buf), "%llu", val);
         sentry_value_set_by_key(
-            param_obj, "value", sentry_value_new_uint64(val));
-        // TODO update once unsigned 64-bit can be sent
+            param_obj, "value", sentry_value_new_string(buf));
```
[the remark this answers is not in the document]
JoshuaMoelans: This is a valid concern, doing something like this
```c
int a = 15454;
sentry_log_debug("I'm the second log message! %X" , a);
```
results in a non-hex value for the `message.parameter.0` attribute. 
Screenshot 2025-10-10 at 10 55 23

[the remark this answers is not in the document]
JoshuaMoelans: fixed in https://github.com/getsentry/sentry-native/pull/1408/commits/681f3d103bd781e7b7db2953391b207df8c46bcf

tests/unit/test_value.c
```diff
@@ -139,7 +139,7 @@ SENTRY_TEST(value_uint64)
     // We don't convert uint64 to int32
     TEST_CHECK(sentry_value_as_int32(val) == 0);
     TEST_CHECK(sentry_value_is_true(val));
-    TEST_CHECK_JSON_VALUE(val, "\"42\"");
+    TEST_CHECK_JSON_VALUE(val, "42");
```
mujacica, line 142: Can we also extend the logging tests to verify the uint64 behavior is also correct there.

src/sentry_logs.c
```diff
@@ -335,9 +335,13 @@ static
     case 'X':
     case 'o': {
         unsigned long long int val = va_arg(*args_copy, unsigned long long int);
+        // TODO update once unsigned 64-bit can be sent as non-string
```
mujacica, line 338: Not sure if that's even a plan

JoshuaMoelans: The logs [attributes docs](https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes:~:text=In%20the%20future%20we%20will%20support%20sending%2064%2Dbit%20unsigned%20integers%20without%20converting%20them%20to%20strings) say the following:
> In the future we will support sending 64-bit unsigned integers without converting them to strings

So I am hopeful that one day we can standardize how we send uint64 🙏

tests/unit/test_value.c
JoshuaMoelans: I added a test which verifies we parse the args into the right `sentry_value_t` types (see https://github.com/getsentry/sentry-native/pull/1408/commits/be248ea3e85d14b0eb4b334ec6ae32b54e9de34b). Do you feel like we should also verify `TEST_CHECK_JSON_VALUE` in that test again, or does the current test suffice?

mujacica: I think that's fine now, thanks!

PR #1416, opened 2025-10-17
feat: Auto-split build ID from version in sentry_library_info

vaind:
## Summary

Automatically extracts build ID from the version string read from `sentry.h` when downstream SDKs modify it to include build metadata in semver format.

Fixes the issue where both `SENTRY_VERSION` and `BUILD` fields contained the full version with build metadata (e.g., `0.11.3+20251016-9e31c9f-dirty`).

## Problem

When downstream SDKs modify `sentry.h` to include the full version with build ID and also set `SENTRY_BUILD_ID` via CMake, the embedded library info duplicates the build ID:

```
SENTRY_VERSION:0.11.3+20251016-9e31c9f-dirty;PLATFORM:scarlett;BUILD:0.11.3+20251016-9e31c9f-dirty;...
```

## Solution

The CMake configuration now:
1. Parses the version from `sentry.h` and extracts any build metadata after the `+` character
2. Uses only the base version (major.minor.patch) for `SENTRY_VERSION` field
3. Uses the extracted build ID for the `BUILD` field
4. Still respects the `SENTRY_BUILD_ID` cache variable if explicitly set (takes precedence)
5. Falls back to timestamp if no build ID is found

## Result

Embedded library info now correctly separates version and build ID:

```
SENTRY_VERSION:0.11.3;PLATFORM:scarlett;BUILD:20251016-9e31c9f-dirty;...
```

## Changes

- **[CMakeLists.txt](CMakeLists.txt#L323-L364)**: Added logic to parse build metadata from `SENTRY_VERSION_FULL` using regex
- **[src/sentry_embedded_info.cpp.in](src/sentry_embedded_info.cpp.in#L9-L11)**: Updated template to use `SENTRY_EMBEDDED_VERSION` and `SENTRY_EMBEDDED_BUILD_ID`
- **[tests/unit/test_embedded_info.c](tests/unit/test_embedded_info.c)**: Updated version test to validate base version format (no `+` character) and added new test for build ID field
- **[tests/unit/tests.inc](tests/unit/tests.inc#L72)**: Registered new `embedded_info_build_id` test

## Testing

- Unit tests verify that `SENTRY_VERSION` contains only base version without build metadata
- New test validates that `BUILD` field is present and non-empty
- Existing tests updated to handle both formats (with and without build metadata in source)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 **Internal**:
 
-- Automatically extract build ID from version string in `sentry_library_info` when downstream SDKs modify `sentry.h` to include build metadata in semver format (e.g., `0.11.3+20251016-9e31c9f-dirty`). The embedded info now correctly separates version (`SENTRY_VERSION:0.11.3`) from build ID (`BUILD:20251016-9e31c9f-dirty`). ([#1416](https://github.com/getsentry/sentry-native/pull/1416))
+- Automatically extract build ID from SENTRY_SDK_VERSION (`sentry.h`) when populating `sentry_library_info`. ([#1416](https://github.com/getsentry/sentry-native/pull/1416))
 
 ## 0.11.3
 


vaind:
@sentry review

vaind:
This is not good enough because it only sets the right values on the second cmake configure invocation (i.e. when the sentry.h has been patched). Instead, I think we'll need to provide a way to override the sdk version

PR #1427, opened 2025-10-22
feat(crashpad): Support optional usage of stack pointer for captured stack frame

mujacica:
* Add option to adjust stack limit based on SP
* Set option for crashpad

Related `crashpad` PR: https://github.com/getsentry/crashpad/pull/137


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - If you use a narrow string path interface (for instance, `sentry_options_set_database_path()`) on _Windows_ rather than one of the wide string variants (`sentry_options_set_database_pathw()`), then the expected encoding is now UTF-8. ([#1413](https://github.com/getsentry/sentry-native/pull/1413))
 
+**Features**:
+
+- Support optional usage of stack pointer for captured stack frame ([#1427](https://github.com/getsentry/sentry-native/pull/1427))
```
supervacuus: I suggest making this changelog entry more specific. Something like
```suggestion
- Add an option to use the stack pointer as an upper limit for the stack capture range in `crashpad` on Windows. This is useful for targets like Proton/Wine, where one can't rely on the TEB-derived upper bound being correctly maintained by the system, leading to overly large stack captures per thread. ([#1427](https://github.com/getsentry/sentry-native/pull/1427), [crashpad#137](https://github.com/getsentry/crashpad/pull/137))
```
highlights how specific this option really is.

include/sentry.h
```diff
@@ -1573,6 +1573,20 @@ SENTRY_API void sentry_options_set_system_crash_reporter_enabled(
 SENTRY_API void sentry_options_set_crashpad_wait_for_upload(
     sentry_options_t *opts, int wait_for_upload);
 
+/**
+ * Enables stack capture adjustment for the Crashpad backend.
+ * When enabled, Crashpad will adjust the stack capture range based on the
+ * current stack pointer instead of using TEB StackLimit/StackBase values.
```
supervacuus: Again, as in the changelog, I would make the wording more concrete here, since "adjust" can mean anything. In contrast, the specific solution is to use the `SP` as the upper bound of the stack capture, once it's been validated against `TEB` limits.

include/sentry.h
```diff
@@ -1573,6 +1573,20 @@ SENTRY_API void sentry_options_set_system_crash_reporter_enabled(
...
+ *
+ * This is useful when running under Wine/Proton where the TEB values may
+ * be incorrect, leading to excessively large stack captures. This is
+ * disabled by default.
+ *
+ * This setting only has an effect when using the `crashpad` backend on Windows.
+ */
+SENTRY_API void sentry_options_set_crashpad_adjust_stack_capture(
+    sentry_options_t *opts, int enabled);
```
supervacuus: There may be a more specific name for that option, too, but I also fail to find one that is similarly short.

supervacuus (approved):

PR #1430, opened 2025-10-28
feat(ndk): expose init return code

hannojg:
Return int from SentryNdk.init reflecting sentry_init result.
This can then be used by sentry-java/android to double check if the Ndk integration could be initialized successfully (which currently isn't happening, we just call the native method and assume it went successful).

Upstream PR in `getsentry/sentry-java` for android ndk:

- https://github.com/getsentry/sentry-java/pull/4842

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@
 - Introduce PAC tests for `arm64e` on macOS. ([#1446](https://github.com/getsentry/sentry-native/pull/1446))
 - For Linux, the SDK now has a vendored "nognu" `libunwind` as the default stack walker and links it statically, but with PIC enabled so it can be used in PIE executables. ([#1446](https://github.com/getsentry/sentry-native/pull/1446))
 
+**Thank you**:
+
+- [hannojg](https://github.com/hannojg)
+
 ## 0.12.8
 
 **Fixes**:
@@ -30,6 +34,8 @@
 - Fix deadlock when re-initializing the SDK while logs or metrics threads are mid-flush. ([#1518](https://github.com/getsentry/sentry-native/pull/1518))
 
 ## 0.12.7
+
+**Features**:
 
 - Add new offline caching options to persist envelopes locally: `sentry_options_set_cache_keep`, `sentry_options_set_cache_max_items`, `sentry_options_set_cache_max_size`, and `sentry_options_set_cache_max_age`. ([#1490](https://github.com/getsentry/sentry-native/pull/1490), [#1493](https://github.com/getsentry/sentry-native/pull/1493))
 


markushi:
@sentry review

ndk/lib/src/main/java/io/sentry/ndk/SentryNdk.java
```diff
@@ -19,9 +22,9 @@ private SentryNdk() {}
    *
    * @param options the SentryAndroidOptions
    */
-  public static void init(@NotNull final NdkOptions options) {
+  public static int init(@NotNull final NdkOptions options) {
     loadNativeLibraries();
-    initSentryNative(options);
+    return initSentryNative(options);
   }
```
markushi: I think we should keep using `void` here and throw an exception within init if the error code is non-0. This keeps API compatibility and ties the error code handling closer to the actual C code. 

We also don't need any extra handling within sentry-java then, as it's covered [by this try-catch here](https://github.com/getsentry/sentry-java/blob/03dca4eeaa9a9f2e2fc79e1e98abfff11827c8ef/sentry-android-core/src/main/java/io/sentry/android/core/NdkIntegration.java#L52-L68). 

```suggestion
  public static void init(@NotNull final NdkOptions options) {
    loadNativeLibraries();
    final int returnCode = initSentryNative(options);
    if (returnCode != 0) {
        throw new IllegalStateException("Failed to initialize sentry-native, return code: " + returnCode);
    }
  }
```

markushi (changes requested):
Thanks again you for contributing, much appreciated! I left a comment on how far the return code should be propagated up. If something is unclear, just let me know.

supervacuus:
Just my 2c here:

* I am fine if we don't throw from the `JNI` layer, however
* if we create a new protocol, then we should make it clear what things mean, especially the negative range currently does not provide any insight into the cause (we could use an enum to represent the failure states from the JNI layer)
* there is not a lot of overlap between the checks done in the JNI layer and the one in sentry_init, since all the checks here are classic client side checks, since for many option interfaces a `NULL` can be a valid value (meaning only the client knows whether that is an error (for instance, setting a transport to `NULL` is a valid configuration)

ndk/lib/src/main/java/io/sentry/ndk/SentryNdk.java
supervacuus: I would even go one step further here and check whether the return code was 1 (in which case there is some native-internal issue which could be resolved via logs and there is no need to render a return code) and having an enum range in the negative return code range, that results in an explanatory expection message for each particular return code.

Even if we don't go that far, we should at least differentiate between `> 0` and  `< 0` `returnCode` exceptions, to clarify where the error happened (in the `JNI` layer or native proper). Since the primary failure mode in the `JNI` layer is about allocation, most exceptions from there would not be actionable to client code. However, having a basic differentiation between the source of the exception, would still be useful.

PR #1431, opened 2025-10-28
chore: fix bump-version.sh matching spaces in sentry.h

JoshuaMoelans:
Fixing the release workflow as became apparent in https://github.com/getsentry/publish/issues/6522#issuecomment-3456923230

Spaces were introduced in https://github.com/getsentry/sentry-native/pull/1417/files#diff-9c77242571e813d699b88663fdc7d7ad7bb5b481a3c1af8d34e1a9e242cbf076R82

#skip-changelog

### DIFF
--- a/scripts/bump-version.sh
+++ b/scripts/bump-version.sh
@@ -14,7 +14,7 @@ NEW_VERSION="$2"
 
 echo "Bumping version: ${NEW_VERSION}"
 
-perl -pi -e "s/^#    define SENTRY_SDK_VERSION.*/#    define SENTRY_SDK_VERSION \"${NEW_VERSION}\"/" include/sentry.h
+perl -pi -e "s/^(#\s*)define SENTRY_SDK_VERSION.*/\$1define SENTRY_SDK_VERSION \"${NEW_VERSION}\"/" include/sentry.h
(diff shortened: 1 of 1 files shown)


scripts/bump-version.sh
```diff
@@ -14,7 +14,7 @@ NEW_VERSION="$2"
 
 echo "Bumping version: ${NEW_VERSION}"
 
-perl -pi -e "s/^#define SENTRY_SDK_VERSION.*/#define SENTRY_SDK_VERSION \"${NEW_VERSION}\"/" include/sentry.h
+perl -pi -e "s/^#    define SENTRY_SDK_VERSION.*/#    define SENTRY_SDK_VERSION \"${NEW_VERSION}\"/" include/sentry.h
```
supervacuus: You could also make the regex independent of the clang-format indent by capturing the spaces between `#` and `define`, then using them in the output string... something like
```suggestion
perl -pi -e "s/^(#\s*)define SENTRY_SDK_VERSION.*/${1}define SENTRY_SDK_VERSION \"${NEW_VERSION}\"/" include/sentry.h
```

JoshuaMoelans: good call, I adapted the suggestion in https://github.com/getsentry/sentry-native/pull/1431/commits/9d8180d0ea0483b3974919b86a1f8ee8c20e66ec & tested it locally 👌

Issue #1405, opened 2025-10-07
Structured Logs: follow up - custom attributes


JoshuaMoelans:
Part of https://github.com/getsentry/sentry-native/issues/1398
The [spec](https://develop.sentry.dev/sdk/telemetry/logs/#log-envelope-item-payload) allows for key-value paired attributes which are added as key : {"value": value, "type": type} to the 'attributes' item in the envelope. This is possible using before_send_log, but not convenient (as one would need to inspect the log to add log-specific attributes).

We could approach this in a few different ways:

## Attributes-only log function

Since downstream does not need string-interpolation (as this is [handled from their side](https://github.com/getsentry/sentry-native/issues/1405#issuecomment-3376626425)), we could provide a simple attributes-only log function

+ Add option to disable string interpolation (both for parsing arguments as well as writing the `body` field).

```cpp
// construct attributes dict
sentry_value_t attributes = sentry_value_new_object();
// create item (e.g. message parameter)
sentry_value_t message_param = sentry_value_new_object();
sentry_value_set_by_key(message_param, "value", sentry_value_new_string("login"));
sentry_value_set_by_key(message_param, "type", sentry_value_new_string("string"));
sentry_value_set_by_key(attributes, "sentry.message.parameter.0", action_attr);

sentry_log_debug("I'm a debug string!", attributes);
```

## Reuse existing API

We can try to parse an additional `sentry_value_t` at the end of the parameters, but since this might or might not be there, this could be tricky.

```cpp
// construct attributes dict
sentry_value_t attributes = sentry_value_new_object();
// create item
sentry_value_t action_attr = sentry_value_new_object();
sentry_value_set_by_key(action_attr, "value", sentry_value_new_string("login"));
sentry_value_set_by_key(action_attr, "type", sentry_value_new_string("string"));
sentry_value_set_by_key(attributes, "user.action", action_attr);

sentry_log_debug("I'm a %s string!", "debug", attributes);
```

## Implement additional API

A simpler solution (although it doubles the API surface) would be to add logs functions that take in the `attributes` as the first argument, such that under the hood we can still call into the existing `sentry_logs_X` functions, and then just add the given attributes to the constructed log.

```cpp
// construct attributes dict
sentry_value_t attributes = sentry_value_new_object();
// create item
sentry_value_t action_attr = sentry_value_new_object();
sentry_value_set_by_key(action_attr, "value", sentry_value_new_string("login"));
[...]

PR #1435, opened 2025-10-30
feat(logs): custom attributes API

JoshuaMoelans:
fixes https://github.com/getsentry/sentry-native/issues/1405

docs PR: https://github.com/getsentry/sentry-docs/pull/15491#issue-3617014475

With the `logs_with_attributes` option set to true, we can still parse the message string as a format string, but we take the first varg passed into `sentry_log_X` as an attributes object. 
If the passed-in value is not a `sentry_value_t` **_object_**, we will just ignore it (as long as it is still a `sentry_value_t` type, otherwise we hit `UNREACHABLE("invalid value type");` when trying to parse it)

```c
sentry_options_set_logs_with_attributes(options, true);
...
sentry_value_t attributes = sentry_value_new_object();
sentry_value_t attr = sentry_value_new_attribute(sentry_value_new_string("my_attribute"), NULL);
sentry_value_t attr_2 = sentry_value_new_attribute(sentry_value_new_int64(INT64_MAX), "fermions");
sentry_value_t attr_3 = sentry_value_new_attribute(sentry_value_new_int64(INT64_MIN), "bosons");
sentry_value_set_by_key(attributes, "my.custom.attribute", attr);
sentry_value_set_by_key(attributes, "number.first", attr_2);
sentry_value_set_by_key(attributes, "number.second", attr_3);

sentry_log_debug("logging with %d custom attributes", attributes, 3);

sentry_log_debug("logging with %s custom attributes", sentry_value_new_object(), "no");
```

Which shows up in the UI as such:

Screenshot 2025-10-31 at 13 35 00

This is also useful for downstream SDKs which might not need string interpolation, but still want to pass down message.parameter.X attributes:
```c
sentry_value_t param_attributes = sentry_value_new_object();
sentry_value_t param_attr = sentry_value_new_attribute(sentry_value_new_string("parameter"), NULL);
sentry_value_set_by_key(param_attributes,"message.parameter.0", param_attr);
sentry_log_fatal("logging with a custom parameter attributes", param_attributes);
```

‼️ custom attributes take precedence over the [default attributes](https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes), except for `sentry.message.parameter.X` values; if you pass in a string with format specifiers, these are used for the message parameter values.
```c
sentry_value_t param_attributes = sentry_value_new_object();
sentry_value_t param_attr = sentry_value_new_attribute(sentry_value_new_string("parameter"), NULL);
sentry_value_t param_attr_2 = sentry_value_new_attribute(sentry_value_new_string("custom-sdk-name"), NULL);
sentry_value_set_by_key(param_attributes, "sentry.message.parameter.0", param_attr);
sentry_value_set_by_key(param_attributes, "sentry.sdk.name", param_attr_2);
sentry_log_fatal("logging with a custom parameter %s attributes", param_attributes, "and format-string");
```
--> the custom-sdk-name shows up, but the `sentry.message.parameter.0` value is `"and format-string"`, not `"parameter"`.
Screenshot 2025-11-03 at 15 11 55

### DIFF
--- a/examples/example.c
+++ b/examples/example.c
@@ -508,6 +508,8 @@ main(int argc, char **argv)
         return EXIT_FAILURE;
     }
 
+    sentry_get_crashed_last_run();
+
     if (has_arg(argc, argv, "log-attributes")) {
         sentry_value_t attributes = sentry_value_new_object();
         sentry_value_t attr = sentry_value_new_attribute(
@@ -520,7 +522,8 @@ main(int argc, char **argv)
         sentry_value_set_by_key(attributes, "number.first", attr_2);
         sentry_value_set_by_key(attributes, "number.second", attr_3);
         // testing multiple attributes
-        sentry_log_debug("logging with %d custom attributes", attributes, 3);
+        sentry_log_debug(
+            "logging with %d custom attributes", attributes, (int64_t)3);
         // testing no attributes
         sentry_log_debug("logging with %s custom attributes",
             sentry_value_new_object(), "no");


src/sentry_logs.c
```diff
@@ -677,25 +677,50 @@ construct_log(sentry_level_t level, const char *message, va_list args)
...
+            va_copy(args_copy_3, args);
+            int len = vsnprintf(NULL, 0, message, args_copy_1) + 1;
+            va_end(args_copy_1);
+            size_t size = (size_t)len;
+            char *fmt_message = sentry_malloc(size);
+            if (!fmt_message) {
+                va_end(args_copy_2);
+                va_end(args_copy_3);
+                return sentry_value_new_null();
+            }
```
JoshuaMoelans, line 691: Currently we take an either-or approach when it comes to format strings VS custom attributes. We could also allow for both with the same option, where we still assume the first argument is a `sentry_value_t`, but then any following `vargs` can still make up the format string pieces.

Then people could call either:
```c
sentry_value_debug("this has %s attributes", sentry_value_new_null(), "no");
// or
sentry_value_t attributes = sentry_value_new_object();
sentry_value_t attr = sentry_value_new_attribute(
    "string", sentry_value_new_string("my_attribute"), NULL);
sentry_value_set_by_key(attributes, "my.custom.attribute", attr);

sentry_value_debug("this has %s attributes", attributes, 1);
```

JoshuaMoelans: attempt at doing both was added in https://github.com/getsentry/sentry-native/pull/1435/commits/b84a65301e9edc39b53f778269ac9da19228aa05

limbonaut:
I gave it a spin in the Godot SDK, and it works fine!

image

examples/example.c
```diff
@@ -503,11 +500,50 @@ main(int argc, char **argv)
...
+        // TODO only sentry.message.parameter.X gets picked up by product
+        //  this will be in the envelope but not searchable!
+        sentry_value_set_by_key(
+            param_attributes, "message.parameter.0", param_attr);
+        sentry_value_set_by_key(
+            param_attributes, "sentry.sdk.name", param_attr_2);
+        // TODO discuss; param_attributes[message.parameter.0] gets overwritten
+        //  by 'and format string'. Is this expected?
+        sentry_log_fatal("logging with a custom parameter %s attributes",
+            param_attributes, "and format-string");
```
JoshuaMoelans: IMO we can overwrite all default attributes except for string format values that are passed in; otherwise there would be a mismatch between the string template + `sentry.message.parameter.X` and the log `body` value

src/sentry_logs.c
```diff
@@ -677,25 +672,66 @@ construct_log(sentry_level_t level, const char *message, va_list args)
...
+            va_copy(args_copy, args);
+            sentry_value_t custom_attributes
+                = va_arg(args_copy, sentry_value_t);
+            va_end(args_copy);
+            if (sentry_value_get_type(custom_attributes)
+                == SENTRY_VALUE_TYPE_OBJECT) {
+                // TODO do we want to inspect that the object is attribute-like?
+                sentry_value_decref(attributes);
+                attributes = sentry__value_clone(custom_attributes);
+            } else {
```
JoshuaMoelans, line 688: I'm not sure how far we need to go to protect from user error. An invalid attribute (no "type" and "value" keys) seems to just get ignored in processing (the log still makes it to the UI)

examples/example.c
limbonaut: I think using both the attributes to pass `message.parameter.0` and the sentry-native interpolation API is already a misuse. I don't think we need to be protective about it. Let it be an undefined behavior.

src/sentry_logs.c
```diff
@@ -708,14 +744,6 @@ construct_log(sentry_level_t level, const char *message, va_list args)
     // to the log
```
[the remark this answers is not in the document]
JoshuaMoelans: There is no exclusive lock over the options, since we only increment the refcount (and the options themselves are read-only). So there is no real performance impact here.

PR #1443, opened 2025-11-04
feat: add runtime API to query user consent requirement

### LINKED ISSUE #1441: Provide way to query user consent requirement at runtime
Sentry Native user consent is configured with options, but once the SDK has been initialized, there's no way to query if user consent is required at runtime. For the SDK and most users, this isn't an issue, but the Unreal plugin needs to expose a lot more context so that games can handle specific scenarios (e.g. telemetry for the number of users that have crash reporting enabled)

tustanivsky:
This PR adds a new public API function `sentry_user_consent_is_required()` that allows applications to query at runtime whether user consent is required. This complements the existing consent management APIs and addresses a gap where the consent requirement could only be queried during initialization with direct access to the options object.

Closes #1441

### DIFF
--- a/tests/unit/test_consent.c
+++ b/tests/unit/test_consent.c
@@ -72,7 +72,15 @@ SENTRY_TEST(query_consent_requirement)
     TEST_ASSERT(!!path);
     sentry__path_remove_all(path);
 
-    // Test when consent is NOT required
+    // Test default behavior when require_user_consent is not set
+    sentry_options_t *opts = sentry_options_new();
+    sentry_options_set_database_path(opts, SENTRY_TEST_PATH_PREFIX ".test-db");
+    sentry_options_set_dsn(opts, "http://foo@127.0.0.1/42");
+    sentry_init(opts);
+    TEST_CHECK_INT_EQUAL(sentry_user_consent_is_required(), 0);
+    sentry_close();
+
+    // Test when consent is explicitly NOT required
     init_not_consenting_sentry();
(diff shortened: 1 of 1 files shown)


tests/unit/test_consent.c
```diff
@@ -11,6 +11,16 @@ init_consenting_sentry(void)
     sentry_init(opts);
 }
 
+static void
+init_not_consenting_sentry(void)
+{
+    sentry_options_t *opts = sentry_options_new();
+    sentry_options_set_database_path(opts, SENTRY_TEST_PATH_PREFIX ".test-db");
+    sentry_options_set_dsn(opts, "http://foo@127.0.0.1/42");
+    sentry_options_set_require_user_consent(opts, false);
```
nwhite-riot, line 20: Should we have a test where `sentry_options_set_require_user_consent` is omitted entirely?

tustanivsky: Yes, that makes sense - added the corresponding check in https://github.com/getsentry/sentry-native/pull/1443/commits/9ad717ab2bee61608d53e170ea69530959384796.

src/sentry_core.c
```diff
@@ -456,6 +456,16 @@ sentry_user_consent_get(void)
     return rv;
 }
 
+int
```
mujacica, line 459: int needed or we could use bool here? require_user_consent is a boolean.

tustanivsky: The existing public API uses `int` for similar boolean-style returns so the idea was just to keep it consistent.

Issue #1448, opened 2025-11-12
Implement Global Attributes API


JoshuaMoelans:
Implement Scope setAttributes API as per: [https://develop.sentry.dev/sdk/telemetry/scopes/#setting-attributes](https://develop.sentry.dev/sdk/telemetry/scopes/#setting-attributes)

Partly implemented by (and waiting on) [https://github.com/getsentry/sentry-native/pull/1435](https://github.com/getsentry/sentry-native/pull/1435) (adding `attribute` creation API)

## ToDos
- [ ] setAttributes API on the global scope
- [ ] apply attributes from global scope to logs
- [ ] check if attributes should support lists already

PR #1450, opened 2025-11-14
feat: implement global attributes API

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/1448

### Todo
- [x] apply attributes to logs
- [x] think about local scope attributes (if there are no local scope log functions?)
- [x] extend & tests attributes to also allow lists of int/string/double
- [x] changelog

### DIFF
--- a/tests/unit/test_value.c
+++ b/tests/unit/test_value.c
@@ -537,7 +537,6 @@ SENTRY_TEST(value_attribute)
         "bytes");
     sentry_value_decref(string_attr_n);
 
-
     // Test list attribute types (arrays)
     sentry_value_t string_list = sentry_value_new_list();
     sentry_value_append(string_list, sentry_value_new_string("foo"));
@@ -546,12 +545,13 @@ SENTRY_TEST(value_attribute)
         = sentry_value_new_attribute(string_list, NULL);
     TEST_CHECK(
         sentry_value_get_type(string_list_attr) == SENTRY_VALUE_TYPE_OBJECT);
-    TEST_CHECK_STRING_EQUAL(sentry_value_as_string(
-                                sentry_value_get_by_key(string_list_attr, "type")),
+    TEST_CHECK_STRING_EQUAL(sentry_value_as_string(sentry_value_get_by_key(
+                                string_list_attr, "type")),
         "string[]");
     sentry_value_t string_list_value
         = sentry_value_get_by_key(string_list_attr, "value");
-    TEST_CHECK(sentry_value_get_type(string_list_value) == SENTRY_VALUE_TYPE_LIST);
+    TEST_CHECK(
+        sentry_value_get_type(string_list_value) == SENTRY_VALUE_TYPE_LIST);
     TEST_CHECK(sentry_value_get_length(string_list_value) == 2);
     sentry_value_decref(string_list_attr);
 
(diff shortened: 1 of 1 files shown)


include/sentry.h
```diff
@@ -1858,6 +1858,29 @@ SENTRY_API void sentry_scope_set_extra_n(sentry_scope_t *scope, const char *key,
...
+SENTRY_API void sentry_remove_attribute(const char *key);
+SENTRY_API void sentry_remove_attribute_n(const char *key, size_t key_len);
+SENTRY_API void sentry_scope_set_attribute(
+    sentry_scope_t *scope, const char *key, sentry_value_t attribute);
+SENTRY_API void sentry_scope_set_attribute_n(sentry_scope_t *scope,
+    const char *key, size_t key_len, sentry_value_t attribute);
+SENTRY_API void sentry_scope_remove_attribute(
+    sentry_scope_t *scope, const char *key);
+SENTRY_API void sentry_scope_remove_attribute_n(
+    sentry_scope_t *scope, const char *key, size_t key_len);
```
JoshuaMoelans: In terms of propagating attributes into a scope, these don't have much effect as of now. The only scope-capture function we provide is [`capture_event_with_scope`](https://github.com/getsentry/sentry-native/pull/1248/files#diff-9c77242571e813d699b88663fdc7d7ad7bb5b481a3c1af8d34e1a9e242cbf076R1532), which in the future will get attributes applied, but for now remains untouched.

For now, we could either introduce a `log_X_with_scope(scope, message, ...)` API, or we only provide attributes APIs for the global scope & create a follow-up issue w.r.t. local scope attributes (and local-scope functionality for logs/traces/metrics).

src/sentry_value.c
```diff
@@ -509,33 +509,65 @@ sentry_value_new_user(const char *id, const char *username, const char *email,
...
+        case SENTRY_VALUE_TYPE_INT64:
+        case SENTRY_VALUE_TYPE_UINT64:
+            return "integer[]";
+        case SENTRY_VALUE_TYPE_DOUBLE:
+            return "double[]";
+        case SENTRY_VALUE_TYPE_STRING:
+            return "string[]";
+        default:
+            return NULL;
+        }
```
JoshuaMoelans, line 554: Not sure if we should check if all list items adhere to the same type here, or maybe we even do this during list append (always check the previous item for its type, and then transitively they must all be the same?)

JoshuaMoelans: We cannot disallow adding different types to the same list without introducing a breaking change (we even have this in our [tests](https://github.com/getsentry/sentry-native/blob/0.12.1/tests/unit/test_value.c#L564-L567)).

IMO we should just document in our attributes API that the values should be the base types OR a list consisting of items with the same base type.

include/sentry.h
JoshuaMoelans: We moved the scope-based API into `sentry_scope.h` for now, until we have a reason to open this to the public.

src/sentry_value.c
JoshuaMoelans: We documented the expected values, so users should adhere to this (tbd if sending malformed attributes break relay, but will test this once relay actually parses list-style attributes)

PR #1452, opened 2025-11-17
chore: parametrize clang/gcc install for ubuntu CI

JoshuaMoelans:
We explicitly `sudo apt install` the `g++-X/gcc-X` or `clang-X` versions specified in our runner matrix.

We also update the gcc-9.5 runners to run on ubuntu-24.04 (since all gcc versions can be covered by this one runner)


.github/workflows/ci.yml
```diff
@@ -197,10 +210,17 @@ jobs:
...
         run: |
           sudo apt update
-          sudo apt install cmake clang-19 llvm g++-12 valgrind zlib1g-dev libcurl4-openssl-dev
+          # Install common dependencies
+          sudo apt install cmake llvm valgrind zlib1g-dev libcurl4-openssl-dev
+          # For GCC, install both gcc-X and g++-X. For Clang, only install clang-X (includes C++ compiler)
+          if [[ "$CC" == gcc-* ]]; then
+            sudo apt install "${CC}" "${CXX}"
+          elif [[ "$CC" == clang-* ]]; then
+            sudo apt install "${CC}"
```
supervacuus, line 234: An else-path that fails here.

.github/workflows/ci.yml
```diff
@@ -197,10 +210,17 @@ jobs:
...
+          fi
           # Install kcov from source
           sudo apt-get install binutils-dev libssl-dev libelf-dev libstdc++-12-dev libdw-dev libiberty-dev
```
supervacuus: This is also a hard-coded libstdc++ version.

.github/workflows/ci.yml
supervacuus: (or a separate step as you did in https://github.com/getsentry/sentry-native/pull/1452/commits/43534802e1c025702f7173b9ef3d65c61446d591)

PR #1319, opened 2025-07-20
ref: crashpad: pre-generate crash event ID

jpnurmi:
Pre-generate an event ID for a potential future crash on the client side to be able to associate feedback with the crash event at the time of the crash. Extracted from #1303.

Before, Crashpad events (`__sentry-event`) were sent without ID. Even if the server generated an ID on demand and sent it as a response, it was too late for the upcoming feedback handler / crash reporter (#1223). It needs an ID before the event is sent because the user might not give consent to send anything at all.

**Before** <details><summary>`msgpack2json -p -i <run>/__sentry-event`</summary>

```json
{
    "level": "fatal",
    "platform": "native",
    "release": "sentry-playground@1.2.3",
    "environment": "development",
    "sdk": {
        "name": "sentry.native",
        "version": "0.9.1",
        "packages": [
            {
                "name": "github:getsentry/sentry-native",
                "version": "0.9.1"
            }
        ],
        "integrations": [
            "crashpad"
        ]
    },
    "tags": {
        "backend": "crashpad"
    },
    "extra": {},
    "contexts": {
        "os": {
            "name": "macOS",
            "version": "15.5.0",
            "build": "24F74",
            "kernel_version": "24.5.0"
        },
        "trace": {
            "trace_id": "be7045b229de43014da2b15ceb4dbedf",
            "span_id": "4a800c2ce1b24ebf",
            "sample_rand": 0.3444084273172463
        }
    }
}
```
</details>

**After** <details><summary>`msgpack2json -p -i <run>/__sentry-event`</summary>

```json
{
    "event_id": "62e3e0f8-d0d1-450a-1f6e-89faf1fe3862",
    "level": "fatal",
    "platform": "native",
    "release": "sentry-playground@1.2.3",
    "environment": "development",
    "sdk": {
        "name": "sentry.native",
        "version": "0.9.1",
        "packages": [
            {
                "name": "github:getsentry/sentry-native",
                "version": "0.9.1"
            }
        ],
        "integrations": [
            "crashpad"
        ]
    },
    "tags": {
        "backend": "crashpad"
    },
    "extra": {},
    "contexts": {
        "os": {
            "name": "macOS",
            "version": "15.5.0",
            "build": "24F74",
            "kernel_version": "24.5.0"
        },
        "trace": {
            "trace_id": "10b740a911374e5d4d918308abed7cd2",
            "span_id": "2f040e2ccb5b4460",
            "sample_rand": 0.502786153155452
        }
    }
}
```

PR #1488, opened 2026-01-22
ref(crashpad): pass predefined report ID

jpnurmi:
This PR is a preparation step for offline caching. It passes a `--report-id=<uuid>` command-line argument to the crashpad handler to make crashpad use predefined report IDs instead of generating random IDs.

This eliminates a discrepancy between sentry-native and crashpad by allowing both operate on the same event/report IDs. Since getsentry/sentry-native#1319, sentry-native pre-generates event IDs (stored in `__sentry-event`) on startup, and this change makes crashpad use the same ID as report ID. This way, sentry-native's crashpad backend can maintain sentry-specific event-related state without having to communicate everything to crashpad, which makes it easier to convert completed crashpad reports back to sentry envelopes for backend-agnostic offline caching.

See also:
- getsentry/crashpad#144
- getsentry/sentry-native#1316
- getsentry/sentry-native#1319

### DIFF
--- a/tests/assertions.py
+++ b/tests/assertions.py
@@ -496,3 +496,14 @@ def assert_failed_proxy_auth_request(stdout):
         and "407 Proxy Authentication Required" in stdout
         and "200 OK" not in stdout
     )
+
+
+def wait_for_file(path, timeout=10.0, poll_interval=0.1):
+    import time
+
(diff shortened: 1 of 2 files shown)


jpnurmi:
Side-note: to be merged after the release

PR #1490, opened 2026-01-23
feat: offline caching (inproc & breakpad)

jpnurmi:
## Summary

Adds offline caching support for the `inproc` and `breakpad` backends, allowing envelopes to be persisted locally for debugging and offline scenarios. When enabled, envelopes are retained in a `cache/` subdirectory within the database path, regardless of whether the send succeeds or fails.

On startup, the cache is pruned based on the configured limits (max items, age, size), removing entries from oldest to newest until constraints are satisfied. The pruning algorithm is based on Crashpad's `prune_crash_reports.cc`.

## API

New options:

- **`sentry_options_set_cache_keep(opts, int enabled)`** - Enables/disables offline caching
- **`sentry_options_set_cache_max_items(opts, size_t items)`** - Maximum number of cache entries (default: 30)
- **`sentry_options_set_cache_max_size(opts, size_t bytes)`** - Maximum cache directory size (no default max size)
- **`sentry_options_set_cache_max_age(opts, time_t seconds)`** - Maximum age for cache entries (no default max age)

### Usage

```c
sentry_options_set_cache_keep(options, true);
sentry_options_set_cache_max_items(options, 10);              // 10 items
sentry_options_set_cache_max_size(options, 8 * 1024 * 1024);  // 8 MB
sentry_options_set_cache_max_age(options, 30 * 24 * 60 * 60); // 30 days
```

## Comparison

| Option | Native | Cocoa/iOS | Java/Android |
|--------|--------|-----------|--------------|
| **Enable/disable** | `cache_keep` (default: off) | Always on | Always on |
| **Max items** | `cache_max_items` (30) | `maxCacheItems` (30) | `maxCacheItems` (30) |
| **Max size** | `cache_max_size` (0) | - | - |
| **Max age** | `cache_max_age` (0) | -  ([90 days, hardcoded](https://github.com/getsentry/sentry-cocoa/blob/18c94bed0a11a2352292ddcad21c2ac2aa02dab3/Sources/Sentry/SentryFileManagerHelper.m#L966-L997)) | - |

## Limitations

- **Crashpad backend**: Not yet supported - coming separately in #1493
- **HTTP retry**: Not implemented in this PR

## Related

- Builds upon #1461 by @JoshuaMoelans
- Partial fix for #1316

## Test Plan

- [x] Unit tests for age-based and size-based cache pruning logic (`test_cache.c`)
- [x] Integration tests for end-to-end caching behavior (`test_integration_cache.py`)

---
Co-authored-by: @JoshuaMoelans
Closes: #1461


src/sentry_database.c
```diff
@@ -295,6 +315,123 @@ sentry__process_old_runs(const sentry_options_t *options, uint64_t last_crash)
...
+            sentry__path_remove_all(entries[i].path);
+        }
+        sentry__path_free(entries[i].path);
+    }
+
+    sentry_free(entries);
+    sentry__path_free(cache_dir);
+}
+
 static const char *g_last_crash_filename = "last_crash";
```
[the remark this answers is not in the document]
jpnurmi: ## Analysis: NOT a valid bug

The AI agent's analysis is **incorrect**.

### The Claim

The agent claims that if `sentry__path_rename()` fails, the original file won't be cleaned up, leading to duplicate submissions on next startup.

### Why It's Wrong

Looking at the full context of `sentry__process_old_runs()`:

**Lines 289-302** (the loop body):

    } else if (sentry__path_ends_with(file, ".envelope")) {
        sentry_envelope_t *envelope = sentry__envelope_from_path(file);

        if (options->cache_keep) {
            sentry_path_t *cached_file = sentry__path_join_str(
                cache_dir, sentry__path_filename(file));
            sentry__path_rename(file, cached_file);
            sentry__path_free(cached_file);
            continue;
        }
    }
    sentry__path_remove(file);

**Line 310** (after the loop):

    sentry__path_remove_all(run_dir);

Even if `sentry__path_rename()` fails and the `continue` skips `sentry__path_remove(file)`, the **entire run directory is deleted** by `sentry__path_remove_all(run_dir)` at line 310. This cleanup happens unconditionally after processing all files.

The file cannot persist across application restarts because the whole `.run` directory is removed.

### Minor Observation

While not a bug causing duplicate submissions, there is a code quality consideration: if rename fails, there's no warning logged, and the caching feature silently degrades (the envelope is sent but not cached). Adding error logging would improve observability, but this is not the critical issue the AI agent described.

src/sentry_database.h
```diff
@@ -78,6 +78,12 @@ bool sentry__run_clear_session(const sentry_run_t *run);
 void sentry__process_old_runs(
     const sentry_options_t *options, uint64_t last_crash);
 
+/**
+ * Cleans up the cache based on options.max_cache_size and
+ * options.max_cache_age.
+ */
+void sentry__cleanup_cache(const sentry_options_t *options);
```
JoshuaMoelans: (typo)
```suggestion
/**
 * Cleans up the cache based on options.cache_max_size and
 * options.cache_max_age.
 */
void sentry__cleanup_cache(const sentry_options_t *options);
```

include/sentry.h
```diff
@@ -1415,6 +1416,51 @@ SENTRY_API void sentry_options_set_symbolize_stacktraces(
 SENTRY_API int sentry_options_get_symbolize_stacktraces(
     const sentry_options_t *opts);
 
+/**
+ * Enables or disables storing envelopes in a persistent cache.
+ *
+ * When enabled, envelopes are written to a `cache/` subdirectory within the
+ * database directory and retained regardless of send success or failure.
+ * The cache is cleared on startup based on the cache_max_items, cache_max_size,
+ * and cache_max_age options.
```
mujacica, line 1425: We could add a default here as well to the comment

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -752,11 +814,29 @@ crashpad_backend_prune_database(sentry_backend_t *backend)
     // We want to eagerly clean up reports older than 2 days, and limit the
     // complete database to a maximum of 8M. That might still be a lot for
     // an embedded use-case, but minidumps on desktop can sometimes be quite
-    // large.
-    data->db->CleanDatabase(60 * 60 * 24 * 2);
+    // large. When offline caching is enabled, the cache_max_* options are
+    // used instead.
+    time_t max_age = 2 * 24 * 60 * 60; // 2 days
```
mujacica, line 792: Maybe extracting these  constants somewhere would make sense (instead of keeping them as crashpad-only defaults in a function)?

mujacica (approved):

src/backends/sentry_backend_crashpad.cpp
jpnurmi: These are just some legacy defaults that were used by the crashpad backend before offline caching was introduced. These defaults are specific to the crashpad backend and are not used anywhere else. The idea is to retain full backwards compatibility by using the old legacy defaults unless the user opts in for offline caching. I updated the comments - I hope it's clear now :)

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -744,19 +744,101 @@ crashpad_backend_last_crash(sentry_backend_t *backend)
...
+        , measured_size_(0)
+    {
+    }
+
+    bool
+    ShouldPruneReport(
+        const crashpad::CrashReportDatabase::Report &report) override
+    {
+        measured_size_ += static_cast<size_t>(report.total_size);
+        return max_size_ > 0 && measured_size_ > max_size_;
```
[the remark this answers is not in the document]
jpnurmi: Both `MaxSizePruneCondition` and `MaxItemsPruneCondition` are stateful. With three conditions in a nested `OR`, any ordering has the same short-circuit property — whichever is on the LHS can cause the others to be skipped. But this only ever happens for reports that are already being pruned by the triggering condition, so the final database state is always correct regardless of order.

Crashpad's own `GetDefault()` puts the size condition first because it only has two conditions (size + age), and age is stateless. With three conditions where two are stateful, there's no "correct" ordering — swapping just trades which stateful condition gets skipped.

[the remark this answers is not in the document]
jpnurmi: https://github.com/getsentry/sentry-native/pull/1490/commits/43450d27923743ba2d42b27852cb9cd861acd4ba

PR #1498, opened 2026-01-29
feat: metrics

### LINKED ISSUE #1453: Sentry Metrics for Native
Closely related to #1210 , we are soon to be adding Metrics as a new type of telemetry for our SDKs. 

These look somewhat similar to logs, but only take in `name, value, options`. They follow the same batching logic (flushing every 100 metrics/5s, have the same default attributes, and a `before_send_metric` hook to filter/modify metrics.

https://develop.sentry.dev/sdk/telemetry/metrics/

## Related work:
- #1210 <- contains the logs batching implementation (to be re-used for metrics)
- #1405 <- introduction of attributes
- #1448  <- global attributes should be applied to metrics as well

jpnurmi:
## Summary
Implements the Sentry Metrics feature for the native SDK, allowing applications to record counter, gauge, and distribution metrics that are batched and sent to Sentry: https://develop.sentry.dev/sdk/telemetry/metrics/.

Closes #1453

## Features
- **Three metric types:** counter, gauge, and distribution
- **Batching:** Metrics are buffered and sent in batches (same logic as logs - every 100 items or 5 seconds)
- **Attributes:** Automatic default attributes (environment, release, SDK info) plus custom user attributes
- **Trace context:** Metrics are automatically associated with the current trace/span
- **Hook:** `before_send_metric` callback for filtering or modifying metrics before sending
- Unit constants: Pre-defined `SENTRY_UNIT_*` constants for common telemetry units (byte, kilobyte, millisecond, second, etc.)
 - Parallel flush/close: Logs and metrics flush and shutdown operations run in parallel for faster `sentry_flush()` and `sentry_close()`

## Public API
```c
// Enable metrics
sentry_options_set_enable_metrics(options, true);

// Record metrics
sentry_metrics_count("api.requests", 1, sentry_value_new_null());
sentry_metrics_gauge("memory.usage", 1024, SENTRY_UNIT_BYTE, sentry_value_new_null());
sentry_metrics_distribution("response.time", 150.5, SENTRY_UNIT_MILLISECOND, attributes);
```

## Considerations
The current API requires passing `sentry_value_new_null()` when no custom attributes are needed. An alternative would be to provide simplified variants without the attributes parameter (e.g., `sentry_metrics_count()`) alongside explicit variants (e.g., `sentry_metrics_count_with_attributes()`). Feedback welcome on which approach is preferred.

## ~~Dependencies~~
- ~~#1495~~
- ~~#1497~~

## Test plan
- [x] Unit tests for all metric types (`tests/unit/test_metrics.c`)
- [x] Integration tests for HTTP transport (`tests/test_integration_http.py`)
- [x] Tests for custom attributes, before_send_metric hook, and disabled state
- [x] Crash-safe flush behavior tested

### DIFF
--- a/src/sentry_batcher.c
+++ b/src/sentry_batcher.c
@@ -282,11 +282,8 @@ batcher_thread_func(void *data)
 }
 
 void
-sentry__batcher_startup(
-    sentry_batcher_t *batcher, sentry_batch_func_t batch_func)
+sentry__batcher_startup(sentry_batcher_t *batcher)
 {
-    batcher->batch_func = batch_func;
-
     // Mark thread as starting before actually spawning so thread can transition
     // to RUNNING. This prevents shutdown from thinking the thread was never
     // started if it races with the thread's initialization.
--- a/src/sentry_metrics.c
+++ b/src/sentry_metrics.c
@@ -29,7 +29,7 @@ static sentry_batcher_t g_batcher = {
     .active_idx = 0,
     .flushing = 0,
     .thread_state = SENTRY_BATCHER_THREAD_STOPPED,
-    .batch_func = NULL,
+    .batch_func = sentry__envelope_add_metrics,
 };
 
 static const char *
@@ -145,7 +145,7 @@ sentry_metrics_distribution(
 void
 sentry__metrics_startup(void)
 {
-    sentry__batcher_startup(&g_batcher, sentry__envelope_add_metrics);
+    sentry__batcher_startup(&g_batcher);
 }
 
 bool
--- a/src/sentry_batcher.h
+++ b/src/sentry_batcher.h
@@ -45,8 +45,7 @@ typedef struct {
 
 bool sentry__batcher_flush(sentry_batcher_t *batcher, bool crash_safe);
 bool sentry__batcher_enqueue(sentry_batcher_t *batcher, sentry_value_t item);
-void sentry__batcher_startup(
-    sentry_batcher_t *batcher, sentry_batch_func_t batch_func);
+void sentry__batcher_startup(sentry_batcher_t *batcher);
 bool sentry__batcher_shutdown_begin(sentry_batcher_t *batcher);
 void sentry__batcher_shutdown_wait(sentry_batcher_t *batcher, uint64_t timeout);
 void sentry__batcher_flush_crash_safe(sentry_batcher_t *batcher);
--- a/src/sentry_logs.c
+++ b/src/sentry_logs.c
@@ -25,7 +25,7 @@ static sentry_batcher_t g_batcher = {
     .active_idx = 0,
     .flushing = 0,
     .thread_state = SENTRY_BATCHER_THREAD_STOPPED,
-    .batch_func = NULL,
+    .batch_func = sentry__envelope_add_logs,
 };
 
 static const char *
@@ -460,7 +460,7 @@ sentry_log_fatal(const char *message, ...)
 void
 sentry__logs_startup(void)
 {
-    sentry__batcher_startup(&g_batcher, sentry__envelope_add_logs);
+    sentry__batcher_startup(&g_batcher);
 }
 
 bool


include/sentry.h
```diff
@@ -2126,6 +2166,91 @@ typedef sentry_value_t (*sentry_before_send_log_function_t)(
...
+ * Metrics are automatically associated with the current trace and span if
+ * available. Default attributes (environment, release, SDK info) are attached
+ * automatically.
+ */
+
+/**
+ * Records a counter metric. Counters track incrementing values like
+ * request counts or error counts.
+ */
+SENTRY_EXPERIMENTAL_API sentry_metrics_result_t sentry_metrics_count(
```
tustanivsky, line 2236: [Android](https://github.com/getsentry/sentry-java/blob/868793533ca77743294a5d616c2678f29f0d45a9/sentry/src/main/java/io/sentry/metrics/IMetricsApi.java#L6) and [Cocoa](https://github.com/getsentry/sentry-cocoa/blob/3b7f30ab2dcfa4905bf98cb5080dd18c521c1941/Sources/Swift/Integrations/Metrics/SentryMetricsApi.swift#L15) SDKs expose a `unit` parameter for `count` metrics - should we add this for native as well?

jpnurmi: Good question. Some SDKs have a unit argument for the counter, while others don't. I also added it there at first, but decided to leave it out because it says in the docs:

> The unit of measurement (only used for `distribution` and `gauge`)

https://develop.sentry.dev/sdk/telemetry/metrics/#method-signatures

include/sentry.h
```diff
@@ -324,6 +324,45 @@ SENTRY_API sentry_value_t sentry_value_new_user_n(const char *id, size_t id_len,
...
+#define SENTRY_UNIT_EXABYTE "exabyte"
+#define SENTRY_UNIT_EXBIBYTE "exbibyte"
+
+/* Fraction units */
+#define SENTRY_UNIT_RATIO "ratio"
+#define SENTRY_UNIT_PERCENT "percent"
+
 /**
  * Creates a new attribute object.
  *  value is required, unit is optional.
```
JoshuaMoelans: (nit): we could add a note clarifying that we only support [specific units](https://develop.sentry.dev/sdk/telemetry/metrics/#method-signatures).
```suggestion
 *  value is required, unit is optional, but has to be one of the `SENTRY_UNIT_X` macros.
```

(for context, there was a thread on Slack about this, saying custom units are a future feature, and for now we should not restrict it with an ENUM. We can just provide these macros, and if users pass in something else, Relay will just put the unit as NONE.)

tests/test_integration_http.py
```diff
@@ -17,6 +17,7 @@
     split_log_request_cond,
     is_feedback_envelope,
     is_logs_envelope,
+    is_metrics_envelope,
```
JoshuaMoelans, line 20: (idea): this integration test file is becoming huge, maybe we should start splitting it up (e.g., additional files for metrics/logs/traces/proxy). I can create a follow-up PR 🧹

tests/unit/test_metrics.c
```diff
@@ -0,0 +1,384 @@
...
+#include "sentry_envelope.h"
+#include <string.h>
+
+#ifdef SENTRY_PLATFORM_WINDOWS
+#    include <windows.h>
+#    define sleep_ms(MILLISECONDS) Sleep(MILLISECONDS)
+#else
+#    include <unistd.h>
+#    define sleep_ms(MILLISECONDS) usleep(MILLISECONDS * 1000)
+#endif
```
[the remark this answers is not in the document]
jpnurmi: https://github.com/getsentry/sentry-native/pull/1498/commits/9478d36f166b9736dcf04eb3e53ce6ff91fbf27c

include/sentry.h
jpnurmi: The unit argument was removed from Cocoa:
- https://github.com/getsentry/sentry-cocoa/pull/7358

> During revision of the Method Signatures of the metrics API in the [develop docs](https://develop.sentry.dev/sdk/telemetry/metrics/#method-signatures) it was surfaced to me that count metrics do not accept a Metric Unit.

src/sentry_metrics.c
```diff
@@ -0,0 +1,195 @@
...
+        },
+        {
+            .index = 0,
+            .adding = 0,
+            .sealed = 0,
+        },
+    },
+    .active_idx = 0,
+    .flushing = 0,
+    .thread_state = SENTRY_BATCHER_THREAD_STOPPED,
```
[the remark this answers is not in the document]
jpnurmi: https://github.com/getsentry/sentry-native/pull/1498/commits/ca4a0e6091fec6e114eb23c64dc81bc583bee936

include/sentry.h
jpnurmi: @tustanivsky Are you ok with leaving this as is? See Sentry Cocoa [9.4.0](https://github.com/getsentry/sentry-cocoa/releases/tag/9.4.0):

> ### Breaking Changes
>
> - `Sentry.metrics.count(..)` does not support units, therefore the API was incorrectly defined. This breaking change applies to a method marked as experimental. (#7358)

src/sentry_batcher.c
```diff
@@ -281,11 +282,8 @@ batcher_thread_func(void *data)
 }
 
 void
-sentry__batcher_startup(
-    sentry_batcher_t *batcher, sentry_batch_func_t batch_func)
+sentry__batcher_startup(sentry_batcher_t *batcher)
```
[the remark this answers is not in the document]
jpnurmi: Both are initialized. The tests wouldn't pass if they weren't.

https://github.com/getsentry/sentry-native/blob/ca4a0e6091fec6e114eb23c64dc81bc583bee936/src/sentry_logs.c#L28

https://github.com/getsentry/sentry-native/blob/ca4a0e6091fec6e114eb23c64dc81bc583bee936/src/sentry_metrics.c#L32

include/sentry.h
tustanivsky: Yes, I think we can proceed without exposing the `unit` param for now.

PR #1504, opened 2026-02-05
chore(test): split up `test_integration_http.py` 

JoshuaMoelans:
makes it slightly easier to find tests & to extend them in the future.

#skip-changelog

### DIFF
--- a/tests/test_integration_logs.py
+++ b/tests/test_integration_logs.py
@@ -18,20 +18,11 @@ from .conditions import has_http, has_breakpad
 
 pytestmark = pytest.mark.skipif(not has_http, reason="tests need http")
 
-# fmt: off
-auth_header = (
-    f"Sentry sentry_key=uiaeosnrtdy, sentry_version=7, sentry_client=sentry.native/{SENTRY_VERSION}"
-)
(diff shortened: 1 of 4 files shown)


tests/test_integration_logs.py
```diff
@@ -0,0 +1,482 @@
...
+    assert_event,
+    assert_logs,
+)
+from .conditions import has_http, has_breakpad
+
+pytestmark = pytest.mark.skipif(not has_http, reason="tests need http")
+
+# fmt: off
+auth_header = (
+    f"Sentry sentry_key=uiaeosnrtdy, sentry_version=7, sentry_client=sentry.native/{SENTRY_VERSION}"
```
jpnurmi: Perhaps we could consider the auth header tested by the http tests, and leave it out from the logs and metrics tests? Feels unnecessary to repeat it here

JoshuaMoelans: yeah that makes sense, thanks for noticing this! updated in https://github.com/getsentry/sentry-native/pull/1504/changes/2ca1f09da099593187924a612757aaeaf2aa677b

PR #1517, opened 2026-02-13
chore(tests): skip tests relying on unsupported APIs on NX and PS platforms

tustanivsky:
Bumping `sentry-native` to `0.12.7` for Switch and PlayStation introduced the following issues:

- The cache tests added in #1490 and #1493 rely on `utime()` to modify file timestamps. This POSIX function is not available in either the Nintendo Switch or PlayStation SDKs which causes build failures when compiling `sentry_test_unit` for these platforms (see the linked [Switch](https://github.com/getsentry/sentry-switch/actions/runs/21978377671/job/63496602787) and [PlayStation](https://github.com/getsentry/sentry-playstation/actions/runs/21978973886/job/63496785253) CI failures).

- A new `path_rename` test exercises `sentry__path_rename` which ultimately calls `rename()` on Unix. The NX filesystem does not support `rename()` so this test also fails on [Switch](https://github.com/getsentry/sentry-switch/actions/runs/21981288830/job/63504281123).

Key changes:

- Guard `<utime.h>` include and `set_file_mtime()` implementation with platform checks to avoid linker errors
- Skip all `test_cache` tests on NX and PlayStation platforms where `utime()` is unavailable
- Skip `path_rename` test on NX, following the same pattern already used for other NX-unsupported tests in that file

#skip-changelog

### DIFF
--- a/tests/unit/test_path.c
+++ b/tests/unit/test_path.c
@@ -288,6 +288,9 @@ SENTRY_TEST(path_mtime)
 
 SENTRY_TEST(path_rename)
 {
+#if defined(SENTRY_PLATFORM_NX)
+    SKIP_TEST();
+#endif
     sentry_path_t *src
         = sentry__path_from_str(SENTRY_TEST_PATH_PREFIX ".rename-src");
     TEST_ASSERT(!!src);


jpnurmi (approved):
Do we run integration tests on consoles? Will `test_integration_cache.py` (using `os.utime`) also create issues?

Issue #1543, opened 2026-02-26
Vendored libunwind clashes with `-static-libgcc` when linking statically on Linux


limbonaut:
In [sentry-godot](https://github.com/getsentry/sentry-godot), we build sentry-native as a static library (`SENTRY_BUILD_SHARED_LIBS=OFF`) and link it into a shared [GDExtension](https://docs.godotengine.org/en/stable/tutorials/scripting/gdextension/what_is_gdextension.html) library. Our toolchain ([godot-cpp](https://github.com/godotengine/godot-cpp)) uses `-static-libgcc -static-libstdc++` for portability.

After upgrading to 0.13.0, the Linux build fails with:

```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/11/libgcc_eh.a(unwind-dw2.o): in function `_Unwind_Resume':
(.text+0x2b40): multiple definition of `_Unwind_Resume'; install/lib/libunwind.a(Resume.c.o): first defined here
collect2: error: ld returned 1 exit status
```

Without `libunwind.a` in the link, we get undefined symbols instead:

```
undefined symbol: _ULx86_64_init_local   (x86_64)
undefined symbol: _ULx86_init_local2     (x86_32)
```

The vendored `libunwind.a` includes C++ exception ABI symbols (`_Unwind_Resume`, etc.) that conflict with `libgcc_eh.a` (pulled in by `-static-libgcc`).

- Downstream PR https://github.com/getsentry/sentry-godot/pull/544

PR #1544, opened 2026-02-26
fix: remove C++ exception ABI from vendored libunwind build.

supervacuus:
fixes https://github.com/getsentry/sentry-native/issues/1543.


jpnurmi (approved):

limbonaut (approved):
All good, thanks for the quick fix!

PR #1547, opened 2026-03-02
chore: Skip `getenv_double` test on PlayStation

tustanivsky:
PlayStation doesn't provide `setenv`/`unsetenv` in its C library, causing linker errors when building the unit tests. The `getenv_double` test relies on these POSIX functions to set up environment variables, so it cannot run meaningfully on this platform.

This PR wraps the test body with `#if !defined(SENTRY_PLATFORM_PS)` to skip it on PlayStation while keeping the test active on all other platforms.

#skip-changelog

### DIFF
--- a/tests/unit/test_utils.c
+++ b/tests/unit/test_utils.c
@@ -430,7 +430,9 @@ SENTRY_TEST(dsn_auth_header_invalid_dsn)
 
 SENTRY_TEST(getenv_double)
 {
-#if !defined(SENTRY_PLATFORM_PS)
+#if defined(SENTRY_PLATFORM_PS)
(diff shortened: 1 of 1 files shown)


tests/unit/test_utils.c
```diff
@@ -430,6 +430,7 @@ SENTRY_TEST(dsn_auth_header_invalid_dsn)
 
 SENTRY_TEST(getenv_double)
 {
+#if !defined(SENTRY_PLATFORM_PS)
```
jpnurmi:
```suggestion
#if defined(SENTRY_PLATFORM_PS)
    SKIP_TEST();
#else
```

PR #1552, opened 2026-03-04
fix(logs): Allow null custom attributes without debug warning

limbonaut:
Passing null as custom attributes to log functions prints a spurious debug message ("Discarded custom attributes on log: non-object sentry_value_t passed in"). Null is a valid way to indicate "no custom attributes" — the metrics path in `construct_metric()` already handles this correctly.

Came up while working on [sentry-godot metrics support](https://github.com/getsentry/sentry-godot/pull/553). The Godot SDK has a shared `dictionary_to_attributes()` helper used by both logs and metrics. Ideally it returns null for empty dictionaries to avoid an unnecessary allocation, but the log path warned about it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 **Fixes**:
 
-- Allow null custom attributes without debug warning ([#1552](https://github.com/getsentry/sentry-native/pull/1552))
+- Allow null custom log attributes without debug warning ([#1552](https://github.com/getsentry/sentry-native/pull/1552))
 
 ## 0.13.1
 


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Fixes**:
+
+- Allow null custom attributes without debug warning ([#1552](https://github.com/getsentry/sentry-native/pull/1552))
```
jpnurmi:
```suggestion
- Allow null custom log attributes without debug warning ([#1552](https://github.com/getsentry/sentry-native/pull/1552))
```

jpnurmi:
Claude suggests using `/clang:-Wall` instead of `-Wall` with clang-cl. Unlike regular clang where `-Wall` is a curated subset, clang-cl's `-Wall` means `-Weverything` which includes `-Wswitch-enum`. 😵

Issue #1516, opened 2026-02-13
Feature request: set release and environment after init


jpnurmi:
Games want to initialize the SDK as early as possible to catch potential startup crashes. At the same time, the release and environment information may not be available immediately on startup, but resolving them requires using asynchronous APIs, for example.

The current workaround is to re-initialize the SDK once the information is available. This results in a whole lot of back and forth, shutting down and restarting everything, and can result in nasty issues:

- https://github.com/getsentry/sentry-native/issues/1515
- https://github.com/getsentry/sentry-unreal/issues/592

Is there a way we could provide a way to set these attributes after init, or allow some kind of "soft" re-initialization? An ongoing session with outdated release/environment attributes would have to be updated or restarted.

PR #1555, opened 2026-03-05
feat: add sentry_set_release/environment

jpnurmi:
Add post-init setters for release and environment, allowing apps to update these values after SDK init, without requiring a full shutdown and re-init.

The new values are stored on the scope (like user), keeping options immutable after init. Sessions pick up the new values when manually started after calling the setters.

Close: #1516

### DIFF
--- a/src/sentry_core.c
+++ b/src/sentry_core.c
@@ -890,12 +890,11 @@ sentry_remove_user(void)
 void
 sentry_set_release_n(const char *release, size_t release_len)
 {
-    SENTRY_WITH_OPTIONS (options) {
-        SENTRY_WITH_SCOPE_MUT (scope) {
-            sentry_free(scope->release);
-            scope->release = sentry__string_clone_n(release, release_len);
-            set_dynamic_sampling_context(options, scope);
-        }
+    SENTRY_WITH_SCOPE_MUT (scope) {
+        sentry_free(scope->release);
+        scope->release = sentry__string_clone_n(release, release_len);
+        sentry_value_set_by_key(scope->dynamic_sampling_context, "release",
+            sentry_value_new_string(scope->release));
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/sentry_core.c
```diff
@@ -885,6 +887,43 @@ sentry_remove_user(void)
...
+sentry_set_environment_n(const char *environment, size_t environment_len)
+{
+    SENTRY_WITH_OPTIONS (options) {
+        SENTRY_WITH_SCOPE_MUT (scope) {
+            sentry_free(scope->environment);
+            scope->environment
+                = sentry__string_clone_n(environment, environment_len);
+            set_dynamic_sampling_context(options, scope);
+        }
+    }
```
supervacuus: Do we truly need this here?
```suggestion
    SENTRY_WITH_SCOPE_MUT (scope) {
        sentry_free(scope->environment);
        scope->environment
            = sentry__string_clone_n(environment, environment_len);
        sentry_value_set_by_key(scope->dynamic_sampling_context,
               "environment", sentry_value_new_string(scope->environment));
    }
```

We could even hide the DSC internals with (if that is a concern)
```C
static void
update_dynamic_sampling_context(
    sentry_scope_t *scope, const char *key, const char *value)
{
    sentry_value_set_by_key(
        scope->dynamic_sampling_context, key, sentry_value_new_string(value));
}
``` 

But no need to repeat the entire DSC construction inside the scope mutation lock, if the only things that could be updated are introduced right here. Or am I missing something?

src/sentry_core.c
```diff
@@ -885,6 +887,43 @@ sentry_remove_user(void)
...
+void
+sentry_set_release_n(const char *release, size_t release_len)
+{
+    SENTRY_WITH_OPTIONS (options) {
+        SENTRY_WITH_SCOPE_MUT (scope) {
+            sentry_free(scope->release);
+            scope->release = sentry__string_clone_n(release, release_len);
+            set_dynamic_sampling_context(options, scope);
+        }
+    }
```
supervacuus: Same as with release.

PR #1566, opened 2026-03-07
feat: add `sentry_log()` for plain-string log messages

jpnurmi:
Add `sentry_log()` — a non-printf API that accepts a plain string body and explicit attributes for downstream SDKs.

The existing `sentry_log_*()` functions treat the message as a printf format string. When callers like sentry-godot pass pre-formatted user text as the message, any `%` characters cause undefined behavior. On MSVC, `%n` kills the process immediately.

Notice that using `"%s"` together with a pre-formatted string as the first parameter is not a viable workaround (getsentry/sentry-godot#577) because it would clobber the `sentry.message.parameter.0` attribute.

The new `sentry_log()` overload skips printf interpretation and parameter extraction entirely, preserving the caller's attributes as-is.

See also:
- getsentry/sentry-godot#576

### DIFF
--- a/tests/unit/test_logs.c
+++ b/tests/unit/test_logs.c
@@ -487,7 +487,7 @@ SENTRY_TEST(logs_plain_string_disabled)
 {
     SENTRY_TEST_OPTIONS_NEW(options);
     sentry_options_set_dsn(options, "https://foo@sentry.invalid/42");
-    // logs disabled by default
+    sentry_options_set_enable_logs(options, false);
 
     sentry_init(options);
 


tests/unit/test_logs.c
```diff
@@ -410,6 +410,96 @@ SENTRY_TEST(logs_custom_attributes_not_modified)
...
+
+    sentry_init(options);
+    sentry__logs_wait_for_thread_startup();
+
+    // Body with printf-dangerous characters stored literally
+    sentry_value_t attrs = sentry_value_new_object();
+    sentry_value_set_by_key(attrs, "my.key",
+        sentry_value_new_attribute(sentry_value_new_string("my_value"), NULL));
+    TEST_CHECK_INT_EQUAL(
+        sentry_log(SENTRY_LEVEL_INFO, "100% done %n %s", attrs), 0);
```
limbonaut: nit: can we use SENTRY_LOG_RETURN_SUCCESS?

tests/unit/test_logs.c
```diff
@@ -410,6 +410,96 @@ SENTRY_TEST(logs_custom_attributes_not_modified)
...
+    sentry_close();
+
+    TEST_CHECK(!validation_data.has_validation_error);
+}
+
+SENTRY_TEST(logs_plain_string_disabled)
+{
+    SENTRY_TEST_OPTIONS_NEW(options);
+    sentry_options_set_dsn(options, "https://foo@sentry.invalid/42");
+    // logs disabled by default
```
limbonaut: Default may change. Should it be explicit?

PR #1572, opened 2026-03-11
fix(android): mask signal during CHAIN_AT_START to survive chained Mono handler re-raises

jpnurmi:
## Summary
- On Android, mask the signal before invoking the chained handler so `SA_NODEFER` doesn't let re-raises kill the process
- Use raw `rt_sigprocmask` syscall to bypass Android's `libsigchain`, whose `sigprocmask` guard is only active inside its own special handlers
- After the chain: restore our handler via `sigaction`, consume any pending signal via `sigtimedwait`, and unmask
- Disables `CHAIN_AT_START` on Android API < 26 where the old debuggerd daemon kills via `SIGKILL`

## Context
With `CHAIN_AT_START` and `SA_NODEFER`, the chained Mono handler can reset the signal handler to `SIG_DFL` (via debuggerd's [`resend_signal()`](https://android.googlesource.com/platform/system/core/+/refs/tags/android-16.0.0_r1/debuggerd/handler/debuggerd_handler.cpp#580)) and re-raise. The re-raised signal is delivered immediately and kills the process before inproc can capture the crash.

This is gated to Android because only the Mono runtime (used on Android) does the reset+raise pattern. On desktop Linux, CoreCLR modifies the ucontext and returns without re-raising.

### DIFF
--- a/src/backends/sentry_backend_inproc.c
+++ b/src/backends/sentry_backend_inproc.c
@@ -1618,13 +1618,8 @@ process_ucontext(const sentry_ucontext_t *uctx)
         }
 
 #    ifdef SENTRY_PLATFORM_ANDROID
-        // Restore our handler via raw syscall to bypass libsigchain.
-        // resend_signal() sets SIG_DFL via libsigchain, which updates
-        // libsigchain's internal state but also the kernel disposition.
-        // A regular sigaction() call goes through libsigchain which may
-        // not propagate to the kernel if its internal state is stale.
-        syscall(SYS_rt_sigaction, uctx->signum, &g_sigaction, NULL,
-            sizeof(sigset_t));
+        // restore our handler after resend_signal() set SIG_DFL
+        sigaction(uctx->signum, &g_sigaction, NULL);
 
         // consume pending signal
         struct timespec timeout = { 0, 0 };


src/backends/sentry_backend_inproc.c
```diff
@@ -507,7 +507,16 @@ startup_inproc_backend(
...
+    // However, SA_NODEFER is incompatible with CHAIN_AT_START: when we
+    // chain to the runtime's signal handler (e.g. Mono), it may reset
+    // the signal to SIG_DFL and re-raise. With SA_NODEFER the re-raised
+    // signal is delivered immediately (killing the process) before our
+    // handler can regain control.
+    g_sigaction.sa_flags = SA_SIGINFO | SA_ONSTACK;
+    if (g_backend_config.handler_strategy
+        != SENTRY_HANDLER_STRATEGY_CHAIN_AT_START) {
+        g_sigaction.sa_flags |= SA_NODEFER;
+    }
```
supervacuus: Tbh, this feels quite like the hammer.

Could you test an approach like this inside the chain-at-start block of `process_ucontext()`

```C
sigset_t mask, old_mask;
sigemptyset(&mask);
sigaddset(&mask, uctx->signum);
sigprocmask(SIG_BLOCK, &mask, &old_mask);

invoke_signal_handler(uctx->signum, uctx->siginfo, (void *)uctx->user_context);

if (ip != get_instruction_pointer(uctx)
    || sp != get_stack_pointer(uctx)) {
    // No need to restore the signal mask here: sigreturn will 
    // restore it from the saved ucontext.
    return;
}

// once we know we own the signal, unmask again
sigprocmask(SIG_SETMASK, &old_mask, NULL);
```

jpnurmi: Thanks for the suggestion. I tested it on an emulator and faced a couple of issues:

1. `sigprocmask` seems to be intercepted by `libsigchain` on Android. A raw `syscall(SYS_rt_sigprocmask, SIG_BLOCK, ...)` works, though.

2. unmasking delivers a pending `SIGSEGV` and kills the process

   Mono restores `SIG_DFL` and re-raises `SIGSEGV`:
   ```c
   sigaction(SIGSEGV, &saved_default_handler, NULL);
   raise(SIGSEGV);
   ```

   With the signal blocked, the `raise()` creates a pending signal instead of being delivered immediately. When unmasked, the pending `SIGSEGV` is delivered, but the handler is now `SIG_DFL`, so it terminates the process before sentry-native can capture the crash.
   
   I guess leaving it unmasked would have more or less the same trade-off as the original `SA_NODEFER` removal approach, basically losing recursive crash detection. What if we would restore our handler, and consume the the pending signal with `sigtimedwait`?

jpnurmi:
```c
sigset_t mask, old_mask;
sigemptyset(&mask);
sigaddset(&mask, uctx->signum);
// bypass libsigchain on Android
syscall(SYS_rt_sigprocmask, SIG_BLOCK, &mask, &old_mask,
    sizeof(sigset_t));

invoke_signal_handler(uctx->signum, uctx->siginfo, (void *)uctx->user_context);

if (ip != get_instruction_pointer(uctx)
    || sp != get_stack_pointer(uctx)) {
    return;
}

// restore our handler
struct sigaction current;
sigaction(uctx->signum, NULL, &current);
if (current.sa_handler == SIG_DFL) {
    sigaction(uctx->signum, &g_sigaction, NULL);
}

// consume pending signal
struct timespec timeout = { 0, 0 };
sigtimedwait(&mask, NULL, &timeout);

// unmask
syscall(SYS_rt_sigprocmask, SIG_SETMASK, &old_mask, NULL,
    sizeof(sigset_t));
```

jpnurmi:
> Yeah, this makes sense, but is there no Mono test in the downstream integration tests? It is quite painful that the two runtimes differ so severely, given that we seem to lack any early warning for that particular config. Or did this only happen with .NET 10?

This happens with both .NET 9 and .NET 10. Unfortunately, the PR that takes chained signal handling into use and would have revealed this problem, has been on hold until now due to other issues in .NET 10.

- getsentry/sentry-dotnet#4676

src/backends/sentry_backend_inproc.c
supervacuus: > 1\. `sigprocmask` seems to be intercepted by `libsigchain` on Android. A raw `syscall(SYS_rt_sigprocmask, SIG_BLOCK, ...)` works, though.

What is the observed effect? If `libsigchain` is the callee of a `sigprocmask()` call and it happens from inside the signal handler, then it normally passes it on to the linked `sigprocmask()`, so yes... this will go through `libsigchain`, but what effect do you see?!

> ```c
> // restore our handler
> struct sigaction current;
> sigaction(uctx->signum, NULL, &current);
> if (current.sa_handler == SIG_DFL) {
>     sigaction(uctx->signum, &g_sigaction, NULL);
> }
> 
> // consume pending signal
> struct timespec timeout = { 0, 0 };
> sigtimedwait(&mask, NULL, &timeout);
> 
> // unmask
> syscall(SYS_rt_sigprocmask, SIG_SETMASK, &old_mask, NULL,
>     sizeof(sigset_t));
> ```

Do we need to restore our handler? It should be enough to just `sigwait()`/`sigtimedwait()` (which are both not async-signal-safe), since they both just block the calling thread until the pending signal arrives (which in our case is already there). Or do you mean for the case of reentering?

jpnurmi: > What is the observed effect? If `libsigchain` is the callee of a `sigprocmask()` call and it happens from inside the signal handler, then it normally passes it on to the linked `sigprocmask()`, so yes... this will go through `libsigchain`, but what effect do you see?!

The intercepted `sigprocmask` call returns 0 indicating success, but the signal isn't blocked. Works fine with a raw syscall.

> Do we need to restore our handler? It should be enough to just `sigwait()`/`sigtimedwait()` (which are both not async-signal-safe), since they both just block the calling thread until the pending signal arrives (which in our case is already there). Or do you mean for the case of reentering?

Yes, it's for trying to preserve the recursive crash detection with `SA_NODEFER`.

jpnurmi: > The intercepted `sigprocmask` call returns 0 indicating success, but the signal isn't blocked. Works fine with a raw syscall.

Could be this: https://cs.android.com/android/platform/superproject/main/+/main:art/sigchainlib/sigchain.cc;l=653-661

supervacuus: Yes, but we shouldn't even be reaching that point, because right before that, `libsigchain` checks whether we are in the signal handler (via TLS), and directly calls the "linked" `sigprocmask`.

The check for claimed signals is only relevant outside of signal handlers, because it could prevent special handlers installed by `libsigchain` from being blocked. 

But `sigprocmask` from within the handler would only defer delivery until the current signal handler leaves. At the point where any of our signal handlers or the .NET signal handler is invoked, all special handlers have already been invoked.

supervacuus: Okay, had a bit deeper look and it seems like the guard is only enabled while in one of the special handlers. So yes, we will enter exactly that block, which means we bypass `libsigchain` rules if we make a direct syscall here. This is not great, but since we keep it short and no special handler should be acting on these signal raises anyway, I think it's still a better solution than disabling `SA_NODEFER` wholesale.

Let's keep the syscall `#ifdef`-gated to Android and use the "real" `sigprocmask` otherwise.

src/backends/sentry_backend_inproc.c
```diff
@@ -1563,6 +1564,19 @@ process_ucontext(const sentry_ucontext_t *uctx)
         uintptr_t ip = get_instruction_pointer(uctx);
         uintptr_t sp = get_stack_pointer(uctx);
 
+        // Mask the signal so SA_NODEFER doesn't let re-raises from the chained
```
supervacuus, line 1583: Now I wonder if we shouldn't do the entire block pre/post only on Android. Because the whole raise topic doesn't seem to happen on any other target, right? The entire machinery exists only because .NET on Android raises the previous handler at the end of its handler. I mean, ideally, we could identify which runtime we run in, but until then, I would isolate it, because there is no reason to do the ceremony if the runtime handler doesn't raise by the end.

src/backends/sentry_backend_inproc.c
```diff
@@ -1563,6 +1564,14 @@ process_ucontext(const sentry_ucontext_t *uctx)
...
+        // handler to kill the process before we regain control.
+        sigset_t mask, old_mask;
+        sigemptyset(&mask);
+        sigaddset(&mask, uctx->signum);
+        // raw syscall to bypass libsigchain on Android
+        syscall(SYS_rt_sigprocmask, SIG_BLOCK, &mask, &old_mask, _NSIG / 8);
```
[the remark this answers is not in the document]
supervacuus: If we do this only on Android, `sizeof(sigset_t)` should be fine, and any weirdness re: platform-specific size and `NSIG` vs `_NSIG` is gone.

src/backends/sentry_backend_inproc.c
jpnurmi: Yeah, makes sense :+1:

src/backends/sentry_backend_inproc.c
```diff
@@ -1578,6 +1593,23 @@ process_ucontext(const sentry_ucontext_t *uctx)
             return;
         }
 
+#    ifdef SENTRY_PLATFORM_ANDROID
+        // restore our handler
```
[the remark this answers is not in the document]
supervacuus: JFC... go read a manpage... you are otherwise not squeamish when it comes to crawling everything. Directly from [`signal(7)`](https://man7.org/linux/man-pages/man7/signal.7.html) Section **Execution of signal handlers**:

```
       (1.3)  Various pieces of signal-related context are saved
                   into a special frame that is created on the stack.
                   The saved information includes:

                   •  the program counter register (i.e., the address of
                      the next instruction in the main program that
                      should be executed when the signal handler
                      returns);

                   •  architecture-specific register state required for
                      resuming the interrupted program;

                   •  the thread's current signal mask;

                   •  the thread's alternate signal stack settings.
```

and later:

```
       (4)  When the signal handler returns, control passes to the signal
            trampoline code.

       (5)  The signal trampoline calls [sigreturn(2)](https://man7.org/linux/man-pages/man2/sigreturn.2.html), a system call that
            uses the information in the stack frame created in step 1 to
            restore the thread to its state before the signal handler was
            called.  The thread's signal mask and alternate signal stack
            settings are restored as part of this procedure.  Upon
            completion of the call to [sigreturn(2)](https://man7.org/linux/man-pages/man2/sigreturn.2.html), the kernel transfers
            control back to user space, and the thread recommences
            execution at the point where it was interrupted by the signal
            handler.
```

i.e.,
1. The thread's current signal mask is saved before entering the signal handler
2. If the handler returns control passes to the trampoline
3. The trampoline calls `sigreturn`
4. `sigreturn` restores the thread's signal mask

src/backends/sentry_backend_inproc.c
```diff
@@ -1578,6 +1593,23 @@ process_ucontext(const sentry_ucontext_t *uctx)
             return;
         }
 
+#    ifdef SENTRY_PLATFORM_ANDROID
```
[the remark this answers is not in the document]
supervacuus: That is a nice catch, but the combination of `sentry_handle_exception()` and `CHAIN_AT_START` is in no way a supported or tested use case. I am happy to make `sentry_handle_exception()` a noop if `CHAIN_AT_START` is active.

src/backends/sentry_backend_inproc.c
```diff
@@ -1550,9 +1551,20 @@ dispatch_ucontext(const sentry_ucontext_t *uctx,
...
+    // CHAIN_AT_START invokes the previous handler and expects to regain
+    // control. On Android API < 26, the old debuggerd daemon kills the
+    // crashing process via SIGKILL after the chained handler triggers
+    // it, so we fall back to DEFAULT which chains at the end instead.
+    if (strategy == SENTRY_HANDLER_STRATEGY_CHAIN_AT_START
+        && android_get_device_api_level() < 26) {
+        strategy = SENTRY_HANDLER_STRATEGY_DEFAULT;
+    }
+#endif
+
```
supervacuus: If we truly want to override this in the Native SDK, then it should happen during the `inproc` initialization rather than `process_ucontext()`.

src/backends/sentry_backend_inproc.c
```diff
@@ -1579,6 +1617,23 @@ process_ucontext(const sentry_ucontext_t *uctx)
             return;
         }
 
+#    ifdef SENTRY_PLATFORM_ANDROID
+        // restore our handler
+        struct sigaction current;
+        sigaction(uctx->signum, NULL, &current);
+        if (current.sa_handler == SIG_DFL) {
+            sigaction(uctx->signum, &g_sigaction, NULL);
+        }
```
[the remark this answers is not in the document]
jpnurmi: Mono uses standard `sigaction()` for handler registration — libsigchain doesn't interfere with `sigaction` calls (unlike `sigprocmask`, which it silently drops outside its own handlers). There's no reason for Mono to bypass it with raw syscalls for handler registration. Our own raw syscall usage is specifically for `sigprocmask` to work around libsigchain's masking behavior, not for `sigaction`.

src/backends/sentry_backend_inproc.c
jpnurmi: I considered adding the check to sentry-dotnet instead, but it felt wrong for the caller to have to know that. And it would get out of sync should we ever want to address the issue with API < 26 devices (though unlikely, given the market share).

supervacuus: Yeah, makes sense. I have no problem with having it here, and now it's in the right place.

src/backends/sentry_backend_inproc.c
```diff
@@ -1579,6 +1617,24 @@ process_ucontext(const sentry_ucontext_t *uctx)
             return;
         }
```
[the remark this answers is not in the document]
jpnurmi: https://github.com/getsentry/sentry-native/pull/1572#discussion_r2923299047

PR #1573, opened 2026-03-11
fix(native): propagate transport options and handler path to crash daemon

mujacica:
The native backend daemon was not receiving SSL certificate, proxy, user agent, or handler path settings from the parent process options. This caused SSL errors (curl code 60) when the daemon tried to upload crash reports, and required users to place the sentry-crash binary next to their executable instead of using sentry_options_set_handler_path.

Pass ca_certs, proxy, user_agent, and handler_path through the shared memory crash context so the daemon can configure its curl transport and find the sentry-crash executable correctly.

### DIFF
--- a/src/backends/native/sentry_crash_daemon.c
+++ b/src/backends/native/sentry_crash_daemon.c
@@ -3164,69 +3164,48 @@ sentry__crash_daemon_start(pid_t app_pid, uint64_t app_tid, HANDLE event_handle,
         char *argv[]
             = { "sentry-crash", pid_str, tid_str, notify_str, ready_str, NULL };
 
-        // 0. If handler_path was explicitly set via options, try it first
+        // If handler_path was explicitly set via options, use it directly.
+        // Otherwise, look for sentry-crash next to the current executable
+        // (matching crashpad's behavior). No fallback chain — fail hard so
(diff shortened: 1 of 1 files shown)


src/backends/native/sentry_crash_daemon.c
```diff
@@ -3147,6 +3164,12 @@ sentry__crash_daemon_start(pid_t app_pid, uint64_t app_tid, HANDLE event_handle,
         char *argv[]
             = { "sentry-crash", pid_str, tid_str, notify_str, ready_str, NULL };
 
+        // 0. If handler_path was explicitly set via options, try it first
+        if (handler_path && handler_path[0] != '\0') {
+            execv(handler_path, argv);
+            // If execv fails, continue to fallback search
```
jpnurmi: For Windows, an explicitly specified handler path is used as-is. Was there some reason we didn't do the same on Unix? I have mixed feelings about the fallback because it might hide configuration issues.

mujacica: You are right, I aligned it now with the way we do it for crashpad - if handler path is set, we are going to use it. If not, we try to execute the binary next to the starting process - otherwise we fail with meaningful errors

PR #1598, opened 2026-03-26
test: reduce and optimize pytest builds

jpnurmi:
- Key build cache on options only (not targets), so builds with the same CMake options reuse the existing build directory and only build missing targets incrementally
  - For example, `cmake(["sentry_example"])`, `cmake(["sentry_example", "sentry_crashreporter"])`, and `cmake(["sentry_screenshot"])` with the same configure options share the same build directory and incrementally build the missing targets
- Skip `embedded_info` tests early (instead of after build) when the `strings` command is unavailable
- Drop redundant `SENTRY_TRANSPORT_COMPRESSION=Off` (the default) to reduce duplicate builds
- Fix `test_unit_with_test_path` to actually build with `-DSENTRY_TEST_PATH_PREFIX`: quote the value for MSVC compatibility, include `CFLAGS` in the build cache key, and prefix wide-char attachment paths for correct deduplication (this actually adds one more build, but it's the right thing to do)

Before (20 min) vs. after (15 min) on an old Surface Laptop 4:
```diff
 (.venv) PS C:\Users\jpnurmi\Projects\sentry\sentry-native> pytest -v .\tests\
 ============================= test session starts =============================
 platform win32 -- Python 3.13.3, pytest-8.1.1, pluggy-1.6.0 -- C:\Users\jpnurmi\Projects\sentry\sentry-native\.venv\Scripts\python.exe
 cachedir: .pytest_cache
 rootdir: C:\Users\jpnurmi\Projects\sentry\sentry-native
 plugins: flaky-3.8.1, pytest_httpserver-1.0.10, repeat-0.9.3, xdist-3.5.0
 collecting ... collected 1142 items
 [...]
-================ 1108 passed, 34 skipped in 1209.85s (0:20:09) ================
+================ 1108 passed, 34 skipped in 784.08s (0:14:28) =================
```

### DIFF
--- a/tests/test_embedded_info.py
+++ b/tests/test_embedded_info.py
@@ -58,7 +58,7 @@ def test_embedded_info_disabled(cmake):
         {
             "SENTRY_BACKEND": "none",
             "SENTRY_TRANSPORT": "none",
-            "SENTRY_EMBED_INFO": "OFF",
+            # "SENTRY_EMBED_INFO": "OFF" (default, cached)
         },
     )
 


supervacuus (approved):
Nice! Seems like Windows builds in CI still around 30 mins though. Well, one step at a time. Thx!

PR #1600, opened 2026-03-26
fix(native): Fix crash daemon premature exit on Windows

tustanivsky:
### Summary

The out-of-process crash daemon (`sentry-crash.exe`) was exiting after exactly 5 seconds (`SENTRY_CRASH_DAEMON_WAIT_TIMEOUT_MS`) on Windows, regardless of whether the parent process is still running.

### Root cause

The `is_parent_alive()` check used `OpenProcess(SYNCHRONIZE, ...)` but `GetExitCodeProcess()` [requires](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getexitcodeprocess#parameters) `PROCESS_QUERY_LIMITED_INFORMATION` access. Without it, `GetExitCodeProcess` always failed causing the daemon to conclude the parent had exited after the first 5-seconds poll cycle.

### Suggested fix

Add `PROCESS_QUERY_LIMITED_INFORMATION` to the `OpenProcess()` flags so `GetExitCodeProcess()` can actually query the parent's exit status.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 **Fixes**:
 
 - Fix external crash reporter to work with the new experimental `native` backend. ([#1589](https://github.com/getsentry/sentry-native/pull/1589))
+- Fix crash daemon premature exit on Windows ([#1600](https://github.com/getsentry/sentry-native/pull/1600))
 
 ## 0.13.3
 


jpnurmi (approved):
Ooh, good catch!

PR #1609, opened 2026-03-30
feat: enable metrics by default

JoshuaMoelans:
Metrics captured by the Public API should now be `opt-out` as per [develop docs](https://develop.sentry.dev/sdk/telemetry/metrics/#initialization-options).

Since we don't have any internal auto-instrumented metrics, we enable it by default so calling `sentry_metrics_X` will just capture instead of requiring setting the option. ⚠️  this does make it so our batching thread starts regardless of whether it actually has any work (see the discussion below).

Also updates the tests to reflect this change; by default it is true, but one can set it to false with `sentry_options_set_enable_metrics(options, false);` (for example, when only wanting metrics for debug and not production).

## follow-up
- update https://docs.sentry.io/platforms/native/metrics/ https://github.com/getsentry/sentry-docs/pull/17207#issue-4185415789
- the same for logs once that goes `opt-out` as well.

---

## Discussion
CI currently fails because we now log from our Metrics Batcher Thread that starts by default; we can fix this either by

a) changing those logging tests to expect the output (but then starting the metrics batcher even if no metric is ever sent) https://github.com/getsentry/sentry-native/pull/1612
b) adding a lazy init to the metrics/logs batcher https://github.com/getsentry/sentry-native/pull/1611

- think about whether we also want the metrics batcher to always be running, since `enable_metrics` just starts it; maybe we only want to do so once the first metric is captured? ⚠️  currently, if we don't do this, this changes the behavior of the SDK (a batching thread starts even if metrics wasn't explicitly enabled, since it is now `on` by default)
  - see https://github.com/getsentry/sentry-native/pull/1611


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Add HTTP retry with exponential backoff, opt-in via `sentry_options_set_http_retry`. ([#1520](https://github.com/getsentry/sentry-native/pull/1520))
 - Store minidump attachments as separate `.dmp` files in the offline cache for direct debugger access. ([#1607](https://github.com/getsentry/sentry-native/pull/1607))
+- Enable metrics by default; metrics are now opt-out via `sentry_options_set_enable_metrics(options, false)`. ([#1609](https://github.com/getsentry/sentry-native/pull/1609))
```
JoshuaMoelans, line 9: I'm not quite sure if we need to disclose the technical detail of it now starting the batcher thread by default.

jpnurmi: I think it's fine to leave it out. An irrelevant internal implementation detail? :)

PR #1613, opened 2026-03-31
feat(android): allow NDK preload for .NET/CoreCLR

jpnurmi:
.NET 10.0 added support for [CoreCLR on Android](https://learn.microsoft.com/en-us/dotnet/maui/whats-new/dotnet-10?view=net-maui-10.0#experimental-coreclr) (experimental). Unlike with the Mono runtime, where we're stuck with `CHAIN_AT_START` (dotnet/runtime#125835), with CoreCLR we can already reorder the signal handlers to the preferred order:

```
┌───────────┐     ┌───────────────┐     ┌────────────────────┐
│ .NET/Mono │────>│ Sentry Native │────>│ System (debuggerd) │
└───────────┘     └───────────────┘     └────────────────────┘
```

This gives the .NET runtime a chance to process managed .NET exceptions without Sentry Native stepping on its toes. This is much safer than `CHAIN_AT_START`, and will be automatically enabled (`io.sentry.ndk.preload=true`) by Sentry.NET when targeting Android .NET 10.0 or later with CoreCLR.

### Changes

Add `sentry__backend_preload()` for installing signal handlers before `sentry_init()`. For the inproc backend on Unix/Android, this allows establishing the correct signal handler chain order when another runtime (e.g. Mono/.NET) installs its own handlers later.

- Backend dispatch: `sentry__backend_preload()` (no-op for non-inproc)
- Extracted `install_signal_handlers()` with idempotency flag
- Preload fallthrough: if handler thread not ready, invoke previous handler
- NDK: `SentryNdkPreloadProvider` [content provider](https://developer.android.com/guide/topics/providers/content-providers) with high init order for early signal handler installation before the CoreCLR runtime
- NDK: Java/JNI `preload()` method on SentryNdk

### See also

- Original issue: getsentry/sentry-dotnet#3954
- Signal handler re-order for iOS: getsentry/sentry-cocoa#6193
- Chain-at-start alternative for Android on Mono: getsentry/sentry-native#1572
- Mono fix for future reorder: dotnet/runtime#125835

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add WinGDK (Gaming.Desktop.x64) platform support ([#1631](https://github.com/getsentry/sentry-native/pull/1631))
 - Track discarded events via client reports. ([#1549](https://github.com/getsentry/sentry-native/pull/1549))
+- Android: allow Sentry.NET to preload the NDK integration to install signal handlers before the .NET runtime. ([#1613](https://github.com/getsentry/sentry-native/pull/1613))
 
 ## 0.13.5
 
@@ -14,7 +15,6 @@
 - Add HTTP retry with exponential backoff, opt-in via `sentry_options_set_http_retry`. ([#1520](https://github.com/getsentry/sentry-native/pull/1520))
 - Store minidump attachments as separate `.dmp` files in the offline cache for direct debugger access. ([#1607](https://github.com/getsentry/sentry-native/pull/1607))
 - Enable metrics by default; metrics are now opt-out via `sentry_options_set_enable_metrics(options, false)`. ([#1609](https://github.com/getsentry/sentry-native/pull/1609))
-- Android: allow Sentry.NET to preload the NDK integration to install signal handlers before the .NET runtime. ([#1613](https://github.com/getsentry/sentry-native/pull/1613))
 
 **Fixes**:
 


ndk/lib/src/main/java/io/sentry/ndk/SentryNdkPreloadProvider.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Preloads the NDK integration before the Mono runtime. This installs and
+ * chains the Native SDK's signal handlers before Mono, allowing Mono to
+ * correctly handle managed exceptions while chaining native crashes to the
+ * Native SDK.
+ *
+ * <p>Enabled by setting {@code io.sentry.ndk.preload} to {@code true} in
+ * AndroidManifest.xml metadata. The high {@code initOrder} ensures this runs
+ * before {@code mono.MonoRuntimeProvider}.
+ */
```
supervacuus, line 29: I would get rid of the mono usage here, since it is kinda misleading, even if technically correct in some aspects. The thing we're introducing here is for `CoreCLR` at this point, and soon for Mono as well. Then we can further simplify the wording. Until then i would prefer something more along these lines:
```suggestion
/**
 * Preloads the NDK integration before the .NET runtime provider on Android.
 *
 * <p>This is intended for downstream SDK integrations that run with CoreCLR on
 * Android. By installing sentry-native before the managed runtime registers
 * its own signal handlers, native crash signals can chain from the runtime
 * back to the Native SDK, while runtime-generated fault signals can still be
 * consumed by the runtime for managed exception handling.
 *
 * <p>This is the preload alternative to CHAIN_AT_START. Mono on Android
 * continues to use CHAIN_AT_START.
 *
 * <p>Enabled by setting {@code io.sentry.ndk.preload} to {@code true} in the
 * app manifest metadata. The high {@code initOrder} ensures this runs before
 * the runtime provider emitted by dotnet/android.
 */
```

src/sentry_backend.h
```diff
@@ -32,6 +32,19 @@ struct sentry_backend_s {
...
+ *
+ * Currently only used from the NDK's SentryNdkPreloadProvider via JNI to
+ * preload the inproc backend on Android. This installs and chains the Native
+ * SDK's signal handlers before the Mono runtime, allowing Mono to correctly
+ * handle managed exceptions while chaining native crashes to the Native SDK.
+ *
+ * If a crash occurs before sentry_init() is called, the handler will fall
+ * through to the previously installed handler.
+ */
+SENTRY_API void sentry__backend_preload(void);
```
supervacuus, line 55: I think this is a good place to further clarify the contract constraints and again tone down the wording towards Mono.
```suggestion
/**
 * Backend-specific pre-initialization that can be called before sentry_init().
 *
 * Currently used by downstream SDKs on Android to preload the inproc backend
 * before the .NET runtime installs its own signal handlers. This is the
 * preload alternative to CHAIN_AT_START for runtimes where the handler order
 * can be established at process startup (e.g., CoreCLR on Android).
 *
 * Preload only establishes signal-chain order. Full crash processing becomes
 * active once sentry_init() starts the handler thread.
 *
 * If a crash/signal occurs before sentry_init() is called, or after a
 * subsequent sentry_close() while the preload chain position is still kept,
 * the preloaded handler will fall through to the previously installed handler.
 *
 * This API is intended for downstream SDK/runtime integrations, not as a
 * general app-level initialization entry point. Callers are expected to gate this to 
 * supported runtime configurations (e.g., no preload together with 
 * CHAIN_AT_START).
 */
```

src/backends/sentry_backend_inproc.c
```diff
@@ -503,39 +538,18 @@ startup_inproc_backend(
...
 shutdown_inproc_backend(sentry_backend_t *backend)
 {
     stop_handler_thread();
 
-    teardown_sigaltstack(&g_signal_stack);
-    reset_signal_handlers();
+    if (!sentry__atomic_fetch(&g_preloaded)) {
+        teardown_sigaltstack(&g_signal_stack);
+        reset_signal_handlers();
+    }
```
supervacuus, line 565: Let's clarify this life-cycle here in particular:
```suggestion
    // In Android preload mode, we intentionally keep our signal handlers
    // installed across shutdown. The handler thread is torn down, but our
    // position in the signal chain must remain stable so that a later
    // sentry_init() can reactivate full crash handling without losing the
    // ordering relative to the managed runtime.
    //
    // While in this state, full inproc crash processing is inactive and any
    // signal we still observe will fall through to the previously installed
    // handler.
    if (!sentry__atomic_fetch(&g_preloaded)) {
        teardown_sigaltstack(&g_signal_stack);
        reset_signal_handlers();
    }
```

tests/test_dotnet_signals.py
```diff
@@ -317,26 +321,51 @@ def run_android(args=None, timeout=30):
...
     reason="needs Android API 26+ (tombstoned)",
 )
-def test_android_signals_inproc(cmake):
+@pytest.mark.parametrize(
+    "runtime,strategy",
+    [
+        ("mono", "chain-at-start"),
+        ("coreclr", "preload"),
+    ],
+)
```
supervacuus, line 360: Make the contract visible for all maintainers.
```suggestion
# Mono on Android keeps using CHAIN_AT_START. Preload is the CoreCLR path,
# where sentry-native can enter the signal chain before the runtime installs
# its own handlers.
@pytest.mark.parametrize(
    "runtime,strategy",
    [
        ("mono", "chain-at-start"),
        ("coreclr", "preload"),
    ],
)
```

tests/test_dotnet_signals.py
```diff
@@ -424,19 +460,23 @@ def has_envelope():
             )
             return bool(result.stdout.strip())
 
+        # Preload installs signal handlers before the runtime, so the
+        # handler strategy should be DEFAULT (0) instead of CHAIN_AT_START.
```
supervacuus: Again, explicit contract for all:
```suggestion
        # Preload replaces CHAIN_AT_START in the CoreCLR path. Once the runtime 
        # chains native crashes back to sentry-native, the app-side handler strategy 
        # should be DEFAULT rather than CHAIN_AT_START.
```

tests/fixtures/dotnet_signal/Program.cs
```diff
@@ -39,14 +42,26 @@ public static void RunTest(string[] args, string? databasePath = null)
 
         // setup minimal sentry-native
         var options = sentry_options_new();
-        sentry_options_set_handler_strategy(options, 1);
+        int strategy = args.Contains("default-strategy") ? 0 : 1;
+        sentry_options_set_handler_strategy(options, strategy);
```
supervacuus, line 49: explicit contract:
```suggestion
        // Preload replaces CHAIN_AT_START in the CoreCLR path. If preload is
        // active, run sentry-native with DEFAULT strategy so the runtime can
        // chain native crashes back to us.
        int strategy = args.Contains("default-strategy") ? 0 : 1;
        sentry_options_set_handler_strategy(options, strategy);
```

src/backends/sentry_backend_inproc.c
```diff
@@ -1643,6 +1657,21 @@ process_ucontext(const sentry_ucontext_t *uctx)
             "multiple recursive crashes detected, bailing out");
         goto cleanup;
     }
+
+    // If signal handlers were installed via sentry__backend_preload() but
+    // sentry_init() hasn't been called yet (no handler thread), skip crash
+    // processing and fall through to the previous handler.
+    if (sentry__atomic_fetch(&g_preloaded)
+        && !sentry__atomic_fetch(&g_handler_thread_ready)) {
```
supervacuus, line 1691:
```suggestion
    // If we were only preloaded (before sentry_init()) or were closed after
    // preload (handlers still installed, but no handler thread), do not attempt
    // full crash capture here. In this state, preload only serves as a
    // placeholder in the signal chain.
    //
    // We therefore remove our placeholder from the chain and forward the
    // signal to the previously installed handler set.
    //
    // This path is expected to be terminal for real crash signals. If it
    // returns, preload mode is consumed for the remainder of the process
    // lifetime until a later sentry_init() reinstalls handlers from the
    // then-current chain.
    //
    // This also covers the window after preload but before the managed runtime
    // has installed its own handlers: a signal seen in that window is forwarded
    // to the pre-preload handler set rather than being captured by Sentry.
    if (sentry__atomic_fetch(&g_preloaded)
```

src/backends/sentry_backend_inproc.c
```diff
@@ -200,6 +200,10 @@ static volatile long g_handler_has_work = 0;
 #define CRASH_STATE_DONE 2
 static volatile long g_crash_handling_state = CRASH_STATE_IDLE;
 
+#ifdef SENTRY_PLATFORM_UNIX
+static volatile long g_preloaded = 0;
+#endif
```
supervacuus, line 209:
```suggestion
// Set once handlers were installed via sentry__backend_preload(). In this
// mode sentry may keep its chain position across sentry_close() until either
// full init reuses it, or a pre-init/post-close signal falls through, and
// consumes the preload state.
#ifdef SENTRY_PLATFORM_UNIX
static volatile long g_preloaded = 0;
#endif
```

tests/test_dotnet_signals.py
```diff
@@ -317,26 +321,51 @@ def run_android(args=None, timeout=30):
...
-def run_android_unhandled_managed_exception():
-    return run_android(["unhandled-managed-exception"])
+def run_android_unhandled_managed_exception(strategy=None):
+    return run_android(["unhandled-managed-exception"], strategy=strategy)
 
 
-def run_android_native_crash():
-    return run_android(["native-crash"])
+def run_android_native_crash(strategy=None):
+    return run_android(["native-crash"], strategy=strategy)
```
supervacuus: We currently prove managed exceptions still stay out of native capture, but we don't prove that native crash capture still works after the close/reinit cycle. So let's use the reinit flag here too:
```suggestion
def run_android_native_crash(strategy=None, reinit=False):
    return run_android(["native-crash"], strategy=strategy, reinit=reinit)
```

tests/test_dotnet_signals.py
```diff
@@ -445,10 +485,23 @@ def has_envelope():
...
+            logcat = run_android_managed_exception(strategy=app_strategy, reinit=True)
+            assert not (
+                "NullReferenceException" in logcat
+            ), f"Managed exception leaked after reinit.\nlogcat:\n{logcat}"
+            assert wait_for(lambda: dir_exists(db)), "No database-path exists"
+            assert not file_exists(
+                db + "/last_crash"
+            ), "A crash was registered after reinit"
+            assert not has_envelope(), "Unexpected envelope found after reinit"
+
```
supervacuus, line 508: ...and add a native crash test with reinit in the preload case too:
```suggestion
            assert not has_envelope(), "Unexpected envelope found after reinit"

            # after close/reinit, native crashes must still be captured through
            # the preloaded chain position
            run_android_native_crash(strategy=app_strategy, reinit=True)
            assert wait_for(
                lambda: file_exists(db + "/last_crash")
            ), "Crash marker missing after reinit"
            assert wait_for(has_envelope), "Crash envelope is missing after reinit"
```

ndk/lib/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+  <application>
+    <!-- initOrder must be higher than AppInitOrder-1 (1999999999) used by the .NET runtime:
+         https://github.com/dotnet/android/blob/3c50ff4573b5/src/Xamarin.Android.Build.Tasks/Utilities/ManifestDocument.cs#L724 -->
```
supervacuus:
```suggestion
    <!-- Must run before the managed runtime provider emitted by dotnet/android
         so that sentry-native is already in the signal chain when the runtime
         installs its own handlers. 
         initOrder must stay higher than AppInitOrder-1 (1999999999) used by the .NET runtime:
         https://github.com/dotnet/android/blob/3c50ff4573b5/src/Xamarin.Android.Build.Tasks/Utilities/ManifestDocument.cs#L724 -->
```

ndk/lib/src/main/java/io/sentry/ndk/SentryNdk.java
```diff
@@ -20,6 +22,12 @@ private SentryNdk() {}
 
   private static native void shutdown();
 
+  /** Preloads the NDK integration */
+  public static void preload() {
+    loadNativeLibraries();
+    preloadSentryNative();
+  }
```
supervacuus: new public API that is restricted in use:

```suggestion
  /**
   * Preloads sentry-native into the process signal chain before full
   * initialization.
   *
   * <p>Intended for downstream SDK/runtime integrations on Android. This does
   * not initialize sentry-native, configure a database path, or start the
   * inproc handler thread; it only establishes signal-handler ordering ahead of
   * the managed runtime.
   *
   * <p>This is intended to be used by downstream integrations that gate the
   * preload flow to supported runtimes and then call {@link #init(NdkOptions)}
   * with the normal DEFAULT handler strategy.
   */
  public static void preload() {
    loadNativeLibraries();
    preloadSentryNative();
  }
```

src/sentry_backend.h
jpnurmi: Nice, thanks! Manually applied in https://github.com/getsentry/sentry-native/pull/1613/commits/ad5004f1ae86892ad3c04e07bba8d3659bd3aef7

supervacuus:
> Big thanks for the suggestions! 

You're welcome!

> And sorry to make you spend so much time writing docs and comments for this change! 😵‍💫

Not at all, I actually think that me writing the code-level docs as part of the review is the right choice here.

PR #1633, opened 2026-04-10
fix(test): replace copytree with symlink in the Windows UTF8_TEST

supervacuus:
As recently discussed, copying the build-dir can be replaced with a far less wasteful symlinking approach. Also moved it into its own tmp-dir, which eliminates that recursive positioning.

#skip-changelog


supervacuus:
The Windows UTF-8 path test is again in the same ballpark as the other Windows tests (without sccache). This means well below 30 minutes (vs. previously often easily 40 minutes, unnecessarily extending the entire CI runtime).

jpnurmi (approved):
Very nice, thank you!

PR #1636, opened 2026-04-10
ref: make SENTRY_HANDLER_STACK_SIZE overridable via env var

jpnurmi:
Allow overriding the compile-time `SENTRY_HANDLER_STACK_SIZE` default at runtime via environment variable. The stack-overflow integration tests previously parametrized `SENTRY_HANDLER_STACK_SIZE` as a CMake option, forcing 5 extra configure+build cycles on Windows (crashpad x2, inproc x1, breakpad x2). Switching to a runtime env var lets every parametrization reuse the cached build, saving a minute or two on Windows.

| | Before (s) | After (s) |
|---|---|---|
| `test_crashpad_dumping_stack_overflow` (default/16/32) | 1.56 / 61.44 / 62.19 | 1.57 / 1.67 / 1.58 |
| `test_inproc_stack_overflow_stdout` (default/32) | 1.61 / 20.33 | 1.86 / 1.95 |
| `test_breakpad_stack_overflow_stdout` (default/16/32) | 0.12 / 21.20 / 21.57 | 0.13 / 0.11 / 0.12 |

#skip-changelog

### DIFF
--- a/src/sentry_os.c
+++ b/src/sentry_os.c
@@ -283,7 +283,8 @@ sentry__handler_stack_size_kib(void)
         }
         val = val * 10 + (size_t)(buf[i] - '0');
     }
-    return val > 0 ? val : SENTRY_HANDLER_STACK_SIZE;
+    return (val > 0 && val <= 0xFFFFFFFFu / 1024) ? val
+                                                  : SENTRY_HANDLER_STACK_SIZE;
 }
 
 void
--- a/src/sentry_os.c
+++ b/src/sentry_os.c
@@ -263,6 +263,29 @@ sentry_set_thread_stack_guarantee(uint32_t stack_guarantee_in_bytes)
     return 1;
 }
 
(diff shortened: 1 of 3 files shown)


supervacuus:
Just a quick heads-up: the reason this is a build-time variable is its use inside `DllMain`, where we must ensure that we don't depend on anything but `kernel32.dll`. That it works in CI _now_ is nice, but having both `getenv()` and `strtod()` usage (both CRT functions) in `DllMain` gives me nightmares, because it creates a latent instability only waiting to raise issues... and that would be again a change in the production code for a win in testing.

If we truly want to improve the test build cache usage, let's not parametrize the build-time parameter in the test; instead, supply the stack size as a CLI parameter in the example and set everything there at runtime. This eliminates multiple configs that miss the cache without destabilizing auto-init by introducing dependencies beyond `kernel32.dll`. Another option would be an envreader/parser that depends solely on `kernel32.dll`.

supervacuus (approved):

JoshuaMoelans (approved):

PR #1651, opened 2026-04-15
fix(native): int32 size and endians on mac

mujacica:


src/backends/native/sentry_crash_daemon.c
```diff
@@ -2092,9 +2092,10 @@ build_native_crash_event(
             sentry_value_set_by_key(
                 image, "image_addr", sentry_value_new_string(addr_buf));
 
-            // Set image_size as int32 (modules > 2GB are extremely rare)
+            // Use double for image_size to handle modules > 2GB (e.g. shared
+            // cache)
             sentry_value_set_by_key(image, "image_size",
-                sentry_value_new_int32((int32_t)mod->size));
+                sentry_value_new_double((double)mod->size));
```
jpnurmi: can it be int64?

src/backends/native/minidump/sentry_minidump_macos.c
```diff
@@ -243,8 +243,31 @@ write_cv_record(
...
+    s[1] = s[2];
+    s[2] = tmp;
+    // Swap Data2 (bytes 4-5)
+    tmp = s[4];
+    s[4] = s[5];
+    s[5] = tmp;
+    // Swap Data3 (bytes 6-7)
+    tmp = s[6];
+    s[6] = s[7];
+    s[7] = tmp;
```
jpnurmi: non-blocking nitpick: some kind of swap/reverse helper macro or function would be nice 🙂

jpnurmi: there's another UUID byte-flipping in `sentry_modulefinder_linux.c`. could be extracted to `sentry_utils`?

https://github.com/getsentry/sentry-native/blob/0b48cd0b403fe6a65d00d4111219c5cf9e47594e/src/modulefinder/sentry_modulefinder_linux.c#L492-L502

mujacica: Oh yeah, will swap that around

src/backends/native/sentry_crash_daemon.c
mujacica: yeah it can, and it should

src/backends/native/sentry_crash_daemon.c
```diff
@@ -2092,9 +2085,8 @@ build_native_crash_event(
             sentry_value_set_by_key(
                 image, "image_addr", sentry_value_new_string(addr_buf));
 
-            // Set image_size as int32 (modules > 2GB are extremely rare)
             sentry_value_set_by_key(image, "image_size",
-                sentry_value_new_int32((int32_t)mod->size));
+                sentry_value_new_int64((int64_t)mod->size));
 
 #if defined(SENTRY_PLATFORM_WINDOWS)
             // Set code_id for PE modules (TimeDateStamp + SizeOfImage)
```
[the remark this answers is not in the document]
mujacica: Linux mod->uuid is populated from the ELF NT_GNU_BUILD_ID note — those are raw hash bytes that Sentry's backend indexes in Windows GUID mixed-endian layout, so the swap at [line 1173](vscode-webview://0mospq6fqbnqb3tp0s60rov4ekkjet5jrgrbuqlv1tkk1b21e1kd/src/backends/native/sentry_crash_daemon.c#L1173) is required.
macOS mod->uuid is populated from LC_UUID, which is already an RFC 4122 UUID. The standard 8-4-4-4-12 hex string form is what Sentry's macOS symbolicator expects — no conversion needed.

jpnurmi (approved):

PR #1655, opened 2026-04-15
fix(modulefinder): handle ENOSYS in read_safely for Kubernetes seccomp

### LINKED ISSUE #1653: `sentry_modulefinder_linux`: empty module list in containers with restricted `process_vm_readv`
**Description**

When running inside a container with a restrictive seccomp profile (e.g. a Kubernetes pod), `sentry_get_modules_list()` always returns an empty list, causing `debug_meta.images` in manually-captured events to be empty. Sentry's symbolicator then marks every frame as `unknown_image`, making stack traces unsymbolicated even when debug files are correctly uploaded.

**Possible issue**

The root cause might be in `read_safely()`: it calls `process_vm_readv()` to read ELF magic bytes from each `/proc/self/maps` entry. There is already a fallback to `memcpy` for `EPERM` and `EINVAL`, but when seccomp blocks the syscall it returns `ENOSYS`, which is not handled — so `read_safely()` returns `false`, `is_valid_elf_header()` returns `false` for every entry, and the module list stays empty.

Note: crash events captured by crashpad are **not** affected — crashpad's handler process enumerates modules independently of `sentry_get_modules_list()`.

**Suggested fix**

Add `ENOSYS` to the existing fallback in `read_safely()`:

```c
if (!rv && (errno == EPERM || errno == EINVAL || errno == ENOSYS)) {
    memcpy(dst, src, size);
    rv = true;
}
```

- [ ] During build
- [x] During run-time

**Environment**

- OS: Ubuntu 24.04.4 LTS (Flatcar Linux kernel 6.12.66), running inside a Kubernetes pod with default seccomp profile
- Compiler: Clang 20
- CMake version and config: `SENTRY_BACKEND=crashpad`
- sentry-native version: `0.13.2`

**Steps To Reproduce**

1. Run the application with sentry-native (crashpad backend) inside a container where `process_vm_readv` is restricted by seccomp.
2. Enable debug logging: `sentry_options_set_debug(options, 1)`.
3. Capture any non-crash event: `sentry_capture_event(...)`.
4. Observe `"read 0 modules from /proc/self/maps"` in the log output.
5. In Sentry, all frames in the event have `symbolicator_status: "unknown_image"` despite debug files being uploaded.

**Log output**
```
[sentry] DEBUG read 0 modules from /proc/self/maps
```

HrMathematiker:
Fixes #1653.

## Problem

When running under Kubernetes with the default `RuntimeDefault` seccomp profile, `process_vm_readv` is blocked and returns `ENOSYS` (errno 38). The existing fallback in `read_safely()` only handles `EPERM` and `EINVAL`, so every ELF module validation fails silently.

The result: `sentry_get_modules_list()` returns 0 modules, `debug_meta.images` is absent from manually-captured events, and every stack frame shows as `unknown_image` in Sentry — even when debug symbols have been uploaded correctly.

## When does this happen?

- [x] Kubernetes pods using the default `RuntimeDefault` seccomp profile
- [ ] Docker containers (Docker default profile returns `EPERM`, already handled)
- [ ] Bare-metal / VMs (syscall is permitted)

Crash events via crashpad are **not** affected because the out-of-process handler enumerates modules from `/proc/<pid>/maps` directly, independent of `sentry_get_modules_list()`.

## Root cause

`read_safely()` in `sentry_modulefinder_linux.c`:

```c
// Before
if (!rv && (errno == EPERM || errno == EINVAL)) {

// After
if (!rv && (errno == EPERM || errno == EINVAL || errno == ENOSYS)) {
```

## Verification

Confirmed with a static test binary copied into a real K8s pod (Flatcar Linux, kernel 6.12, `Seccomp: 2`):

```
FAILED: errno=38 (Function not implemented)
```

Also reproduced locally using Docker with a custom seccomp profile (`SCMP_ACT_ERRNO` / `errnoRet: 38`) — the original code fails, the patched code falls back to `memcpy` and succeeds.

## Fix

One-liner: add `|| errno == ENOSYS` to the existing errno fallback. The address is valid in-process memory, so `memcpy` is safe.

### DIFF
--- a/src/modulefinder/sentry_modulefinder_linux.c
+++ b/src/modulefinder/sentry_modulefinder_linux.c
@@ -141,10 +141,9 @@ read_safely(void *dst, void *src, size_t size)
     // See https://github.com/getsentry/sentry-native/issues/578).
     // Also, the syscall is only available in Linux 3.2, meaning Android 17.
     // In that case we get an `EINVAL`.
-    // Additionally, Kubernetes default seccomp profiles (RuntimeDefault) block
-    // `process_vm_readv` and return `ENOSYS`, causing all ELF module
-    // validations to fail and `debug_meta.images` to be absent from
-    // manually-captured events.
+    // Additionally, in some seccomp-restricted environments,
+    // `process_vm_readv` may be unavailable and fail with `ENOSYS` (see
+    // https://github.com/getsentry/sentry-native/issues/1653).
     //
     // In any of these cases, just fall back to an unsafe `memcpy`.
     if (!rv && (errno == EPERM || errno == EINVAL || errno == ENOSYS)) {
--- a/src/modulefinder/sentry_modulefinder_linux.c
+++ b/src/modulefinder/sentry_modulefinder_linux.c
@@ -141,9 +141,13 @@ read_safely(void *dst, void *src, size_t size)
     // See https://github.com/getsentry/sentry-native/issues/578).
     // Also, the syscall is only available in Linux 3.2, meaning Android 17.
     // In that case we get an `EINVAL`.
+    // Additionally, Kubernetes default seccomp profiles (RuntimeDefault) block
+    // `process_vm_readv` and return `ENOSYS`, causing all ELF module
+    // validations to fail and `debug_meta.images` to be absent from
+    // manually-captured events.
     //
-    // In either of these cases, just fall back to an unsafe `memcpy`.
-    if (!rv && (errno == EPERM || errno == EINVAL)) {
+    // In any of these cases, just fall back to an unsafe `memcpy`.
+    if (!rv && (errno == EPERM || errno == EINVAL || errno == ENOSYS)) {
         memcpy(dst, src, size);
         rv = true;
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Fixes**:
+
+- Linux: handle `ENOSYS` in `read_safely` to fix empty module list in seccomp-restricted environments. ([#1655](https://github.com/getsentry/sentry-native/pull/1655))
+
 ## 0.13.7
 
 **Features**:


jpnurmi:
Hi, thanks for the contribution, and sorry about the bot. 😅 

What do you think, @supervacuus? The proposed fix looks good to me. Just need a changelog entry but we can fix that.

src/modulefinder/sentry_modulefinder_linux.c
```diff
@@ -141,9 +141,13 @@ read_safely(void *dst, void *src, size_t size)
     // See https://github.com/getsentry/sentry-native/issues/578).
     // Also, the syscall is only available in Linux 3.2, meaning Android 17.
     // In that case we get an `EINVAL`.
+    // Additionally, Kubernetes default seccomp profiles (RuntimeDefault) block
+    // `process_vm_readv` and return `ENOSYS`, causing all ELF module
+    // validations to fail and `debug_meta.images` to be absent from
+    // manually-captured events.
```
supervacuus: We can make the comment less specific/assertive about Kubernetes and phrase it in terms of the actual condition. I’d also keep the new comment focused on why `ENOSYS` can occur here. The sentence about failing `ELF` validation and missing `debug_meta.images` is true for the whole per-module path, not specifically for `ENOSYS` or the fallback, so it reads as broader module-level rationale rather than justification for adding this `errno`. Removing it makes the comment more precise.

```suggestion
    // Additionally, in some seccomp-restricted environments,
    // `process_vm_readv` may be unavailable and fail with `ENOSYS`.
```

supervacuus (changes requested):
Thanks!

Please, rebase first, add a changelog entry  and adapt the comment.

As with the other fallbacks, we will keep them as long as they do not lead to unnecessary crashes in other environments. These `errno`s are not specific to the address and as such the fallback should be fine.

Edit: I just saw that the danger comment might not be visible to you. Essentially add to changelog something like:

```markdown
## Unreleased

### Fixes

- Linux: handle `ENOSYS` in the modulefinder's `read_safely()` for seccomp-restricted environments like Kubernetes. ([#1655](https://github.com/getsentry/sentry-native/pull/1655))
```
at the top

src/modulefinder/sentry_modulefinder_linux.c
HrMathematiker: Hello @supervacuus and thanks for your feedback!
I've applied your suggestion with a link to the issue and added an item to the changelog - please, take a look.

supervacuus (approved):

PR #1656, opened 2026-04-16
fix(breakpad): avoid stdio deadlock in macOS exception handler

jpnurmi:
On macOS, breakpad's Mach exception handler runs while all other threads are suspended by the kernel. Any `vfprintf`-based log call from the handler can deadlock waiting for a stdio file lock held by a frozen thread — this is what is currently hanging `macOS 14 (xcode llvm)` CI on #1650 after `test_cache_max_items_with_retry[inproc]`.

Default `enable_logging_when_crashed` to off for the breakpad backend on macOS so the existing `sentry__logger_disable()` path in the callback fires by default. Users who accept the risk can still opt in via `sentry_options_set_logger_enabled_when_crashed(opts, 1)`.

Share the async-signal-safe `SENTRY_SIGNAL_SAFE_LOG` macro via `sentry_logger.h` (previously private to the inproc backend) and convert the five explicit log sites in the breakpad callback plus the crash-safe flush helpers in `sentry_metrics`, `sentry_logs`, and `sentry_batcher`. These sites use `write(STDERR_FILENO, ...)` directly, so they stay visible during crash handling regardless of the logger state.

Captured stack from a local repro (8x CPU `yes` load, 4-iteration `log cache-keep crash` loop against a TCP-RST DSN):

```
Thread sentry-http (suspended):
    curl_easy_perform -> Curl_verboseconnect -> Curl_debug -> fwrite
        -> __sfvwrite -> __write_nocancel        (holds stderr flockfile)

Thread breakpad handler (blocked):
    breakpad_backend_callback -> sentry__logger_log
        -> sentry__logger_defaultlogger -> vfprintf -> flockfile
            -> __psynch_mutexwait                 (waits forever)
```

`sentry_options_set_debug(true)` (via the `log` example arg used in most cache tests) enables `CURLOPT_VERBOSE`, which is what reliably catches the worker mid-`fwrite` on stderr when the crash fires.

A related deadlock class still exists inside breakpad's own `MinidumpGenerator` (`operator new` -> `mfm_alloc`) when another thread is suspended mid-allocation. That is outside sentry-native's reach and not addressed here.

### DIFF
--- a/src/sentry_options.c
+++ b/src/sentry_options.c
@@ -49,12 +49,13 @@ sentry_options_new(void)
     opts->system_crash_reporter_enabled = false;
     opts->attach_screenshot = false;
     opts->crashpad_wait_for_upload = false;
-    // On macOS, breakpad's Mach exception handler runs while all other
-    // threads are suspended by the kernel. Any vfprintf-based log call from
-    // that context can deadlock waiting for a stdio lock held by a suspended
-    // thread (e.g. libcurl's verbose output on stderr). Default off here;
-    // users who accept the risk can still opt in via
-    // `sentry_options_set_logger_enabled_when_crashed(opts, 1)`.
+    // On macOS, breakpad suspends all other threads of the process before
+    // writing the minidump and invokes our backend callback from inside
+    // WriteMinidumpWithException() while those threads are still suspended.
+    // Any vfprintf-based log call from that context can deadlock waiting for
+    // a stdio lock held by a suspended thread (e.g. libcurl's verbose output
+    // on stderr). Default off here; users who accept the risk can still opt
+    // in via `sentry_options_set_logger_enabled_when_crashed(opts, 1)`.
 #if defined(SENTRY_BACKEND_BREAKPAD) && defined(SENTRY_PLATFORM_DARWIN)        \
     && !defined(SENTRY_PLATFORM_IOS)
     opts->enable_logging_when_crashed = false;
--- a/src/sentry_options.c
+++ b/src/sentry_options.c
@@ -49,7 +49,18 @@ sentry_options_new(void)
     opts->system_crash_reporter_enabled = false;
     opts->attach_screenshot = false;
     opts->crashpad_wait_for_upload = false;
+    // On macOS, breakpad's Mach exception handler runs while all other
+    // threads are suspended by the kernel. Any vfprintf-based log call from
+    // that context can deadlock waiting for a stdio lock held by a suspended
+    // thread (e.g. libcurl's verbose output on stderr). Default off here;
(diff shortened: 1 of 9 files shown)


src/sentry_options.c
```diff
@@ -49,7 +49,18 @@ sentry_options_new(void)
     opts->system_crash_reporter_enabled = false;
     opts->attach_screenshot = false;
     opts->crashpad_wait_for_upload = false;
+    // On macOS, breakpad's Mach exception handler runs while all other
+    // threads are suspended by the kernel. Any vfprintf-based log call from
+    // that context can deadlock waiting for a stdio lock held by a suspended
+    // thread (e.g. libcurl's verbose output on stderr). Default off here;
+    // users who accept the risk can still opt in via
+    // `sentry_options_set_logger_enabled_when_crashed(opts, 1)`.
```
supervacuus: The description in `sentry.h` is fine, but I think here we should be much more exact:

* It isn't the kernel that suspended the other threads as a root cause
* Breakpad suspends the other threads before it runs `WriteMinidumpWithException()`
* our handler callback is called from within `WriteMinidumpWithException()`
* so at the point when we are in our backend handler, all other threads are suspended

I know this sounds like a minor correction, but it is important because it means this is entirely caused by the way breakpad structures its callbacks in the context of minidump writing _and_ thread suspension (the `filter` callback runs before the actual minidump but also within suspended threads). It is not related to how macOS handles Mach exception delivery (which only stops the crashing thread) or to the fact that this happens in-process.

supervacuus (approved):
Minor doc correction, but otherwise great. Thanks!

src/sentry_options.c
jpnurmi: Right, thanks for the correction!

src/sentry_options.c
```diff
@@ -49,7 +49,19 @@ sentry_options_new(void)
...
+    // WriteMinidumpWithException() while those threads are still suspended.
+    // Any vfprintf-based log call from that context can deadlock waiting for
+    // a stdio lock held by a suspended thread (e.g. libcurl's verbose output
+    // on stderr). Default off here; users who accept the risk can still opt
+    // in via `sentry_options_set_logger_enabled_when_crashed(opts, 1)`.
+#if defined(SENTRY_BACKEND_BREAKPAD) && defined(SENTRY_PLATFORM_DARWIN)        \
+    && !defined(SENTRY_PLATFORM_IOS)
+    opts->enable_logging_when_crashed = false;
+#else
     opts->enable_logging_when_crashed = true;
```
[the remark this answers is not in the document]
jpnurmi: unit tests are always built with `SENTRY_BACKEND=none`

PR #1666, opened 2026-04-23
feat(native): enable the native backend on Xbox

tustanivsky:
## Summary

This PR enables the `native` (out-of-process) backend for the Xbox GDK (XSX, XboxOne) and WinGDK targets that previously fell back to `breakpad` by default and couldn't select `native` due to build-time link errors plus a handful of runtime-API usages incompatible with the Xbox gaming partition.

Now, `SENTRY_BACKEND=native` produces a working out-of-process daemon that spawns at `sentry_init`, captures minidumps via `MiniDumpWriteDump`, and uploads envelopes over WinHTTP on Xbox.

Enabler for https://github.com/getsentry/sentry-xbox/pull/128.

### Build

- Skip the explicit `dbghelp.lib` link and the `#pragma comment(lib, "dbghelp.lib")` directive when targeting Xbox. On consoles, `MiniDumpWriteDump` is provided by `xgameplatform.lib` (via an API set), while the desktop `dbghelp.lib` is intentionally excluded by the Xbox toolchain’s `/NODEFAULTLIB` list. Verified via the Xbox GDK sample [`AdvancedExceptionHandling`](https://github.com/microsoft/Xbox-GDK-Samples/tree/main/Samples/System/AdvancedExceptionHandling), where XSX configurations link `$(Console_Libs)` (without `Dbghelp.lib`) and successfully call `MiniDumpWriteDump`.

### Thread enumeration

- Guard `enumerate_threads_from_process` with `SENTRY_PLATFORM_XBOX` - the ToolHelp32 API (`CreateToolhelp32Snapshot`, `THREADENTRY32`, `Thread32First/Next`) is not available in the Xbox gaming partition. `MiniDumpWriteDump` still captures all thread contexts from the target PID; the skipped code only populated supplementary per-thread metadata for the Sentry event payload.

### Daemon startup on Xbox

- Call `XGameRuntimeInitialize` at daemon `main` entry - required before any XNetworking calls made by the transport during crash upload. Pair with `XGameRuntimeUninitialize` on exit.
- Pre-warm XNetworking on a detached thread rather than inline. Per the [GDK docs on network initialization](https://learn.microsoft.com/en-us/gaming/gdk/docs/features/console/networking/initialization-connectivity-networking):

  > Network initialization usually takes a couple of seconds on both resume and title launch and varies based on console type and the user's network environment.

Running the warm-up inline could push the daemon’s ready signal past the parent’s `SENTRY_CRASH_DAEMON_READY_TIMEOUT_MS` window on retail cold boots; backgrounding keeps `sentry_init` fast while ensuring the network stack is typically ready by the time a crash occurs.

## Related items
- https://github.com/getsentry/sentry-xbox/pull/128

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Features**:
+
+- Enable experimental `native` backend on Xbox ([#1666](https://github.com/getsentry/sentry-native/pull/1666))
+
 **Fixes**:
 
 - Linux: handle `ENOSYS` in `read_safely` to fix empty module list in seccomp-restricted environments. ([#1655](https://github.com/getsentry/sentry-native/pull/1655))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 **Features**:
 
-- Enable experimental `native` backend on Xbox ([#1666](https://github.com/getsentry/sentry-native/pull/1666))
+- Enable experimental `native` backend on Xbox. ([#1666](https://github.com/getsentry/sentry-native/pull/1666))
 - Linux: support 32-bit ARM. ([#1659](https://github.com/getsentry/sentry-native/issues/1659))
 
 **Fixes**:
--- a/src/backends/native/minidump/sentry_minidump_windows.c
+++ b/src/backends/native/minidump/sentry_minidump_windows.c
@@ -15,7 +15,7 @@
  * On Xbox, MiniDumpWriteDump is provided by xgameplatform.lib
  * (linked by the Xbox toolchain), not dbghelp.lib.
  */
-#    if !defined(_GAMING_XBOX)
+#    if !defined(SENTRY_PLATFORM_XBOX)
 #        pragma comment(lib, "dbghelp.lib")
 #    endif
 
--- a/src/backends/native/minidump/sentry_minidump_windows.c
+++ b/src/backends/native/minidump/sentry_minidump_windows.c
@@ -11,9 +11,11 @@
 #    include "sentry_minidump_writer.h"
 #    include "sentry_string.h"
 
+/**
+ * On Xbox, MiniDumpWriteDump is provided by xgameplatform.lib
+ * (linked by the Xbox toolchain), not dbghelp.lib.
+ */
 #    if !defined(_GAMING_XBOX)
-// On Xbox console MiniDumpWriteDump is supplied by xgameplatform.lib
-// (linked by the Xbox toolchain), not dbghelp.lib.
 #        pragma comment(lib, "dbghelp.lib")
 #    endif
 
--- a/src/backends/native/minidump/sentry_minidump_windows.c
+++ b/src/backends/native/minidump/sentry_minidump_windows.c
@@ -11,7 +11,11 @@
 #    include "sentry_minidump_writer.h"
 #    include "sentry_string.h"
 
-#    pragma comment(lib, "dbghelp.lib")
+#    if !defined(_GAMING_XBOX)
+// On Xbox console MiniDumpWriteDump is supplied by xgameplatform.lib
+// (linked by the Xbox toolchain), not dbghelp.lib.
+#        pragma comment(lib, "dbghelp.lib")
+#    endif
 
 /**
  * Windows minidump writer
--- a/src/backends/native/sentry_crash_daemon.c
+++ b/src/backends/native/sentry_crash_daemon.c
@@ -3610,6 +3610,16 @@ sentry__crash_daemon_start(pid_t app_pid, uint64_t app_tid, HANDLE event_handle,
 // When built as standalone executable, provide main entry point
 #ifdef SENTRY_CRASH_DAEMON_STANDALONE
 
+#    if defined(SENTRY_PLATFORM_XBOX)
+#        include <XGameRuntime.h>
+
+// Forward declared here; implemented by the downstream SDK (sentry-xbox's
+// sentry_transport_xbox.cpp) and linked into the daemon. The same contract
+// is already assumed by sentry_http_transport_winhttp.c under
+// SENTRY_PLATFORM_XBOX.
+extern bool sentry__xbox_ensure_network_initialized(void);
+#    endif
+
 int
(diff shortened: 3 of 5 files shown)


src/backends/native/sentry_crash_daemon.c
```diff
@@ -3642,8 +3667,32 @@ main(int argc, char **argv)
...
+    if (network_prewarm_thread) {
+        CloseHandle(network_prewarm_thread);
+    }
+#        endif
+
+    int rv = sentry__crash_daemon_main(
         app_pid, app_tid, event_handle, ready_event_handle);
+
+#        if defined(SENTRY_PLATFORM_XBOX)
+    XGameRuntimeUninitialize();
```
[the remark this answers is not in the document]
tustanivsky: The `CloseHandle` now runs after a `WaitForSingleObject(thread, INFINITE)` that sits right before `XGameRuntimeUninitialize()`, so the prewarm thread is guaranteed to have exited before we tear down the runtime.                                                                                                                                                                        
                                                                                                                                                                                                
On `INFINITE`: the thread only calls `sentry__xbox_ensure_network_initialized()`, whose blocking wait is bounded (`WaitForSingleObjectEx(..., 60000, FALSE)` in the sentry-xbox transport). So the thread terminates within ~60s in the worst case and the `INFINITE` join can't actually hang.

https://github.com/getsentry/sentry-native/pull/1666/commits/b1b780f314a02c237a26ba0bd6898623910625d7

src/backends/native/minidump/sentry_minidump_windows.c
```diff
@@ -11,7 +11,13 @@
 #    include "sentry_minidump_writer.h"
 #    include "sentry_string.h"
 
-#    pragma comment(lib, "dbghelp.lib")
+/**
+ * On Xbox, MiniDumpWriteDump is provided by xgameplatform.lib
+ * (linked by the Xbox toolchain), not dbghelp.lib.
+ */
+#    if !defined(_GAMING_XBOX)
```
jpnurmi: is `SENTRY_PLATFORM_XBOX` available here?

mujacica (approved):

src/backends/native/sentry_crash_daemon.c
```diff
@@ -3642,8 +3667,34 @@ main(int argc, char **argv)
     unsigned long long ready_event_val = strtoull(argv[4], NULL, 10);
     HANDLE event_handle = (HANDLE)(uintptr_t)event_handle_val;
     HANDLE ready_event_handle = (HANDLE)(uintptr_t)ready_event_val;
-    return sentry__crash_daemon_main(
+
+#        if defined(SENTRY_PLATFORM_XBOX)
+    // Required before any XNetworking call the transport makes at
+    // crash-upload time.
+    HRESULT init_hr = XGameRuntimeInitialize();
```
jpnurmi: is there any way we could export some `sentry__xbox_xxx` helper from sentry_xbox to avoid direct `XGameRuntime` calls?

src/backends/native/sentry_crash_daemon.c
```diff
@@ -1717,10 +1717,19 @@ get_thread_name(HANDLE hThread, sentry_thread_context_windows_t *tctx)
...
+ *
+ * Not available on Xbox (the ToolHelp32 API is not part of the gaming
+ * partition). MiniDumpWriteDump still captures all thread contexts from the
+ * target process independently, so skipping this only loses supplementary
+ * per-thread metadata from the Sentry event payload.
  */
 static void
 enumerate_threads_from_process(sentry_crash_context_t *ctx)
 {
+#    ifdef SENTRY_PLATFORM_XBOX
```
JoshuaMoelans: (nit) consistency with the other places we check for `SENTRY_PLATFORM_XBOX`
```suggestion
#    if defined(SENTRY_PLATFORM_XBOX)
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 **Features**:
 
+- Enable experimental `native` backend on Xbox ([#1666](https://github.com/getsentry/sentry-native/pull/1666))
```
JoshuaMoelans: (nit) changelog entries get a `.` at the end (I can't count how often I've forgotten this too though 😅 ) 
```suggestion
- Enable experimental `native` backend on Xbox. ([#1666](https://github.com/getsentry/sentry-native/pull/1666))
```

JoshuaMoelans (approved):
LGTM, just two very minor nits

src/backends/native/minidump/sentry_minidump_windows.c
tustanivsky: Good catch - replaced it in https://github.com/getsentry/sentry-native/pull/1666/commits/e6884fa410f7265febf6c345b4f54b1d11496ca5.

src/backends/native/sentry_crash_daemon.c
tustanivsky: Added `sentry__xbox_game_runtime_initialize()` / `_uninitialize()` wrappers in sentry-xbox and swapped the daemon over to them - https://github.com/getsentry/sentry-xbox/pull/128/commits/e1a1f349515394ee0212b77ada67326be12f616f

src/backends/native/sentry_crash_daemon.c
```diff
@@ -3642,8 +3669,30 @@ main(int argc, char **argv)
...
+    HANDLE network_prewarm_thread = CreateThread(
+        NULL, 0, sentry__xbox_network_prewarm_thread_proc, NULL, 0, NULL);
+#        endif
+
+    int rv = sentry__crash_daemon_main(
         app_pid, app_tid, event_handle, ready_event_handle);
+
+#        if defined(SENTRY_PLATFORM_XBOX)
+    if (network_prewarm_thread) {
+        WaitForSingleObject(network_prewarm_thread, INFINITE);
```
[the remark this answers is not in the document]
tustanivsky: `sentry__xbox_ensure_network_initialized()` has an internal 60s `WaitForSingleObjectEx` timeout so the prewarm thread provably terminates within ~60s and the `INFINITE` join is bounded by that.

jpnurmi (approved):

PR #1669, opened 2026-04-24
fix: event ownership in sentry_capture_minidump

jpnurmi:
In `sentry_capture_minidump_n`, the `!envelope || sentry_uuid_is_nil(&event_id)` branch incorrectly called `sentry_value_decref(event)`.

After `sentry__prepare_event` returns, `event` is either:
a) already cleaned up by its fail path (`envelope == NULL`), or
b) owned by the envelope via `sentry__envelope_add_event` (`envelope != NULL`).

The `decref` is therefore always a double-decref, and for the non-null-envelope sub-case additionally leaks the envelope. Replace with `sentry_envelope_free(envelope)` — no-op on `NULL`, releases the owned event via item teardown otherwise.

The newly added test fails with the CI jobs that run with sanitizers: https://github.com/getsentry/sentry-native/actions/runs/24885938040

Came up in #1545


jpnurmi:
All tsan and asan jobs fail without the fix: https://github.com/getsentry/sentry-native/actions/runs/24885938040

mujacica (approved):

JoshuaMoelans (approved):

PR #1674, opened 2026-04-27
feat(crashpad): Enable large attachment uploads

jpnurmi:
Add an experimental `enable_large_attachments` option and pass it through to the Crashpad handler. Other backends coming separately in #1545.


jpnurmi:
This is not needed because `/minidump` endpoint does already have built-in support for large user attachments (getsentry/relay#5877). Support for large minidumps is also coming soon.

image

PR #1687, opened 2026-04-29
fix(native): preserve attachments added during crash handling

tustanivsky:
Attachments registered from `on_crash` / `before_send` are not delivered to the daemon under the native backend because the scope-flush path that writes `<run>/__sentry-attachments` is gated by `state->crashed` (#1668) once the crash handler runs. As a result, the daemon assembles the envelope from a stale manifest and the attachments are silently dropped.

This PR narrows the gate to cover only the event JSON write. The manifest write is extracted into a helper and runs on both the pre-crash and post-crash paths inside `native_backend_flush_scope`, so attachments registered during `on_crash` / `before_send` propagate through the existing scope-flush hook and reach the daemon.

Related items:

- https://github.com/getsentry/sentry-xbox/issues/122
- https://github.com/getsentry/sentry-unreal/issues/1365

### DIFF
--- a/src/backends/sentry_backend_native.c
+++ b/src/backends/sentry_backend_native.c
@@ -614,8 +614,8 @@ native_backend_write_attachments(const sentry_path_t *event_path)
                     sentry_value_new_string(it->path->path));
                 const char *filename = sentry__path_filename(
                     it->filename ? it->filename : it->path);
-                sentry_value_set_by_key(attach_info, "filename",
-                    sentry_value_new_string(filename));
+                sentry_value_set_by_key(
+                    attach_info, "filename", sentry_value_new_string(filename));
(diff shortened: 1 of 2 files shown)


src/backends/sentry_backend_native.c
jpnurmi, line 652: what if we'd call `native_backend_write_attachments(state->event_path)` from here instead of unconditionally after on_crash?

tustanivsky: Good call, applied in https://github.com/getsentry/sentry-native/pull/1687/commits/5d0f3c8718c60cf3d82cceea86d5b46e9a0de69b.

mujacica (approved):

jpnurmi (approved):

PR #1694, opened 2026-05-04
fix(native): Fix smaller findings for Linux minidump writer

mujacica:
- Native/Linux: correct `MD_LINUX_MAPS` stream type (was tagged as `MD_LINUX_AUXV`).
- Native/Linux: drop non-ELF mappings (e.g. `/dev/shm/*`, `(deleted)` files) from the minidump module list.
- Native/Linux: merge non-contiguous mappings of the same shared library into a single module, and use the offset==0 mapping as `base_of_image`. Fixes duplicate `ld-linux` entries that confused some debuggers (notably Windows LLDB) reading Linux ARM64 minidumps.
- Native/Linux: log when `uname()` is blocked (sandbox/seccomp) and fall back to `/proc/sys/kernel/osrelease` for the OS version.
- Native/Linux: emit `LinuxAuxv`, `LinuxCpuInfo`, `LinuxLsbRelease`, `LinuxCmdLine`, `LinuxEnviron`, and `LinuxDsoDebug` streams alongside the existing set. LLDB needs `LinuxAuxv` and `LinuxDsoDebug` to identify the dynamic loader and enumerate loaded shared libraries; without them, opening a minidump in LLDB on Linux would only recover one frame per thread.
- Native/Linux: replay each thread's stack memory descriptor into `MemoryListStream`. Previously stack bytes were only referenced from the per-thread record, so debuggers that look up memory by virtual address (LLDB) could not read the stack and unwinding stopped at frame 0 even when `eh_frame` was available.


src/backends/native/minidump/sentry_minidump_format.h
```diff
@@ -468,4 +473,35 @@ typedef struct {
 } PACKED_ATTR minidump_thread_name_list_t;
 PACKED_STRUCT_END
 
+/**
+ * Linux DSO debug structures (MD_LINUX_DSO_DEBUG, 0x4767000A)
+ *
+ * Wire-compatible with Breakpad's MDRawLinkMap64 / MDRawDebug64. Breakpad
```
jpnurmi: are `MDRawLinkMap32` and `MDRawDebug32` worth bringing over?

jpnurmi (approved):

src/backends/native/minidump/sentry_minidump_format.h
mujacica: Yes, added in https://github.com/getsentry/sentry-native/pull/1694/commits/8fd55bec5e389a54d10bb68b367c1b75add62d17

PR #1700, opened 2026-05-05
feat: expose attachment types

jpnurmi:
Allow downstream SDKs to attach files and bytes with standard Sentry attachment types, such as `event.view_hierarchy` or `event.minidump`.

This is more flexible than `sentry_options_add_view_hierarchy()` that a) requires a file on the disk, which must be frequently updated in case of view hierarchy, and b) pushes session management to the downstream SDKs, because they don't have access to the .run directory for multi-process support and automatic cleanup.

### DIFF
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -2757,6 +2757,24 @@ SENTRY_API sentry_attachment_t *sentry_scope_attach_bytesw_n(
     const wchar_t *filename, size_t filename_len);
 #endif
 
+#define SENTRY_ATTACHMENT_TYPE_GENERIC "event.attachment"
+#define SENTRY_ATTACHMENT_TYPE_MINIDUMP "event.minidump"
+#define SENTRY_ATTACHMENT_TYPE_VIEW_HIERARCHY "event.view_hierarchy"
+
+/**
+ * Sets the attachment type.
+ *
+ * Well-known attachment types are exposed as `SENTRY_ATTACHMENT_TYPE_*`
+ * macros. Pass `NULL` or an empty string to clear the attachment type.
+ *
+ * See:
+ * https://develop.sentry.dev/sdk/telemetry/attachments/#attachment-types
+ */
+SENTRY_API void sentry_attachment_set_type(
+    sentry_attachment_t *attachment, const char *type);
+SENTRY_API void sentry_attachment_set_type_n(
+    sentry_attachment_t *attachment, const char *type, size_t type_len);
+
 /**
  * Sets the content type of attachment.
  */
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -2775,6 +2775,8 @@ SENTRY_API sentry_attachment_t *sentry_scope_attach_bytesw_n(
  * Well-known attachment types are exposed as `SENTRY_ATTACHMENT_TYPE_*`
  * macros. Pass `NULL` or an empty string to clear the attachment type.
  *
+ * Also sets the content type for known attachment types unless explicitly set.
+ *
  * See:
  * https://develop.sentry.dev/sdk/telemetry/attachments/#attachment-types
  */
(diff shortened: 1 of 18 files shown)


limbonaut:
Should we deprecate `sentry_options_add_view_hierarchy`? It'll be just a shortcut, not a very useful one.

jpnurmi:
> Should we deprecate `sentry_options_add_view_hierarchy`? It'll be just a shortcut, not a very useful one.

Hmm, yeah, we probably could. Besides sentry-native and its forks and vendored copies, I can only find public references in sentry-godot and sentry-docs.

jpnurmi:
Will get back to this after #1705, which removes the configure-time attachment limitation on macOS and allows deprecating `sentry_options_add_view_hierarchy`.

src/sentry_attachment.c
```diff
@@ -170,15 +201,20 @@ attachment_eq(const sentry_attachment_t *a, const sentry_attachment_t *b)
...
         return false;
     }
     return sentry__path_eq(a->path, b->path);
 }
 
 sentry_attachment_t *
 sentry__attachments_add(sentry_attachment_t **attachments_ptr,
-    sentry_attachment_t *attachment, sentry_attachment_type_t attachment_type,
+    sentry_attachment_t *attachment, const char *attachment_type,
     const char *content_type)
```
limbonaut: It's weird that this function also works as a setter for attachment_type and content_type, considering that there are dedicated setters for it. In most places it's used, the last two parameters are passed as NULL. Only `sentry__attachments_add_path` seem to truly exercise these args?

include/sentry.h
```diff
@@ -2757,6 +2765,24 @@ SENTRY_API sentry_attachment_t *sentry_scope_attach_bytesw_n(
...
+#define SENTRY_ATTACHMENT_TYPE_GENERIC "event.attachment"
+#define SENTRY_ATTACHMENT_TYPE_MINIDUMP "event.minidump"
+#define SENTRY_ATTACHMENT_TYPE_VIEW_HIERARCHY "event.view_hierarchy"
+
+/**
+ * Sets the attachment type.
+ *
+ * Well-known attachment types are exposed as `SENTRY_ATTACHMENT_TYPE_*`
+ * macros. Pass `NULL` or an empty string to clear the attachment type.
+ *
```
limbonaut, line 2777: Should it be mentioned that it also sets `content_type` for recognized attachment types?

src/sentry_attachment.c
jpnurmi: True, the arguments were probably added "for convenience" once upon a time. I dropped them from `sentry__attachments_add` but kept `sentry__attachments_add_path` intact because it's pretty heavily used all over the place. 🙈

limbonaut (approved):
I'd also add a note to doc comments about `content_type` side-effect. Looks good otherwise.

include/sentry.h
jpnurmi: https://github.com/getsentry/sentry-native/pull/1700/commits/6c3add4e5d9f50edf3deb283f0071ce8572a0e97

src/sentry_attachment.c
```diff
@@ -141,7 +173,6 @@ sentry__attachment_is_placeholder(
     const sentry_attachment_t *att, const sentry_options_t *options)
 {
     return options && options->enable_large_attachments && att
```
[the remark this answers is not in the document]
jpnurmi: Sounds like a nice finding and definitely worth investigating. I don't see the connection to these changes, though. 😅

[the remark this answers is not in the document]
jpnurmi: - https://github.com/getsentry/sentry-native/pull/1715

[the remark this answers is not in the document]
mujacica: Nice 🥇

mujacica (approved):

Issue #1688, opened 2026-04-30
Support .dmp caching when using the external crash reporter


dalnoki:
[This PR ](https://github.com/getsentry/sentry-native/pull/1607) added the ability to cache minidumps as separate .dmp files alongside .envelope files in the sentry DatabasePath cache directory. However, this feature only works when sentry-native is handling crashes directly via its own crashpad backend.

When the external Unreal crash reporter is active, crash handling is owned entirely by Unreal, and the sentry-native caching logic in process_completed_reports() is never reached. The external reporter path only writes a small session envelope via flush_external_crash_report(), so no minidump caching occurs.

Expected behavior:
When cache_keep is enabled and the external crash reporter is configured, .dmp files should be written to the DatabasePath cache directory alongside their corresponding .envelope files, consistent with the behavior of the native crashpad path.

jpnurmi:
Thanks for bringing it up! 👍

jpnurmi:
Is the `cache_keep` boolean option for offline caching (on failed sends) enough in general, or would you prefer a more fine-grained enum-like option to choose between a) no cache, b) offline cache, c) cache everything?

dalnoki:
Hey @jpnurmi, here's an update from the customer:

_I also want to clarify something just so there isn’t any confusion. When I said external crash reporter, I was referring to Sentry’s external crash reporter that is enabled with the .ini option “EnableExternalCrashReporter=True” and uses the Crashpad backend, not Unreals crash reporter._

I'll ask them about [this](https://github.com/getsentry/sentry-native/issues/1688#issuecomment-4350877481) as well.

dalnoki:
Hey @jpnurmi, I'm back with a comment from the customer :)

> I’m not sure I fully understand what the boolean `cache_keep` is used for. Based on the context, it sounds like that setting only allows for offline caching of reports that failed to send? If that’s the case, that wouldn’t be what we want. We want to be able to cache every .dmp and have it persist, regardless if it was successfully uploaded or not.

jpnurmi:
Oops, this was accidentally closed. Here's a status update while I'm here:

1. The Native SDK side has been prepared to pass `cache_dir` to the crash reporter, and we also introduced a 3-level cache mode configuration (never, offline, always)
   - https://github.com/getsentry/sentry-native/pull/1698
   - https://github.com/getsentry/sentry-native/pull/1707

2. The next step is to teach `cache_dir` to the external crash reporter:
[...]

PR #1707, opened 2026-05-07
feat: add cache keep modes

jpnurmi:
Turn `cache_keep` into an enum (with backwards compatible values; false=0, true=1):

0. `SENTRY_CACHE_KEEP_NONE` (false): Do not keep envelopes in the persistent cache.
1. `SENTRY_CACHE_KEEP_OFFLINE` (true): The classic Offline Caching mode. Envelopes that cannot be uploaded immediately are cached. Includes send failures and revoked envelopes.
2. `SENTRY_CACHE_KEEP_ALWAYS`: Envelopes are cached regardless of the upload result.

Relates to #1688 but does not fully solve it, as it needs changes in sentry-desktop-crash-reporter, too.


src/sentry_core.c
```diff
@@ -493,7 +493,8 @@ sentry__capture_envelope(sentry_transport_t *transport,
     }
     bool cached = false;
     if (options->cache_keep || options->http_retry) {
-        cached = sentry__run_write_cache(options->run, envelope, 0);
+        int retry_count = options->http_retry ? 0 : -1;
+        cached = sentry__run_write_cache(options->run, envelope, retry_count);
```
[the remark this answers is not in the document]
jpnurmi: yes. intentional, documented behavior

mujacica (approved):

src/sentry_options.c
```diff
@@ -512,7 +512,12 @@ sentry_options_get_symbolize_stacktraces(const sentry_options_t *opts)
 void
 sentry_options_set_cache_keep(sentry_options_t *opts, int enabled)
```
JoshuaMoelans: nit: `enabled` is still a bit boolean-sounding, maybe we change the parameter name to `mode`?

src/sentry_core.c
```diff
@@ -1683,6 +1684,10 @@ sentry__launch_external_crash_reporter(
         return false;
     }
 
+    if (options->cache_keep == SENTRY_CACHE_KEEP_ALWAYS) {
+        sentry__run_write_cache(options->run, envelope, -1);
```
JoshuaMoelans, line 1688: q: should we report on the result of `sentry__run_write_cache` (like we do in [here](https://github.com/getsentry/sentry-native/blob/fc807a12525534f7d041636a7b8688fb1ef56e35/src/sentry_core.c#L494-L506) for example)

tests/unit/test_retry.c
```diff
@@ -425,6 +425,49 @@ SENTRY_TEST(retry_cache)
     sentry_close();
 }
 
+SENTRY_TEST(retry_cache_keep_always)
```
JoshuaMoelans, line 428: q: could we maybe use a test for the path in [sentry_retry.c](https://github.com/getsentry/sentry-native/blob/fc807a12525534f7d041636a7b8688fb1ef56e35/src/sentry_retry.c#L129-L145) where we do the duplicate file checking?

JoshuaMoelans (approved):
LGTM, just a few open q's left

src/sentry_options.c
jpnurmi: https://github.com/getsentry/sentry-native/pull/1707/commits/8532f488241094202c7eacb05d55a741bb16872a

tests/unit/test_retry.c
jpnurmi: https://github.com/getsentry/sentry-native/pull/1707/commits/509b6210346e35650add0b8d0c9062504cabb6cc

src/sentry_options.c
```diff
@@ -510,9 +510,14 @@ sentry_options_get_symbolize_stacktraces(const sentry_options_t *opts)
 }
 
 void
-sentry_options_set_cache_keep(sentry_options_t *opts, int enabled)
+sentry_options_set_cache_keep(sentry_options_t *opts, int mode)
 {
-    opts->cache_keep = !!enabled;
+    if (mode < SENTRY_CACHE_KEEP_NONE) {
+        mode = SENTRY_CACHE_KEEP_OFFLINE;
```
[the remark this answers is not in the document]
jpnurmi: `!= 0` used to be `true` which maps to `SENTRY_CACHE_KEEP_OFFLINE` for backward compatibility

src/sentry_core.c
jpnurmi: hmm, it's indeed tempting to put some informative logging to `sentry__launch_external_crash_reporter`, but it's called during crash handling so we should probably keep it to the minimum and preferably unformatted signal-safe logs... 🤔

actually, there are already a couple of related mkdir/write warnings in `sentry__run_write_cache` - would those suffice?

https://github.com/getsentry/sentry-native/blob/fc807a12525534f7d041636a7b8688fb1ef56e35/src/sentry_database.c#L466

JoshuaMoelans: ah, that should be fine then! I didn't look into the full path, just saw the `INFO` in the earlier handling of the `sentry__run_write_cache` outcome.

JoshuaMoelans (approved):

PR #1708, opened 2026-05-08
fix(native/breakpad): capture abort() on Windows

### LINKED ISSUE #591: Alternative backends on Windows don't catch crashes caused by `abort()`
Using Breakpad or inproc on Windows results in most crashes not being captured.

- [ ] During build
- [ ] During run-time

**Environment**

- OS: Only Windows (in my case, Windows 10, 64-bit in VirtualBox)
- Compiler: MSVC 19
- CMake version and config: 3.21.2 `-D SENTRY_BACKEND=breakpad -D BUILD_SHARED_LIBS=OFF`

**Steps To Reproduce**

Just a simple boilerplate setup does it. Using `abort()` to cause the crash.

```c++
int main() {
  std::cout << "hello world" << std::endl;

  sentry_options_set_dsn(options, 'dsn redacted');

  sentry_init(options);

  std::this_thread::sleep_for(std::chrono::milliseconds(2000));

  abort();

  sentry_close();
}
```

**Log output**

This is the logs when it doesn't work. As you can see, no crash is captured.

```
hello world
[sentry] INFO using database path "C:\Users\vagrant\NonRepos\SentryNativeTutorial\.sentry-native"
[sentry] DEBUG starting transport
[sentry] DEBUG starting background worker thread
[sentry] DEBUG starting backend
[sentry] DEBUG background worker thread started
The terminal process "C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -Command ./install/bin/tutorial.exe" terminated with exit code: 1.
```

However, if I swap out `abort();` for
```
  volatile char* invalid_memory = nullptr;
  *invalid_memory = '\0';  // crash
```
to invoke a different kind of crash, I get

```
hello world
[sentry] INFO using database path "C:\Users\vagrant\NonRepos\SentryNativeTutorial\.sentry-native"
[sentry] DEBUG starting transport
[sentry] DEBUG starting background worker thread
[sentry] DEBUG starting backend
[sentry] DEBUG background worker thread started
[sentry] INFO entering breakpad minidump callback
The terminal process "C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -Command ./install/bin/tutorial.exe" terminated with exit code: 1.
```

That is the first run, and on subsequent runs I see the envelope is uploaded and appears in the Sentry UI.

Similarly, if I change over to use crashpad as the backend, things work fine (but crashpad is not an option for me on this project). Things also work fine on Linux and Mac with the breakpad backend.

In my project, there's a lot of assertions so most crashes are caused in the way the Windows can't handle them.

For now, my hacky workaround is to register my own signal handler like so:

```
[[noreturn]] void windows_signal_handler(int signal) {
  sentry_drop_breadcrumb(fmt::format("signal: {}", signal).c_str(), "windows_signal_handler_hack",
                         SentryBreadcrumbLevel::Info);
  sentry_drop_breadcrumb(
      "invoking Sentry Native Breakpad signal handler by raising SEH exception EXCEPTION_ACCESS_VIOLATION_WRITE",
      "windows_signal_handler_hack", SentryBreadcrumbLevel::Info);
  volatile char* invalid_memory = nullptr;
  *invalid_memory = '\0';  // crash
  // unreachable
}
```

```
std::signal(SIGABRT, windows_signal_handler);
```

This results in the full stack trace appearing on the UI, albeit with an extra frame or two.

jpnurmi:
The handler was directly copied from inproc (#1446):

https://github.com/getsentry/sentry-native/blob/104513c21092163eab0c3e913ab0e4da38ee9360/src/backends/sentry_backend_inproc.c#L593-L629

Example event: https://sentry-sdks.sentry.io/issues/7467638795

Close: #591

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,10 @@
 **Features**:
 
 - Auto-populate `event.user.id` with a persistent per-installation UUID when no explicit user ID is set. ([#1661](https://github.com/getsentry/sentry-native/pull/1661))
-- Native/Windows: added capability to capture abort(). ([#1708](https://github.com/getsentry/sentry-native/pull/1708))
+
+**Fixes**:
+
+- Native/Breakpad/Windows: fixed capturing abort(). ([#1708](https://github.com/getsentry/sentry-native/pull/1708))
 
 ## 0.14.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Features**:
 
 - Auto-populate `event.user.id` with a persistent per-installation UUID when no explicit user ID is set. ([#1661](https://github.com/getsentry/sentry-native/pull/1661))
+- Native/Windows: added capability to capture abort(). ([#1708](https://github.com/getsentry/sentry-native/pull/1708))
 
 ## 0.14.0
 
--- a/src/backends/native/sentry_crash_handler.c
+++ b/src/backends/native/sentry_crash_handler.c
@@ -25,6 +25,7 @@
 #    include <sys/types.h>
 #    include <unistd.h>
 #elif defined(SENTRY_PLATFORM_WINDOWS)
+#    include <signal.h>
 #    include <tlhelp32.h>
 #    include <windows.h>
 #    include <winnt.h>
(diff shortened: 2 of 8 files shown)


jpnurmi:
A) Do we care to share the handler code somewhere?
B) Is it a feature, or rather a fix? 😅

supervacuus:
> A) Do we care to share the handler code somewhere?

I don't think this is a high-value module for reuse or downstream maintenance. Maybe one could even argue that keeping implementations separate allows the backends to add backend-specific context that should be distinct with the underlying UEF implementation.

> B) Is it a feature, or rather a fix? 😅 

I would call it a fix, but I don't think it matters much.

supervacuus (approved):

jpnurmi:
> I don't think this is a high-value module for reuse or downstream maintenance. Maybe one could even argue that keeping implementations separate allows the backends to add backend-specific context that should be distinct with the underlying UEF implementation.

I ended up with a shared helper after all, because including `<signal.h>` conflicts with breakpad's definitions:
https://github.com/getsentry/breakpad/blob/47d70322c848012ed801e36841767d7ffb79412d/src/google_breakpad/common/minidump_exception_linux.h#L52

mujacica (approved):

PR #1704, opened 2026-05-05
feat(screenshot): allow downstream Xbox SDK to supply capture implementation

tustanivsky:
This PR allows the downstream sentry-xbox extension to provide its own implementation of `sentry__screenshot_capture` by skipping the bundled Windows screenshot module when targeting Xbox.

Related items:
- https://github.com/getsentry/sentry-xbox/pull/135

#skip-changelog

PR #1709, opened 2026-05-08
feat(session-replay): add `attach_session_replay` option with platform extension point

### LINKED ISSUE #1172: video clips on Xbox dev kit
Relates to:
* https://github.com/getsentry/sentry-native/issues/693

On Xbox. See GDK's APIs: [`XAppCaptureRecordDiagnosticClip`](https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/reference/system/xappcapture/functions/xappcapturerecorddiagnosticclip)

> Calling this function does not trigger a UI notification. Clips are saved to a game accessible partition. The [XAppCaptureRecordDiagnosticClipResult](https://learn.microsoft.com/en-us/gaming/gdk/_content/gc/reference/system/xappcapture/structs/xappcapturerecordclipresult) contains a file path which you can use to access the clip. On Xbox, this function will only be available on development kits. On retail kits this api will return an error immediately.

> On PC, you must enable background recording for your app for this function to work. The best way to do this is to bring up Game Bar by pressing Win+G, then click the Settings icon, go to the Capturing tab, and enable "Record in the background while I'm playing a game". Then, you must mark your app as a game. The best way to do this is by running your game first. While the game is in focus, bring up Game Bar by pressing Win+G, then click the Settings icon, go to the General tab, and check "Remember this is a game".

tustanivsky:
This PR adds two new options to enable attaching a short retroactive video clip to captured crash event:

- `sentry_options_set_attach_session_replay(opts, int val)` - toggle (default off)
- `sentry_options_set_session_replay_duration(opts, uint32_t)` - requested window (default 5000 ms)

It mirrors the existing screenshot architecture: a public option, an internal `sentry_session_replay.h` interface, a no-op default impl, and an extension point so downstream SDKs (e.g. sentry-xbox using `XAppCaptureRecordDiagnosticClip`) can supply a real implementation.

Closes #1172 

Related items:
- https://github.com/getsentry/sentry-xbox/pull/136

#skip-changelog


tustanivsky:
Console checks are failing for the same reason as in the case with Xbox screenshot capturing (see https://github.com/getsentry/sentry-native/pull/1704#issuecomment-4380245962).

src/backends/sentry_backend_breakpad.cpp
```diff
@@ -219,6 +220,18 @@ breakpad_backend_callback(const google_breakpad::MinidumpDescriptor &descriptor,
...
+                sentry_attachment_t *clip = sentry__attachment_from_path(
+                    sentry__replay_clip_get_path(options));
+                if (clip
+                    && sentry__replay_clip_capture(
+                        clip->path, options->replay_clip_duration_ms, 0)) {
+                    sentry__envelope_add_attachment(envelope, clip);
+                } else {
+                    sentry__attachment_free(clip);
+                }
+            }
```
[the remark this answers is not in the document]
tustanivsky: Fixed via https://github.com/getsentry/sentry-native/pull/1709/commits/b08dd9ca879e04d14be8152b074674129c26d54e

JoshuaMoelans (approved):

src/sentry_options.c
```diff
@@ -656,6 +658,19 @@ sentry_options_set_before_screenshot(sentry_options_t *opts,
...
+    opts->attach_replay_clip = !!val;
+}
+
+void
+sentry_options_set_replay_clip_duration_ms(
+    sentry_options_t *opts, uint32_t duration_ms)
+{
+    opts->replay_clip_duration_ms = duration_ms;
+}
+
```
JoshuaMoelans, line 673: maybe one small q: are we calling it `replay_clip` for a specific reason (instead of using something more `session_replay`-y like we have in the other SDKs?)

jpnurmi:
Closes: #1172 🙂

src/sentry_options.c
tustanivsky: Changed naming from "replay clip" to "session replay" in https://github.com/getsentry/sentry-native/pull/1709/commits/fd02e93f92a58c695668a943adc05671c23011ed.

include/sentry.h
```diff
@@ -1656,6 +1656,27 @@ SENTRY_EXPERIMENTAL_API void sentry_options_set_before_screenshot(
...
+    sentry_options_t *opts, int val);
+
+/**
+ * Sets the requested duration of the retroactive session replay in
+ * milliseconds.
+ *
+ * The resulting replay can be shorter than the requested duration if it hasn't
+ * accumulated enough buffered frames yet. Defaults to 5000 ms.
+ */
+SENTRY_EXPERIMENTAL_API void sentry_options_set_session_replay_duration_ms(
```
jpnurmi: Q: Do we want the unit in the function name here?

I have mixed feelings because the suffix makes things clearer, but at the same time, it feels a bit inconsistent because none of the existing API uses such suffixes. 🤔

tustanivsky: Consistency with the rest of the API probably wins here - let's drop the "ms" suffix.

tustanivsky: This should be addressed via https://github.com/getsentry/sentry-native/pull/1709/commits/e06478531ba7b87c602e2bde1ad17b9215130c5a.

jpnurmi (approved):

PR #1711, opened 2026-05-11
wip: allow fetching `event_id` from last crash

### LINKED ISSUE #1447: Allow fetching `event_id` from last crash
We already write the crash timestamp to the crash-marker in `sentry__write_crash_marker`. Since we now have an `event_id` for all backends (since https://github.com/getsentry/sentry-native/pull/1319), we could additionally write this to the same file & provide a method to fetch this value.

Depending on how this value is to be used, we could additionally continue with https://github.com/getsentry/sentry-native/issues/885#issuecomment-3073596212 and the `sentry_options_set_on_crashed_last_run()` proposal, allowing users to add data to an envelope. This is probably out-of-scope for the `event_id` issue, and we might split this issue up in two.

jpnurmi:
Expose `sentry_get_last_crash` so callers can inspect the previous crash timestamp and event ID after initialization.

```c
sentry_last_crash_t crash;
if (sentry_get_last_crash(&crash, sizeof(&crash))) {
    char event_id[37];
    sentry_uuid_as_string(&crash.event_id, event_id);
    fprintf(stderr, "event_id: %s\n", event_id); // "event_id: 006be96a-0100-0000-68a3-320701000000"
    fprintf(stderr, "timestamp: %llu\n", crash.timestamp); // "timestamp: 1778520001725508"
}
```

The crash event ID is persisted in the same crash marker file next to the existing timestamp, while keeping the old timestamp-only marker format readable.

Close: #1447

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Features**:
+
+- Add `sentry_get_last_crash` to allow fetching `event_id` from last crash. ([#1711](https://github.com/getsentry/sentry-native/pull/1711))
+
 **Fixes**:
 
 - Reject overly deep msgpack payloads during deserialization. ([#1727](https://github.com/getsentry/sentry-native/pull/1727))
--- a/include/sentry.h
+++ b/include/sentry.h
@@ -3563,6 +3563,39 @@ SENTRY_EXPERIMENTAL_API void sentry_transaction_iter_headers(
 SENTRY_EXPERIMENTAL_API int sentry_get_crashed_last_run(void);
 
 /**
+ * Information about the crash from the last run.
+ */
+typedef struct sentry_last_crash_s {
+    /**
+     * The crash timestamp in microseconds since the Unix epoch, or 0 if it is
+     * unavailable.
+     */
+    uint64_t timestamp;
+    /**
+     * The crash event ID, or a nil UUID if it is unavailable.
+     */
+    sentry_uuid_t event_id;
+} sentry_last_crash_t;
+
+/**
+ * Returns whether the application crashed on the last run, and populates
+ * `crash` with available information.
+ *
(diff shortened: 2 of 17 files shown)


mujacica (approved):

JoshuaMoelans (approved):
LGTM!

jpnurmi:
While I like the `sentry_get_last_crash` and `sentry_last_crash_t` naming in isolation, I'm not sure if it's the right choice provided that we already have:

```c
SENTRY_EXPERIMENTAL_API int sentry_get_crashed_last_run(void);
SENTRY_EXPERIMENTAL_API int sentry_clear_crashed_last_run(void);
```

Do you have any better naming ideas? Perhaps something like `sentry_get_crashed_last_info`, `sentry_get_crashed_last_run_info`, or `sentry_get_crashed_last_run_s` to keep it consistent with the "crashed last run" terminology, which seems also consistent with other SDKs?

Alternatively, since the API is tagged experimental, would a break be acceptable to add optional/nullable arguments to the existing getter? Then we could name the struct `sentry_crashed_last_run_t`... :thinking:

jpnurmi:
After digesting this for a few days, I'm leaning towards not exposing specific crash details like this via a getter because it would be ambiguous in case of multiple crashes.

Exposing a plain status of whether crashes were registered, the way `get_last_crashed_run` currently does, is fine. However, I think `on_crashed_last_run` as proposed in #885 is a better way to expose the details of specific crashes, because
- it gets unambiguously called at the right time during the processing of old runs,
- can handle potentially multiple unprocessed crashes in multi-process and early startup crash scenarios, and
- seems more consistent with other SDKs.

Please let me know if you disagree with this assessment. 🙏 I'll reject this proposal for now, and will try to solve `on_crashed_last_run` in a native backend/daemon-compatible way sometime later.

PR #1723, opened 2026-05-14
fix: exclude WER code from Xbox builds in native backend

tustanivsky:
This PR fixes Xbox build failures in the native backend caused by WER (Windows Error Reporting) code being compiled on Xbox, where `werapi.h`, registry APIs and WER APIs are not available in the GDK.

Xbox defines `SENTRY_PLATFORM_WINDOWS` (because it also defines `_WIN32`), which caused all existing `SENTRY_PLATFORM_WINDOWS` guards around WER-related code to evaluate as true. This PR adds exclusions to those guards to match the existing platform handling patterns already used elsewhere in the codebase.

Additionally, the CMake condition for the `sentry-wer` target has been updated.

Failing CI run: https://github.com/getsentry/sentry-xbox/actions/runs/25810318958/job/75848636085

#skip-changelog

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -879,7 +879,7 @@ elseif(SENTRY_BACKEND_NATIVE)
 
 	# Make sentry library depend on crash daemon so it's always built together
 	add_dependencies(sentry sentry-crash)
-	if(WIN32)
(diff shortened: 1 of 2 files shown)


jpnurmi (approved):
Oh no, not again… 😭

PR #1747, opened 2026-05-22
feat(native): symbolicated stacktraces for non-crashed threads on Linux

jpnurmi:
Add libunwind remote unwinding support to the native backend's crash daemon, enabling DWARF-based stack walking and symbol name resolution for non-crashed threads on Linux. Since #1764, only the crashing thread received a stacktrace; all other threads appeared without one.

- Re-add upstream libunwind `src/ptrace/` (`_UPT_*` accessors)
- Add `unwind_remote` CMake target with G-prefix + ptrace sources, only built when `SENTRY_BACKEND=native` on Linux
- New `sentry_remote_unwind.c` using `unw_init_remote()` + `unw_get_proc_name()`
- Integrate into daemon's `build_stacktrace_for_thread()` with fallback to pre-captured backtrace and FP-walking

| Before | After |
|---|---|
| before | after |

See also:
- https://github.com/getsentry/sentry-desktop-crash-reporter/pull/212


src/backends/native/sentry_crash_daemon.c
```diff
@@ -900,10 +903,71 @@ build_stacktrace_for_thread(
...
+                    if (remote_frames[i].ip == 0
+                        || !is_valid_code_addr(remote_frames[i].ip)) {
+                        continue;
+                    }
+                    temp_frames[frame_count] = sentry_value_new_object();
+                    sentry_value_set_by_key(temp_frames[frame_count],
+                        "instruction_addr",
+                        sentry__value_new_addr(remote_frames[i].ip));
+                    sentry_value_set_by_key(temp_frames[frame_count], "trust",
+                        sentry_value_new_string(i == 0 ? "context" : "cfi"));
```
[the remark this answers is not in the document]
jpnurmi: The `trust` value describes how the frame was obtained, not its position after filtering. `i == 0` is the frame produced from the initial cursor/context; later indices are produced after unwinding. If the initial context frame is filtered out, the next emitted frame was still obtained via unwind metadata, so `"cfi"` is correct rather than `"context"`.

src/unwinder/sentry_unwinder_libunwind_remote.c
```diff
@@ -0,0 +1,175 @@
...
+        return 0;
+    }
+
+    int status;
+    if (waitpid(tid, &status, __WALL) < 0) {
+        SENTRY_WARNF(
+            "remote_unwind: waitpid failed for %d: %s", tid, strerror(errno));
+        ptrace(PTRACE_DETACH, tid, NULL, NULL);
+        return 0;
+    }
```
supervacuus: Maybe worth noting that this could block indefinitely (threads can get stuck before reaching a stop point), which would change the client contract. This can be a fair tradeoff, but worth noting.

vendor/libunwind/src/ptrace/_UPT_access_reg.c
```diff
@@ -0,0 +1,402 @@
...
+  if (!write)
+    Debug (16, "%s[%u] -> %lx\n", unw_regname (reg), (unsigned) reg, (long) *val);
+#endif
+  return 0;
+
+ badreg:
+  Debug (1, "bad register %s [%u] (error: %s)\n", unw_regname(reg), reg, strerror (errno));
+  return -UNW_EBADREG;
+}
+#elif defined(HAVE_DECL_PT_GETREGS) && defined(__linux__)
```
supervacuus: Ugh... so I had a quick look at the `config.h.cmake.in` for the vendored `libunwind` because I remembered it included a couple of `PTRACE`/`PT` defs. And I guess I got lucky 😓 

This particular def seems to be an upstream bug because it should use the value rather than check whether it is defined. I also checked the autoconf script again, and other uses of the `ptrace` defs seem to indicate that this is primarily a leftover. 

Just to be clear, this is likely a non-issue on most targets, but for maximum compatibility:

* Let's replace this `defined()` (and the one for `FreeBSD` below, even if we don't support it) with the value (like in [other users of these defs](https://github.com/getsentry/sentry-native/pull/1747/changes#diff-bbfd8ded60ed41cd3dccffe6c53a53a20dba5da4d596bf6bb93a22cdf0876683R69)). I leave this up to you whether to raise this with upstream.
* In `config.h.cmake.in`, use `#cmakedefine01 HAVE_DECL_PT_GETREGS` etc., not hardcoded 0. This is on me, I only did this because I was "sure we never needed it(tm)". The comment should also be removed, because it is only true for `aarch64`.
* In `vendor/libunwind/CMakeLists.txt`, let's check the `PT_*` declarations just like the `PTRACE_*` declarations under `# Check for PTRACE declarations`. Same, mea culpa.

supervacuus (approved):
LGTM. Just one documentation feedback and one build-cleanup.

vendor/libunwind/src/ptrace/_UPT_access_reg.c
jpnurmi: very nice, thanks! https://github.com/getsentry/sentry-native/pull/1747/commits/8c90ccfca9df06347399deec66d1f215aad72ac4

src/unwinder/sentry_unwinder_libunwind_remote.c
jpnurmi: The clanker didn't like it either: https://github.com/getsentry/sentry-native/pull/1747#discussion_r3318226520

If we'd go with a non-hanging wait loop, does 50 x 100ms sound like a reasonable limit?

supervacuus: I didn't say I don't like it 😅 

I would align it with whatever you consider a fair wait time for a stack trace. 5 seconds seems like a lot to me, given that even our shutdown limit defaults to only 2 seconds. But you might have a better sense of your scenario.

mujacica (approved):

Issue #1398, opened 2025-10-02
Structured Logs: follow-up


JoshuaMoelans:
The following items are all potential follow-ups of the initial Structured Logs implementation introduced in https://github.com/getsentry/sentry-native/pull/1271 . 

- **add debug output for structured logs when `debug` option is enabled**
  - https://github.com/getsentry/sentry-native/pull/1466
- **increase logs queue size**
  - https://github.com/getsentry/sentry-native/issues/1469
- **length format specifier handling**
  - see [comment](https://github.com/getsentry/sentry-native/pull/1271#discussion_r2344290456)
  - https://github.com/getsentry/sentry-native/issues/1401
- **replace `cond_wake` with level-triggered flag**
  -  see [comment](https://github.com/getsentry/sentry-native/pull/1271#discussion_r2344303173)
  - https://github.com/getsentry/sentry-native/issues/1402
- **Set `sentry.origin` for logs coming from integrations**
  - For now, only if created from downstream (e.g. Unreal), since native itself just has the manual API.
  - https://develop.sentry.dev/sdk/telemetry/logs/#sdk-integration-attributes
-  **Logger integrations**
    - logging `sentry__logger_func()` calls can be added as an option, but need to disable logging when in the `sentry_logs.c` code to avoid recursion.
    - community requests for C/C++ logging libraries (feel free to add requests below!)
  - **Named parameter API**
    - to allow for better searchability for `sentry.message.parameter.X` where X is no longer confined to being numerical, we can add an alternative API that allows passing in a `sentry_value_t` list containing (key, value) pair objects. Proof-of-concept already lives here https://github.com/getsentry/sentry-native/pull/1297

## Done

- Using `SENTRY_SDK_NAME` instead of always writing `sentry.native` when setting the sdk name attribute
  - this changes to `native.android` or `native.xbox` in our SDK, and gets set to `native.unreal/unity/godot` from the respective GDX SDKs
  - https://github.com/getsentry/sentry-native/pull/1399
- **ensure logs get sent/cached to disk on crash**
  - inspired by https://github.com/getsentry/rfcs/pull/148
  - https://github.com/getsentry/sentry-native/issues/1403
  - https://github.com/getsentry/sentry-native/pull/1404
- **Allow passing in custom attributes**
[...]

PR #1752, opened 2026-05-25
fix(logs): respect printf length modifiers in params

### LINKED ISSUE #1401: Structured Logs: follow up - length format specifier handling
For context https://github.com/getsentry/sentry-native/pull/1271#discussion_r2344290456

We currently skip all length format specifiers, which leads to limitations in what our API can take in as arguments. We should properly handle these.

Part of https://github.com/getsentry/sentry-native/issues/1398

jpnurmi:
Fix structured log parameter extraction to respect printf argument widths.

The previous parser always consumed integer variadic arguments as 64-bit values, even for default-width specifiers like `%d` and `%u`. On 32-bit platforms this can read adjacent stack data, desynchronize later arguments, and serialize leaked or corrupted values into log attributes.

This also consumes dynamic width and precision arguments so formats like `%*d` stay aligned with the underlying `va_list`.

Closes https://github.com/getsentry/sentry-native/issues/1401 , part of https://github.com/getsentry/sentry-native/issues/1398


JoshuaMoelans (approved):
Very nice, LGTM! 🚀

PR #1753, opened 2026-05-26
build: add `SENTRY_LIBUNWIND_SYSTEM` option

### LINKED ISSUE #1714: [feature] Add option to consume unvendored libunwind
Greetings! 

As the project Sentry native is packaged in [Conan Center](https://conan.io/center/recipes/sentry-native?version=0.12.6) already, and during our last review https://github.com/conan-io/conan-center-index/pull/29690 was spotted a room for improvement, I'm sending this feature request / suggestion. 

As Conan and other package managers can provide not only Sentry native, but also other dependencies, it would be great if we could manage libunwind as well, so consumers would not need to build that dependency from source too, when needed to use Sentry native. 

As a suggestion, you could add a CMake option like `WITH_EXTERNAL_LIBUNWIND`, then use a logic like:

https://github.com/getsentry/sentry-native/blob/master/CMakeLists.txt#L648

```cmake
if(SENTRY_WITH_LIBUNWIND AND LINUX)
	if(WITH_EXTERNAL_LIBUNWIND)
        find_package(libunwind REQUIRED)
        target_link_libraries(sentry PRIVATE libunwind::libunwind)
    else()
		# Use vendored libunwind
		add_subdirectory(vendor/libunwind)
		target_link_libraries(sentry PRIVATE unwind)
		if(NOT SENTRY_BUILD_SHARED_LIBS)
			sentry_install(TARGETS unwind EXPORT sentry
				LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
				ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
			)
		endif()
	endif()
endif()
```

As a result, it would be possible to combine libunwind from the system, or the package manager. It's true libunwind does not provide an official CMake config file, but it could be easily wrapped by the consumer or the package manager (Conan provides one automatically). 

Regards!

jpnurmi:
Add `SENTRY_LIBUNWIND_SYSTEM` as a Linux-only opt-in for using system `libunwind` via pkg-config instead of the vendored copy.

Verified with the Alpine CI builds:
```
-- Checking for module 'libunwind'
--   Found libunwind, version 1.8.1
```

Close: #1714


uilianries (commented):
LGTM ✅

supervacuus (approved):
This is fine, but it requires clear documentation under compile options in the README.

In this proposed setup, there is a dramatic asymmetry that this documentation should highlight:

* The vendored `libunwind` _will always be built as a static archive_ and either linked into the resulting shared library or colocated with the other resulting static artifacts as output. This is what most Native SDK consumers will expect.
* when `SENTRY_LIBUNWIND_SYSTEM=On`, the decision of what kind of library artifact will be included is entirely deferred to the host distribution/sysroot/package manager. This path only considers `SENTRY_BUILD_SHARED_LIBS` in terms of how it makes an artifact available to a consuming CMake project, not in terms of how contained the resulting deployable install will be. The latter will have to be solved externally (either via a package manager configuration or through a very careful setup of packages on the build and target machines).

The latter, besides being different from the former, is also in stark contrast to what we had before we vendored `libunwind` as the default unwinder on Linux.

This is neither good nor bad, but it requires clear documentation of expectations, since it is now part of the build and package contract. The story is entirely different for users who can configure their build from within a source-level package manager like `conan` or `vcpkg`, and those who build and deploy against binary distro packages.

README.md
```diff
@@ -259,6 +259,16 @@ using `cmake -D BUILD_SHARED_LIBS=OFF ..`.
 - `SENTRY_BREAKPAD_SYSTEM` (Default: `OFF`):
   This instructs the build system to use system-installed breakpad libraries instead of the in-tree version.
 
+- `SENTRY_LIBUNWIND_SYSTEM` (Default: `OFF`, only for Linux):
+  This instructs the build system to use a system-installed `libunwind` (found via `pkg-config`) instead of the
+  vendored copy in `vendor/libunwind`.
+
+  In contrast to the vendored `libunwind` is always built as a static archive and either linked into the resulting
```
supervacuus: In contrast to the vendored `libunwind` **_which_** is always built [...]

jpnurmi: thanks, good catch 😅

PR #1755, opened 2026-05-26
ci: Bump kcov and re-enable crash exit checks

jpnurmi:
Pin the CI-built kcov revision to the upstream merge that preserves signal exit codes after crashes.

Remove kcov-specific crash exit fallbacks and per-test skips for paths that can run under kcov, while keeping ptrace-based Crashpad and Breakpad tests disabled.

See also:
- https://github.com/SimonKagstrom/kcov/pull/495


supervacuus (approved):
Nice!

PR #1763, opened 2026-05-27
ref(value)!: atomic decref result check

jpnurmi:
> [!NOTE]
> SC/BC break (low impact)

Fixes TOCTOU races in `sentry__transaction_context_free`, `sentry__transaction_decref` and `sentry__span_decref` by adding return value for `sentry_value_decref` indicating whether the value still has references, has been freed, or is primitive that needs no tracking:
```c
// before
if (sentry_value_refcount(value) <= 1) {
  sentry_value_decref(value); // race, may have changed
  sentry_free(related_resources);
}

// after
if (!sentry_value_decref(value)) {
  sentry_free(related_resources);
}
```

While taking the BC/SC break hit, make `sentry_value_incref` return the incref'd value for convenience, to help reduce such boilerplate:
```c
// before
sentry_value_incref(value);
do_something(value);

// after
do_something(sentry_value_incref(value));
```

<details><summary>Transaction and span refcount decrement has TOCTOU race</summary>

## Details
sentry__transaction_decref (lines 352-364) and sentry__span_decref (lines 375-388) implement non-atomic check-then-act on the reference count. The code reads the refcount, compares it to 1, and if equal, proceeds to free the object. Between the read and the free, another thread could increment the refcount (via incref), creating a use-after-free. Alternatively, two concurrent decref calls could both read refcount=2, both decrement to 1, and neither triggers the free, causing a memory leak. Or both read refcount=1 and both free, causing a double-free.

## Location
[src/sentry_tracing.c:352](https://github.com/getsentry/sentry-native/blob/master/src/sentry_tracing.c#L352)

## Impact
Double-free or use-after-free from concurrent refcount operations.

## Reproduction steps
1. Thread A calls sentry__transaction_decref, reads refcount=1, and is about to free. Thread B calls sentry__transaction_incref, incrementing refcount to 2. Thread A proceeds to free the transaction. Thread B now holds a dangling pointer. When Thread B later decrefs, it accesses freed memory, causing a use-after-free that could lead to code execution.

## Recommended fix
Use atomic decrement (e.g., __atomic_sub_fetch) and check the result atomically, ensuring the free decision is based on the atomic operation's return value.

---
**Severity:** MEDIUM
**Status:** Open
**Category:** Race Condition
**Repository:** getsentry/sentry-native
**Branch:** master

</details>


mujacica (approved):

JoshuaMoelans (approved):
nice 👌  LGTM

PR #1765, opened 2026-05-28
feat(apple): use `os_sync` for waitable flag on modern Darwin

supervacuus:
This aligns macOS batcher implementations with Windows and Linux, at least for certain versions of macOS (14.4+)/iOS (17.4+).

* Uses weak linking with a runtime fallback to the POSIX implementation if the target is below the minimum
* If the target matches, it uses hard linking
* If the build does not have `os_sync_wait_on_address` accessible, it just builds for a generic POSIX as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Native: add opt-in async crash upload mode so crashed apps can exit early after crash data is captured, while the crash daemon finishes potentially large uploads in the background. ([#1739](https://github.com/getsentry/sentry-native/pull/1739))
 - Add a `transfer_timeout` option for SDK-managed HTTP transports. ([#1741](https://github.com/getsentry/sentry-native/pull/1741))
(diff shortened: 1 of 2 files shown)


jpnurmi (approved):
Pity that the oldest macOS 14 CI runners are 14.8

supervacuus:
> Pity that the oldest macOS 14 CI runners are 14.8

Tbh, this is less of a runtime question, especially given that the fallback for that platform isn't the new thing in this PR, but rather a toolchain question.

So if one wanted to be glorious, one could run the batcher in isolation on an old iOS emulator image. But pretty much all conditionals could be tested from additional toolchain configurations:

* use [outdated SDKs](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md#installed-sdks) to trigger the build-time fallback to POSIX
* use a more recent SDK, but target a macOS version below the threshold to trigger the weak linking path (that could also then test the runtime conditional of actually falling back on an outdated iOS emulator run)
* use a recent SDK + a recent target macOS and stress the hard linking path

But that is a can of worms for another time, since that feature is by no means the only one that could benefit from that level of rigor.

PR #1742, opened 2026-05-22
poc: add heartbeat functionality to native backend

bitsandfoxes:
Resolves https://github.com/getsentry/team-gdx/issues/155

> [!WARNING]
> ### WIP&nbsp;&nbsp;&nbsp;&nbsp;🚧🔨⏳⛔

PR #1769, opened 2026-05-29
ref(native): scope application to event

bitsandfoxes:
Enables https://github.com/getsentry/sentry-native/pull/1742

This PR does not change the current output at all. For #1742 the daemon needs to be able to create non-crash events (i.e. app hangs).

- `build_native_crash_event` -> `build_native_event` with updated signature. This is what the app hang feature will then also use
- `native_backend_apply_scope` is the "source of truth" for all paths (update file & actual event)
- The way the snapshot that gets applied to events gets updated should be unified. This means the Windows specific duplication gets removed

#skip-changelog

### DIFF
--- a/src/backends/sentry_backend_native.c
+++ b/src/backends/sentry_backend_native.c
@@ -839,6 +839,9 @@ native_backend_flush_scope(
     // Keep the on-disk base event complete and current, so the daemon has the
     // full scope even if a crash beats the in-process handler to the file.
     sentry_value_t event = sentry_value_new_object();
+    // Default to `FATAL` for all paths, i.e. minidump mode.
+    sentry_value_set_by_key(
+        event, "level", sentry__value_new_level(SENTRY_LEVEL_FATAL));
     native_backend_apply_scope(event, options);
 
     size_t json_len = 0;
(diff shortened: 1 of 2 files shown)


src/backends/sentry_backend_native.c
```diff
@@ -767,9 +767,61 @@ native_backend_write_attachments(const sentry_path_t *event_path)
     }
 }
 
+#if defined(SENTRY_PLATFORM_WINDOWS)
+// Sentry's symbolicator needs `contexts.device.arch` to process PE modules. If
+// the scope already carries a device context with arch (host SDKs like Unity
+// provide one), leave it; otherwise synthesize a minimal one so native-only
+// consumers still symbolicate.
+static void
+native_backend_ensure_device_arch(sentry_value_t event)
```
jpnurmi: The `native_backend_` prefix is used only for native backend "methods" that implement the `sentry_backend_t` interface. It could be left out of both internal helpers; this and `native_backend_apply_scope`.

jpnurmi (approved):

mujacica:
Maybe a unit-test extension, or making sure there is a native unit-test for checking the breadcrumbs

bitsandfoxes:
> Maybe a unit-test extension, or making sure there is a native unit-test for checking the breadcrumbs

I'll follow up on the breadcrumbs in https://github.com/getsentry/sentry-native/pull/1775

PR #1774, opened 2026-06-01
fix(native): prevent IPC shared memory spoofing on Linux

jpnurmi:
Create Linux crash IPC shared memory with a random nonce and avoid reusing pre-existing segments.

<details><summary>Shared memory name prediction enables same-user IPC spoofing</summary>

## Details
The shared memory segment name is derived from `getpid() ^ pthread_self()` at line 36, formatted as `/s-%08x`. The PID is observable via /proc and pthread_self() is often deterministic for the main thread. When `shm_open` with O_EXCL fails (EEXIST) at lines 48-53, the code falls back to opening the existing segment with no ownership verification (no fstat uid check). The daemon reads unvalidated DSN, proxy URL, database_path, and external_reporter_path from this shared memory. A same-user attacker can pre-create the segment with controlled content. Although the app overwrites the segment after opening, the attacker retains a writable mapping and can race to overwrite values between the app's writes and the daemon's reads.

## Location
[src/backends/native/sentry_crash_ipc.c:36](https://github.com/getsentry/sentry-native/blob/master/src/backends/native/sentry_crash_ipc.c#L36)

## Impact
Same-user attacker redirects crash data to attacker-controlled server or executes arbitrary binary.

## Reproduction steps
1. A same-user attacker observes the target process PID via /proc, predicts the shared memory name, and pre-creates `/dev/shm/s-XXXXXXXX` with attacker-controlled content including a malicious DSN pointing to attacker's server and an external_reporter_path pointing to attacker's binary. When the target application initializes Sentry and hits the EEXIST fallback, it opens the attacker's segment. The attacker races to overwrite the configuration after the app writes it but before the daemon reads it, redirecting crash reports (potentially containing sensitive stack data) to the attacker's server.

## Recommended fix
The EEXIST fallback must verify ownership (fstat st_uid) of the existing segment before using it, and the shared memory name should include additional entropy beyond PID and TID.

---
**Severity:** LOW
**Status:** Open
**Category:** Race-condition
**Repository:** getsentry/sentry-native
**Branch:** master

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Fix a potential out-of-bounds read when parsing non-NUL-terminated `sentry-trace` headers. ([#1749](https://github.com/getsentry/sentry-native/pull/1749))
 - Fix division by zero when breadcrumbs are disabled. ([#1767](https://github.com/getsentry/sentry-native/pull/1767))
 - Handle memory allocation failures during JSON serialization to prevent truncated output. ([#1772](https://github.com/getsentry/sentry-native/pull/1772))
+- Native/Linux: prevent IPC shared memory spoofing. ([#1774](https://github.com/getsentry/sentry-native/pull/1774))
 
 ## 0.14.2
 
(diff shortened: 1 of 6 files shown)


jpnurmi:
gets annoyingly noisy :(

PR #1775, opened 2026-06-01
fix(native): daemon constructs breadcrumbs from file

bitsandfoxes:
Follows https://github.com/getsentry/sentry-native/pull/1769

### The Goal

The goal is to have the daemon create non-crashing events that have the same scope as crashing ones. The non-crashing event created by the [app hang detection](https://github.com/getsentry/sentry-native/pull/1742) was missing breadcrumbs.

### Context

The crash capture works by the in-process part of the SDK maintaining a base event that the daemon later can pick up, enrich and send. This base event also contains the flushed scope.

Breadcrumbs live on the scope but do not get flushed when added:
https://github.com/getsentry/sentry-native/blob/259c59a34c085823aa5f9e41ccee3d9d19075ae1/src/sentry_core.c#L936-L941

So the `native_backend_except` running in-process has to flush the scope including the breadcrumbs to have them in the base event for them to be available in the event sent by the daemon.
https://github.com/getsentry/sentry-native/blob/259c59a34c085823aa5f9e41ccee3d9d19075ae1/src/backends/sentry_backend_native.c#L1027-L1031

That also means that the daemon never touched the breadcrumb ringfiles.

### What This Does

Instead of relying on the `native_backend_except` to sync the breadcrumbs to the base event (which won't be happening in case of non-crashing events) we make the daemon pick up the breadcrumbs that are written to file anyway.

#skip-changelog


src/backends/sentry_backend_native.c
```diff
@@ -867,18 +868,20 @@ native_backend_add_breadcrumb(sentry_backend_t *backend,
         return;
     }
 
-    // Serialize to JSON (so it can be deserialized on next start)
```
bitsandfoxes, line 870: That comment is misleading, could not find who or what was reading the JSON on the next start.

src/backends/native/sentry_crash_context.h
```diff
@@ -289,6 +289,7 @@ typedef struct {
     uint64_t shutdown_timeout;
     uint64_t transfer_timeout;
     bool system_crash_reporter_enabled;
+    uint32_t max_breadcrumbs;
```
bitsandfoxes, line 293: Since it's the daemon now constructing the breadcrumbs from files it needs the `max_breadcrumbs`.

src/backends/native/sentry_crash_daemon.c
```diff
@@ -2731,21 +2800,43 @@ write_envelope_with_minidump(const sentry_options_t *options,
...
+                    sentry_value_get_by_key(event, "event_id")));
+                event_json = sentry__value_to_json(event, &event_size);
+                sentry_value_decref(event);
+                if (!event_json) {
+                    // Re-serialization failed (e.g. OOM). Fall back to the raw
+                    // event bytes so the crash report is preserved, losing only
+                    // the merged breadcrumbs rather than the whole event.
+                    event_json = sentry__string_clone_n(base_json, base_size);
+                    event_size = event_json ? base_size : 0;
+                }
```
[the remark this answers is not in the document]
bitsandfoxes: In case things go sideways and we're in fallback/recovery mode. I'd rather lose the breadcrumbs than the entire event.

mujacica:
Let's also write some unit and integration tests to verify this works end-to-end on all 3 platforms

mujacica (approved):

PR #1796, opened 2026-06-05
fix(tus): remove redundant Upload-Metadata header from PATCH request

JoshuaMoelans:
The TUS spec only defines Upload-Metadata on the creation request (POST), not on the upload request (PATCH). Remove the duplicate header from the PATCH path and drop the now-unused attachment_type parameter from `prepare_tus_upload_request`.

follow-up of #1795 

#skip-changelog


tobias-wilfert (approved):
Looks ok to me as a c-novice.

mujacica (approved):

PR #1800, opened 2026-06-10
fix(native): refrain from widening `crashed_tid` to `uint64_t` on macOS

bitsandfoxes:
As far as I can tell this came through this being implemented for Linux first where the thread ID genuinely is a `pid_t` (32-bit). Windows has this as a `DWORD` which is 32-bit too. 
Later on macOS got retrofit, and `thread_identifier_info.thread_id` returning a 64-bit caused the squeeze via `(pid_t)target_tid`.
But since the minidump format in the end has it as 32-bit anyway, it doesn't make sense to neither widen `crashed_tid` neither in the context nor in the comparison.

#skip-changelog


bitsandfoxes:
nevermind

PR #1806, opened 2026-06-16
feat: in proc App Hang capture

bitsandfoxes:
Resolves https://github.com/getsentry/team-gdx/issues/155
Alternative or parallelization of https://github.com/getsentry/sentry-native/pull/1780

### The Goal

The goal is to have a mechanism that allows capture of frozen/stuck main threads. The resulting event should provide the same level of detail and insights as actual errors. This means contexts, tags, breadcrumbs, trace, and stack trace.

### The How

With an in-proc thread stackwalker, we pause (lol) the hung thread, get to the stack, and create a proper event. 
Whatever is considered the main thread periodically calls `heartbeat` on the SDK. This reset a timestamp. The monitor captures the state of the stuck process should the timestamp exceed the configured threshold.

For this the PR adds the following new API:
- `sentry_options_set_app_hang_enabled`:
- `sentry_options_set_app_hang_timeout_ms`
- `sentry_app_hang_heartbeat`

### Limitations & Followups:
- Limited to x64 on Windows. Needs a follow up for ARM64.
- Letting a variable number of threads emit a heartbeat and monitor those

#### macOS
Screenshot 2026-06-16 at 17 51 59

#### Windows
Screenshot 2026-06-16 at 17 55 52


src/sentry_app_hang_latch.h
```diff
@@ -0,0 +1,31 @@
...
+
+uint64_t sentry__app_hang_now_ms(void);
+
+typedef struct {
+    uint64_t target_tid;
+    uint64_t last_heartbeat_ms;
+} sentry_app_hang_latch_t;
+
+uint64_t sentry__app_hang_current_tid(void);
+void sentry__app_hang_latch_read(sentry_app_hang_latch_t *out);
```
limbonaut: Seeing that it's a fairly small 16 byte POD struct, return by value?

include/sentry.h
```diff
@@ -2651,6 +2651,46 @@ SENTRY_EXPERIMENTAL_API void sentry_options_set_enable_metrics(
...
+ *
+ * The first call latches the calling thread as the monitored thread.
+ * Call this regularly from the thread you want watched for hangs. If the
+ * watchdog does not receive a heartbeat within the configured timeout, it
+ * captures an app-hang event.
+ *
+ * This function is a no-op unless app-hang detection is enabled via
+ * `sentry_options_set_enable_app_hang_tracking`.
+ */
+SENTRY_EXPERIMENTAL_API void sentry_app_hang_heartbeat(void);
```
limbonaut, line 2697: Should probably mention that it's designed to work with a single thread being monitored. Otherwise, somebody may use it with several threads.

src/sentry_sync.h
```diff
@@ -431,6 +431,32 @@ sentry__atomic_compare_swap(volatile long *val, long expected, long desired)
...
+ * a full 64-bit atomic (e.g. a monotonic timestamp that must not tear on a
+ * 32-bit platform) use these instead.
+ */
+static inline void
+sentry__atomic_store_u64(uint64_t *val, uint64_t value)
+{
+#ifdef SENTRY_PLATFORM_WINDOWS
+    InterlockedExchange64((volatile LONG64 *)val, (LONG64)value);
+#else
+    __atomic_store_n(val, value, __ATOMIC_SEQ_CST);
```
limbonaut, line 446: AFAIR not lock-free on ARMv7. Is that ok?

limbonaut (commented):
Great stuff! This was a good read - learned a lot of new things about sentry-native. Surprisingly concise PR for what it does. Got my brain cooked trying to understand the nuances between the various unwinder implementations 😅 You may want to have others take a look too, esp someone who knows unwinders well. 

Let's try integrating this into Godot. In the context of the Godot SDK, Godot Engine doesn't provide an API for inspecting script stack traces from other threads, so this would only be of use for issues triggered in the native code (engine or C++ extension). Now I'm wondering what would be required to make script traces possible for App Hangs. And same for CoreCLR .NET.

tustanivsky (approved):
LGTM - tested this in Unreal on Windows, Linux and macOS (https://github.com/getsentry/sentry-unreal/pull/1427) and it works as expected. I also confirmed that in-process hang tracking allows us to bypass the IPC limitations in sandboxed Mac applications.

src/sentry_app_hang_latch.h
bitsandfoxes: Good call!

src/sentry_sync.h
bitsandfoxes: Yes, true, but fine. On ARMv7 this ends up being an implicit lock. But it still guarantees a tear-free 64-bit timestamp, and the lock is off any signal path so it can't deadlock the hot path.

src/sentry_thread_stackwalk_posix.c
```diff
@@ -0,0 +1,248 @@
...
+    }
+    g_park_uctx = NULL;
+    // watchdog wrote g_ips/g_count before releasing us
+    n = (size_t)__atomic_load_n(&g_count, __ATOMIC_RELAXED);
+#    else
+    (void)ucontext;
+#    endif
+    __atomic_store_n(&g_count, (sig_atomic_t)n, __ATOMIC_RELAXED);
+    __atomic_store_n(&g_active, 0, __ATOMIC_RELEASE);
+    sem_post(&g_done);
```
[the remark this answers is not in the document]
bitsandfoxes: Every path of the watchdog returns with a `g_unwind_done`. And parking it it the whole point.

JoshuaMoelans (approved):
Nice work 🚀 LGTM

Issue #1803, opened 2026-06-16
Envelope writer reports success on partial/failed disk write, producing truncated envelopes


Kobby-Bawuah:
**Description**

When persisting an envelope to disk, the SDK treats the write as successful as long as *any* bytes were written, rather than verifying the full envelope was written. If a write fails partway (for example, the disk is full / `ENOSPC`), the per-item write results are not checked, so a truncated envelope is left on disk and reported as a successful write — no error is returned and nothing is logged.

The resulting envelope is malformed (e.g. a blank/missing header line and a truncated minidump attachment) and is later rejected by ingest. Because the failure is swallowed, there is no signal pointing at the real cause (the failed write).

The affected logic is in `envelope_write_to_path` in `src/sentry_envelope.c`, which determines success from the total byte count written rather than from whether each write completed. The underlying file-write helpers also do not log when a write fails or is truncated.

This is platform-independent (the serializer is shared code), though it was observed on a Windows-family target.

**When does the problem happen**

- [ ] During build
- [x] During run-time
- [x] When capturing a hard crash

**Environment**

- OS: Windows-family target (platform-independent code path)
- Compiler: MSVC
- CMake version and config: `SENTRY_BACKEND=breakpad`

**Steps To Reproduce**

1. Configure the SDK to persist crash envelopes to disk.
2. Induce a write failure while an envelope is being written (e.g. fill the target volume so writes fail with `ENOSPC`).
3. Observe that a truncated/invalid envelope is written to disk, the write is reported as successful, and no error or warning is logged.

**Log output**

No error or warning is emitted on the partial-write path; the write is reported as successful. (A specific customer case, including the captured malformed envelope, is tracked in the attached Linear ticket.)

PR #1807, opened 2026-06-16
ref(envelope): track streaming write failures in shared writer

supervacuus:
This PR builds on https://github.com/getsentry/sentry-native/pull/1804, but actually tackles the underlying layer refactorings.

While #1804 is a fine immediate fix for #1803, it actually makes it much clearer that the layers below have outgrown their clothing:

* A client has to check every return value to ensure the correctness of the entire write process. 
* It relies on the `failed` flag, which was only recently introduced, but which is accidentally being reset by the `reset()` of the JSON writer (which should only reset JSON state and not sticky failure flags)
* It does not check `fflush()`, `fclose()`, or `close()` errors (meaning partial writes could still exist, when they appear during the flush of previously buffered writes)

These are all easy to make mistakes in the current setup, and it is partially the result of a conflated layering below.

This PR adds

* byte-writer layer that can be used by both the raw envelope payload writes and the JSON serialization
  * the byte-writer layer relieves the JSON serialization of handling the lower-level writers
  * it offers a contract that failure is sticky: once a failure is raised, any following operation immediately returns, meaning you only need to check for return values for optimization, not for correctness
  * jsonwriter stops caring about where it writes to
* a failure handling on each layer with different responsibilities at each layer:
  * `sentry_filewriter_t` tracks platform/file-space write/close/flush failures and tracks the "closed" (and is only used by any consuming file-backed writers)
  * `sentry_writer_t` as a generic byte-sink, offers sticky failures regardless of backend and covers the stringbuilder, file path, and the existing file writer.
  * `sentry_jsonwriter_t` only tracks JSON-level failure and delegates any other failure tracking to `sentry_writer_t`. `reset()` no longer clears the failure.

I missed the opportunity to split `sentry_jsonwriter_t` into a separate byte-sink abstraction back when I introduced streaming envelope writing, which made failure tracking unnecessarily complicated and unprincipled, and in turn made it way too easy to miss important checks (in the client code and on each layer involved).

With this PR, we have one authoritative check at the end that is not simply based on checking the written bytes, but pervasive failure tracking on each layer. We can still have earlier checks to limit unnecessary processing in the failure case, but we no longer need to check every write solely for correctness.


src/sentry_envelope.c
```diff
@@ -916,48 +916,72 @@ typedef bool (*envelope_item_writer_fn)(
...
             }
-            const char newline = '\n';
-            sentry__filewriter_write(fw, &newline, sizeof(char));
 
+            sentry__writer_write_char(output, '\n');
             sentry__jsonwriter_write_value(jw, item->headers);
             sentry__jsonwriter_reset(jw);
+            if (sentry__jsonwriter_has_failed(jw)) {
+                goto done;
+            }
```
supervacuus, line 958: The byte writer underneath `jw` has sticky failure semantics. That means this streaming code does not need to check every newline, header, and payload write. Any partial write or late close/flush error is remembered by `output` and observed once at the end.

The early exits are only performance shortcuts to avoid serializing or attempting large payload writes after the sink has already failed; the final checks in `done` remain authoritative, and the code would still be correct without them.

src/sentry_json.c
```diff
@@ -23,195 +23,113 @@
...
 sentry__jsonwriter_reset(sentry_jsonwriter_t *jw)
 {
+    if (!jw) {
+        return;
+    }
+
     jw->want_comma = 0;
     jw->depth = 0;
-    jw->last_was_key = 0;
-    jw->failed = false;
```
supervacuus, line 210: We shouldn't be clearing `failed` here. `reset()` is used between envelope parts _to restart JSON grammar state_, but output errors must remain sticky so a caller can safely check once after the whole stream has been produced.

supervacuus:
Btw, since I simply hijacked your changelog, @JoshuaMoelans: neither your PR nor this one addresses the issue of leaving partial envelopes on disk. This should likely be resolved in a different PR, because there should be consistent handling across the SDK.

But let me know if we should add this here. We can simply delete a failed envelope (a 3-line change) as a best-effort approach, but given the classic failure cases this involves, a proper solution should be atomic (write to a temp and rename on success).

src/sentry_envelope.c
```diff
@@ -916,64 +916,70 @@ typedef bool (*envelope_item_writer_fn)(
...
+    if (!sentry__writer_close(output)) {
+        failed = 1;
+    }
+    if (jw && sentry__jsonwriter_has_failed(jw)) {
+        failed = 1;
+    }
+    if (sentry__writer_has_failed(output)) {
+        failed = 1;
+    }
+
```
JoshuaMoelans, line 974: (nit) since we already check in sentry_writer_close() whether `writer->failed` is true and whether the writer is NULL, checking `has_failed` here again seems redundant.

```suggestion
    if (!sentry__writer_close(output)) {
        failed = 1;
    }
    if (jw && sentry__jsonwriter_has_failed(jw)) {
        failed = 1;
    }
```

src/sentry_writer.c
```diff
@@ -0,0 +1,299 @@
...
+
+static char *
+file_writer_into_string(void *state, bool UNUSED(failed), size_t *len_out)
+{
+    if (len_out) {
+        *len_out = 0;
+    }
+    file_writer_free(state);
+    return NULL;
+}
```
JoshuaMoelans, line 127: q: the old `into_string_file` had an `UNREACHABLE` block, should we add the same here (or at express a warning?)

JoshuaMoelans (approved):
LGTM, just one small nit and open question (neither are blocking)

src/sentry_envelope.c
supervacuus: Agreed, that's no longer necessary.

src/sentry_writer.c
supervacuus: The old file-backed JSON writer had no meaningful `into_string` path, so `UNREACHABLE` made sense there. The new, more generic `sentry_writer_t` explicitly documents that `into_string` returns `NULL` for non-string-backed sinks:

https://github.com/getsentry/sentry-native/blob/c923da1e3cb665d7e7fdfdc8cff153aaec14da81/src/sentry_writer.h#L47-L51

So this is now a defined misuse/unsupported conversion, not an impossible code path. I'll add a comment in the concrete implementation nonetheless.

PR #1814, opened 2026-06-23
fix: app hang option name convention

bitsandfoxes:
Aligned name with the rest of the `_timeout` options.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 **Features**:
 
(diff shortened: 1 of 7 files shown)


JoshuaMoelans (changes requested):
missed a rename in
https://github.com/getsentry/sentry-native/blob/fix/app-hang-timeout-naming/tests/unit/test_app_hang.c#L112
https://github.com/getsentry/sentry-native/blob/fix/app-hang-timeout-naming/tests/unit/test_app_hang.c#L176

JoshuaMoelans (approved):

PR #1816, opened 2026-06-23
fix: disable app hang capture during crash capture

bitsandfoxes:
To avoid (best case) duplicate capture.

#skip-changelog


src/sentry_core.c
```diff
@@ -830,6 +831,9 @@ prepare_user_feedback(const sentry_options_t *options,
 void
 sentry_handle_exception(const sentry_ucontext_t *uctx)
 {
+    // Disarm the app-hang monitor to avoid dupliace capture.
```
JoshuaMoelans: (nit:typo)
```suggestion
    // Disarm the app-hang monitor to avoid duplicate capture.
```

src/sentry_app_hang_latch.c
```diff
@@ -61,6 +61,19 @@ sentry__app_hang_set_active(bool active)
...
+    return sentry__atomic_fetch(&g_app_hang_active) != 0;
+}
+
+void
+sentry__app_hang_disarm(void)
+{
+    // Called from the crash handler, so this must stay async-signal-safe.
+    sentry__atomic_store(&g_app_hang_active, 0);
+}
+
```
JoshuaMoelans: q: why do we need `sentry__app_hang_disarm(void)` if we could just call the existing `sentry__app_hang_set_active(false)`?

bitsandfoxes: good catch. initially, i was doing more in there than just toggling the flag but now it's truly redundant. removed.

JoshuaMoelans (approved):
nice, LGTM!

PR #1825, opened 2026-06-26
feat(consoles): allow Xbox to override sentry_init for custom device context

JoshuaMoelans:
Add SENTRY_PLATFORM_XBOX to the preprocessor guard that exposes sentry__native_init, enabling the Xbox console SDK to provide its own sentry_init with device model detection.

Related to https://github.com/getsentry/sentry-xbox/issues/149 and https://github.com/getsentry/sentry-xbox/issues/151

#skip-changelog

### DIFF
--- a/src/sentry_core.c
+++ b/src/sentry_core.c
@@ -100,7 +100,8 @@ generate_propagation_context(sentry_value_t propagation_context)
         sentry_value_get_by_key(propagation_context, "trace"));
 }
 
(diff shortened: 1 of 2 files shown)


JoshuaMoelans:
(xbox on main will fail because the xbox side of this lives in https://github.com/getsentry/sentry-xbox/pull/151 )

bitsandfoxes (commented):
:shipit:

bitsandfoxes (approved):
🚢

PR #1835, opened 2026-07-01
feat: add `sentry_update_context`

JoshuaMoelans:
Adding this for downstream SDKs, so they can avoid fully overwriting pre-set context from our native SDK init. 

Popped up as part of https://github.com/getsentry/sentry-xbox/pull/151 where the unreal integration test calls [AddDeviceContext](https://github.com/getsentry/sentry-unreal/blob/main/plugin-dev/Source/Sentry/Private/SentrySubsystem.cpp#L975) which does a plain `set_by_key`, making us lose `device.X` context set by the console SDK.

Now with `sentry_update_context` we only overwrite existing keys' values, and also add the new keys from the passed-in object to the context. For example

```c
ctx={a:0, b:1, c:2}
sentry_update_context(ctx, {a:-1, d:3})
// ctx becomes {a:-1, b:1, c:2, d:3}
```


src/sentry_core.c
```diff
@@ -1075,6 +1075,35 @@ sentry_set_context_n(const char *key, size_t key_len, sentry_value_t value)
     }
 }
 
+void
+sentry_merge_context(const char *key, sentry_value_t value)
+{
+    SENTRY_WITH_SCOPE_MUT (scope) {
```
jpnurmi, line 1081: did you intentionally choose not to call `sentry_merge_context_n`? it's so little repeat that either way is fine, imo. just curious :)

include/sentry.h
```diff
@@ -2300,6 +2300,14 @@ SENTRY_API void sentry_scope_set_context(
     sentry_scope_t *scope, const char *key, sentry_value_t value);
 SENTRY_API void sentry_scope_set_context_n(sentry_scope_t *scope,
     const char *key, size_t key_len, sentry_value_t value);
+/**
+ * Merges a context object with the existing context object.
+ * Existing keys within the context object will not be overwritten.
```
jpnurmi: Hmm, shouldn't downstream values take precedence? For example, passing "Xbox" should override "PC", right? 🤔

src/sentry_core.c
JoshuaMoelans: good catch, I remembered there was some pattern with the `_n` variants but from looking at the `set_context(_n)` above I didn't figure out what it was exactly 😅  fixed in https://github.com/getsentry/sentry-native/pull/1835/commits/4d36ebabb3d71de88f524eb2805722ef70325a0e

include/sentry.h
JoshuaMoelans: In general there are definitely values that should take precedence (e.g., on [detecting steamOS](https://github.com/getsentry/sentry-unreal/blob/main/plugin-dev/Source/Sentry/Private/Linux/LinuxSentrySubsystem.cpp#L38) for example), but these could then still use the normal `SetContext`. That would lose the 'merge' for other OS context though, so maybe alternatively we could add a flag to our merge called "overwrite_dst" that says whether to keep values of existing keys, or overwrite them 🤔

jpnurmi: What if we made the OS/Device context extensible by allowing downstream SDKs to provide the initial device context as a starting point, and then internally merge sentry-native's OS context into that?

jpnurmi: Never mind, I didn't quite understand the desired merge flow when dropping the comment. I thought it was all about providing the defaults, but they are coming from multiple places and need to be merged at runtime.

jpnurmi (approved):
LGTM 👍

PR #1843, opened 2026-07-06
fix(scopes): don't drop propagation context when event already has contexts

limbonaut:
When a scope is applied to an event, the trace context created by `sentry_set_trace()` was only added if the event had no existing contexts. This creates a problem when a local scope is applied before the global scope, because it adds contexts to event and the trace infomation is skipped so the event loses its trace id. This PR fixes this by ensuring the propagation context is applied whenever an event doesn't already have a trace or active span. Also, adding tests to verify correct behavior.


src/sentry_scope.c
```diff
@@ -468,11 +468,13 @@ sentry__scope_apply_to_event(const sentry_scope_t *scope,
...
+        sentry__value_merge_objects(contexts, scope->propagation_context);
+    }
     if (sentry_value_is_null(event_contexts)) {
-        // only merge in propagation context if there is no scoped span
-        if (!is_transaction && sentry_value_is_null(scope_trace)) {
-            sentry__value_merge_objects(contexts, scope->propagation_context);
-        }
         PLACE_VALUE("contexts", contexts);
     } else {
         sentry__value_merge_objects(event_contexts, contexts);
```
limbonaut, line 480: Notice that the else branch skips adding propagation context altogether. So, just having pre-existing contexts omits adding trace completely.

JoshuaMoelans (approved):

jpnurmi (approved):

Issue #1838, opened 2026-07-03
Make scopes more aligned to Sentry specs to ease support in the downstream SDKs


limbonaut:
## Context

I'm working on implementing [current scopes](https://develop.sentry.dev/sdk/foundations/state-management/scopes/#current-scope) in the Godot SDK, which requires scope lifetime to be fixed to specific functions in GDScript or active span. The scopes are a prerequisite to implementing performance API in Godot.

Take a look at API I'm aiming to implement across all supported platforms (it's GDScript code):

```gdscript
SentrySDK.with_scope(func(scope):
  scope.set_tag(...)
  do_some_work()  # can cause errors that should be associated with this scope
  do_more_work()  # can cause more errors (Godot errors are not terminating)
)
```

It's modeled after [three-scope model](https://develop.sentry.dev/sdk/foundations/state-management/scopes/#three-scope-types) (global scope + current scope minus isolation scope because underlying SDKs  don't support isolation scopes yet). So it's a simplified "two-scope" model: global + current scopes.

The performance API will need this too:
```gdscript
SentrySDK.with_span("game.some_work", func(span):  
  # with_span() forks current scope for the duration of the active span (i.e. this function)
  SentrySDK.get_current_scope().set_tag(...)
  do_some_work()
)
```

The recent Scope API spec mandates: 
> When a new span is started, the current scope of the parent span MUST be forked (duplicated), transferring all data from the parent scope to the new scope.

The current scope stack will be managed internally in the Godot SDK.

## Proposal(s)

A) Allow for scope lifetime to be owned by users. Currently, `capture_event_with_scope` takes ownership of "local" scope, making it a one-time construct for our use-case. This multiplies the work needed to maintain current scopes at runtime in the Godot SDK (Godot errors are non-terminal - application continues running).
  
This could be implemented as a separate constructor: `sentry_scope_new` (creates user-owned scope) and `sentry_scope_free`, and it tags scope as "user-owned".

B) Add `sentry_scope_clone()` so the parent span's current scope cloning could be streamlined in the downstream SDKs. This should perform a deep copy (unless CoW). This operation exists in other SDKs such as cocoa, android, js, etc.

C) Implement `sentry_scope_capture_log/metric` so integrations can associate logs with specific scope-bound active span (considering that the scope model and stack would be owned by Godot SDK) and apply scope attributes. 

D) Add `sentry_scope_set_span` and `sentry_scope_set_transaction` - so the events can be associated with the current scope's active span/transaction.

[...]

PR #1855, opened 2026-07-10
feat(scopes): Add user-owned scope lifecycle

limbonaut:
Adds a user-owned scope lifecycle so downstream SDKs can hold and reuse a scope across many captures, which is needed to implement current scopes and the performance API in the Godot SDK. See rationale in #1838.

- Refs #1838
- Needed for https://github.com/getsentry/sentry-godot/pull/807


src/sentry_scope.c
```diff
@@ -192,6 +193,69 @@ sentry__scope_free(sentry_scope_t *scope)
...
+    if (scope) {
+        scope->user_owned = true;
+    }
+    return scope;
+}
+
+void
+sentry_scope_free(sentry_scope_t *scope)
+{
+    sentry__scope_free(scope);
```
jpnurmi: is `sentry__scope_free` used in many places? would it get noisy if we renamed and promoted the existing function instead?

src/sentry_scope.c
```diff
@@ -192,6 +193,69 @@ sentry__scope_free(sentry_scope_t *scope)
...
+{
+    if (!scope) {
+        return NULL;
+    }
+
+    sentry_scope_t *clone = SENTRY_MAKE(sentry_scope_t);
+    if (!clone) {
+        return NULL;
+    }
+    memset(clone, 0, sizeof(sentry_scope_t));
```
jpnurmi: `SENTRY_MAKE` was changed to use `calloc` a while back, so these `memset(0)`'s are no longer needed 🙂

src/sentry_scope.c
```diff
@@ -192,6 +193,69 @@ sentry__scope_free(sentry_scope_t *scope)
     sentry_free(scope);
 }
 
+sentry_scope_t *
+sentry_scope_new(void)
+{
+    sentry_scope_t *scope = sentry_local_scope_new();
```
jpnurmi: a minor sore in the eye 🙂 i'm tempted to propose flipping the semantics of the flag around (e.g. `is_local`) and make `sentry_local_scope_new` reuse `sentry_scope_new` instead. would also avoid setting it in `sentry_scope_clone` :)

src/sentry_scope.c
limbonaut: Around 15-20 places, not overly much. Mostly tests.

jpnurmi: sorry for generating more work for you, but then i'd go for it to avoid leaving an unnecessary indirection behind 😎

src/sentry_scope.c
limbonaut: Good to know, thanks!

src/sentry_scope.c
limbonaut: How about `one_shot`?

jpnurmi (approved):

jpnurmi:
Perfect 👌

mujacica (approved):

PR #1858, opened 2026-07-13
feat(tracing): Add public discard APIs for spans and transactions

jpnurmi:
Downstream SDKs can own native span and transaction handles with lifetimes that are independent from explicit finish calls. When such wrappers are destroyed, invalidated, or reject a span in a callback, they need to release the native handle without recording the span or sending the transaction.

The existing public finish APIs cannot model that case because they take ownership while also finishing the item. The only available non-sending release path was the private tracing decref functions, which forced downstream SDKs to include private sentry-native headers or leak unfinished handles.

Expose public experimental discard functions for transactions and spans. They take ownership, clear matching active scope references, and release the handle without sending or appending span data. Cover both paths with tracing unit tests so the discard behavior stays distinct from finish semantics.


limbonaut:
Mm 🤔 Nice, this should be useful for spans/performance work in Godot SDK 👍

JoshuaMoelans (approved):

PR #1861, opened 2026-07-13
feat(scope): Scoped logs and metrics capture

limbonaut:
This PR makes scope attributes a public, working part of the API. It promotes the previously-internal `sentry_scope_set_attribute/_n` and `sentry_scope_remove_attribute/_n`  to public, and adds `sentry_scope_capture_log` and `sentry_scope_capture_metric` so a log or metric can be captured against a specific scope and pick up that scope's attributes and trace. Both trace and attributes now flow onto logs and metrics with spec-correct precedence, resolving most-specific-first: per-call attributes, then the passed scope, then the global scope.

- Refs #1838
- Needed for https://github.com/getsentry/sentry-godot/pull/807

Needed for current scopes and performance API in the Godot SDK. See rationale in https://github.com/getsentry/sentry-native/issues/1838.

I plan to add `sentry_scope_set_span/transaction` and also `sentry_scope_clear` as follow-ups.


src/sentry_scope.h
```diff
@@ -143,7 +133,10 @@ void sentry__scope_update_dsc(
 void sentry__scope_freeze_dsc(sentry_scope_t *scope, sentry_value_t incoming);
 
 /**
- * Adds scoped attributes to the telemetry attributes object.
+ * Adds scoped attributes to the telemetry attributes object, and the scope's
+ * trace to the telemetry item itself. Existing attributes are kept, so this can
+ * be called for each scope in a precedence chain, most specific first (first
+ * write wins).
  */
 void sentry__scope_apply_attributes(const sentry_scope_t *scope,
```
limbonaut: IMO needs a better name, since it's not just the attributes.

limbonaut:
To check how it would be used, you can see my WIP implementation over here (see `NativeSDK::metrics_add_X` and `project/test/isolated/` tests):
- https://github.com/getsentry/sentry-godot/pull/807

src/sentry_scope.c
```diff
@@ -823,40 +833,41 @@ sentry__scope_apply_attributes(const sentry_scope_t *scope,
 {
     sentry__value_merge_objects(attributes, scope->attributes);
 
+    // a span on the scope MUST take precedence over the propagation context
```
jpnurmi, line 836: the rest seems mostly mechanical rewiring and making things public, but here we have some scary-looking (behavioral?) changes. are they something that could be taken separately, to avoid hiding them in the larger mechanical change? or do these only make sense together with the rest?

limbonaut: Oh yeah, it's a bug in the current implementation that loses scope trace on scope "re-application" (we apply current scope first, then the global scope, and the global overwrites the trace). But it only surfaces if you add scoped logs and metrics (we didn't have scope chain before this PR). So in practice, it's not a bug until this PR is introduced.

limbonaut: For more context, the function used to be called once, with the global scope. But now it's called with a sequence of scopes (local then global). The "scary" change could be summarized to:
> if telemetry already has trace pre-populated - dont overwrite it (i.e. first writer wins).

But because the code was structured the way it was, I had to lightly restructure (to avoid creating trace object). I would completely rewrite this function, if I'm being honest - I chose to introduce minimal changes instead.

This only happens if a scope was applied to telementry previously (local scope, then global). Which in current `master` never happens.

src/sentry_scope.h
jpnurmi: Would `scope_apply_to_telemetry` sound better alongside `scope_apply_to_event`? 🤔

limbonaut: Applied rename in f9cb8a1, plus the parent function.

src/sentry_scope.c
```diff
@@ -818,45 +828,46 @@ sentry_scope_attach_bytesw_n(sentry_scope_t *scope, const char *buf,
...
     sentry_value_t trace_id = sentry_value_get_by_key(
         sentry_value_get_by_key(scope->propagation_context, "trace"),
         "trace_id");
-    sentry_value_incref(trace_id);
-    sentry_value_set_by_key(telemetry, "trace_id", trace_id);
 
     sentry_value_t parent_span_id = sentry_value_new_object();
-    sentry_value_t scoped_span_trace_id = sentry_value_new_null();
```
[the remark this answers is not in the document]
limbonaut: Hmm, this looks like a pre-existing problem. If I understand it correctly, on `master` if a global scope would contain an attribute with the same key and with a unit, a unit would be written over per-call attribute if it didn't have a unit (deep merge = pollution here).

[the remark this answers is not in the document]
limbonaut: I'll do a fix in a separate PR tomorrow. Affects `master` too.

src/sentry_scope.c
jpnurmi: Makes sense, thanks for the clarification. :pray:

src/sentry_scope.c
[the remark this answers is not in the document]
limbonaut: - PR https://github.com/getsentry/sentry-native/pull/1879

jpnurmi (approved):

PR #1875, opened 2026-07-16
feat(session-replay): embed crash event breadcrumbs into replay recordings

tustanivsky:
This PR adds breadcrumbs to session replay recordings: breadcrumbs carried by the crash event that fall inside the replay window are converted into rrweb `breadcrumb` custom events and embedded into the `replay_recording` payload, matching how the Cocoa and Android SDKs surface breadcrumbs on the replay timeline.

### Key changes

- New internal `sentry__value_from_msgpack_stream` decodes append-only msgpack streams (like the breadcrumb ring files) into a list even when the buffer holds a single value; the crash daemon's ring-file reader and `report_to_envelope` now use it. `sentry__value_from_msgpack` rejects buffers holding concatenated values instead of implicitly wrapping them in a list, so each call site declares its intent. This also fixes `report_to_envelope` dropping breadcrumbs when a ring file holds exactly one entry.
- `sentry__session_replay_flush_pending` embeds the crash event's in-window breadcrumbs into the rrweb recording as type-5 custom events with `tag: "breadcrumb"` (outer timestamps in ms, payload timestamps in seconds, chronological order). The breakpad and inproc backends pick this up without further changes since their crash event already carries the scope breadcrumbs.
- Crashpad: the first-chance handler merges the `__sentry-breadcrumb1/2` ring files onto the in-memory crash event before flushing the staged replay; the on-disk `__sentry-event` and the regular crash delivery path are unaffected.
- Native crash daemon: applies its existing ring-file merge to the crash event copy used for the replay flush (the `__sentry-event` snapshot is scope-applied without breadcrumbs).

### Testing

- New `value_from_msgpack_stream` unit test.
- The shared replay-envelope assertion now verifies that the example's breadcrumbs are embedded in the recording with the expected payload fields, timestamp units, replay-window bounds, and ordering. This extends the existing replay integration tests covering the crashpad, breakpad, inproc, and native daemon backends.
- Verified end-to-end via sentry-unreal on Windows: manually added breadcrumbs show up on the replay timeline.


src/sentry_value.h
```diff
@@ -131,13 +131,29 @@ void sentry__value_add_attribute(sentry_value_t attributes,
 /**
  * Deserialize a sentry value from msgpack.
  *
- * If the buffer contains multiple sequential msgpack values (as in flat buffers
- * like breadcrumb files), they are automatically wrapped in a list.
+ * A single msgpack value spanning the whole buffer decodes to that value. If
+ * the buffer contains multiple sequential msgpack values (as in flat buffers
+ * like breadcrumb files), they are wrapped in a list; use
+ * `sentry__value_from_msgpack_stream` when a list is expected regardless of
+ * the number of values.
```
jpnurmi: the implicit list conversion was clearly not so well thought out. what if we would simply disallow multiple values in `sentry__value_from_msgpack` now that we have `sentry__value_from_msgpack_stream`?

mujacica (approved):

src/sentry_value.h
tustanivsky: Good call - `sentry__value_from_msgpack` now requires a single value spanning the whole buffer and returns null if there's trailing data (https://github.com/getsentry/sentry-native/pull/1875/commits/41e48b5d7557da6b64017737eed87bbeac629e87). Also, the crashpad breadcrumb ring-file reads moved to `sentry__value_from_msgpack_stream` which fixes crash events losing breadcrumbs when a ring file held exactly one entry.

src/backends/sentry_backend_crashpad.cpp
```diff
@@ -446,6 +424,22 @@ crashpad_handler(int signum, siginfo_t *info, ucontext_t *user_context)
...
+                    = read_msgpack_stream_file(state->breadcrumb1_path);
+                sentry_value_t b2
+                    = read_msgpack_stream_file(state->breadcrumb2_path);
+                sentry_value_t merged = sentry__value_merge_breadcrumbs(
+                    b1, b2, options->max_breadcrumbs);
+                if (!sentry_value_is_null(merged)) {
+                    sentry_value_set_by_key(crash_event, "breadcrumbs", merged);
+                }
+                sentry_value_decref(b1);
+                sentry_value_decref(b2);
```
jpnurmi: hmm, can't we get the list of breadcrumbs straight from the scope instead of reading them back from files? 🤔

PR #1880, opened 2026-07-17
fix: prefix vendored mpack symbols

### LINKED ISSUE #1878: Vendored mpack exports unprefixed extern "C" symbols — silent symbol collision with any other mpack in the process causes stack corruption / crash at sentry_init (Unreal monolithic builds)
NOTE: This report has been drafted using a LLM, as a summary of 5 days of debugging efforts.  I've pretty thoroughly reviewed it, but the basic summary of it is:

Please namespace your external libraries.

# Environment

- sentry-native 0.15.3 (via sentry-unreal 1.16.0), static sentry.lib
- Windows x64, MSVC, Unreal Engine 5.7 monolithic game target
- Both backends affected (crashpad and native — shared serialization path)

# Summary

sentry-native's vendored mpack (vendor/mpack.c/h) compiles as C and exports the full mpack API as unprefixed global symbols (mpack_writer_init, mpack_writer_init_growable, …). If the host binary statically links any other copy of mpack — common in game engines, where multiple plugins vendor serialization libs; mpack's own API is extern "C" even when built as C++ — all copies export identical symbol names. The linker silently resolves sentry's mpack calls into the foreign copy. There is no multiply-defined-symbol error (lazy lib resolution), so the hijack is invisible at build time.

If the two copies differ in version/config, mpack_writer_t's layout differs. In our case the foreign copy was mpack with a date of 2021 (compared to yours with a 2018 date) with MPACK_BUILDER=1 (the default there), whose mpack_writer_t embeds mpack_builder_t. sentry's sentry_value_to_msgpack stack-allocates its (smaller) mpack_writer_t, then calls mpack_writer_init_growable, which lands in the foreign copy. Its init executes writer->builder.current_page = NULL; — an offset past the end of sentry's struct — which happens to zero the caller's saved return address. The function completes normally and rets to address 0.

# Observed failure (why this is brutal to diagnose)

Access violation executing location 0x0000000000000000, no walkable callstack, RAX = the msgpack buffer (return value in flight).
Crashpad backend: crashes inside sentry_init (first crashpad_backend_flush_scope → scope serialization). Native backend: crashes at the first breadcrumb flush. Same root cause, same function.
The apparent crash point wanders between runs (interaction with the first-chance handler's sentry__logger_disable() truncating logs at a timing-dependent point), which mimics heap/stack corruption and defeats ordinary debugging. We root-caused it only via a hardware data breakpoint on the return-address slot, which halted inside the other library's mpack_writer_init.
Modular (editor) builds are unaffected — DLL symbol isolation — so the SDK appears fine in development and dies only in packaged/monolithic builds and on CI.

# Repro sketch

Static-link sentry-native (any backend) into a binary that also links another mpack (≥1.1, default config, extern "C" API — stock mpack).
Ensure the other mpack's objects precede sentry.lib in link order (direct object inputs vs searched lib — typical in UE).
Call sentry_init → ret-to-0 during scope flush / first breadcrumb.
Suggested fix
Prefix the vendored mpack's symbols so they're private to sentry-native — e.g., a rename header (#define mpack_writer_init sentry__mpack_writer_init …) included ahead of vendor/mpack.h/c, or upstream mpack gaining a MPACK_SYMBOL_PREFIX. Prefixing vendored dependencies is standard hardening for SDKs that ship static libs into host applications they don't control; sentry-native is exactly such an SDK, and game engines are exactly such hosts.

# Workaround we applied

We owned the other mpack copy, so we wrapped it in a C++ namespace (dropping its extern "C"). Users who don't control the colliding copy have no good options short of rebuilding sentry-native with renamed symbols.

jpnurmi:
Similar to https://github.com/getsentry/crashpad/pull/143, but in C, we don't have namespaces, so the symbol rename comes with some less-nice define trickery... Includes a test to catch unprefixed mpack symbols:
```sh
$ tests/test_build_static.py::test_static_mpack
...
FAILED tests/test_build_static.py::test_static_mpack - AssertionError: libsentry.a contains 322 unprefixed MPack symbols: mpack_discard, mpack_done_array, mpack_done_bin, mpack_done_map, mpack_done_str, mpack_done_type, mpack_error_to_string, mpack_expect_array, mpack_expect_array_alloc_impl, mpack_expect_array_match
```

Close: #1878


src/sentry_mpack.h
jpnurmi, line 1: this is quite ugly but i still find it better than the alternative of prefixing all symbols in the vendored mpack sources. we should upgrade to the latest mpack, and then it gets painful if ours is heavily modified...

> [!NOTE]
> Sentry Native vendors **MPack 1.0**, released upstream on **October 20, 2018**—about **7 years and 9 months old** today: https://github.com/ludocode/mpack/commit/e150cb7455cefdab0df6359414e9e4c792723d73
>
> It was imported into Sentry Native on **February 24, 2020**, with the Crashpad integration—about **6 years and 5 months ago**: https://github.com/getsentry/sentry-native/commit/e4bed02184fa12b2002faf7b129e0912336393b6
>
> It has never been upgraded from 1.0. The vendored files only received two local patches:
> - September 15, 2020: work around missing `_Static_assert` support in MSVC: https://github.com/getsentry/sentry-native/commit/f1437333d4ea1eca2f8a3664d57efff081b4dad6
> - July 23, 2024: use Sentry’s allocator for MPack allocations: https://github.com/getsentry/sentry-native/commit/e58b655b54ef89795c838507872cbb92c7339085
>
> The latest upstream release remains MPack 1.1.1 from January 2023.

JoshuaMoelans (approved):
nice, lgtm

PR #1881, opened 2026-07-17
feat(scopes): Add sentry_scope_clear()

limbonaut:
Adds `sentry_scope_clear`, which resets the given scope data (user, tags, breadcrumbs, attributes, and so on) to a fresh state while keeping its trace, so a reusable scope can be cleared and used again.
- Refs https://github.com/getsentry/sentry-native/issues/1838


src/sentry_scope.c
```diff
@@ -211,6 +211,33 @@ sentry_local_scope_new(void)
...
 
+void
+sentry_scope_clear(sentry_scope_t *scope)
+{
+    if (!scope) {
+        return;
+    }
+
+    // Keep the propagation and dynamic sampling contexts across clears so
+    // telemetry captured afterwards continues on the same trace.
```
limbonaut, line 222: I aligned this implementation with Cocoa and Android. I didn't find much details on scope.clear() in the specs, so borrowed this part from other SDKs.
> [spec](https://develop.sentry.dev/sdk/foundations/state-management/scopes/#scope-data-methods-1): Reset scope to defaults, keeping event processors.

It makes sense that the scope.clear() doesn't drop trace defaults.

jpnurmi (approved):

src/sentry_scope.c
limbonaut: In the current shape, this is close to no-op, because user-created scopes carry no default trace, and it's picked up from the global scope instead. But this may change if we introduce propagation context mutators.

PR #1882, opened 2026-07-17
ref(scopes): Rename `sentry_capture_event_with_scope` => `sentry_scope_capture_event`

limbonaut:
Renames `sentry_capture_event_with_scope` to `sentry_scope_capture_event` to match the `sentry_scope_capture_x`  family, with the scope as the first argument. The old name stays as a deprecated forwarder that keeps the original (event, scope) order, so existing callers keep working and get a warning pointing them to the new function.


jpnurmi:
@tustanivsky heads-up: a small deprecation/rename coming. results in nice and consistent `sentry_scope_capture_xxx`

limbonaut:
@jpnurmi should I frame this as a breaking change in the changelog?

limbonaut:
~~I added an inline "Breaking" marker in the changelog, since we have a compatibility forwarder.~~ 

~~Reframed as breaking change.~~

Under "Deprecations" section now.

jpnurmi:
I wouldn't necessarily consider a deprecation breaking (even if it technically can break `-Werror` builds) as long as the old API remains functional. For example, it would be hard to deprecate APIs planned for removal in the next major version ahead of time if the deprecation itself needed another major release. :)

jpnurmi (approved):

PR #1883, opened 2026-07-17
feat(internal): thread pool

jpnurmi:
Introduce a generic thread pool that runs tasks in parallel in a predetermined number of pooled threads and invokes completion callbacks in submission order. The primary use case is the telemetry serialization pool coming in #1946, but it could be used for offloading other heavy tasks, too.

See usage in #1946:

- Pool creation:
  https://github.com/getsentry/sentry-native/blob/1134ee21f0dc6befd23a8b47e226e3652e9c85f0/src/sentry_telemetry.c#L27-L28

- Batch submission:
  https://github.com/getsentry/sentry-native/blob/1134ee21f0dc6befd23a8b47e226e3652e9c85f0/src/sentry_batcher.c#L477-L478

#skip-changelog (internal)


src/sentry_telemetry.c
```diff
@@ -0,0 +1,83 @@
...
+
+    SENTRY_DEBUG("shutting down telemetry");
+    sentry__logs_shutdown(timeout);
+    sentry__metrics_shutdown(timeout);
+    sentry__threadpool_shutdown(g_telemetry_pool);
+    sentry__threadpool_free(g_telemetry_pool);
+    g_telemetry_pool = NULL;
+    SENTRY_DEBUG("telemetry shutdown complete");
+
+    sentry__mutex_unlock(&g_telemetry_lock);
```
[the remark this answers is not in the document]
jpnurmi: Not valid for `sentry_flush()`.

`src/sentry_telemetry.c` serializes both paths with `g_telemetry_lock`:

- `sentry__telemetry_force_flush()` holds the lock from `*_force_flush_begin()` through `*_force_flush_wait()`.
- `sentry__telemetry_shutdown()` takes the same lock before swapping batchers out and freeing `g_telemetry_pool`.

So the interleavings are:

- flush wins first: shutdown waits until force-flush is done, then frees the pool.
- shutdown wins first: it swaps logs/metrics batchers to `NULL`, frees the pool, then a later flush sees no batchers.

The reported stale `batcher->threadpool` access through `sentry_flush()` should not happen. No fix needed for this finding.

src/sentry_batcher.c
```diff
@@ -167,27 +178,250 @@ rotate_buffer(sentry_batcher_t *batcher, long old_idx)
...
+        sentry_envelope_free(task->envelope);
+    } else if (!sentry__run_should_skip_upload(task->batcher->run)) {
+        // Normal operation: use transport for HTTP transmission
+        sentry__transport_send_envelope(
+            task->batcher->transport, task->envelope);
+    } else {
+        sentry_envelope_free(task->envelope);
+    }
+    task->envelope = NULL;
+}
```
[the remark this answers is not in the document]
jpnurmi: The batcher can indeed miss a task that is concurrently completing, but crash handling suspends the transport before invoking telemetry crash flush:

- Breakpad: https://github.com/getsentry/sentry-native/blob/ba24ac69032df31f5e570ee149ef1f1d3bb82544/src/backends/sentry_backend_breakpad.cpp#L168-L171
- Crashpad: https://github.com/getsentry/sentry-native/blob/ba24ac69032df31f5e570ee149ef1f1d3bb82544/src/backends/sentry_backend_crashpad.cpp#L478-L481
- Inproc: https://github.com/getsentry/sentry-native/blob/ba24ac69032df31f5e570ee149ef1f1d3bb82544/src/backends/sentry_backend_inproc.c#L1087-L1090
- Native: https://github.com/getsentry/sentry-native/blob/ba24ac69032df31f5e570ee149ef1f1d3bb82544/src/backends/sentry_backend_native.c#L990-L993

Once suspended, any later `sentry__transport_send_envelope()` writes directly to the run directory instead of using the network transport:

https://github.com/getsentry/sentry-native/blob/ba24ac69032df31f5e570ee149ef1f1d3bb82544/src/sentry_transport.c#L92-L100

So even if `batch_task_complete()` races with `batch_task_dump_pending_all()`, the completed task's send path is still crash-routed to disk after transport suspension. If the send had already entered the transport before suspension, the post-send suspended check dumps the transport queue as well.

src/sentry_telemetry.c
```diff
@@ -22,19 +50,32 @@ sentry__telemetry_shutdown(const sentry_options_t *options)
...
 {
+    SENTRY__MUTEX_INIT_DYN_ONCE(g_telemetry_lock);
+    sentry__mutex_lock(&g_telemetry_lock);
+
     uintptr_t ltoken = sentry__logs_force_flush_begin();
     uintptr_t mtoken = sentry__metrics_force_flush_begin();
     sentry__logs_force_flush_wait(ltoken);
     sentry__metrics_force_flush_wait(mtoken);
+
+    sentry__mutex_unlock(&g_telemetry_lock);
```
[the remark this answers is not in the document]
jpnurmi: **Verdict**

No issue as reported.

The thread-pool wait includes `batch_task_complete()`, since `pending` is decremented only after completion returns ([sentry_sync.c:L163-L178](https://github.com/getsentry/sentry-native/blob/ba24ac69032df31f5e570ee149ef1f1d3bb82544/src/sentry_sync.c#L163-L178)). However:

- The built-in HTTP transport callback only enqueues work onto its background worker ([sentry_http_transport.c:L852-L858](https://github.com/getsentry/sentry-native/blob/ba24ac69032df31f5e570ee149ef1f1d3bb82544/src/transports/sentry_http_transport.c#L852-L858)). Network I/O occurs later on that separate worker.
- A custom transport callback can block because it is invoked synchronously ([sentry_transport.c:L81-L102](https://github.com/getsentry/sentry-native/blob/ba24ac69032df31f5e570ee149ef1f1d3bb82544/src/sentry_transport.c#L81-L102)), but that does not create a lock cycle. If the callback returns, both calls progress.
- If the callback never returns, `sentry_close()` would also block without concurrent `sentry_flush()`: batcher shutdown flushes the same pool ([sentry_batcher.c:L643-L670](https://github.com/getsentry/sentry-native/blob/ba24ac69032df31f5e570ee149ef1f1d3bb82544/src/sentry_batcher.c#L643-L670)), and pool shutdown joins its workers ([sentry_sync.c:L343-L361](https://github.com/getsentry/sentry-native/blob/ba24ac69032df31f5e570ee149ef1f1d3bb82544/src/sentry_sync.c#L343-L361)).
- Releasing `g_telemetry_lock` around the wait is unsafe because shutdown can free the shared pool while force-flush still uses it ([sentry_telemetry.c:L47-L78](https://github.com/getsentry/sentry-native/blob/ba24ac69032df31f5e570ee149ef1f1d3bb82544/src/sentry_telemetry.c#L47-L78)).

No fix is warranted for the alleged deadlock. The unbounded telemetry waits are a separate timeout/liveness concern.

tests/unit/test_sync.c
```diff
@@ -581,6 +581,486 @@ SENTRY_TEST(bgworker_delayed_shutdown)
...
+        == 0);
+
+    sentry__threadpool_flush(pool);
+
+    TEST_CHECK_INT_EQUAL(sentry__atomic_fetch(&state.flush_calls), 3);
+    TEST_CHECK_INT_EQUAL(sentry__atomic_fetch(&state.shutdown_calls), 1);
+    TEST_CHECK_INT_EQUAL(sentry__atomic_fetch(&state.free_calls), 1);
+
+    sentry__threadpool_shutdown(pool);
+    sentry__threadpool_free(pool);
```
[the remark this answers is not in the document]
jpnurmi: Not valid for the current code.

`callback_free` runs as a completion callback on a pooled thread. `sentry__threadpool_free` now checks that case before freeing:

```c
if (sentry__atomic_fetch(&pool->running)) {
    sentry__mutex_lock(&pool->lock);
    if (is_pooled_thread(pool)) {
        sentry__mutex_unlock(&pool->lock);
        SENTRY_WARN("cannot free thread pool from a pooled thread");
        return;
    }
    sentry__mutex_unlock(&pool->lock);
}
```

So the callback increments `free_calls`, but it does not deallocate `pool`. The later test cleanup:

```c
sentry__threadpool_shutdown(pool);
sentry__threadpool_free(pool);
```

is necessary to avoid leaking the pool after confirming the guard rejected callback-side free.

I would not remove those lines. The bot is assuming the old behavior, but the test is specifically verifying the new guard behavior.

src/sentry_sync.c
```diff
@@ -107,6 +107,346 @@ sentry__thread_setname(sentry_threadid_t thread_id, const char *thread_name)
...
+} sentry_threadpool_task_t;
+
+struct sentry_threadpool_s {
+    sentry_threadid_t *threads;
+    char *thread_name;
+    size_t thread_count;
+    size_t started_threads;
+    sentry_mutex_t lock;
+    sentry_cond_t work_signal;
+    sentry_cond_t state_signal;
```
limbonaut, line 127: Got a little confused on the purpose of this condition due to its naming. Would `idle_signal` be a better name?

src/sentry_sync.c
```diff
@@ -107,6 +107,346 @@ sentry__thread_setname(sentry_threadid_t thread_id, const char *thread_name)
...
+
+        sentry__mutex_unlock(&pool->lock);
+        if (task->complete_func) {
+            task->complete_func(task->task_data);
+        }
+        threadpool_task_free(task);
+        sentry__mutex_lock(&pool->lock);
+
+        sentry__atomic_fetch_and_add(&pool->pending, -1);
+        sentry__cond_wake(&pool->state_signal);
```
limbonaut: Three places wait on `state_signal` with different predicates (one checks `pending`, the others `running`!) but we only wake one listener at a time. Since this signal is not for the worker threads, and the audience are all external waiters, it seems to me it would be more robust to `wake_all` instead after completing full drain (or even consider moving it into the `pending == 0` block below). One scenario where this would be important is if someone called `sentry_flush()` from two different threads. One of the calls could block until the next batch retires, because we wake one at a time. I'm not sure if it's a realistic scenario, but I also don't see a reason why we shouldn't broadcast here. What do you think?

I see a bot flagged the same issue below, but I'm not sure why it got resolved.

limbonaut (commented):
Epic work!🦸 The chart from the stacked follow-up shows exactly how useful this can be. I really like the idea of a flexible ordered worker queue with pooled threads. Quite a brain-crack to review 😄 I only have one concern, and a naming nit.

src/sentry_sync.c
jpnurmi: it's supposed to signal the `running`, `stopping`, or `pending` "states". if they were representable as a single enum-like value, i'd probably have named it something like `sentry_threadpool_state_t`. :upside_down_face:

src/sentry_sync.c
jpnurmi: I think you're right :pray: https://github.com/getsentry/sentry-native/pull/1883/commits/e7f44ddc32add0f435e5967b285bac39f2954832

Not sure which commit Cursor considered. I like how Sentry edits the review feedback with "Resolved in &lt;commit&gt;"

src/sentry_sync.c
limbonaut: Aha, I guess the name is alright then

limbonaut (approved):

src/sentry_sync.c
```diff
@@ -107,6 +107,346 @@ sentry__thread_setname(sentry_threadid_t thread_id, const char *thread_name)
...
+
+void
+sentry__threadpool_setname(sentry_threadpool_t *pool, const char *thread_name)
+{
+    if (!pool) {
+        return;
+    }
+    sentry_free(pool->thread_name);
+    pool->thread_name = sentry__string_clone(thread_name);
+}
```
JoshuaMoelans, line 273: q: to avoid a potential path to a use-after-free (if one would execute this alongside the thread pool starting function call of `threadpool_thread`), should we lock and check for `pool->running` to ensure this is indeed only executed before it starts?

src/sentry_sync.c
```diff
@@ -107,6 +107,346 @@ sentry__thread_setname(sentry_threadid_t thread_id, const char *thread_name)
...
+    }
+    sentry__mutex_unlock(&pool->lock);
+    return 0;
+}
+
+int
+sentry__threadpool_submit(sentry_threadpool_t *pool,
+    void (*exec_func)(void *task_data), void (*complete_func)(void *task_data),
+    void (*cleanup_func)(void *task_data), void *task_data)
+{
```
JoshuaMoelans, line 326: q: IIUC we can keep submitting unboundedly to the threadpool, should there be some maximum of tasks we keep accepting? (bgworker has the same semantics afaict, but maybe worth considering). Could also be something we do on the calling side (and that we should then document here)

src/sentry_sync.c
```diff
@@ -107,6 +107,346 @@ sentry__thread_setname(sentry_threadid_t thread_id, const char *thread_name)
...
+        while (sentry__atomic_fetch(&pool->running)) {
+            sentry__cond_wait(&pool->state_signal, &pool->lock);
+        }
+        sentry__mutex_unlock(&pool->lock);
+        return;
+    }
+    pool->stopping = true;
+    const size_t started_threads = pool->started_threads;
+    threadpool_wake_all(pool);
+    sentry__cond_wake(&pool->state_signal);
```
JoshuaMoelans: (nit): this might be an unnecessary `cond_wake` call, since we don't touch `pool->running` or `pool->pending` (only `pool->stopping`) and those are the only two that are directly gating progress even after waking on the `state_signal` ([here](https://github.com/getsentry/sentry-native/pull/1883/changes#diff-05d284250cb83958aeae2f79dc691448f3c65d2d85b5d37e28ac31f26796b496R275), [here](https://github.com/getsentry/sentry-native/pull/1883/changes#diff-05d284250cb83958aeae2f79dc691448f3c65d2d85b5d37e28ac31f26796b496R372) and [here](https://github.com/getsentry/sentry-native/pull/1883/changes#diff-05d284250cb83958aeae2f79dc691448f3c65d2d85b5d37e28ac31f26796b496R392))

JoshuaMoelans (approved):
This is a very cool feature, learned a lot reading up on thread-pooling! LGTM, with just a nit and two questions remaining.

src/sentry_sync.c
jpnurmi: good sanity check :+1:  https://github.com/getsentry/sentry-native/pull/1883/commits/96890c2b423c8a3e0db56453f9a01faff1032650

src/sentry_sync.c
jpnurmi: i added `max_pending` in https://github.com/getsentry/sentry-native/pull/1883/commits/9bcbc6c0fc6198e0c6384ce2d8e30d403dff4f44

in the next PR, we can make it respect the hard max limit of 1000 buffered items:
https://develop.sentry.dev/sdk/telemetry/logs/#buffering

src/sentry_sync.c
jpnurmi: thanks, confirmed and removed in https://github.com/getsentry/sentry-native/pull/1883/commits/7f9c7b598282fd1a419a394c34336050181b94e2. also, double-checked that it seems to have been there ever since the initial implementation and was not brought up by any earlier AI review comments in this PR

PR #1906, opened 2026-07-23
fix(consent): honor checks before launching crash reporter

d2anamaria:
* External reporter bypassed the consent checks usually done in capture_envelope
* Check consent early and skip reporter + materialize when blocked
* Fall through to capture_envelope so the crash is cached instead


jpnurmi:
Amazing, thank you for opening my eyes!  🙏 Do you want to fix the other backends too, while at it? They all behave this way because I always thought it was the correct behavior to delegate user consent to the Crash Reporter UI. 🤦‍♂️

jpnurmi (commented):
What if we would put the consent check somewhere early in `sentry__launch_external_crash_reporter`? Would it then work for all backends without necessarily needing any changes at call sites in the crash handlers? :)

d2anamaria:
@jpnurmi From what I'm seeing, the external reporter only gets access to the envelope it’s supposed to send, and that doesn't contain user consent. Would we want to expose that? I’m not sure exactly how much overhead the whole sentry__envelope_materialize process brings in this case, or if other backends have other steps like this one that could be avoided with the early check.

jpnurmi:
> @jpnurmi From what I'm seeing, the external reporter only gets access to the envelope it’s supposed to send, and that doesn't contain user consent.

Since we pass options to `sentry__launch_external_crash_reporter`, I was hoping we could do `sentry__run_should_skip_upload(options->run)` there and return false if consent is revoked. All backends and the daemon should fallback to `sentry__capture_envelope`.

> I’m not sure exactly how much overhead the whole sentry__envelope_materialize process brings in this case, or if other backends have other steps like this one that could be avoided with the early check.

It would be fine to move the `sentry__envelope_materialize` call from the native daemon to `sentry__launch_external_crash_reporter` to keep the code simple. Technically, materialization is only needed for the daemon that reads a "raw" envelope from disk, which is why it was placed where it currently is, but materialization would be a no-op for the other backends that pass a proper, freshly constructed envelope. So it would be fine to call `sentry__envelope_materialize` after checking the consent in `sentry__launch_external_crash_reporter`.

jpnurmi (approved):
Thanks, LGTM 👍

Bonus points if we can have a similar integration test for the other backends to verify the behavior

d2anamaria:
> Thanks, LGTM 👍
> 
> Bonus points if we can have a similar integration test for the other backends to verify the behavior

will do

PR #1911, opened 2026-07-24
feat(native): support custom Windows minidump flags

### LINKED ISSUE #1335: Allow configuring Windows minidump type via sentry_options_t API in C++ SDK
### Problem Statement

The Sentry C++ SDK (sentry-native) currently generates minimal crash dumps on Windows using Crashpad. While this works for basic crash diagnostics, it lacks deeper memory-related information, such as heap state, handles, or full memory regions. In debugging complex issues (e.g., heap corruption, access violations, resource leaks), this limitation makes crash analysis significantly harder.

Although Crashpad internally supports various minidump levels via MINIDUMP_TYPE (e.g., MiniDumpWithFullMemory, MiniDumpWithHandleData), the current Sentry C++ API does not provide any mechanism to configure or override this behavior. As a result, users are forced to fork or wrap Crashpad themselves to generate richer dumps, which is error-prone and difficult to maintain.

### Solution Brainstorm

Expose a new API in sentry_options_t to configure the desired Windows minidump level (e.g., via flags or an enum representing MINIDUMP_TYPE values). This option could be transparently passed to Crashpad at initialization time via command-line arguments or inter-process communication, similar to how other options are forwarded to the crashpad_handler.

Example proposal (pseudo-API):

void sentry_options_set_windows_minidump_type(sentry_options_t* opts, unsigned int minidump_type);
Where minidump_type is a bitmask compatible with Windows' MINIDUMP_TYPE enum.

This would allow users to enable richer dump types like:

MiniDumpWithFullMemory

MiniDumpWithHandleData

MiniDumpWithThreadInfo

...

without needing to fork or modify Crashpad directly.

### Product Area

Settings

HuzaifaAbdulRehman:
## Summary

Fixes #1335.

- add `sentry_options_set_minidump_flags()` for the Native/Windows backend
- store custom flags in a single signed field, preserving `-1` as the unset sentinel and `0` as `MiniDumpNormal`
- make explicit Windows flags take precedence over `sentry_options_set_minidump_mode()`, regardless of call order
- propagate raw flags through the native crash context and derive `PROCESS_DUP_HANDLE` from the effective dump type
- document the supported `MINIDUMP_TYPE` values using the official Windows API reference
- add focused coverage for the default sentinel and mode, zero and maximum valid flags, both setter orders, and crash-context propagation

## Validation

- Native/Windows MinGW build with the Native backend completed successfully (119 build steps)
- `git diff --check` passes
- Windows unit tests require a PDB-capable MSVC or Clang toolchain and will run in CI; the repository explicitly rejects the installed MinGW toolchain for unit tests

## Notes

The platform-neutral flags API is currently guarded for Native/Windows. Existing preset behavior is unchanged unless the new setter is called; support for other platforms can be added later without another API rename.

Implementation and validation were assisted by OpenAI Codex. I reviewed the resulting design, code, and tests.


HuzaifaAbdulRehman:
@jpnurmi The implementation discussed in #1335 is ready in #1911. The available checks have passed; the remaining workflows are awaiting approval. PTAL when convenient. Thank you.

src/sentry_options.c
```diff
@@ -639,6 +653,20 @@ sentry_options_set_minidump_mode(
...
+sentry_options_set_windows_minidump_flags(
+    sentry_options_t *opts, uint32_t minidump_type_flags)
+{
+    if (opts->minidump_mode_explicitly_set) {
+        opts->minidump_configuration_overridden = true;
+    }
+    opts->windows_minidump_flags = minidump_type_flags;
+    opts->windows_minidump_flags_active = true;
+    opts->minidump_mode_explicitly_set = false;
+}
```
jpnurmi: Since we don't have getters for these that would force us to keep the original values around, could we simply clear/reset the other value to an invalid/negative/zero value and warn about a conflict here right away? Then we shouldn't need all that bookkeeping with multiple additional flags. 🤔 What do you think?

HuzaifaAbdulRehman: Good point. I can use an internal negative `minidump_mode` value for custom Windows flags, while retaining the raw `uint32_t` so `0` remains valid. This removes `windows_minidump_flags_active` and the deferred override state.

I think we still need one bit to distinguish the default `SMART` mode from an explicit mode setter call, so the first flags call does not warn.

Since these setters normally run before `sentry_init()` installs the options logger globally, should the warning invoke the options logger directly, or is `SENTRY_WARN` expected to work here?

jpnurmi: Good point about the logger! My bad, sorry. I think we've made that same mistake before.

Hmm, I might change my mind about the precedence if the other alternative, that explicit minidump type flags override the abstract minidump mode, looks significantly cleaner implementation-wise. Would we get rid of all the flags and clutter that way?

HuzaifaAbdulRehman: Yes, this would be substantially cleaner. The trade-off is that a later `sentry_options_set_minidump_mode()` call would not affect Windows once explicit flags have been configured. I think that is defensible if we clearly document that the lower-level Windows flags override the abstract mode regardless of call order.

Internally, I'd store `windows_minidump_flags` as an `int64_t` initialized to `-1`. The flags setter stores the incoming `uint32_t`, and the Windows writer uses it whenever non-negative; otherwise it derives the dump type from `minidump_mode`. This preserves `0` and the complete `uint32_t` input range.

That removes all three bookkeeping booleans and the warning path. Does this match the simplification you had in mind?

jpnurmi: Yes, more or less! :+1: Although we might not need 64 bits just for -1 because `MiniDumpValidTypeFlags = 0x01ffffff` fits into `int32_t`?

include/sentry.h
```diff
@@ -1875,12 +1875,34 @@ SENTRY_API void sentry_options_set_system_crash_reporter_enabled(
...
+#ifdef SENTRY_PLATFORM_WINDOWS
+/**
+ * Sets the raw Windows `MINIDUMP_TYPE` flags for the native backend.
+ *
+ * This setting and `sentry_options_set_minidump_mode` are mutually exclusive.
+ * When both are called, the most recently called function takes precedence and
+ * a warning is logged during initialization.
+ *
+ * `minidump_type_flags` is a bitmask compatible with the Windows
+ * `MINIDUMP_TYPE` enum. A value of zero selects `MiniDumpNormal`.
```
jpnurmi: Let's link to https://learn.microsoft.com/en-us/windows/win32/api/minidumpapiset/ne-minidumpapiset-minidump_type somewhere here

src/sentry_options.c
HuzaifaAbdulRehman: Good catch. Since `MiniDumpValidTypeFlags` fits comfortably in the positive `int32_t` range, `int32_t` with `-1` as the unset sentinel is sufficient. I'll use that and document that the input must be a valid `MINIDUMP_TYPE` bitmask. Thanks!

include/sentry.h
HuzaifaAbdulRehman: Will do. I’ll add the official MINIDUMP_TYPE documentation link here.

src/sentry_options.c
```diff
@@ -639,6 +642,15 @@ sentry_options_set_minidump_mode(
     opts->minidump_mode = mode;
 }
 
+#ifdef SENTRY_PLATFORM_WINDOWS
+void
+sentry_options_set_windows_minidump_flags(
+    sentry_options_t *opts, uint32_t minidump_type_flags)
+{
+    opts->windows_minidump_flags = (int32_t)minidump_type_flags;
```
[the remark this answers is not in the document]
HuzaifaAbdulRehman: Good catch. Documented `MINIDUMP_TYPE` values fit within positive `int32_t`, so valid inputs are unaffected. However, the narrowing conversion means an out-of-contract high-bit value could be misinterpreted as the unset sentinel and silently select the preset.

We can avoid that without restoring any bookkeeping flags: use the originally proposed `int64_t` field with `-1` as the sentinel and assign the `uint32_t` value without narrowing. That preserves the single-field design and the complete input representation.

@jpnurmi, would you prefer switching the internal field to `int64_t` after all, or retaining `int32_t` based on the documented valid-mask contract? I'd lean toward `int64_t` since it removes the representation gap entirely with no added complexity, but happy to go either way.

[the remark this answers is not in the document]
jpnurmi: What about simple masking, considering how `sentry_options_set_minidump_mode` fixes up the value, too? Also, I'd already take a cosmetic "preparation" step for the flags type for potential future cross-platform extension:

```c
typedef uint32_t sentry_minidump_flags_t;
```

```c
sentry_options_set_windows_minidump_flags(sentry_options_t *opts, sentry_minidump_flags_t flags)
{
    opts->minidump_flags = (int32_t)(flags & 0x01ffffff);
}
```
EDIT: Oh, and even drop the `windows_` prefix because the ultimage goal is to support other platforms where we can.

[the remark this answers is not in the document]
HuzaifaAbdulRehman: That makes sense. Masking before the cast keeps the single-field `int32_t` sentinel design, avoids the narrowing problem, and is consistent with `sentry_options_set_minidump_mode()` normalizing its input. Since this intentionally follows that precedent, I'm comfortable with unsupported bits being silently discarded.

I'll add `sentry_minidump_flags_t` alongside the public minidump mode type, use it in the setter signature, rename `windows_minidump_flags` to the platform-neutral `minidump_flags` throughout the internal state and crash context, mask with `0x01ffffff` before storing, document the normalization, and add coverage confirming that `UINT32_MAX` is normalized to `MiniDumpValidTypeFlags`.

One clarification: does dropping the `windows_` prefix also apply to the public setter, making it `sentry_options_set_minidump_flags()`, or should only the internal field become platform-neutral for now?

src/backends/native/sentry_crash_context.h
```diff
@@ -275,6 +275,9 @@ typedef struct {
 
     // Configuration (set by app during init)
     sentry_minidump_mode_t minidump_mode;
+#ifdef SENTRY_PLATFORM_WINDOWS
+    int32_t windows_minidump_flags;
```
jpnurmi: Just in case you missed the edited note at https://github.com/getsentry/sentry-native/pull/1911/changes#r3645900347, let's rename this to just `minidump_flags` because the ultimate goal is to support it for other platforms where we can.

HuzaifaAbdulRehman: Thanks, I saw the edited note. I'll rename the internal field to minidump_flags throughout the options state, crash context, propagation, writer, and tests.

src/sentry_options.c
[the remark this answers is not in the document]
jpnurmi: I would drop `windows_` everywhere. The `SENTRY_PLATFORM_WINDOWS` guards are sufficient for now. Once we add support for other platforms, we can lift the restriction by simply removing the guards, and won't need to rename it again at that point.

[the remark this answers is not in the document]
HuzaifaAbdulRehman: Understood, thanks for clarifying. I will rename the public setter and related test identifiers to sentry_options_set_minidump_flags() as well, while keeping the current Windows guards.

jpnurmi (approved):
Looks great, thank you for the contribution!

Sorry for the back-and-forth. It took a bit different shape than originally discussed, but I'm happy with where it landed. Appreciate you bearing with me through the nitpicks.

HuzaifaAbdulRehman:
Thank you, @jpnurmi! I really appreciate the thoughtful review and design guidance throughout the process. The back-and-forth helped me better understand the API trade-offs, and I learned a lot from working through them with you. I'm glad we arrived at a cleaner, extensible design. I'd also be happy to explore support for additional backends or platforms in the future when there is a suitable direction to pursue.

PR #1915, opened 2026-07-26
feat(feedback): Enrich user feedback with scope data

### LINKED ISSUE #1886: Apply relevant scope data to user feedback
Currently user feedback does not carry scope data, but it should. Expected: `environment`, `tags`, `release` etc.
Check [feedback spec](https://develop.sentry.dev/sdk/telemetry/feedbacks) for more details.

> SDKs MUST apply scope data to the feedback event, including tags, user, trace, and contexts.

Currently, user feedback is sent without trace, so it's not connected to the same trace as errors.

limbonaut:
Previously, user feedback was sent as a bare event payload carrying no scope data, so it had no trace tying it to errors from the same session, no release, environment, etc, which left Sentry defaulting the environment to `production` and mislabeling feedback coming from other builds. This PR applies the global scope to the feedback event before it is enveloped, so release, environment, tags, user, and contexts including the trace all come along. Breadcrumbs, modules, stacktraces are deliberately left off.

- Closes #1886
- Before this PR: [feedback item](https://sentry-sdks.sentry.io/issues/feedback/?feedbackSlug=sentry-godot%3A7633961032&project=6680910&referrer=feedback_list_page&statsPeriod=1h) 
- After this PR: [feedback item](https://sentry-sdks.sentry.io/issues/feedback/?feedbackSlug=sentry-godot%3A7634109749&project=6680910&referrer=feedback_list_page&statsPeriod=1h)

It also fixes a pre-existing leak where capturing feedback before the SDK is initialized dropped the feedback value without releasing it.

~~Scope attachments are excluded from this change.~~ 
The [spec](https://develop.sentry.dev/sdk/telemetry/feedbacks/) requires scope data, such as tags, user, trace, and contexts, to be included with feedback. It doesn't enumerate scope attachments specifically, but most SDKs that support user feedback do attach them. The only exception I found was Java/Android which doesn't include scope attachments.


src/sentry_core.c
```diff
@@ -831,11 +831,20 @@ static sentry_envelope_t *
...
-    if (!envelope
-        || !sentry__envelope_add_user_feedback(envelope, user_feedback)) {
+    sentry_value_set_by_key(
+        event, "level", sentry__value_new_level(SENTRY_LEVEL_INFO));
+
+    SENTRY_WITH_SCOPE (scope) {
+        sentry__scope_apply_to_event(scope, options, event, SENTRY_SCOPE_NONE);
+    }
+
+    sentry_envelope_t *envelope = sentry__envelope_new();
```
limbonaut, line 844: One decision point is whether to include scope attachments in user feedback. ~~Currently, scope attachments are excluded from this change.~~ The [spec](https://develop.sentry.dev/sdk/telemetry/feedbacks/) requires scope data, such as tags, user, trace, and contexts, to be included with feedback, but only says attachments MAY be sent. The spec doesn't explicitly forbid scope attachments either. Java/Android doesn't include scope attachments, but .NET and Cocoa do. So it seems that existing implementations drifted on this. Per-feedback attachments remain available via `sentry_capture_feedback_with_hint`.

limbonaut: I decided to go with attachments here: it seems only Java/Android does not attach scope attachments, others do. It's probably just a drift in the Java SDK.

jpnurmi (approved):
LGTM :+1:

PR #1916, opened 2026-07-26
feat(scope): Scoped user feedback capture

limbonaut:
Adds scoped capture of user feedback by introducing the `sentry_scope_capture_feedback` function, joining the existing family of `sentry_scope_capture_x` functions.  It layers the given scope on top of the global scope when building the feedback event, so a single capture can carry its own tags, user, and contexts without mutating global state.

- Depends on #1915
- Refs #1838


include/sentry.h
```diff
@@ -3749,6 +3750,22 @@ SENTRY_API sentry_attachment_t *sentry_hint_attach_bytesw_n(sentry_hint_t *hint,
...
+ * will be freed automatically. The hint parameter can be NULL if no additional
+ * context is needed.
+ *
+ * If `scope` is a local scope (`sentry_local_scope_new`), this takes ownership
+ * of it and frees it. If `scope` is user-owned (`sentry_scope_new` or
+ * `sentry_scope_clone`), it is applied but not freed, so it can be reused; free
+ * it yourself with `sentry_scope_free`.
+ */
+SENTRY_API sentry_uuid_t sentry_scope_capture_feedback(
+    sentry_scope_t *scope, sentry_value_t user_feedback, sentry_hint_t *hint);
```
limbonaut, line 3767: I opted to go with the `hint` argument here, instead of forking it to a separate `sentry_scope_capture_feedback_with_hint()` function. The top-level functions do have such a split though. So, I'm still not 100% sure if it's the right call. Let me know what you think.

jpnurmi: I think it’s fine. We can probably drop the old hintless overload in 1.0

src/sentry_core.c
```diff
@@ -839,22 +840,44 @@ prepare_user_feedback(const sentry_options_t *options,
...
+                envelope, scope->attachments, options);
+        }
         if (options->run) {
-            sentry__cache_attachment_refs(envelope, hint->attachments, options,
+            sentry__cache_attachment_refs(envelope,
+                all_attachments ? all_attachments : scope->attachments, options,
                 options->run->cache_path, options->run->run_path);
         }
     }
+    sentry__attachments_free(all_attachments);
```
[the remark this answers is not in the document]
limbonaut: This needs to be decided in #1915

jpnurmi (approved):

PR #1924, opened 2026-07-27
fix: close fd 0 when mmap fails

fallintoplace:
When stdin is closed, `open()` can return file descriptor 0. The failure cleanup in `sentry__mmap_file` only closed descriptors greater than 0, which leaked the file descriptor if a later step failed.

Close every non-negative descriptor and add a Linux regression test that exercises the empty-file failure path with stdin closed.


jpnurmi (approved):
LGTM, thank you!

PR #1925, opened 2026-07-27
fix(docker): expand PIPX_BIN_DIR when adding it to PATH

fallintoplace:
The Alpine Dockerfile currently adds the literal string `PIPX_BIN_DIR` to `PATH`, so executables installed by pipx into `/opt/pipx_bin` are not discoverable.

Expand the existing environment variable when constructing `PATH`.

#skip-changelog


jpnurmi (approved):
Good catch, thank you!

PR #1927, opened 2026-07-29
(WIP) feat(client-reports): report log and metric byte outcomes

JoshuaMoelans:
Record the byte size of discarded logs and metrics under the `log_byte` and `trace_metric_byte` categories, alongside the existing `log_item` and `trace_metric` counts, as required by
https://develop.sentry.dev/sdk/telemetry/client-reports/#log-byte-outcomes

Byte sizes come from the serialized payload where an item is already serialized, and from a new allocation-free `sentry_value_t` size estimator on the `before_send_log` / `before_send_metric` and batcher overflow paths. The hooks take ownership of the value, so it is sized before the call.

This also fixes the item count for batched envelope items: discarding a `log` or `trace_metric` item now reports every log in the batch via its `item_count` header instead of counting the whole item as one discard.

Following https://develop.sentry.dev/sdk/telemetry/client-reports/#log-byte-outcomes


src/sentry_value.c
```diff
@@ -1158,6 +1158,45 @@ sentry__value_foreach_key_value(sentry_value_t value,
...
+        return strlen(thing->payload._ptr) + 2; // surrounding quotes
+    case THING_TYPE_DOUBLE:
+    case THING_TYPE_INT64:
+    case THING_TYPE_UINT64:
+    default:
+        // upper bound for any 64-bit integer or double rendering
+        return 24;
+    }
+}
+
```
JoshuaMoelans, line 1199: Q: I'm a bit afraid of the overhead on this being called for every single log item, especially the recursive nature of the function... I have no insights into what size logs people are usually sending/discarding, so will need to do some benchmarking before committing to this...

any other approaches to keep track of the size are welcome; we could maybe store these things upon creating the `sentry_value_t` types and adding to the objects, but this is a larger rework of our internals so worth discussing

JoshuaMoelans: (potentially we make this PR just fix the log count numbers, and do the rework with log_bytes counting separately)

JoshuaMoelans:
It might make sense to surface LogItems in our product instead of providing this sort of scuffed pre-serializing size counting in sentry-native. We can already show them in product, but this is currently gated behind a feature flag ([turned on for Sentry-SDKs](https://sentry-sdks.sentry.io/settings/stats/?clientDiscard=true&dataCategory=logItems&project=4511738974961664&statsPeriod=7d))

[Flag lives here](https://github.com/getsentry/sentry/blob/master/src/sentry/features/permanent.py#L67)
Checked in [organizationStats](https://github.com/getsentry/sentry/blob/dd98e6cdc1da0899f536361ce2c4c9ff1816c8a5/static/app/views/organizationStats/index.tsx#L273-L275) and [metricFlags](https://github.com/getsentry/sentry/blob/dd98e6cdc1da0899f536361ce2c4c9ff1816c8a5/static/app/views/explore/metrics/metricsFlags.tsx#L10-L15)

src/sentry_value.c
jpnurmi: it would be interesting to benchmark how much time such a "dry-run" serialization really takes for complex objects. even if it needs to recurse the entire tree and scan all strings, it doesn't need to allocate anything 🤔

jpnurmi: with the upcoming telemetry thread-pool, we could probably dispatch these as tasks processed in the background, just like we do when serializing logs and metrics for real
- https://github.com/getsentry/sentry-native/pull/1883

JoshuaMoelans: hmm yeah true, I'll try it out when I get some downtime

jpnurmi: we could even consider adding an internal `string_t` alongside `list_t` and `obj_t`, for storing string lengths. `sentry_value_new_string` calls  `sentry_value_new_string_n`, so lengths are known at construction time anyway.

PR #1928, opened 2026-07-29
feat: added `app_hang_pause`

bitsandfoxes:
Huge footgun for anything that can be put on the background. Like a game.

Calling `pause` pauses the app hang detection. The next `heartbeat` resumes the detection again.


src/sentry_app_hang_monitor.c
```diff
@@ -181,3 +187,12 @@ sentry__app_hang_monitor_stop(void)
...
 #endif // SENTRY_HAS_THREAD_STACKWALK
+
+void
+sentry_app_hang_pause(void)
+{
+    if (!sentry__app_hang_is_active()) {
+        return;
+    }
+    sentry__app_hang_set_paused(true);
+}
```
JoshuaMoelans: q: should this live here or in `app_hang_latch.c` next to the `app_hang_beartbeat()` function?

CHANGELOG.md
```diff
@@ -2,7 +2,12 @@
 
 ## Unreleased
 
+**Features**
```
JoshuaMoelans: smol :

```suggestion
**Features**:
```

include/sentry.h
```diff
@@ -2829,6 +2829,15 @@ SENTRY_EXPERIMENTAL_API uint64_t sentry_options_get_app_hang_timeout(
  */
 SENTRY_EXPERIMENTAL_API void sentry_app_hang_heartbeat(void);
 
+/**
+ * Pauses app-hang detection without changing the watched thread.
+ *
+ * While paused, the watchdog does not capture app hangs. A subsequent heartbeat
+ * from the watched thread automatically resumes detection. Heartbeats from
+ * other threads continue to be ignored.
+ */
```
JoshuaMoelans, line 2841: (nit) consistent no-op documentation akin to the one for the heartbeat function above
```suggestion
 * from the watched thread automatically resumes detection. Heartbeats from
 * other threads continue to be ignored.
 * 
 * This function is a no-op unless app-hang detection is enabled via
 * `sentry_options_set_enable_app_hang_tracking`.
 */
```

tests/unit/test_app_hang.c
```diff
@@ -43,6 +43,38 @@ SENTRY_TEST(app_hang_latch)
     sentry__app_hang_set_active(false);
 }
 
+SENTRY_TEST(app_hang_pause_resumes_on_heartbeat)
+{
+#if !SENTRY_HAS_THREAD_STACKWALK
+    SKIP_TEST();
+#endif
```
JoshuaMoelans, line 50: q: do we also want a full test that does the full SDK init as in `app_hang_disarm_prevents_capture` , not just the latch-only unit test? Currently we don't hit the `worker` function path that checks the 'paused' state.

src/sentry_app_hang_latch.c
```diff
@@ -67,6 +68,23 @@ sentry__app_hang_is_active(void)
...
 
+void
+sentry__app_hang_set_paused(bool paused)
+{
+    if (paused) {
+        // Clear the old timestamp before publishing the paused state, so a
+        // concurrent heartbeat cannot leave detection armed with stale data.
+        sentry__atomic_store_u64(&g_last_heartbeat_ms, 0);
+    }
+    sentry__atomic_store(&g_app_hang_paused, paused ? 1 : 0);
```
JoshuaMoelans, line 74: q: because the `heartbeat` function stores the `g_last_heartbeat_ms` before setting paused to false again, is this 0-ing necessary? The stale data always gets overwritten before unpausing.

Would make more sense if we had a `resume` call that itself doesn't trigger a heartbeat, so if we want to keep that avenue open we could keep this in (but that could also be added on top later on IMO)

JoshuaMoelans (approved):
LGTM, few nits/comments/questions, main ones are about resetting the last timestamp captured to 0 on pause & a full test that actually invokes the worker which checks the paused state

src/sentry_app_hang_monitor.c
bitsandfoxes: That makes sense.

src/sentry_app_hang_latch.c
bitsandfoxes: Good catch!

PR #1931, opened 2026-07-30
chore: virtual clock for app hang tests

bitsandfoxes:
Instead of having to call `sleep_ms` and adding timeouts to wait for app hangs to be detected this introduces an abstraction for the clock for unit tests. This can then be expanded to be used in other tests as well.

This also extracts the `checking` part out of the worker for unit tests. Checking the full worker functionality i.e. scheduling and wake up are a follow up.

#skip-changelog


src/sentry_app_hang_monitor.c
```diff
@@ -85,43 +85,49 @@ app_hang_capture(uint64_t hang_time_ms, uint64_t tid)
     return true;
 }
 
+void
+sentry__app_hang_monitor_check(uint64_t *last_fired_heartbeat_ms)
```
bitsandfoxes: Extracted for unit test.

src/sentry_app_hang_monitor.c
```diff
@@ -85,43 +85,49 @@ app_hang_capture(uint64_t hang_time_ms, uint64_t tid)
...
 SENTRY_THREAD_FN
 worker(void *arg)
 {
     (void)arg;
     uint64_t last_fired_hb = 0;
     while (sentry__app_hang_is_active()) {
         sentry__mutex_lock(&g_wait_mutex);
+        if (!sentry__app_hang_is_active()) {
+            sentry__mutex_unlock(&g_wait_mutex);
+            break;
```
bitsandfoxes, line 98: Covers race between outer while condition and acquiring `g_wait_mutex`.

tests/unit/test_app_hang.c
```diff
@@ -281,25 +308,37 @@ SENTRY_TEST(app_hang_end_to_end)
...
-        sleep_ms(10);
+    sentry__mutex_lock(&spinner_lock);
+    for (int i = 0; i < 10 && !g_spinner_started; i++) {
+        sentry__cond_wait_timeout(&g_spinner_ready, &spinner_lock, 100);
     }
+    const bool spinner_started = g_spinner_started;
+    sentry__mutex_unlock(&spinner_lock);
+    TEST_ASSERT(spinner_started);
+
+    sentry__test_clock_advance(1000);
```
bitsandfoxes: The changes revolve around this. No more sleeping and waiting something is finished spinning up, or waking up. Allows for deterministic testing.

PR #1932, opened 2026-07-30
feat(scopes): add `sentry_scope_remove_fingerprint`

limbonaut:
Adds `sentry_scope_remove_fingerprint` to complement pre-existing `sentry_remove_fingerprint` and `sentry_scope_set_fingerprint`.

- Refs https://github.com/getsentry/sentry-godot/pull/834#discussion_r3682642311

Once a fingerprint was set on a scope, there was no way to remove it. Passing an empty list doesn't clear the fingerprint, it still counts as one. Because the local scope is applied to an event before the global scope, that empty list overrides the global scope fingerprint instead of allowing it to be inherited.


JoshuaMoelans (approved):
nice ☝️ 🖨️

PR #1934, opened 2026-07-31
fix(attachments): flush scope on updating attachments

### LINKED ISSUE #1933: Native backend uses physical attachment filename for hard crashes, instead of the name passed to USentryLibrary::CreateSentryAttachmentWithData()
**Description**

On Windows, the native backend can use the wrong attachment filename after a hard crash.

The application adds a file attachment and then sets a custom filename:

```cpp
sentry_attachment_t *attachment =
    sentry_attach_file("C:\\temp\\archive-2026-07-31.zst");
sentry_attachment_set_filename(attachment, "game.log.zst");
```

A normal event shows `game.log.zst`, as expected. A hard-crash event shows `archive-2026-07-31.zst`, which is the basename of the physical path.

UE ensure events do not have this problem. They use the normal event path and read the current in-memory attachment. Only hard-crash events use the stale attachment manifest.

The cause is the order of these operations:

1. `sentry_attach_file(path)` adds the attachment to the global scope.
2. `SENTRY_WITH_SCOPE_MUT` immediately flushes the scope (`sentry_core.c:2040-2068`, `sentry_scope.h:162-164`, and `sentry_scope.c:179-186`).
3. The native backend writes `__sentry-attachments` before the custom filename is set (`sentry_backend_native.c:729-780`, especially lines 757-760).
4. `sentry_attachment_set_filename` changes the in-memory attachment, but it does not flush the scope (`sentry_attachment.c:63-80`).
5. A normal event reads the current in-memory attachment and uses the custom filename (`sentry_attachment.c:163-169` and `sentry_envelope.c:722-751`).
6. The crash daemon reads the earlier attachment manifest and uses the physical basename (`sentry_crash_daemon.c:3614-3658`).

A later scope change can rewrite the manifest and hide the problem. Therefore, the result can depend on timing.

The Sentry UE wrapper calls these APIs in this order in `GenericPlatformSentrySubsystem.cpp:411-419`. The Windows implementation has the same order in `MicrosoftSentrySubsystem.cpp:77-84`. Project code only supplies the physical path and custom filename. The affected processing after that point is Sentry code.

- [ ] During build
- [x] During run-time

**Environment**

- OS: All platforms where we use the Sentry native backend
- Compiler: MSVC 14.38.33130
- sentry-native version: 0.16.0
- Sentry UE5 plugin version: 1.18.0

**Steps To Reproduce**

1. Initialize sentry-native with `SENTRY_BACKEND=native`.
2. Create `C:\temp\archive-2026-07-31.zst`.
3. Add the attachment and set a different filename:

```cpp
sentry_attachment_t *attachment =
    sentry_attach_file("C:\\temp\\archive-2026-07-31.zst");
sentry_attachment_set_filename(attachment, "game.log.zst");
```

4. Capture a normal event. Confirm that its attachment filename is `game.log.zst`.
5. Repeat the setup.
6. Do not make another scope change after `sentry_attachment_set_filename`.
7. Cause a hard crash.
8. Inspect the attachment filename in the hard-crash event. It is `archive-2026-07-31.zst` instead of `game.log.zst`.

**Log output**

No relevant error is written to the log.

Expected: Ensure events, other normal events, and hard-crash events use `game.log.zst`.

Actual: Ensure events and other normal events use `game.log.zst`. Hard-crash events use `archive-2026-07-31.zst`.

JoshuaMoelans:
Fixes https://github.com/getsentry/sentry-native/issues/1933

Updates to attachments weren't flushed to the scope (only when adding it with `sentry_attach_file`). This caused updates to attachments to only apply on non-fatal events, since the native backend only knows about the state of the attachment from right before attaching it.

This fixes both questions in the original issue; attachment names are synced & so are the content_types, which initially made Sentry backend interpret the file as the default `application/octet-stream` which gets a Preview button (whereas the `application/ztsd` would not)

From the flow that UE hits on adding attachments, we see the same attach -> set_filename -> set_content_type flow that we now fix with scope syncing: 

```c++
void FGenericPlatformSentrySubsystem::AddFileAttachment(TSharedPtr<ISentryAttachment> attachment)
{
	TSharedPtr<FGenericPlatformSentryAttachment> platformAttachment = StaticCastSharedPtr<FGenericPlatformSentryAttachment>(attachment);

	sentry_attachment_t* nativeAttachment =
		sentry_attach_file(TCHAR_TO_UTF8(*platformAttachment->GetPath()));

	if (!platformAttachment->GetFilename().IsEmpty())
		sentry_attachment_set_filename(nativeAttachment, TCHAR_TO_UTF8(*platformAttachment->GetFilename()));

	if (!platformAttachment->GetContentType().IsEmpty())
		sentry_attachment_set_content_type(nativeAttachment, TCHAR_TO_UTF8(*platformAttachment->GetContentType()));

	platformAttachment->SetNativeObject(nativeAttachment);

	attachments.Add(platformAttachment);
}
```


JoshuaMoelans:
@sentry review

src/sentry_attachment.c
```diff
@@ -22,6 +28,19 @@ sentry_attachment_set_type_n(
         return;
     }
 
+    SENTRY_WITH_SCOPE_MUT (scope) {
+        sentry__attachment_set_type_n(attachment, type, type_len);
+    }
```
JoshuaMoelans, line 33: open q: both places where we now `WITH_SCOPE` are additional overhead, although these (probably) don't get called that often. This overhead is acceptable for the attachments that actually are on the global scope, but we also do it for those that are in the local scopes (or hint attachments). 
We could walk the global attachments list to check if a scope lock is necessary, but this also seems like overhead (albeit less so than a full scope flush), so it might be worth considering.

jpnurmi: There are quite a few changes ahead if we want to do that, but I'd eventually
- replace heap-allocated `sentry_attachment_t` objects with `sentry_value_t`-backed value types,
- replace global scope mutex with fine-grained read-write locks, and
- replace backend flush functions with scope observers.

jpnurmi (approved):
I think this is fine for now.

There's an increasing number of reasons for moving towards sentry_value-based attachments and potentially getting rid of the whole backend flush-function altogether, but such changes would not happen overnight. This change helps until then.

tests/test_integration_native.py
```diff
@@ -285,6 +287,58 @@ def test_native_overflow_breadcrumbs(cmake, httpserver, crash_mode):
...
+        if not paths:
+            return None
+        try:
+            return json.loads(paths[0].read_text())
+        except (OSError, json.JSONDecodeError):
+            return None
+
+    try:
+        renamed = wait_for(
+            lambda: (read_manifest() or [{}])[0].get("filename") == "custom-name.log"
```
jpnurmi, line 329: should it wait for `content_type` as well? if the timing is unlucky, this could end up being true between two flushes:
```c
sentry_attachment_set_filename(attachment, "custom-name.log");
sentry_attachment_set_content_type(attachment, "application/zstd");
```

```
FAILED tests/test_integration_native.py::test_native_attachment_manifest_is_current - KeyError: 'content_type'
```
https://github.com/getsentry/sentry-native/actions/runs/30884798825/job/91913533861

PR #1938, opened 2026-08-03
ref(transport): move crash-time disk dumping into transport

jpnurmi:
Let crash backends suspend the transport's network submission and automatically route crash-time envelopes through its disk-dump path.

This removes the need for allocating fallback disk transports in the backends, and more importantly, prepares for the upcoming thread-pool telemetry module, which cannot safely drain the transport queue from a crash handler (see https://github.com/getsentry/sentry-native/pull/1883#discussion_r3604766425 and https://github.com/getsentry/sentry-native/pull/1883#discussion_r3631459231 for more details). This change abstracts away the problem so that callers no longer need to care whether the SDK has already started dumping in-flight envelopes to disk.

### Before (pseudo)

```c
void backend_except_func()
{
    // ...
    sentry_transport_t *disk_transport = sentry_new_disk_transport(options->run);
    sentry__capture_envelope(disk_transport, envelope, options);
    sentry__transport_dump_queue(disk_transport, options->run);
    sentry_transport_free(disk_transport);
}
```

### After (pseudo)

```c
void backend_except_func()
{
    sentry__transport_suspend(options->transport);
    // ...
    sentry__capture_envelope(options->transport, envelope, options);
}
```

#skip-changelog (internal refactor)


JoshuaMoelans (approved):
very nice refactor 👌

PR #1946, opened 2026-08-04
perf(telemetry): serialize batches on shared thread pool

### LINKED ISSUE #1862: Logs: optimize performance
In sentry-unreal, `UE_LOG` can be wired into sentry-native logs. That means the native SDK may receive not only explicit game/application logs, but also high-volume Unreal Engine logs from many engine and game threads.

This makes log ingestion a hot path: in busy Unreal-like workloads, every log reads the global scope to apply attributes and then passes through the native log batcher. Under bursty multi-threaded logging, that exposes two main bottlenecks:

- logger/producer threads: global scope lock for applying attributes (`#ff0000`)
- batcher/consumer thread: log item serialization (`#0000ff`) and cleanup (`#00ff00`)

Image

jpnurmi:
Let the telemetry module own a shared [thread pool](https://github.com/getsentry/sentry-native/pull/1883) for telemetry batchers, and use it for log and metric batch serialization.

image

**Before**: on the left, the batcher thread is unable to keep up with serializing incoming logs before handing them off to the transport thread. Serialization takes too much time, queues fill up, logs overflow, and get discarded. The batcher thread continues working for a good while after logs stop coming in.

**After**: on the right, the batcher thread only acts as a gateway for incoming logs. It has no problem keeping up with the flood because they are only quickly organized into batches and immediately handed over to the telemetry pool, which then distributes the heavy serialization work to its pooled threads, resulting in no discarded logs.

[before-vs-after.ftrace.zip](https://github.com/user-attachments/files/30353388/before-vs-after.ftrace.zip)

A partial, batcher-side fix to:
- #1862


src/sentry_batcher.c
```diff
@@ -167,27 +178,322 @@ rotate_buffer(sentry_batcher_t *batcher, long old_idx)
...
+
+    if (sentry__threadpool_submit(batcher->threadpool, batch_task_exec,
+            batch_task_complete_and_cleanup, NULL, task)
+        != 0) {
+        SENTRY_WARN("serializing telemetry batch synchronously: "
+                    "serialization pool unavailable or out of memory");
+        batch_task_unlink(task);
+        process_batch_sync_ordered(batcher, envelope, items);
+        sentry_free(task);
+        return;
```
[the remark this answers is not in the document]
jpnurmi: Not valid.

The failure path does not pass a cleanup callback to `sentry__threadpool_submit`: `cleanup_func` is `NULL` in the call here:
https://github.com/getsentry/sentry-native/blob/543773586e635c59cc325c7a363fb1d86c0ff5dc/src/sentry_batcher.c#L478-L485

`submit` only invokes submitted-data cleanup through `cleanup_func`; on shutdown/rejection it frees only the internal threadpool task wrapper:
https://github.com/getsentry/sentry-native/blob/543773586e635c59cc325c7a363fb1d86c0ff5dc/src/sentry_sync.c#L293-L320

So after submit failure, `process_batch` still owns `task`, `envelope`, and `items`. The fallback path can safely unlink the batch task, serialize/send synchronously, and free the task. This would be a real bug only if `batch_task_cleanup` were passed as the cleanup callback.

src/sentry_telemetry.c
```diff
@@ -22,19 +50,33 @@ sentry__telemetry_shutdown(const sentry_options_t *options)
...
+    SENTRY__MUTEX_INIT_DYN_ONCE(g_telemetry_lock);
+    sentry__mutex_lock(&g_telemetry_lock);
+
     SENTRY_DEBUG("shutting down telemetry");
     sentry__logs_shutdown(options->shutdown_timeout);
     sentry__metrics_shutdown(options->shutdown_timeout);
+    sentry__threadpool_flush(g_telemetry_pool);
+    sentry__threadpool_shutdown(g_telemetry_pool);
+    sentry__threadpool_free(g_telemetry_pool);
+    g_telemetry_pool = NULL;
```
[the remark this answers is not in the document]
jpnurmi: we probably should try to do something about it, but the timeout is already ignored in master:
https://github.com/getsentry/sentry-native/blob/776adaefd6cd9eefbe4e9e41a4ef7929387d84ee/src/sentry_batcher.c#L426-L429

jpnurmi:
Update: I'm investigating why TSAN jobs started hanging in the CI after the latest rebase and `max_pending` changes. I haven't been able to reproduce locally... :(

jpnurmi:
> Update: I'm investigating why TSAN jobs started hanging in the CI after the latest rebase and `max_pending` changes. I haven't been able to reproduce locally... :(

The problem turned out to be the shared spinlock promoted in #1964. :cry: Apparently, the Unix page allocator and telemetry batcher require different locking semantics and should not share the same implementation. Thus, I've taken a step back and reverted it in https://github.com/getsentry/sentry-native/pull/1946/commits/c64b85012c2af000a003df7ffbd9a9b2195cba30. The CI is back to green.

mujacica (approved):

src/sentry_batcher.c
```diff
@@ -167,27 +182,337 @@ rotate_buffer(sentry_batcher_t *batcher, long old_idx)
...
+{
+    sentry_batch_task_t *task = task_data;
+    sentry_batcher_t *batcher = task->batcher;
+
+    lock_tasks(batcher);
+    if (!sentry__atomic_compare_swap(&task->state, SENTRY_BATCH_TASK_PENDING,
+            SENTRY_BATCH_TASK_RUNNING)) {
+        unlock_tasks(batcher);
+        return;
+    }
```
limbonaut, line 370: A small nit: Can we add a comment for a future reader why this path is expected to execute during crash handling? (crash path got there first and dumped it)

src/sentry_batcher.c
```diff
@@ -167,27 +182,337 @@ rotate_buffer(sentry_batcher_t *batcher, long old_idx)
...
+batch_task_wait_all(sentry_batcher_t *batcher)
+{
+    while (true) {
+        lock_tasks(batcher);
+        const bool done = batcher->tasks == NULL;
+        unlock_tasks(batcher);
+        if (done) {
+            return;
+        }
+        sentry__cpu_relax();
```
limbonaut: Q: Should we wait on a condvar here or use `crash_safe_sleep_ms`? Or do we expect this to be a relatively short duration? This part is not on a crash path if I'm not mistaken.

src/sentry_batcher.c
```diff
@@ -435,40 +745,47 @@ sentry__batcher_shutdown(sentry_batcher_t *batcher, uint64_t timeout)
...
     // Perform final flush to ensure any remaining items are sent
     sentry__batcher_flush(batcher, false);
+    batch_task_wait_all(batcher);
 }
 
 void
 sentry__batcher_flush_crash_safe(sentry_batcher_t *batcher)
 {
     // Check if batcher is initialized
+    sentry__atomic_store(&batcher->crash_flush, 1);
```
limbonaut, line 822: Potential use-after-free here: write through a pointer to a possibly freed batcher. Two things can happen at the same moment: the app is shutting down sdk cleanly via `sentry_close()` and the app crashing. The shutdown path frees the batcher. The crash path at the same time takes that same object through `sentry__batcher_peek` without refcount.

Since it involves a narrow race, here is a clanker-reproduced pinned sequence that demonstrates the problem.

```c
SENTRY_TEST(batcher_crash_flush_after_release)
{
    // Models the shutdown-vs-crash race in sentry__logs_flush_crash_safe():
    // the crash thread peeks the batcher without taking a reference, then
    // sentry_close() swaps the ref, shuts the batcher down and releases it
    // before the crash thread dereferences the pointer it already holds.
    sentry_batcher_ref_t ref = SENTRY_BATCHER_REF_INIT;
    sentry_batcher_t *batcher = sentry__batcher_new(pending_batch_func, NULL);
    TEST_ASSERT(!!batcher);

    sentry_path_t *database_path = NULL;
    sentry_run_t *run = new_test_run(
        SENTRY_TEST_PATH_PREFIX ".batcher-crash-after-release", &database_path);
    batcher->run = run;

    sentry__batcher_swap(&ref, batcher);

    // crash thread: sentry__batcher_peek() takes no reference
    sentry_batcher_t *stale = sentry__batcher_peek(&ref);
    TEST_CHECK(stale == batcher);

    // shutdown thread: sentry__logs_shutdown()
    sentry_batcher_t *old = sentry__batcher_swap(&ref, NULL);
    sentry__batcher_shutdown(old, 0);
    sentry__batcher_release(old); // refcount 1 -> 0, batcher is freed

    // crash thread resumes on the pointer it peeked before the swap
    sentry__batcher_flush_crash_safe(stale);

    free_test_run(run, database_path);
}
```

limbonaut (approved):
Super well-polished. This should increase batching throughput quite noticeably. As far as I understood the workings, I didn't spot any issues except one narrow race between closing and crashing paths (see comments). Looks good otherwise!

src/sentry_batcher.c
jpnurmi: Resulted in a small side quest:
- https://github.com/getsentry/sentry-native/pull/2004

src/sentry_batcher.c
jpnurmi: Hah, awesome. Thanks for the test case! Failed nicely with TSAN. `sentry__batcher_peek` became `sentry__batcher_pin` + `unpin` :)

https://github.com/getsentry/sentry-native/pull/1946/commits/fbec2888329b0ee6bdd3f054604bedc2e9522531

src/sentry_batcher.c
jpnurmi: https://github.com/getsentry/sentry-native/pull/1946/changes/936acfa9e8c879cb154d34a1c49f6d3c97ffb6df

src/sentry_batcher.c
jpnurmi: 936acfa9e8c879cb154d34a1c49f6d3c97ffb6df

jpnurmi: I should've stress-tested https://github.com/getsentry/sentry-native/commit/936acfa9e8c879cb154d34a1c49f6d3c97ffb6df harder with TSAN. :slightly_frowning_face: The very first [CI run](https://github.com/getsentry/sentry-native/actions/runs/32364700595/job/96411581752) in master failed...

- https://github.com/getsentry/sentry-native/pull/2008

PR #1947, opened 2026-08-04
test(attachments): cleanup scope flush assertions

JoshuaMoelans:
follow-up to #1934 , more robustly fetching the keys using `get()` and folded the assertion into the function that we wait for. No longer should fail on unflushed `content_type` as in https://github.com/getsentry/sentry-native/actions/runs/30884798825/job/91913533861#step:31:63880

#skip-changelog


jpnurmi (approved):
🙏

Issue #1852, opened 2026-07-09
Native: optimize backend startup on Linux


jpnurmi:
crashpad vs. native startup should be comparable

### Local

```sh
$ pytest -v tests/benchmark.py::test_benchmark_backend -k "crashpad or native"
...
tests/benchmark.py::test_benchmark_backend[crashpad] PASSED
Min 1.465ms, Max 1.637ms, Mean 1.536ms, StdDev 0.079ms, Median 1.506ms, CPU 0.357ms
tests/benchmark.py::test_benchmark_backend[native] PASSED
Min 5.710ms, Max 7.042ms, Mean 6.445ms, StdDev 0.617ms, Median 6.579ms, CPU 2.677ms
```

### CI

Note: the Y-axis is logarithmic

<img width="2598" height="753" alt="Image" src="https://github.com/user-attachments/assets/fae0c09f-ae97-44bf-b6ad-d2da677ffdbb" />

https://getsentry.github.io/sentry-native/

jpnurmi:
Worst offenders (~9ms startup):
- zeroing 8MB of shared memory, dominated by modules (3ms, ~30%)
- loading curl and its deps (`LD_DEBUG=statistics`) before sentry-daemon main() (4ms, ~40%)

jpnurmi:
Fixed by #1966 and #1955 

<img width="2078" height="1014" alt="Image" src="https://github.com/user-attachments/assets/5192df23-0bb5-4cfe-bd57-846a1467d88a" />

<img width="2078" height="1014" alt="Image" src="https://github.com/user-attachments/assets/4411102b-6813-4aa5-a76f-51735d45c471" />

https://getsentry.github.io/sentry-native/

PR #1966, opened 2026-08-06
perf(native): avoid zeroing large shmem modules array

jpnurmi:
When creating shared memory for the native backend and daemon, zero the crash context only up to `module_count`. Leaving the trailing ~8 MiB `modules[]` array uninitialized is safe, because the array access is guarded by `module_count`. 

This shaves off ~40% of the current native backend startup time on Linux:

### Before
```
tests/benchmark.py::test_benchmark_backend[native] PASSED
Min 6.029ms, Max 7.111ms, Mean 6.659ms, StdDev 0.475ms, Median 6.736ms, CPU 2.968ms
```

### After
```
tests/benchmark.py::test_benchmark_backend[native] PASSED
Min 3.760ms, Max 4.572ms, Mean 4.076ms, StdDev 0.376ms, Median 3.831ms, CPU 0.461ms
```

Partial fix to:
- #1852


limbonaut (approved):

JoshuaMoelans (approved):

jpnurmi:
Quite a bit better

image

The rest will be caught up in
- https://github.com/getsentry/sentry-native/pull/1955

Issue #1945, opened 2026-08-04
ref(attachments): rework to be `sentry_value_t`-based


JoshuaMoelans:
Follow-up to #1934 as brought up by @jpnurmi  ; the fix we provided there works for now, but looking forward we'd ideally end up with a more robust attachments-handling system by implementing the following:

- replace heap-allocated `sentry_attachment_t` objects with `sentry_value_t`-backed value types,
- replace global scope mutex with fine-grained read-write locks, and
- replace backend flush functions with scope observers (see #1839 )

PR #1968, opened 2026-08-07
ref(value): cache string length

jpnurmi:
Since `sentry_value_new_string` calls `strlen` + `sentry_value_new_string_n` anyway, we might as well store the length:

https://github.com/getsentry/sentry-native/blob/ff209d555ab75f203821d312ee6c8f4dbd31b014/src/sentry_value.c#L517-L522

The pre-determined length nicely flows into serialization etc. It not only saves us from calling `strlen` wherever we have `_n` APIs available, but also implicitly adds the capability to handle embedded `NUL` bytes for `sentry_value_t`-backed attachments in the future.

To avoid paying for an extra allocation (`char[]` AND `blob_t`), the proposal uses a flexible array member so the length and string bytes share a single allocation. This way, the number of allocations doesn't increase, but we just allocate `size_t` + string together in one go.

See also:
- #1945

### DIFF
--- a/src/sentry_json.c
+++ b/src/sentry_json.c
@@ -22,6 +22,7 @@
 #include "sentry_value.h"
 
 #define SENTRY_JSON_MAX_DEPTH 64
+#define SENTRY_JSON_STRLEN ((size_t)-1)
 
 struct sentry_jsonwriter_s {
     sentry_writer_t *writer;
@@ -201,12 +202,15 @@ static unsigned char needs_escaping[256] = {
 };
 
 static void
-write_json_str(sentry_jsonwriter_t *jw, const char *str)
+write_json_str(sentry_jsonwriter_t *jw, const char *str, size_t str_len)
 {
     // using unsigned here because utf-8 is > 127 :-)
     const unsigned char *ptr = (const unsigned char *)str;
     const unsigned char *start = ptr;
-    for (; *ptr && !sentry__jsonwriter_has_failed(jw); ptr++) {
+    const unsigned char *end
+        = str_len == SENTRY_JSON_STRLEN ? NULL : ptr + str_len;
+    for (; (end ? ptr < end : *ptr) && !sentry__jsonwriter_has_failed(jw);
+        ptr++) {
         if (!needs_escaping[*ptr]) {
             continue;
         }
@@ -353,13 +357,20 @@ sentry__jsonwriter_write_double(sentry_jsonwriter_t *jw, double val)
 void
 sentry__jsonwriter_write_str(sentry_jsonwriter_t *jw, const char *val)
 {
+    sentry__jsonwriter_write_str_n(jw, val, SENTRY_JSON_STRLEN);
+}
+
+void
+sentry__jsonwriter_write_str_n(
+    sentry_jsonwriter_t *jw, const char *val, size_t val_len)
+{
     if (!val) {
         sentry__jsonwriter_write_null(jw);
         return;
     }
     if (can_write_item(jw)) {
         write_char(jw, '"');
-        write_json_str(jw, val);
+        write_json_str(jw, val, val_len);
         write_char(jw, '"');
     }
 }
@@ -390,7 +401,7 @@ sentry__jsonwriter_write_key(sentry_jsonwriter_t *jw, const char *val)
 {
     if (can_write_item(jw)) {
         write_char(jw, '"');
-        write_json_str(jw, val);
+        write_json_str(jw, val, SENTRY_JSON_STRLEN);
         write_char(jw, '"');
         write_char(jw, ':');
         jw->last_was_key = true;
--- a/src/sentry_json.c
+++ b/src/sentry_json.c
@@ -470,7 +470,7 @@ read_escaped_unicode_char(const char *buf)
 }
 
 static bool
-decode_string_inplace(char *buf)
+decode_string_inplace(char *buf, size_t *len_out)
 {
     const char *input = buf;
     char *output = buf;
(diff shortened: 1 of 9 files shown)


src/sentry_json.c
```diff
@@ -628,7 +641,7 @@ tokens_to_value(jsmntok_t *tokens, size_t token_count, const char *buf,
 
             char *key = sentry__string_clone_n_unchecked(
                 buf + token->start, (size_t)(token->end - token->start));
-            if (decode_string_inplace(key)) {
+            if (decode_string_inplace(key, NULL)) {
```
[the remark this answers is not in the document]
jpnurmi: This is a known limitation of object keys, not a regression this change should address. `sentry_value` string values are now length-aware, but object keys are still represented and accessed as NUL-terminated strings throughout the public API. Supporting `\u0000` in keys would require a separate length-aware key representation, which is out of scope here.

src/sentry_value.c
```diff
@@ -1600,7 +1651,7 @@ sentry__value_new_uuid(const sentry_uuid_t *uuid)
     }
     sentry_uuid_as_string(uuid, buf);
     buf[36] = '\0';
-    return sentry__value_new_string_owned(buf);
+    return sentry__value_new_string_owned_n(buf, 36);
```
JoshuaMoelans: (suggestion): Since `sentry_value_new_string_n` already does a malloc to store a memcpy of the buffer, and we free the buffer in `sentry__value_new_string_owned_n` directly after, could we simplify this (and the other static-sized `new_internal_uuid` and `new_span_uuid`) to stack-allocate the buffer before passing them?
```suggestion
    char buf[37];
    sentry_uuid_as_string(uuid, buf);
    return sentry_value_new_string_n(buf, 36);
```

JoshuaMoelans (approved):
LGTM, small suggestion and two Qs:
- should we have a changelog entry since we preserve `NUL` bytes now, which could cause some confusion as to `sentry_value_get_length` returning blob->len (which accounts for these NULs), while `strlen(sentry_value_as_string(...))` will just return the C-string-length? IIUC the JSON/msgpack values we write will also contain these escape characters from now on, which are kinda bloating the envelopes we send. This was introduced afaict for the future attachments rework, but maybe this is something we should reconsider for strings (e.g., maybe on `value_new_string_n` we check where the first `\0` is and shorten the length we pass to `new_string_value`?)
- Should we update the [docstring](https://github.com/getsentry/sentry-native/blob/master/include/sentry.h#L473-L478) for `value_get_length` to also include mention of returning the length of strings?

src/sentry_value.c
jpnurmi: Thanks. Yeah, we should update them all wherever we can avoid allocations: https://github.com/getsentry/sentry-native/pull/1968/commits/123a625d8467f80688b61df5020f4722f73e8bb8

I had to also fix the `sentry__internal_uuid_as_string` declaration to match the expected array length...

```
/home/jpnurmi/Projects/sentry/sentry-native-blob/src/sentry_value.c: In function ‘sentry__value_new_internal_uuid’:
/home/jpnurmi/Projects/sentry/sentry-native-blob/src/sentry_value.c:1633:5: warning: ‘sentry__internal_uuid_as_string’ accessing 37 bytes in a region of size 33 [-Wstringop-overflow=]
 1633 |     sentry__internal_uuid_as_string(uuid, buf);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/jpnurmi/Projects/sentry/sentry-native-blob/src/sentry_value.c:1633:5: note: referencing argument 2 of type ‘char[37]’
In file included from /home/jpnurmi/Projects/sentry/sentry-native-blob/src/sentry_value.c:40:
/home/jpnurmi/Projects/sentry/sentry-native-blob/src/sentry_uuid.h:10:6: note: in a call to function ‘sentry__internal_uuid_as_string’
   10 | void sentry__internal_uuid_as_string(const sentry_uuid_t *uuid, char str[37]);
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

jpnurmi:
> LGTM, small suggestion and two Qs:
> 
> * should we have a changelog entry since we preserve `NUL` bytes now, which could cause some confusion as to `sentry_value_get_length` returning blob->len (which accounts for these NULs), while `strlen(sentry_value_as_string(...))` will just return the C-string-length? IIUC the JSON/msgpack values we write will also contain these escape characters from now on, which are kinda bloating the envelopes we send. This was introduced afaict for the future attachments rework, but maybe this is something we should reconsider for strings (e.g., maybe on `value_new_string_n` we check where the first `\0` is and shorten the length we pass to `new_string_value`?)

Changelog entry added in https://github.com/getsentry/sentry-native/pull/1968/commits/e95382a0402f98fa1ba4ed624fb53cd6de697699.

I'd expect all `_n` APIs to respect the explicitly given length and not quietly truncate by calling `strlen`. Arguably, it was a bug. :wink: Notice that the most commonly used `sentry_value_new_string` API works the same as before. The only difference is that we cache `strlen`.

What comes to attachments, `sentry_value_t` would be (ab)used as safe ref-counted storage. Those values are never directly serialized into events. We'd pass `sentry_value_t` instead of `sentry_attachment_t*` to `envelope_add_attachment` and it'd serialize the data the same way as before.

> * Should we update the [docstring](https://github.com/getsentry/sentry-native/blob/master/include/sentry.h#L473-L478) for `value_get_length` to also include mention of returning the length of strings?

Docs updated in https://github.com/getsentry/sentry-native/pull/1968/commits/668470ca0f34199146b885fff4ee35d032831406. `get_length` actually worked for strings before, it just doesn't need `strlen` anymore.

PR #1980, opened 2026-08-12
feat!: remove enable_logs/metrics

jpnurmi:
> [!IMPORTANT]
> ### BREAKING CHANGE
> Remove `sentry_options_get/set_enable_logs` and `sentry_options_get/set_enable_metrics`
>
>> Structured logs and metrics have been enabled by default since `0.13`, and the options were deprecated and made no-ops in `0.16`.
>>
>> We recognize that this change may inconvenience applications that rely on the opt-out. Use `sentry_options_set_before_send_log` or `sentry_options_set_before_send_metric` to filter logs or metrics. We made this tradeoff deliberately because consistent behavior across SDK integrations will help most users successfully adopt these features.


limbonaut:
Consensus changed towards making `enable_logs` noop, and removing it in the next major. Although, from compile-time safety perspective, it's less important for native.

src/sentry_telemetry.c
```diff
@@ -7,21 +7,13 @@
...
 sentry__telemetry_startup(const sentry_options_t *options)
 {
-    if (options->enable_logs) {
-        sentry__logs_startup(options);
-    }
-    if (options->enable_metrics) {
-        sentry__metrics_startup(options);
-    }
+    sentry__logs_startup(options);
+    sentry__metrics_startup(options);
```
JoshuaMoelans: one of my only concerns here is starting the batcher threads for every SDK user now, even if they never call a single log/metric API. but think there's probably no real way around this 🤔

JoshuaMoelans:
I created a follow-up docs PR https://github.com/getsentry/sentry-docs/pull/19041 , to be merged after we release this one

src/sentry_telemetry.c
jpnurmi: Yeah, these unconditional threads are a bit annoying, but fortunately quite cheap because they don't do anything besides waiting on the condition variable.

Perhaps lazy batcher creation/start could be one option? It might introduce some latency for the first log/metric call, but it would be justified because it speeds up SDK init for everyone, and moves the cost for those who actually use logs and metrics...

JoshuaMoelans: I had a prototype of this approach in https://github.com/getsentry/sentry-native/pull/1611 that we ended up not doing back then, but might be worth reconsidering 👀

limbonaut (approved):
Nice!

mujacica (approved):

JoshuaMoelans (approved):

PR #1987, opened 2026-08-12
feat(native): add public custom HTTP transport client interface

### LINKED ISSUE #1979: Custom HTTP transport client interface
## Problem

The public custom transport API only allows implementing a generic envelope callback function through `sentry_transport_new`. Once an application replaces the built-in HTTP transport this way, it also bypasses the HTTP transport’s shared behavior, including:

- HTTP retry and exponential backoff configured with `sentry_options_set_http_retry`
- Offline caching configured with `sentry_options_set_cache_keep`
- Rate-limit handling
- Client report attachment and restoration
- The standard background worker, flushing, and shutdown behavior

This makes it difficult for downstream SDKs to use platform- or framework-native HTTP clients (.NET, Dart, Qt, ...) without reimplementing a substantial part of `sentry-native`’s HTTP transport logic.

## Proposed solution

Expose a supported public API for creating a custom HTTP transports, in addition to the existing generic function transport API.

The new API should allow users to implement the HTTP "client" interface currently used internally by `sentry_http_transport`. The custom client would be responsible only for executing HTTP requests and returning  HTTP responses (status code & headers).

The shared HTTP transport would remain responsible for:

- Preparing and serializing requests
- Queueing work and preserving envelope order
- HTTP retry and exponential backoff
- Offline caching and cache cleanup
- Rate-limit handling
- Client reports
- Flush and shutdown behavior

Conceptually, this would be a public, stable equivalent of the existing internal `sentry__http_transport_new` client interface and its lifecycle callbacks.

## Notes

The API design should account for potential future multi-threading.

The HTTP transport currently runs on a single transport thread. With recently introduced support for large attachments, it may eventually be useful to be able to process requests using multiple transport threads, to avoid blocking the transport pipeline with a long-lasting large attachment upload.

In a multi-threaded transport model, each transport thread would need its own HTTP client instance. The public interface should therefore avoid assuming that one client instance is permanently shared by the entire transport. For example, it could expose a client factory with per-thread startup, shutdown, and cleanup callbacks.

Thread-safety, request ordering, and client ownership should be documented so that adding multiple transport threads later does not require a breaking API change.

HuzaifaAbdulRehman:
Closes #1979.

## Problem

The only public custom transport API today is `sentry_transport_new`, which takes a raw envelope-send function. Using it means giving up everything the built-in HTTP transport already provides: retry with exponential backoff, offline caching, rate-limit handling, client report attachment, and the background worker/flush/shutdown machinery. Downstream SDKs that want to use a platform-native HTTP client (Qt, .NET, Dart, ...) currently have to reimplement all of that themselves.

## Solution

Adds `sentry_http_transport_new`, a public equivalent of the internal `sentry__http_transport_new` client interface. A custom HTTP client only has to execute a request and report back the status code and headers; sentry-native keeps owning request preparation/serialization, queueing, envelope ordering, retry/backoff, offline caching, rate-limit handling, client reports, and flush/shutdown.

New public surface:

- `sentry_http_request_t` / `sentry_http_response_t`, opaque handles, following the same forward-declared-in-`sentry.h` pattern already used for `sentry_transport_t`.
- Request accessors: `get_method`, `get_url`, `get_header_count`/`get_header`, `get_body`, `get_body_file_path` (the last one covers TUS large-attachment streaming).
- Response setters: `set_status_code`, `set_header`.
- `sentry_http_client_factory_func_t` + `sentry_http_transport_new(factory, factory_data, send_func, client_free_func)`.
- `sentry_http_transport_set_client_start_func` / `set_client_shutdown_func`.

All marked `SENTRY_EXPERIMENTAL_API`.

### Following up on the design discussion in the issue

- **Header parsing centralization**: `sentry_http_response_set_header` matches the signature you proposed. curl and WinHTTP no longer parse `retry-after`/`x-sentry-rate-limits`/`location` independently, they both funnel through this one function now, so they're effectively reference implementations of the public interface rather than a separate code path.
- **Factory over shared instance**: `sentry_http_client_factory_func_t` is called once today, but nothing in its signature assumes that. A future multi-threaded transport could call it once per worker thread without a breaking change.
- **Thread-safety/ownership/ordering**: documented directly on the factory and send-function doc comments, since that was called out as a prerequisite before finalizing the shape.

### A note on client ownership

Initially I had `client_free_func` as a separate post-construction setter (mirroring the existing `start`/`shutdown` hooks). That leaves a leak window: if `factory()` succeeds but transport construction then fails, there's no free hook registered yet to clean up the client. Fixed by folding `client_free_func` into the `sentry_http_transport_new` constructor itself, matching how curl.c/winhttp.c already co-locate client creation with their free function.

## Testing

- 15 new unit tests covering the header setter, the status setter, request accessors (in-memory and file-backed bodies), and the factory constructor (success, ownership verification, null-argument handling, factory failure).
- Full existing test suite passes under both backends, including against a real curl-linked binary, confirming curl/WinHTTP behavior is unchanged.
- Formatting checked against the repo's pinned clang-format (20.1.5) via `scripts/check-clang-format.py`, clean.

## Checklist

- [x] Code formatting checked (clang-format, matching `make style`)
- [x] Tests run and passing
- [x] New unit tests added for the change
- [x] CHANGELOG.md entry added


include/sentry.h
```diff
@@ -972,6 +972,187 @@ SENTRY_EXPERIMENTAL_API void sentry_transport_retry(
...
+SENTRY_EXPERIMENTAL_API int sentry_http_request_get_header(
+    const sentry_http_request_t *req, size_t index, const char **key_out,
+    const char **value_out);
+
+/**
+ * Returns the in-memory request body of `req` and writes its length to
+ * `*len_out`. Returns `NULL` if `req` has no in-memory body, which happens
+ * for large-attachment uploads that stream a file instead -- see
+ * `sentry_http_request_get_body_file_path`. Exactly one of the two ever
+ * returns non-`NULL` for a given request.
```
jpnurmi: The exactly-one guarantee is false for the initial TUS creation `POST`, which leaves both `body` and `body_path` null:

https://github.com/getsentry/sentry-native/blob/2d9e98ab303d0d949e28f63f4fb739c405ab453d/src/transports/sentry_http_transport.c#L509-L515

A custom client following this contract may try to open a null file path and fail large-attachment uploads. Document that both accessors may return `NULL` for bodyless requests, and cover that case in the accessor tests.

include/sentry.h
```diff
@@ -972,6 +972,187 @@ SENTRY_EXPERIMENTAL_API void sentry_transport_retry(
...
+
+/**
+ * Marks `resp` as having failed because the transport is shutting down,
+ * rather than because of a regular request error. Only meaningful when the
+ * client's `sentry_http_client_send_func_t` call is about to return `0`
+ * (failure); sentry-native uses this to skip retry/caching for requests
+ * that failed only because shutdown interrupted them.
+ */
+SENTRY_EXPERIMENTAL_API void sentry_http_response_set_shutdown(
+    sentry_http_response_t *resp, int is_shutdown);
```
jpnurmi: Hmm, this looks like a flag that should probably stay internal, or rather be moved from the HTTP client to the HTTP transport host, if possible. There's a potential danger of introducing shutdown races, though.

Not sure how this even ended up in the client layer. The author, \*cough\*, probably didn't care because it was internal... :grin: Anyway, I _think_ individual HTTP clients shouldn't have to communicate it back to the HTTP transport.

include/sentry.h
```diff
@@ -972,6 +972,187 @@ SENTRY_EXPERIMENTAL_API void sentry_transport_retry(
...
+ * `sentry_http_transport_new`, unchanged and still owned by the caller.
+ *
+ * sentry-native currently calls this exactly once per transport, on the
+ * thread that calls `sentry_http_transport_new`, to create the single
+ * client the transport's background thread will use for every request. A
+ * future multi-threaded transport may call this once per worker thread
+ * instead, each time from that worker thread; implementations should not
+ * assume they are called from any particular thread, or exactly once.
+ */
+typedef void *(*sentry_http_client_factory_func_t)(void *factory_data);
```
jpnurmi: Idea: what about a cosmetic `sentry_http_client_t` typedef to help distinguish between factory/user data and the client, which are both void pointers? The typedef could slightly improve readability by making it stand out more clearly which void pointers are client pointers.

```c
typedef void sentry_http_client_t;

/* ... */
typedef sentry_http_client_t *(*sentry_http_client_factory_func_t)(
    void *factory_data);

/* ... */
typedef int (*sentry_http_client_send_func_t)(sentry_http_client_t *client,
    sentry_http_request_t *req, sentry_http_response_t *resp);

/* ... */
```

src/transports/sentry_http_transport.c
```diff
@@ -960,6 +1071,44 @@ sentry__http_transport_set_shutdown_client(
...
+{
+    if (!factory || !send_func) {
+        return NULL;
+    }
+    void *client = factory(factory_data);
+    if (!client) {
+        return NULL;
+    }
+    sentry_transport_t *transport
+        = sentry__http_transport_new(client, send_func);
```
jpnurmi, line 1107: Should we get rid of `sentry__http_transport_new` and migrate `curl` and `winhttp` to use `sentry_http_transport_new` instead to back the claim that they are reference implementations of the interface? :)

include/sentry.h
```diff
@@ -972,6 +972,187 @@ SENTRY_EXPERIMENTAL_API void sentry_transport_retry(
...
+
+/**
+ * Represents a single HTTP request prepared by sentry-native, to be
+ * executed by a custom HTTP client set up via `sentry_http_transport_new`.
+ *
+ * The pointer is only valid for the duration of the
+ * `sentry_http_client_send_func_t` call it was passed to; the client must
+ * not retain it afterwards.
+ */
+struct sentry_prepared_http_request_s;
```
jpnurmi: Nitpick: I'd rename the `sentry_prepared_http_request_s` tag to `sentry_http_request_s` while at it, to avoid "leaking" the old internal name in the public API. The rename will cause a bit extra churn, but I think it's worth it to leave it in a clean and consistent state.

include/sentry.h
```diff
@@ -972,6 +972,187 @@ SENTRY_EXPERIMENTAL_API void sentry_transport_retry(
...
+ * on success; a non-zero return bubbles up to `sentry_init`.
+ */
+SENTRY_EXPERIMENTAL_API void sentry_http_transport_set_client_start_func(
+    sentry_transport_t *transport,
+    int (*start_func)(void *client, const sentry_options_t *options));
+
+/**
+ * Sets the hook that tells the client the transport is shutting down, e.g.
+ * to unblock or cancel an in-flight request. Called at most once, when the
+ * transport is shut down.
```
jpnurmi, line 1175: There should be a warning that this is called from the thread that calls `sentry_close`, which is not the same thread that runs the HTTP transport and client.

include/sentry.h
```diff
@@ -972,6 +972,187 @@ SENTRY_EXPERIMENTAL_API void sentry_transport_retry(
...
+ * sentry-native's retry/caching logic; call
+ * `sentry_http_response_set_shutdown` first if the failure is because the
+ * transport is shutting down rather than a request error.
+ *
+ * sentry-native calls this from a single background thread, once per
+ * request, in the order requests were queued, and never invokes it again
+ * for the same client until the previous call returns. A future
+ * multi-threaded transport may drive different client instances from
+ * different threads concurrently, but will still only ever call into one
+ * client instance from one thread at a time.
```
jpnurmi: Actually, the shutdown hook is called from the thread that calls `sentry_close`, which unfortunately imposes a nasty requirement that the client must be thread-safe even in the current single-instance use...

jpnurmi (commented):
Nice, thanks! Awesome start, looking good so far. I left a few suggestions and discussion points.

include/sentry.h
HuzaifaAbdulRehman: Oh good catch, totally missed that case. Updated the docs to say both can come back null for bodyless requests like the TUS creation POST, and added a test for it.

include/sentry.h
HuzaifaAbdulRehman: Yeah that's fair. I think your comment below kind of already answers it though, since the shutdown hook can fire on a different thread while send_func is still running, clients already have to deal with being called from two threads at once anyway. So having them also set shutdown on the response from inside send_func doesn't really add anything new on top of that.

I don't think we can move this fully to the transport side though. When http_transport_shutdown starts, it doesn't actually know yet whether the send_func call that's in flight is going to fail because of shutdown or for some other reason. Only the client really knows why it failed.

Maybe an alternative could work: have the transport set an internal flag when shutdown starts, and if send_func fails after that point just treat it as shutdown no matter what the client says. That would let us drop sentry_http_response_set_shutdown entirely. Downside is a request that happens to genuinely fail for some other reason right as shutdown kicks in would get lumped in as a shutdown failure too, and skip retry/caching it maybe should've gotten. Client-reported is more accurate but it does put more burden on every client to be thread safe. 

Let me know which way you'd rather go, I'm fine either way, just want to build the right one.

include/sentry.h
HuzaifaAbdulRehman: Yeah that's nicer, added it in.

src/transports/sentry_http_transport.c
HuzaifaAbdulRehman: I like the idea, would be a nice way to actually back up the "reference implementation" claim lol. I'd rather do it as a separate PR though if that's ok? This one's already pretty big with the new types plus the header parsing changes, and moving curl/winhttp onto the new factory/send func shape is its own chunk of work since their client lifecycle stuff would need to change quite a bit. Can open a follow-up for it once this lands if that works for you.

include/sentry.h
HuzaifaAbdulRehman: Agreed, renamed it. Wasn't too bad since it's just the one tag.

include/sentry.h
HuzaifaAbdulRehman: Oh good catch, you're right, checked it and yeah, the shutdown func gets called from whatever thread calls sentry_close, not the background thread that's actually sending stuff. Not the same thread at all. Updated the docs to say that now, and also fixed the other comment that claimed it's only ever "one thread at a time" since that's not quite true because of this case. Thanks for catching this, would've been a nasty one for someone to run into

include/sentry.h
HuzaifaAbdulRehman: Added a warning about that on the doc comment, thanks.

src/transports/sentry_http_transport.c
jpnurmi: Sure, a separate follow-up PR is fine. It could even be an opportunity to use GitHub's new stacked PRs :)

src/transports/sentry_http_transport.h
```diff
@@ -12,7 +12,11 @@ typedef struct sentry_prepared_http_header_s {
     char *value;
 } sentry_prepared_http_header_t;
 
-typedef struct sentry_prepared_http_request_s {
+// The public `sentry_http_request_t` (see sentry.h) is a forward-declared
+// opaque handle for this same tag; this is the only place its layout is
+// defined. Internal code and the public accessors in sentry_http_transport.c
+// both operate on this concrete struct via that shared tag.
```
jpnurmi: Nitpick: feels unnecessary to write it out here because the same pattern is used for many structs (envelope, options, transport, scope, ...) across the codebase. 😉

src/transports/sentry_http_transport.h
```diff
@@ -35,23 +39,29 @@ sentry_prepared_http_request_t *sentry__prepare_tus_upload_request(
 
 void sentry__prepared_http_request_free(sentry_prepared_http_request_t *req);
 
-typedef struct {
+// The public `sentry_http_response_t` (see sentry.h) is a forward-declared
+// opaque handle for this same tag; this is the only place its layout is
+// defined. Internal code and the public accessors in sentry_http_transport.c
+// both operate on this concrete struct via that shared tag.
```
jpnurmi: Same here, I think we can leave the comment out.

jpnurmi (commented):
Very nice, LGTM 👍 

I'll give it a try with Qt just to see how it feels and that we didn't miss anything

src/transports/sentry_http_transport.h
HuzaifaAbdulRehman: Fair enough, removed it.

src/transports/sentry_http_transport.h
HuzaifaAbdulRehman: Removed this one too.

jpnurmi:
> I'll give it a try with Qt just to see how it feels and that we didn't miss anything

The only thing that stands out is that our `send_func` is synchronous vs. Qt's networking APIs are asynchronous. Therefore, a hypothetical Qt-based `send_func` implementation needs a bit of an annoying event loop trick to hold off until the reply is received:

```c++
static int qt_http_client_send(sentry_http_client_t *client, sentry_http_request_t *request, sentry_http_response_t *response)
{
    QNetworkAccessManager *nam = static_cast<QtHttpClient *>(client)->nam;
    QNetworkReply *reply = nam->sendCustomRequest(...);

    QEventLoop loop;
    QObject::connect(reply, &QNetworkReply::finished, &loop, &QEventLoop::quit);
    loop.exec();

    auto status = reply->attribute(QNetworkRequest::HttpStatusCodeAttribute);
    sentry_http_response_set_status_code(response, status.toInt());

    ...
}
```

However, I'm hesitant to complicate things and split the flow into separate
- sentry-native → client: request
- client → sentry-native: response

calls because it would create a whole range of new issues... :thinking:

HuzaifaAbdulRehman:
Yeah I ran into the same thought. I'd still keep it synchronous though. The retry, ordering and caching all depend on knowing how a request went before deciding what's next, and splitting it up would mean rebuilding all that around pending state. curl and winhttp are both synchronous too, so it fits the rest of the code.

The Qt side might be less scary than it looks though, since send_func runs on the background worker thread, not the main one. So that nested QEventLoop is spinning on a thread with no UI and no user code, which avoids most of the usual re-entrancy headaches.

One gotcha I noticed while checking: the factory and the start hook both run on the calling thread, but send_func runs on the worker thread. So a QNetworkAccessManager created in either of those would end up with the wrong thread affinity. Creating it lazily on the first send_func call should sort it.

Happy to document that thread split if you think it's worth calling out, feels like an easy one to trip over.

jpnurmi:
I think the current flow is ok. Let's ignore the Qt-related finding - not a real issue.

The only remaining issue is the dubious shutdown flag, which I'm frankly not quite sure what to do with...

include/sentry.h
jpnurmi: > Maybe an alternative could work: have the transport set an internal flag when shutdown starts, and if send_func fails after that point just treat it as shutdown no matter what the client says. That would let us drop sentry_http_response_set_shutdown entirely. Downside is a request that happens to genuinely fail for some other reason right as shutdown kicks in would get lumped in as a shutdown failure too, and skip retry/caching it maybe should've gotten. Client-reported is more accurate but it does put more burden on every client to be thread safe.

Sounds like worth exploring. Wanna give it a try and let's see where it brings us? :pray: (Sorry for the little hole in the plan)

HuzaifaAbdulRehman:
Happy to just drop it from the public API if you're unsure. Easier to add it back later than take it away, and it means clients don't have to think about it at all.

I had a look at what the flag actually does and it's not much. It only really changes things for envelopes with attachment refs, and even then the envelope still gets persisted either way, just through a different path. So if the transport just assumed any failure after shutdown starts is a shutdown failure, I don't think we'd lose anything meaningful.

Would be a small change, just an atomic flag on the transport state like the bgworker already does. Want me to push that, or leave it as is?

jpnurmi:
> Would be a small change, just an atomic flag on the transport state like the bgworker already does. Want me to push that, or leave it as is?

Sure, go for it. And thanks for willing to iterate, very much appreciated!

HuzaifaAbdulRehman:
Pushed it. sentry_http_response_set_shutdown is gone from the public API, the transport now sets an atomic flag at the start of http_transport_shutdown and treats any send_func failure after that as a shutdown failure. Set before the client is ever asked to stop, so there's no window where a shutdown-caused failure gets read as a normal error.

Also rebased on master to pick up the changelog fix.

src/transports/sentry_http_transport.c
```diff
@@ -330,6 +334,107 @@ http_response_cleanup(sentry_http_response_t *resp)
...
+        *len_out = req->body_len;
+    }
+    return req->body;
+}
+
+const char *
+sentry_http_request_get_body_file_path(
+    const sentry_http_request_t *req, size_t *len_out)
+{
+    if (!req || !req->body_path) {
```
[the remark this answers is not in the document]
HuzaifaAbdulRehman: Good catch, this one's real. On Windows sentry_path_t keeps both the UTF-8 narrow path and the wide one, and the SDK's rule is to use the wide Win32 APIs at the boundary since the narrow ones read char* as ANSI rather than UTF-8. Both built-in transports already do that, CreateFileW in winhttp and _wfopen in curl, so a custom client had no way to match them and would break on non-ASCII paths.

Added sentry_http_request_get_body_file_pathw following the same w suffix convention as sentry_options_set_database_pathw and friends, and documented the encoding on the narrow one.

src/transports/sentry_http_transport.c
```diff
@@ -960,6 +1091,45 @@ sentry__http_transport_set_shutdown_client(
...
+    }
+    sentry__http_transport_set_free_client(transport, client_free_func);
+    return transport;
+}
+
+void
+sentry_http_transport_set_client_start_func(sentry_transport_t *transport,
+    int (*start_func)(
+        sentry_http_client_t *client, const sentry_options_t *options))
+{
```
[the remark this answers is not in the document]
HuzaifaAbdulRehman: The mechanism is right but I think this is working as intended, it's the tradeoff we landed on with @jpnurmi above when moving this out of the client layer.

One thing though, there isn't really an immediate retry to lose here. sentry__retry_shutdown runs at the top of http_transport_shutdown and drops the pending poll, so the compare_swap from POLL_IDLE in retry_enqueue fails and nothing gets scheduled anyway. Both paths just persist to disk for the next run. It also only applies to envelopes with attachment refs, since that branch is gated on ref_paths being non-empty.

Setting the flag before interrupting the client is deliberate as well. The other order is the risky one, since a request aborted by the shutdown hook could return before the flag was visible and look like a normal network error.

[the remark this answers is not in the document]
jpnurmi: Agreed, it's an acceptable tradeoff

jpnurmi (approved):
This looks great, thank you so much for contributing a significant feature!

This will become very handy for downstream SDKs (e.g. https://github.com/getsentry/sentry-dart/issues/3551) wishing to replace the curl dependency without losing all the functionality provided by the built-in HTTP transport. 🎉 

@JoshuaMoelans would appreciate 👀 if you have time 🙏

HuzaifaAbdulRehman:
Thanks! Glad it'll be useful for the Dart side, I hadn't seen that issue before. Happy to address anything Joshua flags.

mujacica (approved):

PR #1988, opened 2026-08-13
fix(logging): drop noisy envelope-serialization debug log

tustanivsky:
The `serializing envelope into buffer` debug line is emitted on every envelope serialization. Since #1841 the crashpad backend serializes the external crash report envelope on every scope mutation (`flush_external_crash_report` → `sentry_envelope_serialize` → `sentry__envelope_serialize_into_stringbuilder`), so with debug logging enabled this floods the logs. It was reported by an Unreal SDK user who wants to keep debug logging on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 **Fixes**:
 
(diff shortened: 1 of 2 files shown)


jpnurmi (approved):
Sorry about the accidental debug flood and thanks for fixing it!

PR #1993, opened 2026-08-14
feat: add `set_tags` for updating multiple tags in one go

jpnurmi:
Makes a significant difference in terms of performance when updating multiple tags, because out-of-process handlers only need to flush the scope once.


JoshuaMoelans (approved):
very nice, simple, yet elegant improvement 🚀  did you happen to benchmark how much this enhances performance 👀

jpnurmi:
> very nice, simple, yet elegant improvement 🚀 did you happen to benchmark how much this enhances performance 👀

Very rough numbers measured on Windows using `QueryPerformanceFrequency` + `QueryPerformanceCounter`:

### Before (~2-4 ms)
```c
sentry_set_tag("foo", "111");
sentry_set_tag("bar", "222");
sentry_set_tag("baz", "333");
sentry_set_tag("qux", "444");
```

### After (~0.5-1 ms)
```c
sentry_value_t tags = sentry_value_new_object();
sentry_value_set_by_key(tags, "foo", sentry_value_new_string("111"));
sentry_value_set_by_key(tags, "bar", sentry_value_new_string("222"));
sentry_value_set_by_key(tags, "baz", sentry_value_new_string("333"));
sentry_value_set_by_key(tags, "qux", sentry_value_new_string("444"));
sentry_set_tags(tags);
```

* with external crash reporter enabled, causing extra work during scope flush

jpnurmi:
Q: `set_tags` vs. `update_tags`? 

- sentry-unreal has `USentryScope::SetTags`
- sentry-native has `update_context` which is not the same thing but there's some similarity

tustanivsky (approved):

jpnurmi:
> Q: `set_tags` vs. `update_tags`?
> 
> * sentry-unreal has `USentryScope::SetTags`
> * sentry-native has `update_context` which is not the same thing but there's some similarity

let's keep `set_tags` because it calls `set_tag` in a loop, so it's literally a plural version of `set_tag`

PR #2000, opened 2026-08-18
chore: deprecate enable_logs/metrics

jpnurmi:
A softer alternative to #1980, to be released with the current features and fixes. This provides a heads-up for the actual removal coming soon, together with the other upcoming breaking changes (e.g. value-based attachments for RW-locking).


src/sentry_logs.c
```diff
@@ -14,6 +14,17 @@
 
 static sentry_batcher_ref_t g_batcher = SENTRY_BATCHER_REF_INIT;
 
+static bool
+sdk_is_initialized(void)
```
limbonaut: Other SDKs usually provide `is_enabled()` to check whether it's initialized. Just an idea, and probably out of scope for this PR.

limbonaut (commented):
The more I think about it, the more I like the previous version, because it's more honest. If someone calls `sentry_options_set_enable_logs(options, false)` on this version, they'll get a warning, but application will compile and run, and produce a different result than before. It's a sneaky footgun.

jpnurmi:
> The more I think about it, the more I like the previous version, because it's more honest. If someone calls `sentry_options_set_enable_logs(options, false)` on this version, they'll get a warning, but application will compile and run, and produce a different result than before. It's a sneaky footgun.

Good point! A bare deprecation without ripping out the functionality just yet makes more sense, indeed.

limbonaut (approved):

mujacica (approved):

PR #2009, opened 2026-08-20
ref: add sentry_attachment_t accessors

jpnurmi:
This is a **non-breaking** preparation step for the breaking stacked PR:
- #1974

This adds internal accessors for attachment properties, reducing direct access to the `sentry_attachment_s` data structure and moving mechanical churn out of #1974. This makes the large value-based attachment migration slightly smaller, easier to review, and less conflict-prone without changing behavior, and helps with partial pre-migration of console repos.

#skip-changelog (internal)

### DIFF
--- a/src/sentry_envelope.c
+++ b/src/sentry_envelope.c
@@ -791,7 +791,7 @@ sentry_envelope_item_t *
 sentry__envelope_add_from_path(
     sentry_envelope_t *envelope, const sentry_path_t *path, const char *type)
 {
-    if (!envelope) {
+    if (!envelope || !path) {
         return NULL;
     }
     size_t buf_len;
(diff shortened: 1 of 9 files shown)


src/sentry_envelope.c
```diff
@@ -729,24 +729,29 @@ sentry__envelope_add_attachment(
...
-    if (attachment->buf) {
+    if (bytes) {
         item = sentry__envelope_add_from_buffer(
-            envelope, attachment->buf, attachment->buf_len, "attachment");
+            envelope, bytes, bytes_len, "attachment");
     } else {
-        item = sentry__envelope_add_from_path(
-            envelope, attachment->path, "attachment");
+        sentry_path_t *path = sentry__attachment_make_path(attachment);
+        item = sentry__envelope_add_from_path(envelope, path, "attachment");
```
limbonaut, line 740: Should we check if path allocation was successful?

limbonaut (approved):
Looks good! Just need a null check in one place.

PR #2011, opened 2026-08-21
feat(scope): add missing public mutators

jpnurmi:
Make the scope API more symmetric with the global scope and scope observer APIs by adding the following missing scope-level mutators: `set_release`, `set_environment`, `set_transaction`, `remove_tag`, `remove_extra`, `remove_context`, and `remove_attachment`.

Console SDKs currently update some of these fields through direct private `sentry_scope_s` access, coupling them with internal implementation details, and making scope refactoring (RW-locks) harder than it needs to be. :)


mujacica (approved):

limbonaut (approved):
Sweet! 🍻

PR #2028, opened 2026-08-26
fix(unwind): prevent section table size overflow

jpnurmi:
Calculate malformed ELF section table bounds without overflowing the promoted header fields or their offset sum.

Fixes: https://github.com/getsentry/sentry-native/security/code-scanning/5

> [!WARNING]
> ### Multiplication result converted to larger type
> Multiplication result may overflow 'int' before it is converted to 'unsigned long'.

https://codeql.github.com/codeql-query-help/cpp/cpp-integer-multiplication-cast-to-long/


jpnurmi:
Submitted upstream instead:
- https://github.com/libunwind/libunwind/pull/1044

PR #2032, opened 2026-08-28
feat(value): add foreach APIs for lists and objects

### LINKED ISSUE #2017: Add ability to iterate object keys
**Context**: In the Godot SDK, we need to provide access to event contexts carried by events inside `before-send` handler. The Godot's API would look like this: `SentryEvent.get_context("os")`. With `sentry_value_t` object, we can only ask for a specific value by key, but in order to get all the data for specific context, we need to know which keys that context carries. Ideally, we should be able to iterate over key-value pairs contained in the given object.

**Proposal**: add API to iterate object key-value pairs (or just the keys).

jpnurmi:
Add interruptible callbacks for visiting the values of a list and the key/value pairs of an object. A non-zero callback return stops iteration, and the same value is returned to the caller.

Note: callbacks may modify the container freely. Those changes are applied, but the current iteration continues over the entries present when it began.

Close: #2017


jpnurmi:
@limbonaut Would this kind of `foreach` solution work for sentry-godot? I was hoping to get away without implementing a full-blown iterator API. :sweat_smile:

limbonaut (approved):
Nice! Visitor will totally work for what we need in Godot.

mujacica (approved):

PR #2038, opened 2026-09-01
feat: Add per-event level setter

jpnurmi:
The Native ["Set the Level"](https://docs.sentry.io/platforms/native/usage/set-level/) docs say the level can be overridden within an event, but the example calls `sentry_set_level`. That API sets a scope default; there is no API for assigning a level to a specific event, while native crash events are always assigned a fatal level.

The [Godot guide](https://docs.sentry.io/platforms/godot/usage/set-level/) demonstrates the intended per-event flow: create an event, assign its level, and capture it. sentry-unreal and sentry-godot both expose this operation while constructing events or pre-processing them in hooks, but currently implement it by writing the protocol "level" field manually.

Add `sentry_event_set_level` to support that operation directly and clarify its distinction from the scope default. This also gives the Native guide a per-event API to use in its example:

```c
#include <sentry.h>

sentry_event_set_level(event, SENTRY_LEVEL_WARNING);
sentry_capture_event(event);
```


JoshuaMoelans (approved):

mujacica (approved):

mujacica:
@jpnurmi - nice, can we also extend the docs with the explanation now?

PR #2045, opened 2026-09-02
feat: add `sentry_is_enabled()`

jpnurmi:
Adds `sentry_is_enabled()` as a thread-safe public API for checking whether the SDK has completed initialization and has not yet been closed. It returns `0` before initialization, after initialization failure, and after `sentry_close()`.

This is inspired by many other Sentry SDKs, and motivated by #1980, which needs a way to determine whether the SDK is initialized to avoid duplicating private state checks in the logging and metrics implementations.

_Originally proposed by @limbonaut in https://github.com/getsentry/sentry-native/pull/2000#discussion_r3806903122_

> Other SDKs usually provide `is_enabled()` to check whether it's initialized. Just an idea, and probably out of scope for this PR.


JoshuaMoelans (approved):
nice addition 🚀

limbonaut (approved):

PR #2052, opened 2026-09-02
fix(native): read frame records at pointer width in the crash daemon's FP walk

GLinnik21:
## Summary

`read_stack_value` in the native backend's crash daemon bounds-checks and copies `sizeof(uint64_t)` bytes regardless of the target's pointer size, and the frame-pointer walk steps to the return address with the same constant. On a 32-bit target every read therefore spans two stack slots: the high half of each `saved_fp` / `return_addr` is the neighbouring word, the return address is read from the wrong slot, and the last legitimate frame record fails the bounds check.

This reads a `uintptr_t` and widens it to `uint64_t` (so the value is endian-independent), and steps by pointer size. On 64-bit targets the size and offsets are unchanged. While here, the range check is done by subtraction rather than `addr + size`, since `addr` is a frame pointer from a crashed process and a corrupted value near the top of the address space would wrap the sum past the naive comparison.

## Scope

This is the daemon's frame-pointer **fallback**. On Linux the signal handler captures a libunwind backtrace first and `build_stacktrace_for_thread` returns that when it is non-empty; the FP walk runs when that capture is unavailable or empty (for example a build that disables in-process libunwind, or `unw_init_local2` failing). Within that fallback, the pointer-width reads are wrong on every 32-bit target. On ARM32 this alone is not sufficient for GCC-built frames — see #2053, which is stacked on this PR.

## Verification

- ARM32 hardware (LG webOS 4.x television, Cortex-A9, glibc 2.24, in-process libunwind disabled so the FP fallback is exercised): with this fix alone, a SIGABRT in a rustc-built application walks `gsignal ← plex_run ← __libc_start_main` (3 named frames); a GCC-built chain with frame pointers still stops after one real frame followed by a stack address, which is the record-shape problem #2053 addresses.
- `sentry-crash` cross-compiles for `arm-linux-gnueabi` (GCC 12); the SDK and daemon build on macOS arm64 with `SENTRY_BACKEND=native`; `clang-format` clean.
- No automated 32-bit coverage of the native backend runs in CI today (the ARM32 job runs under `TEST_QEMU`, which skips the native integration suite), and `read_stack_value` is a static function inside the daemon with no unit-test seam, so this PR adds no test. Happy to add one if there is a preferred shape for daemon-internal tests.


GLinnik21:
Addressed the review: `read_stack_value` now range-checks by subtraction (no `addr + size` that could wrap for a corrupted frame pointer) and reads into a `uintptr_t` before widening to `uint64_t`, so the value is endian-independent. Re-verified on ARM32 hardware; macOS native-backend build and `arm-linux-gnueabi` cross-compile still clean.

jpnurmi (approved):
Thank you! LGTM :+1:

PR #2055, opened 2026-09-03
fix(minidump): use sentry__elf_find_note for build ids

HuzaifaAbdulRehman:
#1773 switched the crash daemon and module finder to `sentry__elf_find_note`. `extract_elf_build_id` in the Linux minidump writer still did `(n_descsz + 3) & ~3` on a `uint32_t`, so the aligned size can wrap to 0.

This uses the same helper.

I ran `make style` and `make test-unit` in WSL (Ubuntu). Unit tests passed, including `elf_find_note`. I did not run the full pytest integration suite.


src/backends/native/minidump/sentry_minidump_linux.c
```diff
@@ -936,45 +936,21 @@ extract_elf_build_id(const char *elf_path, uint8_t *build_id, size_t max_len)
...
-                    }
-
-                    ptr += aligned_namesz;
-                    ptr += aligned_descsz;
+                // Find the GNU build-id note using the shared bounds-checked
+                // iterator. It performs overflow-safe arithmetic and verifies
+                // the descriptor fits within the buffer, unlike the previous
+                // inline walk whose `(n_descsz + 3) & ~3` alignment wrapped for
+                // near-UINT32_MAX sizes and let the memcpy read past the note.
+                // SHT_NOTE uses 4-byte alignment.
```
jpnurmi: I think it's unnecessary and potentially confusing (for a future reader without the historical context) to leave behind such a detailed description of the issue this change already fixed. Perhaps something like this is sufficient? 🙂 

```suggestion
                // Find the GNU build-id note. SHT_NOTE uses 4-byte alignment.
```

jpnurmi (commented):
Great finding, thank you! 🙏

src/backends/native/minidump/sentry_minidump_linux.c
HuzaifaAbdulRehman: Applied, thanks.

jpnurmi (approved):

PR #2073, opened 2026-09-07
fix(native): Resolve the WER module relative to `handler_path`

tustanivsky:
The native backend looks for `sentry-wer.dll` only next to the running executable. The crashpad backend resolves `crashpad_wer.dll` against `handler_path` instead, so an embedder that installs the handler elsewhere gets a working WER module on crashpad and a silent `"Native WER module not found"` on native.

In Unreal Engine plugin runtime dependencies are staged under the plugin's own `Binaries` directory, never next to the game executable:

```
Game/Binaries/Win64/Game.exe                        <- searched
Game/Plugins/sentry/Binaries/Win64/sentry-crash.exe <- handler_path
Game/Plugins/sentry/Binaries/Win64/sentry-wer.dll   <- ships with the handler
```

Since plugin doesn't place the DLL next to the executable WER-delivered crashes (`__fastfail`, heap corruption, stack buffer overrun) are lost on every native-backend build.

`wer_module_path()` now tries `handler_path`'s directory first and falls back to the executable directory, matching crashpad. Existence checking moves into `wer_path_in_dir()` so the fallback is actually reached when the handler directory holds no module.

Related items:
- https://github.com/getsentry/sentry-unreal/pull/1568


jpnurmi (approved):
Thanks!

PR #2083, opened 2026-09-10
ref: prevent direct access to scope data

jpnurmi:
Move scope data out of the header to prevent direct access, and provide internal API tailored for protecting the data with RW-locks in #2042. This offloads a lot of mechanical churn out of #2042, lets it focus on getting the RW-locks right, and allows migrating the console SDKs.

See also:
- https://github.com/getsentry/sentry-playstation/pull/171
- https://github.com/getsentry/sentry-switch/pull/190
- https://github.com/getsentry/sentry-xbox/pull/177

### DIFF
--- a/src/sentry_scope.c
+++ b/src/sentry_scope.c
@@ -1519,9 +1519,9 @@ sentry__scope_ref_span(const sentry_scope_t *scope)
 }
 
 sentry_value_t
-sentry__scope_ref_span_or_transaction(const sentry_scope_t *scope)
+sentry__scope_load_span_or_transaction(const sentry_scope_t *scope)
 {
-    return sentry_value_incref(get_span_or_transaction(scope));
+    return sentry__value_clone(get_span_or_transaction(scope));
 }
 
 void
--- a/src/sentry_scope.c
+++ b/src/sentry_scope.c
@@ -13,6 +13,7 @@
 #include "sentry_sync.h"
 #include "sentry_tracing.h"
 #include "sentry_transport.h"
+#include "sentry_uuid.h"
 #include "sentry_value.h"
 
 #include <stdlib.h>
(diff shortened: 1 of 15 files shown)


src/sentry_scope.c
```diff
@@ -1184,6 +1448,158 @@ sentry_scope_add_attachment(sentry_scope_t *scope, sentry_value_t attachment)
...
+    sentry_span_t *span = scope->data->span;
+    sentry__span_incref(span);
+    return span;
+}
+
+sentry_value_t
+sentry__scope_ref_span_or_transaction(const sentry_scope_t *scope)
+{
+    return sentry_value_incref(get_span_or_transaction(scope));
+}
```
limbonaut, line 1525: There is a trap here. A transaction consists of two allocations: wrapper `sentry_transaction_t` and inner `sentry_value_t`. The wrapper uses the inner's refcount to decide when to clean up (so it basically borrows its refcount mechanism). This new accessor returns a reference, but if the caller ever keeps it long enough to outlive the last transaction reference (e.g. across `finish()`) the subsequent value decref will not free the leaked transaction wrapper.

I realize that this function is only used in testing, as things currently stand. Would it be more prudent to expose a cloning function instead, to avoid this pitfall in the future?
```c
sentry__scope_load_span_or_transaction() {
    return sentry__value_clone(get_span_or_transaction(scope));
}
```

limbonaut (approved):
I went through the whole change set. It's well structured and mostly mechanical shift, but because of the sheer size I might have missed some tiny speck. Ran a scan too. Looks good with one place worth considering a small alteration in behavior to avoid a potential future trap.

src/sentry_scope.c
jpnurmi: Yeah, cloning is better. Cheap and safe CoW clones coincidentally became the cornerstone of RW-locking, too. :star_struck:

mujacica (approved):

PR #2087, opened 2026-09-10
feat: allow initial scope configuration before crash handler startup

### LINKED ISSUE #2086: Allow tags to be configured before crash handler startup
Tags can currently only be set after `sentry_init()`. This creates a window where an immediate crash or fast-fail can be captured without essential correlation tags, especially with out-of-process Native and Crashpad handlers.

Add an options API for initial scope tags so they are included in crash snapshots before handler activation and propagated to integrations such as Windows Error Reporting (WER).

nathan-white-mojang:
Adds `sentry_options_set_initial_scope` to configure tags, contexts, users, breadcrumbs, and attachments before crash handlers start. Native, Crashpad, and WER preserve this data for immediate crashes, fast-fails, and retained crash envelopes.

Fixes #2086.


jpnurmi:
I like the simple and targeted solution to the problem, but I wish it was a bit more flexible and would allow early initialization of other scope data too, not just tags, and preferably without duplicating the scope API in options.

What if we would take some inspiration from the Cocoa SDK that allows [configuring the initial scope](https://docs.sentry.io/platforms/apple/guides/macos/configuration/options/#initialScope)?

```c
static void
configure_initial_scope(sentry_scope_t *scope, void *user_data)
{
    sentry_scope_set_tag(scope, "tag", "initial");

    sentry_value_t context = sentry_value_new_object();
    sentry_value_set_by_key(context, "foo", sentry_value_new_string("bar"));
    sentry_scope_set_context(scope, "initial", context);

    sentry_value_t user
        = sentry_value_new_user("1", "user", "initial@example.com", NULL);
    sentry_scope_set_user(scope, user);
}

sentry_options_set_initial_scope(options, configure_initial_scope, NULL);
```

What do you think?

nathan-white-mojang:
@jpnurmi, love the suggestion, thank you. Please let me know if the changes I've made align with your thoughts!

jpnurmi:
I took a quick Saturday-morning-coffee glance at this, and it looks great! Thanks for the PR! :pray:

mujacica (approved):

mujacica:
This looks amazing! Thanks @nathan-white-mojang for the contribution and @jpnurmi for taking a look!

Issue #2050, opened 2026-09-02
Align and flesh out trace propagation API


limbonaut:
The [trace propagation spec's Public API section](https://develop.sentry.dev/sdk/foundations/trace-propagation/#public-api) wants three methods: `continue_trace`, `get_trace_data`, and `start_new_trace` (naming can vary). Native already has most of the underlying behavior, but only the trace reset API is directly reachable from public.

**`continue_trace` is not provided without starting a transaction.** 
`sentry_transaction_context_update_from_header()` already parses `sentry-trace` and `baggage` and checks `sentry-org_id` for strict continuation, but it writes into a transaction context and is only applied by the transaction-start path. The spec requires a public continuation API that updates trace propagation state directly and **MUST NOT** create a segment span (aka transaction or topmost span).

> The method MUST NOT create a new segment span on its own.

**`get_trace_data` is not provided without an active span.** 
`sentry_span_iter_headers()` and `sentry_transaction_iter_headers()` both require a live span or transaction. Similar gap as in the previous issue.

**`start_new_trace` exists as `sentry_regenerate_trace()`.** 
Already correct. Consider renaming for clarity.

Without spans there is no way to propagate a trace at all. The spec says an SDK with no `traces_sample_rate` must still continue incoming traces and send `sentry-trace` and `baggage` ([default propagation](https://develop.sentry.dev/sdk/foundations/trace-propagation/#default-propagation-no-tracing-config)), and [deferred sampling](https://develop.sentry.dev/sdk/foundations/trace-propagation/#deferred-sampling) describes what to send when no span is available.

## Proposal

> I marked API that we need in the Godot SDK with `// [G]`.

```c
// New API for continuing a trace from headers.
SENTRY_API void sentry_continue_trace(
    const char *sentry_trace, const char *baggage); // [G]
SENTRY_API void sentry_continue_trace_n(
    const char *sentry_trace, size_t sentry_trace_len,
    const char *baggage, size_t baggage_len);
// Same, for a given scope, so concurrent units of work can hold distinct traces.
SENTRY_API void sentry_scope_continue_trace(sentry_scope_t *scope,
    const char *sentry_trace, const char *baggage); // [G]
SENTRY_API void sentry_scope_continue_trace_n(sentry_scope_t *scope,
    const char *sentry_trace, size_t sentry_trace_len,
    const char *baggage, size_t baggage_len);

// The missing scope twin of sentry_set_trace.
[...]

PR #2095, opened 2026-09-14
feat(tracing): Add start new trace API

limbonaut:
Adds `sentry_start_new_trace()` as the public name for starting a new trace, matching Sentry’s [`startNewTrace` specification](https://develop.sentry.dev/sdk/foundations/trace-propagation/#startnewtrace). Keeps `sentry_regenerate_trace()` as a deprecated compatibility alias and updates tracing tests to use the new API.

- Refs #2050


tests/unit/test_tracing.c
```diff
@@ -2071,6 +2071,27 @@ SENTRY_TEST(propagation_context_init)
     sentry_close();
 }
 
+SENTRY_TEST(regenerate_trace_deprecated_alias)
```
jpnurmi: not sure if this is worth the noise 🥲

jpnurmi (approved):

tests/unit/test_tracing.c
limbonaut: You mean the test? Sure, we can drop it.

PR #2099, opened 2026-09-15
feat!: Add hint support to scope-level event capture

### LINKED ISSUE #2098: Feature request: hints
Allow callers to pass `sentry_hint_t` to event capture and use it in `before_send` callbacks, matching the hint support already available for feedback. This enables event-specific attachments, including attachments added by callbacks before sending.

Hints are particularly useful for downstream SDK integrations. Applications that do not need them should ideally retain convenient capture calls. The design therefore needs to balance migration costs, everyday usability, and the number of API variants.

The tables below show call signatures; return types remain unchanged. All hint arguments accept `NULL`.

## Option 1: Preserve existing capture signatures

Add `_with_hint` alternatives for global and scoped event capture.

| Before | After |
|---|---|
| `sentry_capture_event(event)` | Unchanged |
| — | Add `sentry_capture_event_with_hint(event, hint)` |
| `sentry_scope_capture_event(scope, event)` | Unchanged |
| — | Add `sentry_scope_capture_event_with_hint(scope, event, hint)` |
| `sentry_capture_feedback(feedback)` | Unchanged |
| `sentry_capture_feedback_with_hint(feedback, hint)` | Unchanged |
| `sentry_scope_capture_feedback(scope, feedback, hint)` | Unchanged |

**Breaking consequences:** No existing capture calls change. The shared `before_send` callback change described below still applies.

**Tradeoffs:**

- Preserves convenient event capture without hints.
- Lets integrations opt into hints explicitly.
- Adds two functions.
- Leaves scoped event capture with two variants, while scoped feedback has one function accepting a hint.

Sentry Go provides a precedent for this approach:

| Without a hint | With a hint |
|---|---|
| `Hub.CaptureEvent(event)` | `Hub.CaptureEventWithHint(event, hint)` |

[`Hub.CaptureEvent` delegates to `Hub.CaptureEventWithHint(event, nil)`](https://github.com/getsentry/sentry-go/blob/2e882527398d211a63cfcc93b9f8a152c108a3de/hub.go#L214-L219), preserving convenient capture for users who do not need hints.

## Option 2: Keep `_with_hint` alternatives only for global capture

Preserve global capture signatures, but add a hint argument directly to scoped event capture, matching scoped feedback. This is the current implementation.

| Before | After |
|---|---|
| `sentry_capture_event(event)` | Unchanged |
| — | Add `sentry_capture_event_with_hint(event, hint)` |
| `sentry_scope_capture_event(scope, event)` | `sentry_scope_capture_event(scope, event, hint)` |
| `sentry_capture_feedback(feedback)` | Unchanged |
| `sentry_capture_feedback_with_hint(feedback, hint)` | Unchanged |
| `sentry_scope_capture_feedback(scope, feedback, hint)` | Unchanged |

**Breaking consequences:** Existing scoped event callers must add a hint or `NULL`. Global capture calls remain compatible. The shared callback change also applies.

**Tradeoffs:**

- Preserves convenient global capture without hints, using the separate-alternative pattern shown in Sentry Go above.
- Makes scoped event and feedback capture consistent.
- Adds only one function.
- Requires scoped event callers to pass `NULL` even when they only want to apply a scope.
- Limits capture signature changes to the relatively recently introduced scoped event API.

## Option 3: Accept hints directly in all event and feedback capture functions

Use one function for each capture operation, always accepting a hint argument.

| Before | After |
|---|---|
| `sentry_capture_event(event)` | `sentry_capture_event(event, hint)` |
| `sentry_scope_capture_event(scope, event)` | `sentry_scope_capture_event(scope, event, hint)` |
| `sentry_capture_feedback(feedback)` | `sentry_capture_feedback(feedback, hint)` |
| `sentry_capture_feedback_with_hint(feedback, hint)` | Remove; use `sentry_capture_feedback(feedback, hint)` |
| `sentry_scope_capture_feedback(scope, feedback, hint)` | Unchanged |

No `sentry_capture_event_with_hint` function would be introduced.

**Breaking consequences:** Global event, global feedback, and scoped event callers must add a hint or `NULL`. Existing feedback `_with_hint` callers must rename their calls. The shared callback change also applies.

**Tradeoffs:**

- Provides consistent signatures without `_with_hint` alternatives.
- Requires migration of the widely used `sentry_capture_event` API.
- Introduces permanent friction for applications that do not need hints: even basic capture requires `sentry_capture_event(event, NULL)`.

Go's lower-level client API illustrates accepting a hint directly:

| API | Signature |
|---|---|
| `Client.CaptureEvent` | `CaptureEvent(event *Event, hint *EventHint, scope EventModifier)` |

[Source](https://github.com/getsentry/sentry-go/blob/2e882527398d211a63cfcc93b9f8a152c108a3de/client.go#L632). Go also retains the convenient Hub functions shown in Option 1, so this is precedent for explicit hints at a lower level, rather than requiring them throughout the public capture API.

## Callback changes shared by all options

Change the `before_send` hint parameter from `void *` to `sentry_hint_t *`. Feedback already uses the typed hint.

| Callback | Before | After |
|---|---|---|
| `before_send` | `(sentry_value_t event, void *hint, void *user_data)` | `(sentry_value_t event, sentry_hint_t *hint, void *user_data)` |
| `before_send_feedback` | `(sentry_value_t feedback, sentry_hint_t *hint, void *user_data)` | Unchanged |

This is a breaking callback signature change. Return types and callback registration functions remain unchanged.

## External usage and migration impact

The relevant migration burden is external applications and libraries, excluding Sentry projects, downstream SDKs, forks, and vendored SDK code.

The earlier manually filtered search reported:

| Existing API | External caller files | External repositories | Capture calls affected by |
|---|---:|---:|---|
| `sentry_capture_event` | 75 | 69 | Option 3 |
| `sentry_capture_feedback` | 3 | 2 | Option 3 |
| `sentry_capture_feedback_with_hint` | 1 | 1 | Option 3 |

These are provisional historical counts, not an exhaustive census; the saved results are no longer available to reverify. They count files containing callers, not individual calls. No verified external counts were established for scoped event capture or `before_send`.

[Inspect `sentry_capture_event` usage on GitHub](https://github.com/search?q=%22sentry_capture_event%22+NOT+org%3Agetsentry+NOT+org%3Asentry-demos+NOT+is%3Afork+NOT+path%3Asentry.h+NOT+path%3Asentry_core.c&type=code). This query excludes Sentry organizations, forks, `sentry.h`, and `sentry_core.c`; downstream SDKs and other SDK copies still require manual filtering. The browser query has not been successfully verified and does not reproduce the counts above.

## Comparison

| Consideration | Option 1 | Option 2 | Option 3 |
|---|---|---|---|
| Existing global capture calls | Preserved | Preserved | Must migrate |
| Existing scoped event calls | Preserved | Must migrate | Must migrate |
| Global capture without hints | No extra argument | No extra argument | Must pass `NULL` |
| Scoped event capture without hints | No extra argument | Must pass `NULL` | Must pass `NULL` |
| API variants | Global and scoped event alternatives | Global alternatives only | No alternatives |
| `before_send` callback signature | Breaking change | Breaking change | Breaking change |

Option 1 prioritizes compatibility and convenience. Option 2 preserves global convenience while making scoped event and feedback capture consistent. Option 3 reduces API variants but imposes migration work and ongoing extra arguments on callers who do not need hints.

jpnurmi:
Accept hints in `sentry_scope_capture_event` and pass them to `before_send`. Keep `sentry_capture_event` unchanged.

Merge hint and scope attachments before `before_send` and `before_send_feedback`, then use the resulting hint attachments in the envelope. Add `sentry_hint_remove_attachment` and
`sentry_hint_clear_attachments` to allow filtering attachments without modifying scopes.

> [!CAUTION]
> BREAKING CHANGE: `before_send` callbacks now take `sentry_hint_t *` instead of `void *` for the hint parameter. `sentry_scope_capture_event` now requires a hint argument; pass `NULL` when no hint is needed. Scope attachment changes inside either callback no longer affect the current event; modify the hint instead.

Close: #2098


src/sentry_core.c
```diff
@@ -2023,8 +2026,10 @@ capture_minidump(sentry_path_t *dump_path)
         sentry_value_t event = sentry_value_new_event();
         sentry_value_set_by_key(
             event, "level", sentry__value_new_level(SENTRY_LEVEL_FATAL));
-        sentry_envelope_t *envelope
-            = sentry__prepare_event(options, event, &event_id, true, NULL);
+        sentry_hint_t hint = { sentry_value_new_null() };
+        sentry_envelope_t *envelope = sentry__prepare_event(
+            options, event, &event_id, true, NULL, &hint);
+        sentry_value_decref(hint.attachments);
```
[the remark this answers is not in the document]
jpnurmi: The hint is borrowed for the callback; retaining its pointer is unsupported.

Heap allocation would not make retention safe either: regular event capture frees its hint after capture completes. The stack-allocated minidump hint remains valid throughout the callback and envelope preparation.

mujacica (approved):

include/sentry.h
```diff
@@ -1486,6 +1493,9 @@ SENTRY_API void sentry_options_set_send_default_pii(
  * call `sentry_value_decref` on the provided event and return a
  * `sentry_value_new_null()` instead.
  *
+ * The hint is always provided and can be used to modify attachments on the
```
limbonaut, line 1496: We need to document a gap: crash backends currently do not pass a hint.

limbonaut (approved):
Looks good!

tustanivsky (commented):
Just to clarify - do we support hints in `before_send` for crash events as well? All four backends seem to pass `NULL` as the hint there 🤔

jpnurmi:
> Just to clarify - do we support hints in `before_send` for crash events as well? All four backends seem to pass `NULL` as the hint there 🤔

Hmm, but sentry-unreal sets `on_crash` which replaces `before_send` for crash events. :confused: 

EDIT: Which means, do we then need `hint` for `on_crash`, too?

tustanivsky:
> Hmm, but sentry-unreal sets on_crash which replaces before_send for crash events

Right, for Unreal it’s `on_crash` that matters so a `before_send` fix alone wouldn’t help there. But it still makes sense for users who rely solely on `before_send`.

> Which means, do we then need hint for on_crash, too?

Yes, I think it would be good to align what both callbacks carry so downstream SDKs get the same capabilities regardless of which hook handles the event.

include/sentry.h
jpnurmi: Coming in https://github.com/getsentry/sentry-native/pull/2112

